PR #2, opened 2016-02-01
fix(layout): SD-3658, SD-3948, SD-3965 Removed eye from package, layout issues in task, name in user unlock popup

pavlovicnemanja:
- [x] SD-3658 Delete eye button from package edit-> right menu
- [x] SD-3948 Layout issue in the tasks info section
- [x] SD-3965 The unlock popup window does not display the correct name of the user that unlocks the item

### DIFF
--- a/scripts/superdesk-packaging/packaging.js
+++ b/scripts/superdesk-packaging/packaging.js
@@ -804,18 +804,6 @@
             backend: {rel: 'archived'}
         });
     }])
-    .config(['authoringWidgetsProvider', function(authoringWidgetsProvider) {
-        authoringWidgetsProvider
-            .widget('search', {
-                icon: 'view',
-                label: gettext('Search'),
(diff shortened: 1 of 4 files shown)


scripts/superdesk-packaging/packaging.js
```diff
@@ -804,18 +804,6 @@
             backend: {rel: 'archived'}
         });
     }])
-    .config(['authoringWidgetsProvider', function(authoringWidgetsProvider) {
```
petrjasek: I think it's used in some e2e tests, so this will require more changes

pavlovicnemanja: Yep, I also deleted e2e test that was testing this widget.

petrjasek:
:+1:

ioanpocol:
:+1:

PR #5, opened 2016-02-03
fix(authoring): close item when user editing it is removed from desk

petrjasek:
currently it asks user to save his work into his private space,
but after reload user has still that item opened for edit - but is not able to save.

after fix user won't be on edit item page anymore, instead will be on item list page.

SD-2822


mdhaman:
:+1:

mugurrus:
:+1:

PR #7, opened 2016-02-03
Item actions menu is closed on scroll

Aca-jov:
SD-3070
SD-3853 - I think this is now finally fixed
SD-3917 - Small fix

### DIFF
--- a/scripts/superdesk-monitoring/monitoring.js
+++ b/scripts/superdesk-monitoring/monitoring.js
@@ -245,7 +245,8 @@
      *
      * it's a directive so that it can be put together with authoring into some container directive
      */
-    function MonitoringViewDirective() {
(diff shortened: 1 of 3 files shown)


mugurrus:
:+1:

scripts/superdesk-monitoring/monitoring.js
```diff
@@ -253,6 +254,19 @@
             scope: {
                 type: '=',
                 state: '='
+            },
+            link: function(scope, elem) {
+                var containerElem = elem.find('.content-list');
+                containerElem.on('scroll', handleContainerScroll);
```
petrjasek: i think you should call `containerElem.off('scroll');` on `$destroy` event, otherwise this will leak memory

ioanpocol:
:+1:  (except comment)

PR #13, opened 2016-02-05
[sd-3986]-(fix) - Keyboard operation and focus issues in headers's me…

sjunaid:
…ta-data drop-downs

### DIFF
--- a/scripts/superdesk-authoring/metadata/metadata.js
+++ b/scripts/superdesk-authoring/metadata/metadata.js
@@ -352,8 +352,8 @@ function MetadataWordsListEditingDirective($timeout) {
  * @param {String} unique - specify the name of the field, in list item which is unique (qcode, value...)
  *
  */
-MetadataListEditingDirective.$inject = ['metadata', '$filter'];
-function MetadataListEditingDirective(metadata, $filter) {
+MetadataListEditingDirective.$inject = ['metadata', '$filter', '$timeout'];
+function MetadataListEditingDirective(metadata, $filter, $timeout) {
     return {
         scope: {
             item: '=',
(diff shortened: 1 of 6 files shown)


scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -457,8 +457,10 @@ function MetadataListEditingDirective(metadata, $filter) {
                         scope.activeTree = _.without(scope.activeTree, term);
                     }
 
-                    scope.postprocessing();
-                    scope.change({item: scope.item});
+                    $timeout(function() {
+                        scope.postprocessing();
+                        scope.change({item: scope.item});
+                    }, 50);
```
petrjasek: probably there is no need to trigger full digest, so it would be better to use `50, false);` and call `scope.applyAsync` in timeout

scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -469,6 +471,7 @@ function MetadataListEditingDirective(metadata, $filter) {
                             scope.activeTree = scope.tree[null];
                         } else {
                             scope.terms = _.clone(scope.activeTree) || [];
+                            scope.allSelected = (scope.item[scope.field].length === scope.list.length) ? true : false;
```
petrjasek: `scope.allSelected = scope.item[scope.field].length === scope.list.length;`  should be enough

spec/users_spec.js
```diff
@@ -223,6 +224,43 @@ describe('users', function() {
...
+                monitoring.openMonitoring();
+                authoring.navbarMenuBtn.click();
+                authoring.newPlainArticleLink.click();
+
+                browser.sleep(100);
+                //Open subject metadata dropdown field
+                authoring.getCategoryMetadataDropdownOpened();
+                browser.sleep(100); //wait a bit
+
+                var cat = element(by.id('category-setting'));
```
ioanpocol: Any element selection should be moved on helper files.

petrjasek:
:+1:

mdhaman:
:+1:

PR #16, opened 2016-02-07
feat interactive macros

petrjasek:
when calling a macro it can return a `diff` object which will be then used by client for highlighting text and replacing matches one by one.


mdhaman:
:+1:

pavlovicnemanja:
:+1:

mugurrus:
:+1:

PR #17, opened 2016-02-09
fix(spike): only display current desk spiked items

petrjasek:
instead of reading it from monitoring config on desk it should
only display desk spiked items. on custom workspace keep using
the workspace settings.

SD-4002


mdhaman:
:+1:

ioanpocol:
:+1:

PR #21, opened 2016-02-10
[SD-4010] On Updating the role of an user it is not properly reflecting privileges in the privileges tab.

mdhaman:


akintolga:
:+1:

scripts/superdesk-users/users.js
```diff
@@ -1390,30 +1390,50 @@
...
+                            notify.error(gettext('User not found.'));
+                            console.log(error);
+                            return $q.reject(error);
+                        });
+                    }
 
-                    // the last user privileges that were saved on the server
-                    scope.origPrivileges = angular.copy(scope.user.privileges);
+                    function getPrivileges() {
+                        api('privileges').query().then(function(result) {
```
petrjasek: seems like missing return? without this it won't be able to stop the promise chain

mdhaman: yes

mugurrus:
:+1: (a regression test would be good to have)

PR #23, opened 2016-02-10
[sd-3986]-(fix:[SD-3994]) - Keyboard operation and focus issues in su…

sjunaid:
…bject metadata drop-down in search filter panel

### DIFF
--- a/scripts/superdesk-search/views/item-search.html
+++ b/scripts/superdesk-search/views/item-search.html
@@ -19,6 +19,29 @@
     </fieldset>
     <fieldset ng-show="repo.search === 'local'">
         <div class="field">
+            <label for="search-slugline">
+                {{:: 'Slugline' | translate}}
+            </label>
+            <div class="control">
+                <input type="text" id="search-slugline" ng-model="meta.slugline">
+            </div>
+        </div>
+        <div class="field">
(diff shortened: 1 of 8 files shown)


scripts/superdesk-search/views/item-search.html
```diff
@@ -19,23 +19,38 @@
...
             <div class="control">
-                <input type="text" id="search-headline" ng-model="meta.headline">
+                <input type="text" id="search-slugline" ng-model="meta.slugline">
             </div>
         </div>
-
         <div class="field">
-            <label for="search-slugline">
-                {{:: 'Slugline' | translate}}
+            <label for="search-subject">
```
petrjasek: just wondering why moving those fields around?

sjunaid: just arranged in similar order that appears in header section of authoring, slugline > subject > headline and so on.

mdhaman: would be worth confirming the order with Andrew.

sjunaid: just confirmed, no issues.

mdhaman:
:+1:

petrjasek:
:+1:

PR #28, opened 2016-02-11
fix(search): set explicitly when there should be aggregations

petrjasek:
server will be able to skip this if not required.
+ in search only refresh once per sec


akintolga:
:+1:

akintolga:
This PR addresses the second part of this ticket: https://dev.sourcefabric.org/browse/SD-3991

mdhaman:
:+1:

PR #29, opened 2016-02-11
fix(e2e): fix e2e tests

petrjasek:


scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -630,6 +630,10 @@ function MetadataService(api, $q) {
             });
         },
         removeSubjectTerm: function(term) {
+            if (!this.subjectScope) {
```
petrjasek: @sjunaid had to add this - in case there is no authoring opened there is an error

scripts/superdesk-search/search.js
```diff
@@ -2020,10 +2020,9 @@
                         var listComponent = ReactDOM.render(itemList, elem[0]);
 
                         $document.on('keydown', function(evt) {
-                            if (evt.target.className.match('-view list-view') != null) {
```
petrjasek: @sjunaid this prevents keyboard controlling in content lists - you have to put focus on dropdown and stop event propagation instead

petrjasek:
update tests according to design + fix a few bugs

mdhaman:
:+1:

akintolga:
:+1:

PR #33, opened 2016-02-16
[sd-3860]-(fix)-Highlighting two items out of which first is already …

sjunaid:
…highlighted does not effect the second


ioanpocol:
:+1:

petrjasek:
:+1:

PR #35, opened 2016-02-17
[SD-4044] - UI: Load the crop sizes from vocabularies dynamically

sjunaid:


petrjasek:
hey, so what is the idea? set width/height or aspect ratio?

sjunaid:
yes it works by setting width/height in vocab, and 4-3 and 16-9 are just the names.

akintolga:
:+1:

mdhaman:
:+1:

PR #36, opened 2016-02-17
fix(authoring): prevent dropdown rendering if not visible

petrjasek:
it does 2 things atm:
- limits number of cities visible in dropdown - user has to type more
- stops rendering of some dropdowns with subjectcodes if not needed

I think this terms dropdown with typeahead will need refactoring, it's still rendering some items even when not visible, but at least not all of them now.

### DIFF
--- a/scripts/superdesk-authoring/metadata/metadata.js
+++ b/scripts/superdesk-authoring/metadata/metadata.js
@@ -184,8 +184,8 @@ function MetadataCtrl(
     resolvePublishScheduleAndEmbargoTS();
 }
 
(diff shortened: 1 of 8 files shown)


akintolga:
:+1:

mdhaman:
:+1:

petrjasek:
fixed some more issues - like empty category dropdown by default, auto scrolling in dateline dropdown when using keyboard. tried to refactor meta-dropdown but it would take more time. but imo the way how it's working together with sd-typeahead and dropdown-focus directives should be fixed at some point.

PR #38, opened 2016-02-19
[SD-3361] - Improvement in Client Notification intervals for re-connect attempts on websocket disconnection

sjunaid:


mdhaman:
:+1:

scripts/superdesk/notification/notification.js
```diff
@@ -71,10 +71,15 @@
...
                 $interval.cancel(connectTimer);
                 $rootScope.$broadcast('connected');
+                notified = null;
             };
 
             ws.onclose = function(event) {
-                $rootScope.$broadcast('disconnected');
+                if (!notified) {
+                    $rootScope.$broadcast('disconnected');
+                    notified = true;
```
mdhaman: if somebody closes the notification, will the user see it again.

sjunaid: sorry missed that, working on it, so if its closed by click, then notification would re-appear on next timeout, if disconnection persist.

sjunaid: fixed.

scripts/superdesk/notification/notification.js
```diff
@@ -96,23 +97,26 @@
...
         _this.message = null;
 
         $rootScope.$on('disconnected', function(event) {
-            _this.message = 'Disconnected to Notification Server, attempting to reconnect ...';
-            $timeout.cancel(alertTimeout);
-            alertTimeout = $timeout(function() {
-                notify.error(gettext(_this.message));
-            }, 100);
+            if (!notify.disconnectionNotified) {
+                _this.message = 'Disconnected to Notification Server!';
```
petrjasek: gettext call should be here, otherwise it won't be picked when extracting strings

petrjasek:
if I get this right - it seems like the idea is to have there only single message of given type? maybe it can avoid this `disconnectionNotified` variable and use that context param - when you add a notification it should check if there is any other with that value, and if so ignore the new one. but I would probably rename this context to unique or id param. what do you think?

sjunaid:
@petrjasek , right that's make sense, so as discussed, it's now more simplified by removing context and disconnectionNotified at all, and simply checking the existence of message before adding it.
Thanks.

scripts/superdesk/notify/notify.js
```diff
@@ -46,11 +54,15 @@
                 this.stopSaving = function() {
                     this.pop();
                 };
+
+                this.stopWarning = function() {
```
petrjasek: this is probably not needed, you can call `pop` instead

scripts/superdesk/notify/notify.js
```diff
@@ -11,8 +11,12 @@
                     error: 5000
                 };
 
+                var messageTypes = ['info', 'success', 'error', 'warning'];
+
                 this.messages = [];
 
+                //this.disconnectionNotified = null;
```
petrjasek: pls remove

petrjasek:
looks good now, added a few minor things

mugurrus:
:+1:

akintolga:
:+1:

PR #40, opened 2016-02-22
[sd-4064]-(fix) - Arrow keys do not work in the editor

sjunaid:


petrjasek:
not sure, but this might break some global keyboard shortcuts, like trigger a spellcheck when not set to automatic, or is that fine?

sjunaid:
@petrjasek, you are right, sorry i missed that, thanks. its fixed in update.

scripts/superdesk/editor/editor.js
```diff
@@ -219,6 +228,12 @@ function EditorService(spellcheck, $rootScope, $timeout, $q) {
             return true;
         }
 
+        // ignore editing keys
+        if (self.EDITING[event.keyCode]) {
+            event.stopPropagation();
+            return true;
```
petrjasek: I think it shouldn't be here, because this `shouldIgnore` returns true when there is no change to content, but if you press enter or backspace there is a change so it should return false and trigger autosave..
probably better would be to move this outside this function right into event handler..

ioanpocol:
:+1:

mugurrus:
:+1:

PR #41, opened 2016-02-22
merge pa branch

petrjasek:
merging client changes from pa-poc branch to master

it should be usable for pa branch, will start doing changes in main repo to make use of it. on master we need to merge changes on server side which are renaming `description` to `description_text`.

new editor is now in `editor2` folder and `superdesk.editor2` module, not replacing current editor,
so we can toggle between those

@hlmnrmr
there is one issue for pa - in search it's using react which doesn't have draggable implemented, but shouldn't be a big deal and this can be done later this week/sprint.

@vied12 @eozan @naomiaro @ioanpocol
pls no more client related commits on superdesk/superdesk pa-poc branch, you can start working from here, your changes will be available later. in case there is something urgent you can do it but then you should also do it in client core.
also pls review your features you've done for pa-poc, it might not be there anymore ;)


vied12:
Wow, well done !
I will move my current PRs from `superdesk/superdesk` to this repo now.

scripts/superdesk-desks/views/desk-config-modal.html
```diff
@@ -45,6 +45,17 @@ <h3 ng-show="desk.edit._id" translate>Edit "{{desk.edit.name}}" Desk</h3>
                                     <option value=""></option>
                                 </select>
                             </div>
+                            <div class="field">
```
mdhaman: Is this default categories per desk?

scripts/superdesk-workspace/content/content.js
```diff
@@ -84,6 +84,51 @@
...
+
+        /**
+         * Create new item using given content type
+         *
+         * @param {Object} contentType
+         * @return {Promise}
+         */
+        this.createItemFromContentType = function(contentType) {
+            var item = {
+                type: TEXT_TYPE,
```
mdhaman: We are treating `text only` and `rich text (with images, embeds)` items as same. I like the idea.
But, what would be the profile of ingested items? Can we change the profile after ingestion? How do we treat the associated item in NewsML/NITF?

mdhaman:
:+1: awesome.

akintolga:
:+1:

scripts/superdesk-desks/views/desk-config-modal.html
petrjasek: yes

scripts/superdesk-workspace/content/content.js
petrjasek: I think we will have to have some default that would be used for ingest. And should be editable imo. regarding export - I'm not sure yet. we will have to discuss this. but if it's not possible to output this with some format (like anpa) - we will just ignore it imo.

mugurrus:
:+1:

PR #55, opened 2016-02-24
feat: make it possible to override template from core

petrjasek:
it will pick templates from app/scripts folder and put it into
templates cache. it has to have same filename there, so to override
`core/scripts/superdesk/menu/views/menu.html` you have to name it
`app/scripts/superdesk/menu/views/menu.html`.


mugurrus:
:+1:

akintolga:
:+1:

PR #57, opened 2016-02-24
feat(fields): Created directive for controlling required fields UI [SDNTB-21]

pavlovicnemanja:
SDNTB-21

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -1142,6 +1142,7 @@
 
                 function publishItem(orig, item) {
                     var action = $scope.action === 'edit' ? 'publish' : $scope.action;
+                    validate(orig, item);
                     authoring.publish(orig, item, action)
                     .then(function(response) {
                         if (response) {
@@ -1172,6 +1173,21 @@
                             }
                         } else {
                             notify.error(gettext('Unknown Error: Item not published.'));
+                        }
+                    });
+                }
+
+                function validate(orig, item) {
+                    $scope.error = {};
+                    angular.forEach(_.extend(orig, item), function (value, key) {
+                        if (value && value !== '' && value !== null) {
+                            if (typeof value === 'object' && !value.length) {
+                                $scope.error[key] = true;
+                            } else {
+                                $scope.error[key] = false;
+                            }
+                        } else {
+                            $scope.error[key] = true;
                         }
                     });
                 }
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -1180,7 +1180,7 @@
                 function validate(orig, item) {
                     $scope.error = {};
                     angular.forEach(_.extend(orig, item), function (value, key) {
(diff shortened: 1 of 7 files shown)


ioanpocol:
:+1:

scripts/superdesk/ui/ui.js
```diff
@@ -1425,6 +1425,52 @@
...
+                if (_.has(elem[0].attributes, 'sd-meta-terms')) {
+                    $timeout(function () {
+                        if (elem.find('.terms-list').html()) {
+                            validate(elem.find('.terms-list').html());
+                        }
+
+                        elem.find('.dropdown-toggle').on('blur', function () {
+                            validate(elem.find('.terms-list').html());
+                        });
+                    }, 5000);
```
petrjasek: you can add false here..

scripts/superdesk/ui/ui.js
```diff
@@ -1425,6 +1425,52 @@
...
+                } else if (_.has(elem[0].attributes, 'sd-text-editor')) {
+                    elem.find('.text-editor').on('blur', function () {
+                        validate($(this).html());
+                    });
+                } else if (elem[0].nodeName === 'INPUT' || elem[0].nodeName === 'TEXTAREA') {
```
petrjasek: imo would be better if it uses ngModel when possible

ioanpocol:
please rebase

scripts/superdesk-authoring/authoring.js
```diff
@@ -1176,6 +1177,21 @@
                     });
                 }
 
+                function validate(orig, item) {
+                    $scope.error = {};
+                    angular.forEach(_.extend(orig, item), function (value, key) {
+                        if (value && value !== '' && value !== null) {
```
petrjasek: fyi if value is `true` it can't be `''` or `null` so those checks are extra

petrjasek:
not sure why are you using `ngRequired` which works with `ngModel`, it should probably have different name

ioanpocol:
:+1:

pavlovicnemanja:
Hey @petrjasek, ng-required read's from content_types and define witch fields are required and witch not.

scripts/superdesk-authoring/authoring.js
pavlovicnemanja: Never be 100% sure! :smile:

scripts/superdesk-authoring/authoring.js
```diff
@@ -1176,6 +1177,21 @@
                     });
                 }
 
+                function validate(orig, item) {
+                    $scope.error = {};
+                    angular.forEach(_.extend(orig, item), function (value, key) {
+                        if (value) {
```
petrjasek: btw this function could be ```$scope.error[key] = _.isEmpty(value);```

petrjasek:
:+1:

scripts/superdesk-authoring/authoring.js
pavlovicnemanja: :+1: I really like underscore! :smile: tnx

PR #65, opened 2016-02-26
Updated metadata dropdown to manage urgency.

eozan:
SDPA-76
SDPA-77

### DIFF
--- a/scripts/superdesk-authoring/metadata/metadata.js
+++ b/scripts/superdesk-authoring/metadata/metadata.js
@@ -245,15 +245,16 @@ function MetaDropdownDirective($filter, keyboardManager) {
             field: '@',
             icon: '@',
             label: '@',
-            change: '&'
+            change: '&',
+            key: '@'
         },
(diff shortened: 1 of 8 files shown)


scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -252,8 +252,14 @@ function MetaDropdownDirective($filter, keyboardManager) {
             scope.select = function(item) {
                 var o = {};
 
-                if (angular.isDefined(item)) {
-                    o[scope.field] = (scope.field === 'place' || scope.field === 'genre') ? [item] : item.value;
+                if (item) {
+                    if (scope.field === 'place' || scope.field === 'genre') {
```
petrjasek: imo it would be better to have some param for this on directive scope, and then it should use qcode if item has one , if not then value.
so avoiding to hardcode some field names

vied12:
:+1:

actionless:
:+1:

mugurrus:
:+1:

PR #69, opened 2016-02-29
[sd-4097]-(fix) - When a story with company codes re-opened company c…

sjunaid:
…odes fields gets broken.

### DIFF
--- a/scripts/superdesk-authoring/metadata/metadata.js
+++ b/scripts/superdesk-authoring/metadata/metadata.js
@@ -20,7 +20,8 @@ function MetadataCtrl(
         $scope.metadata = metadata.values;
         return preferencesService.get();
     })
-    .then(setAvailableCategories);
+    .then(setAvailableCategories)
+    .then(setAvailableCompanyCodes);
 
     $scope.processGenre = function() {
         $scope.item.genre = _.map($scope.item.genre, function(g) {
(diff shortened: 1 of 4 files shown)


scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -470,23 +499,29 @@ function MetadataListEditingDirective(metadata, $filter, $timeout) {
...
 
-                    scope.selectedTerm = '';
-
-                    if (!reloadList) {
-                        // Remove the selected term from the terms
-                        scope.terms = _.without(scope.terms, term);
-                        scope.activeTree = _.without(scope.activeTree, term);
-                    }
+                    // wait for pending request before applying change (e.g: while autosaving)
+                    var waitForPendingRequests = function() {
```
petrjasek: what is the idea here? wait for autosave to finish? then it would be better imo to broadcast an event from autosave..
also seems like there is some code duplication later

ioanpocol:
Please rebase

scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -91,6 +92,34 @@ function MetadataCtrl(
...
+
+        itemCompanyCodes.forEach(function (companyCode) {
+            assigned[companyCode.qcode] = true;
+        });
+
+        filtered = _.filter(all, function (companyCode) {
+            return !assigned[companyCode.qcode];
+        });
+
+        $scope.availableCompanyCodes = _.sortBy(filtered, 'name');
```
sivakuna-aap: Can we use `superdesk filter sort` instead of `sortBy`?

ioanpocol:
please rebase

ioanpocol:
What is the status of this PR? It is very old one.

PR #70, opened 2016-02-29
Pa image crop

eozan:


eozan:
Closing this PR as it will be worked on on pa-image-crop feature branch

PR #75, opened 2016-02-29
Rebase pa-poc from master

ioanpocol:


petrjasek:
replaced via https://github.com/superdesk/superdesk-client-core/pull/81

PR #88, opened 2016-03-02
chore(ntb): Merge ntb related changes on master

ioanpocol:


pavlovicnemanja:
:+1:

app/scripts/translations.js
```diff
@@ -1,5 +1,13 @@
 angular.module('gettext').run(['gettextCatalog', function (gettextCatalog) {
 /* jshint -W100 */
 	gettextCatalog.setStrings('de', {"My Profile":"DE - My Profile"});
+    gettextCatalog.setStrings('en', {
```
petrjasek: this part shouldn't go to master. we need to keep this project specific

scripts/superdesk-authoring/views/authoring-header.html
```diff
@@ -138,10 +139,10 @@
             </div>
           </div>
         </div>
-
-        <div class="authoring-header__row" ng-if="metadata.subjectcodes && schema.subject">
+
+        <div class="authoring-header__row" ng-if="metadata.subjectcodes && schema.subject && !metadata.subjects">
```
petrjasek: where/how is `metadata.subjects` set?

ioanpocol: This is an alternative list of subjects code for ntb. If it is defined, the current one will not be showed.

app/scripts/translations.js
```diff
@@ -1,5 +1,3 @@
 angular.module('gettext').run(['gettextCatalog', function (gettextCatalog) {
-/* jshint -W100 */
-	gettextCatalog.setStrings('de', {"My Profile":"DE - My Profile"});
```
petrjasek: this should probably stay there otherwise it will break an e2e test

petrjasek:
:+1:

PR #49, opened 2016-02-23
Extracting date/time formats out and fixing SD-4103 SDPA-146

naomiaro:
porting this over to the new client-core

PR #90, opened 2016-03-02
SD-4103 SDPA-146 : Extract date/time formats into config

naomiaro:
Reopening #49 to go into master. Will rebase.


naomiaro:
@akintolga Yeah, metadata-view is confusing, (also created at, last updated etc) since timezone can be specified per article (and it's shown there), it's unclear which timezone all these fields are in (It's displayed in local time of the user regardless of shown timezone) A Timezone offset should be added here for clarity I think.

@petrjasek @hlmnrmr @ioanpocol 
any thoughts about preferred display? Embargo/Publish/Created/Last Updated at metadata times

screen shot 2016-03-02 at 1 53 02 pm

petrjasek:
imo it should display what was set, not necessary local timezone

akintolga:
Timezones are specified only for publish_schedule and embargo fields, it doesn't apply to other fields.

I agree with Petr, no localization needed here for embargo and publish_schedule because those are processed on server side and converted into the local date and time of the selected timezone in a separate field. If no timezone is selected they are regarded as UTC. So we should show just whatever the client set the date/time + timezone (if any). 

For other dates, like expiry, last updated, created etc. they still need to do localization which is how it is at the moment so we don't need to change anything for those fields in the metadata.

naomiaro:
OK, I'm just concerned that with a timezone shown in text as part of the metadata it will look kind of weird if some of the times in this column are displayed in 1 timezone, while others aren't.

akintolga:
I agree, we can always show the timezone if either publish_schedule or embargo is defined and we can even how it on the same line if the space permits.

petrjasek:
:+1:

mdhaman:
:+1:

PR #94, opened 2016-03-02
SDPA-251 Fix typos and duplicate strings in localisation file 

danielhjames:
Fixes for various issues found when reviewing English strings on Transifex. If this pull request is merged, please regenerate the templates.pot file afterwards.

### DIFF
--- a/scripts/superdesk-search/search.js
+++ b/scripts/superdesk-search/search.js
@@ -2167,7 +2167,7 @@
                     scope.save = function(editSearch) {
 
                         function onSuccess() {
-                            notify.success(gettext('Saved search is saved successfully'));
+                            notify.success(gettext('Search was saved successfully'));
                             scope.cancel();
                             scope.sTab = false;
                             scope.edit = null;
@@ -2178,7 +2178,7 @@
                             if (angular.isDefined(error.data._message)) {
                                 notify.error(error.data._message);
                             } else {
-                                notify.error(gettext('Error. Saved search could not be saved.'));
+                                notify.error(gettext('Error. Search could not be saved.'));
                             }
                         }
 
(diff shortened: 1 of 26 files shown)


scripts/superdesk-vocabularies/vocabularies.js
```diff
@@ -73,7 +73,7 @@
         }
 
         function onSuccess(result) {
-            notify.success(gettext('Vocabulary saved succesfully'));
+            notify.success(gettext('Vocabulary saved successfully'));
```
akintolga: 'is' missing here imo...

scripts/superdesk-search/search.js
```diff
@@ -2167,7 +2167,7 @@
                     scope.save = function(editSearch) {
 
                         function onSuccess() {
-                            notify.success(gettext('Saved search is saved successfully'));
+                            notify.success(gettext('Search was saved successfully'));
```
akintolga: Present tense would be better: Search is saved ..

akintolga:
:+1:

scripts/superdesk-vocabularies/vocabularies.js
danielhjames: Or we could use 'was' since the notification takes place immediately after the save.

scripts/superdesk-search/search.js
danielhjames: I'm happy to use present tense for notifications, if you prefer it.

danielhjames: Having looked through the strings again, apart from a few instances of "There is an error", "Password is changed" and "User is not found", we generally use the past tense for notifications. I'll add a commit to make these consistent.

akintolga: Sounds good...

mdhaman:
:+1:

petrjasek:
:+1:

PR #91, opened 2016-03-02
react item lists in monitoring

petrjasek:
- [x] moved react part into its own file
- [x] handle highlights with react
- [x] put menu out of item dom so that it can be used withing scrolled areas
- [x] use react lists in monitoring
- [x] update react views with changes to media-box-list
- [x] fix multi select in monitoring
- [x] fix/implement add to package
- _remove old media-boxes code_ - will do this one later

PR #101, opened 2016-03-03
[SD-4093] - Search results didn't display the last corrected item

akintolga:


scripts/superdesk-search/search.js
```diff
@@ -2050,13 +2070,14 @@
                             var itemsById = angular.extend({}, listComponent.state.itemsById);
 
                             items._items.forEach(function(item) {
-                                if (!itemsById[item._id] || !isSameVersion(itemsById[item._id], item)) {
-                                    itemsById[item._id] = item;
+                                var itemId = search.generateTrackByIdentifier(item);
```
mdhaman: We get `_id` of the document in the item:lock and item:unlock notification. 
https://github.com/akintolga/superdesk-client-core/blob/cd-4093/scripts/superdesk-search/search.js#L2098
we won't be able to find the item in the `itemsById[itemId]` to update the state.

akintolga: That's correct. Now when we lock we use the `id:version` and when an unlock happens we use `updateAllItems` function to update all previous versions to unlock. This is required in published collection where different versions of the same story is visible and when an update happens, returned version is different than the locked one.

akintolga:
depends on https://github.com/superdesk/superdesk-core/pull/210

petrjasek:
either this one or https://github.com/superdesk/superdesk-client-core/pull/91 will need an update, lets see which one gets merged first ;)

petrjasek:
I have a question in general - is there a usecase when there would be multiple versions of same item visible at once?

akintolga:
Multiple versions of the same item are visible when the search is done against the "published" collection. They have the same id (backend does the flicking...)

mdhaman:
:+1:

marwoodandrew:
:+1:

akintolga:
Hey Petr, I'll leave "which one goes in first decision" to you... I'm pretty sure you'll do the right thing :)

petrjasek:
well seems like @ioanpocol decided ;) I will update my version

ioanpocol:
@petrjasek  In the morning your PR already needed a rebase so I merged the other one.

petrjasek:
@ioanpocol sure no problem

PR #104, opened 2016-03-03
Images from PA api, don't have a baseImage rendition

naomiaro:
I had only tested this dialog on an image I uploaded myself. There's no baseImage rendition for images from the PA images API.

### DIFF
--- a/scripts/superdesk-archive/views/single-item-preview.html
+++ b/scripts/superdesk-archive/views/single-item-preview.html
@@ -7,7 +7,7 @@
 
      <!-- picture preview -->
     <div class="full-preview" ng-if="item.type == 'picture'" sd-ratio-calc>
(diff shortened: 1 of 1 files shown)


actionless:
i think it's defined in the settings:
https://github.com/superdesk/superdesk/blob/master/server/settings.py#L136-L146

actionless:
i am concerned what it could slow down the browser in the case if images will be too big

naomiaro:
Ah OK, I'm still not sure why a baseImage rendition has not been created for a PA img then. It has a ype picture, and includes the other types of renditions.

actionless:
papoc don't have renditions defined:
https://github.com/superdesk/superdesk/blob/pa-poc/server/settings.py

but default ones had `baseImage` there:
https://github.com/superdesk/superdesk-core/blob/master/superdesk/factory/default_settings.py#L307-L317

naomiaro:
changing pa img datalayer instead

PR #109, opened 2016-03-03
SDNTB-97 Fixed duplicate native language

vladnicoara:


mdhaman:
:+1:  a test would be good to have.

akintolga:
:+1:

vladnicoara:
hi @mdhaman What I did here is just an insurance to not show the same language more times in the selector. This can be caused by using the translation system to modify button texts for example from 'Next' to 'Take me further' without changing the language, or something similar. I have talked this through with @hlmnrmr when originally doing the work and he agreed that this should not be a translation functionality. The base language that SD is using now is plain English( 'en' ISO 639-1 code). If adaptations are needed then the proper codes should be used en-au, en-ca etc., or, of course, the default text should be changed if the original it's not right :). 
Maybe the guys were just playing with the translation feature on a branch that I was checking out, but I thought it should be verified anyway. If we should have a test for this, there should be a consensus/decision on how the translation should behave. 
Thanks for the thumbs up!

nistormihai:
:+1:

ioanpocol:
:+1:  (one unit tests can help)

mdhaman:
thanks for the clarification @vladnicoara

PR #110, opened 2016-03-03
Fix creating new desks in core master.

naomiaro:


mdhaman:
Please don't merge this as it will break desk creation. Vocabularies are not updated https://github.com/superdesk/superdesk-core/blob/master/apps/prepopulate/data_initialization/vocabularies.json#L272

naomiaro:
Desk creation already doesn't work though in the UI, I'm not sure about these vocabulary though.

scripts/superdesk-desks/views/desk-config-modal.html
```diff
@@ -41,7 +41,7 @@ <h3 ng-show="desk.edit._id" translate>Edit "{{desk.edit.name}}" Desk</h3>
                                 <label translate>Desk Type</label>
                                 <select id="deskType" ng-model="desk.edit.desk_type" required>
                                     <option value=""></option>
-                                    <option id="{{type.value}}" ng-repeat="type in deskTypes" value="{{type.value}}" ng-selected="desk.edit.desk_type === type.value"
+                                    <option id="{{type.qcode}}" ng-repeat="type in deskTypes" value="{{type.qcode}}" ng-selected="desk.edit.desk_type === type.qcode"
```
petrjasek: maybe you can make it compatible with both versions? like `{{ type.qcode || type.value }}`

naomiaro:
Why don't they have the same vocab?

mdhaman:
@naomiaro desk type definition was defined earlier using `value`. Then on PA branch `desk_type` vocab was changed to use `qcode` (which is the right thing). As PA changes are merged, we need to ensure that it works on both `PA` and `master`. 

@marwoodandrew has already started changing vocab on master to avoid this confusion.

For the time being, you can change as @petr suggested.

vied12:
:+1:

petrjasek:
:+1:

PR #118, opened 2016-03-04
add rights to uploaded image metadata

vied12:
Requires https://github.com/superdesk/superdesk-core/pull/208


petrjasek:
:+1: but you can add `translate` to `option`s so those can be translated

PR #120, opened 2016-03-07
[SD-3670] fixed issues relate to interactive macros

mdhaman:
- diff returned from the server was not clear after the item is closed and was visible if a new item is opened.
- replace was not highlighting text.

### DIFF
--- a/scripts/superdesk-authoring/macros/macros.js
+++ b/scripts/superdesk-authoring/macros/macros.js
@@ -2,10 +2,8 @@
 
 'use strict';
 
-MacrosService.$inject = ['api', 'autosave', 'notify', 'editor'];
-function MacrosService(api, autosave, notify, editor) {
-
-    var self = this;
+MacrosService.$inject = ['api', 'notify'];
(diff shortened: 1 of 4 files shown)


mdhaman:
Depends on [PR-211](https://github.com/superdesk/superdesk-core/pull/211)

scripts/superdesk-authoring/macros/macros.js
```diff
@@ -82,11 +81,13 @@ function MacrosReplaceDirective(macros, editor) {
...
+            scope.diff = null;
 
-            scope.$watch(function() {
-                return macros.diff;
-            }, function(diff) {
+            scope.$on('macro:diff', function(evt, diff) {
                 scope.diff = diff;
+            });
+
+            scope.$watch('diff', function(diff) {
```
petrjasek: seems like this watch can be avoided altogether

petrjasek:
:+1:

akintolga:
:+1:

PR #137, opened 2016-03-08
feat(editor): Added indentation for paragraphs in editor [SDNTB-27]

pavlovicnemanja:
SDNTB-27

### DIFF
--- a/scripts/superdesk/editor2/editor.js
+++ b/scripts/superdesk/editor2/editor.js
@@ -1172,6 +1172,11 @@ angular.module('superdesk.editor2', [
                         cancelTimeout(event);
                         updateTimeout = $timeout(updateModel, 800, false);
                     });
+
+                    editorElem.on('paste', function(event) {
+                        updateIndent();
+                    });
+
                     editorElem.on('contextmenu', function(event) {
                         if (editor.isErrorNode(event.target)) {
                             event.preventDefault();
@@ -1207,6 +1212,7 @@ angular.module('superdesk.editor2', [
                         spellcheck.setLanguage(null);
                     });
                     scope.cursor = {};
+                    updateIndent();
                     render(null, null, true);
                 };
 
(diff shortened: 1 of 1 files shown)


pavlovicnemanja:
Wrong branch! Closed!

ioanpocol:
I think that it is the right branch

Aca-jov:
:+1:

ioanpocol:
:+1:

ioanpocol:
please rebase

vied12:
This can't be done in CSS instead ?

pavlovicnemanja:
@vied12 It can, but how to display it on output? On sending item, we are not sending CSS :)

vied12:
From my point of view, provided content should be neutral in terms of style.
I think this is up to the view to add style to it where it will appear.
Because maybe everybody don't want (the same) indentation. This choice can also depends where you plan to show the text.

pavlovicnemanja:
This is a convention on text formatting in Norway which is required by NTB customers. Here is description: https://dev.sourcefabric.org/browse/SDNTB-27

This is NTB request.

ioanpocol:
I think we can add a setting and perform this indentation only if it is required by deployed instance.
You can add the setting on superdesk repo->  client/app/index.html
        config.features = {
            preview: 1,
            alchemy: 1
        };

petrjasek:
@pavlovicnemanja afaik the idea was that they will add those spaces on output, and we only do some css to make it look like there are spaces

vied12:
I don't believe we should hard code NTB's special convention in the Superdesk master branch.
To satisfy their needs, I would instead create a special filter on the Public Content API level that would customize the output for them.

I'm not a big fan for the settings solution neither. I don't think that we should save the style in the database at this step. And I believe that having too much settings and options will make the code harder to maintain and debug.

/ping @petrjasek and @hlmnrmr

pavlovicnemanja:
First implementation was with pure CSS, just on frontend side... But after discussion on morning scrum with @petrjasek we agreed to add styles on element so it will be visible on output. Maybe I misunderstand something.

petrjasek:
output updates should be handled by their system, so afaik we should not do any changes to content, just make it look like those spaces are there. and that css change should go into superdesk/superdesk ntb branch

pavlovicnemanja:
Ok, so just to be clear, we should add this indentation without any changing of content (adding styles, adding classes...). So this will be visible only on frontend in our editor and preview.

Also, this acceptance criteria will be ignored:
"We should distinguish between the different "paragraph types" in the output format"

vied12:
This is how I see it :)
And if they really need it in the output, the Content API should be in theory a better place for that

PR #147, opened 2016-03-10
[SD-3097] - Case sensitive dictionary lookup

akintolga:
- Search in the editor should be case-sensitive
- Search in dictionary settings page should be case-insensitive


petrjasek:
:+1:

mugurrus:
:+1:

PR #155, opened 2016-03-10
chore(e2e): Fix e2e tests

ioanpocol:


mugurrus:
:+1:

ioanpocol:
Peer test PR: https://github.com/superdesk/superdesk/pull/1828

spec/send_spec.js
```diff
@@ -114,8 +106,6 @@ describe('send', function() {
         monitoring.showHideList();
 
         authoring.sendTo('Politic Desk');
-        //Spell check confirmation modal save action
-        authoring.confirmSendTo();
```
petrjasek: it should stay there, problem is the implementation not working with promises..
it's fixed here https://github.com/superdesk/superdesk-client-core/pull/91/files#diff-5c9045fd902f440d5b002c381e24165cR77

ioanpocol: I think that is better to add back on your PR and in this way it will be validated that the issue is fixed.

petrjasek: ok

PR #156, opened 2016-03-10
[SD-4033] Display the Urgency field as News Value

marwoodandrew:


petrjasek:
can we keep it urgency and change it in superdesk repo master branch via [translations](https://github.com/superdesk/superdesk/blob/pa-poc/client/app/translations.js) ? seems like it's aap specific, in ninjs it's urgency so it would be less confusing imo

marwoodandrew:
Yes a better approach I think.

PR #160, opened 2016-03-11
Fixed keyword UI.

eozan:
SDPA-258

### DIFF
--- a/scripts/superdesk-authoring/metadata/metadata.js
+++ b/scripts/superdesk-authoring/metadata/metadata.js
@@ -294,8 +294,8 @@ function MetaDropdownDirective($filter, keyboardManager) {
     };
 }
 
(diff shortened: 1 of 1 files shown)


scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -328,6 +329,14 @@ function MetaTagsDirective(api) {
                 scope.adding = false;
             };
 
+            scope.$watch('adding', function() {
+                if (scope.adding) {
+                    $timeout(function() {
```
petrjasek: this will trigger full digest which is not needed for elem focus, better add `0, false` arguments

eozan: Fixed.

PR #164, opened 2016-03-14
[SD-4033] use translate to rename Urgency to News Value

marwoodandrew:


app/scripts/translations.js
```diff
@@ -1,5 +1,11 @@
 angular.module('gettext').run(['gettextCatalog', function (gettextCatalog) {
 /* jshint -W100 */
 	gettextCatalog.setStrings('de', {"My Profile":"DE - My Profile"});
+    gettextCatalog.setStrings('en', {
+        "URGENCY": "News Value",
+        "Urgency": "News Value",
+        "Urgency stats": "News Value stats",
+        "urgency": "News Value"
```
ahilles107: Why we have so many (3) keys (with just different size of letters) keys for this same translation string? Translators will need translate  "News Value" at least 3 times that string.

petrjasek: because it's case sensitive

ahilles107: I see, but value isn't - so we will have 3 this same translations for different keys. Which is bad.

marwoodandrew:
@petrjasek The thing I don't get is why gettext would not translate the value in seach.js for the sort order, is there something else I need to do there?

app/scripts/translations.js
```diff
@@ -1,5 +1,11 @@
...
 	gettextCatalog.setStrings('de', {"My Profile":"DE - My Profile"});
+    gettextCatalog.setStrings('en', {
+        "URGENCY": "NEWS VALUE",
+        "Urgency": "News Value",
+        "urgency": "news value",
+        "Urgency stats": "News Value stats"
+    });
 /* jshint +W100 */
-}]);
+}]);
```
ahilles107: missing empty line at end of file

ioanpocol:
please rebase

petrjasek:
:+1:

app/scripts/translations.js
petrjasek: right, good point

mdhaman:
:+1:

PR #168, opened 2016-03-14
Editor3 SDPA-288

vied12:
- [x] hide the empty text fields on mouse out
- [x] "Add text here" replaced by "Add content here"
- [x] A (+) button replaces the `Embed` and `Picture` toolbar buttons

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -102,7 +102,7 @@
         abstract: {order: 3, formatOptions: ['bold', 'italic', 'underline', 'anchor', 'removeFormat']},
         body_html: {
             order: 6,
-            formatOptions: ['h2', 'bold', 'italic', 'underline', 'quote', 'anchor', 'embed', 'picture', 'table', 'removeFormat']
+            formatOptions: ['h2', 'bold', 'italic', 'underline', 'quote', 'anchor', 'embed', 'picture', 'removeFormat']
         },
         byline: {order: 4},
         dateline: {order: 5},
--- a/scripts/superdesk/editor2/views/block-text.html
+++ b/scripts/superdesk/editor2/views/block-text.html
@@ -1,11 +1,12 @@
 <div class="dropdown" dropdown>
+    <sd-add-content ng-if="config.multiBlockEdition" class="text-editor__add-content"></sd-add-content>
     <pre class="text-editor editor-type-text"
         ng-show="type=='preformatted'"
         spellcheck="false">
     </pre>
     <!-- Placeholder text is temporary, ideally it should come from content type -->
     <div class="text-editor editor-type-html"
-        placeholder="Add text here"
+        placeholder="{{ ::'Add content here' | translate }}"
         ng-show="type=='text' || !type"
         spellcheck="false">
     </div>
--- a/scripts/superdesk/editor2/views/editor.html
+++ b/scripts/superdesk/editor2/views/editor.html
@@ -6,12 +6,14 @@
 </a>
 <div ng-repeat="block in vm.blocks track by vm.generateBlockId(block)"
      class="block__container"
-     ng-class="{'block--selected': vm.blockToMoveIndex === $index}">
+     ng-class="{'block--selected': vm.blockToMoveIndex === $index,
+                'block--not-alone-and-empty': vm.blocks.length > 1 && !block.body}">
     <div class="block__actions" ng-if="vm.config.multiBlockEdition && !vm.reorderingMode">
         <div class="block__remove" ng-click="vm.removeBlock(block)">
             <i class="icon-close-small"></i>
         </div>
-        <div class="block__move" ng-click="vm.enableReorderingMode($index, $event)">
+        <div class="block__move" ng-click="vm.enableReorderingMode($index, $event)"
+             ng-if="vm.blocks.length > 1 && block.body">
             <i class="svg-icon-move"></i>
         </div>
     </div>
--- a/bower.json
+++ b/bower.json
@@ -18,7 +18,7 @@
     "ng-file-upload": "11.2.3",
     "raven-js": "1.1.11",
     "jquery-ui": "1.11.2",
-    "medium-editor": "~5.13.0",
+    "medium-editor": "~5.15.0",
     "ment.io": "0.9.22",
     "angular-moment": "~0.10.3",
     "classnames": "~2.2.0",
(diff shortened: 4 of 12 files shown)


scripts/superdesk/editor2/views/block-text.html
```diff
@@ -5,7 +5,7 @@
     </pre>
     <!-- Placeholder text is temporary, ideally it should come from content type -->
     <div class="text-editor editor-type-html"
-        placeholder="Add text here"
+        placeholder="Add content here"
```
ahilles107: how it can be translated?

scripts/superdesk-authoring/authoring.js
```diff
@@ -61,9 +61,9 @@
 
     var DEFAULT_SCHEMA = Object.freeze({
         slugline: {maxlength: 24},
-        headline: {maxlength: 64, order: 1, editor: ['h1', 'h2', 'bold', 'italic', 'underline', 'quote', 'anchor']},
-        abstract: {maxlength: 160, order: 2, editor: ['h1', 'h2', 'bold', 'italic', 'underline', 'quote', 'anchor']},
-        body_html: {order: 5, editor: ['h1', 'h2', 'bold', 'italic', 'underline', 'quote', 'anchor', 'embed', 'picture', 'table']},
```
Aca-jov: Why this 3 have been removed from here?

vied12: It's explained here: https://github.com/superdesk/superdesk-client-core/commit/9af018a9b45b03cb1f9a68f5ff5f803ee84430f4#commitcomment-16783689
:)

Aca-jov: Yes, but that's the way it works, it's fine. All settings are made in json file from where it takes settings, and default_schema is there just in case, as some default settings. If you want to make some changes, they are changed in:
https://github.com/superdesk/superdesk/blob/ntb/server/data/content_types.json

vied12: what is fine ? the toolbar that disappear ? :)
Sorry I don't understand the logic here. Why the value of `body_html.editor` contains first the buttons list and few second after it disappears ?

Aca-jov: I mean code is fine.
If you want to change toolbar settings, it is changed here: https://github.com/superdesk/superdesk/blob/pa-poc/server/data/content_types.json, not in default_schema. default_schema is defined in case content_types.json is not defined (like on master)

vied12: I did not want to change the toolbar settings, I want to repair what has been broken :)
Do you have another idea on how we can fix this issue ?

vied12: It is possible that the default values are set before the `content_types.json` was loaded ? And then this might be why the toolbar buttons disappear when `content_types.json` is loaded
It would be maybe more clear to use the default values if and only if there is no custom values. What do you think?

Aca-jov: Well, I commented on this 3 lines you changed, this changes removes toolbar settings.

This idea in previous comments is just fine, just check it out :)

vied12: Which idea is fine ?
Let's not discuss only about these lines that belongs to a bigger change (https://github.com/superdesk/superdesk-client-core/pull/168/commits/c8b7b5d2147cc6ec20c7e8a356acdd8dd0be85e9)

What I understand now is:
* the buttons available in the toolbar are now customizable since https://github.com/superdesk/superdesk-client-core/commit/9af018a9b45b03cb1f9a68f5ff5f803ee84430f4
* for each client project, we need now to define these toolbar buttons in `content_types.json`.

Is that right?

But a problem remains: the value of `schema.abstract.editor` in `data-editorformat="schema.abstract.editor"` for instance (https://github.com/superdesk/superdesk-client-core/commit/9af018a9b45b03cb1f9a68f5ff5f803ee84430f4#diff-245b0820e8afbb398822c4050a9cb7aaR36) is not stable. the value changes in the time. What can we do for that ?

Aca-jov: Yes, you understood it well.

Yes, values change over time and values are changed by values that are set in content_type.json. 

https://github.com/superdesk/superdesk-client-core/blob/e8ab4e3046eed2aa4a3444b005f93b095bf55ef2/scripts/superdesk-authoring/authoring.js#L2314, this type.chema is setting from content_type.json.

vied12: haha, we understand each other better and better, but I still don't know if you agree that this a problem that values of buttons can change.
My first thought is that it shouldn't change. If you define them in `content_type.json` then it should show them at the first rendering. Because now I have the feeling it uses the default values, and if the directive is reloaded again, it will use the values from `content_type.json` (for PA it means empty)
Maybe this occurs only in the case when you didn't specified the buttons list in `content_type.json`...
Is it the wanted behaviour when `content_type.json` doesn't provide the button list ?

Aca-jov:
This should be rebased.

vied12:
@Aca-jov done :bowtie:

scripts/superdesk/editor2/views/editor.html
```diff
@@ -6,12 +6,14 @@
 </a>
 <div ng-repeat="block in vm.blocks track by vm.generateBlockId(block)"
      class="block__container"
-     ng-class="{'block--selected': vm.blockToMoveIndex === $index}">
+     ng-class="{'block--selected': vm.blockToMoveIndex === $index,
+                'block--not-alone-and-empty': vm.blocks.length > 1 && (!block.body || block.body === '')}">
```
Aca-jov: ` !block.body` is enough.

Here should be `vm.blocks.length > 1 && !block.body` ;)

scripts/superdesk/editor2/views/editor.html
```diff
@@ -6,12 +6,14 @@
...
     <div class="block__actions" ng-if="vm.config.multiBlockEdition && !vm.reorderingMode">
         <div class="block__remove" ng-click="vm.removeBlock(block)">
             <i class="icon-close-small"></i>
         </div>
-        <div class="block__move" ng-click="vm.enableReorderingMode($index, $event)">
+        <div class="block__move" ng-click="vm.enableReorderingMode($index, $event)"
+             ng-if="vm.blocks.length > 1 && block.body && block.body !== ''">
```
Aca-jov: Same,  only `block.body` is needed.

Aca-jov:
Just small fix needed added in coments, the rest is fine.
:+1:

PR #172, opened 2016-03-15
chore(core): Update server core dependency

ioanpocol:


mugurrus:
:+1: 
But we need to impose a rule to check all PRs on superdesk-core against client-core before merging.

PR #169, opened 2016-03-14
copyrightholder added into metadata SDPA-268

vied12:
requires https://github.com/superdesk/superdesk-core/pull/229

PR #174, opened 2016-03-15
`alt_text` field added into "edit image" view SDPA-332

vied12:
If empty, `description_text` will be used

Depends on https://github.com/superdesk/superdesk-core/pull/234

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -886,12 +886,7 @@
         // should show the metadata form in the view
         $scope.data.showMetadataEditor = $scope.data.showMetadataEditor === true;
         // initialize metadata from `item`
-        if ($scope.data.showMetadataEditor) {
-            $scope.data.metadata = {};
-            ['headline', 'description_text', 'slugline', 'byline', 'usageterms', 'copyrightnotice'].forEach(function(key) {
-                $scope.data.metadata[key] = $scope.data.item[key];
-            });
-        }
+        $scope.data.metadata = angular.copy($scope.data.item);
 
         $scope.selectedRendition = null;
 
--- a/scripts/superdesk-authoring/views/change-image.html
+++ b/scripts/superdesk-authoring/views/change-image.html
@@ -16,6 +16,10 @@
                 <input type="text" ng-model="data.metadata.description_text" ng-change="data.isDirty = true" class="line-input">
             </div>
             <div>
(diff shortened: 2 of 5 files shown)


scripts/superdesk-authoring/views/change-image.html
```diff
@@ -16,6 +16,10 @@ <h3 class="pull-left">{{ activity.label|translate }}</h3>
                 <input type="text" ng-model="data.metadata.description_text" ng-change="data.isDirty = true" class="line-input">
             </div>
             <div>
+                <label translate>Alt Text</label>
```
ahilles107: It shouldn't be located before required fields

scripts/superdesk-authoring/authoring.js
```diff
@@ -886,12 +886,7 @@
         // should show the metadata form in the view
         $scope.data.showMetadataEditor = $scope.data.showMetadataEditor === true;
         // initialize metadata from `item`
-        if ($scope.data.showMetadataEditor) {
-            $scope.data.metadata = {};
-            ['headline', 'description_text', 'slugline', 'byline', 'usageterms', 'copyrightnotice'].forEach(function(key) {
-                $scope.data.metadata[key] = $scope.data.item[key];
-            });
-        }
+        $scope.data.metadata = angular.copy($scope.data.item);
```
ahilles107: :+1:

scripts/superdesk-authoring/views/change-image.html
ahilles107: + i don't see in this PR edit of that field in Image edit view.

vied12: (1) I will move it one place below.
(2) We only need this `alt_text` in the context of a story for now.
Anyway, this image edit view needs to be fully upgraded since the new cropping feature (https://dev.sourcefabric.org/browse/SD-4211)

ahilles107: But there is any way (in master) to edit it after setting?

vied12: Exactly at the same place where you set it.
On master it should appear only for featured images in a story. So you can edit it by clicking on the image.

eozan:
:+1:

vied12:
will be merged with https://github.com/superdesk/superdesk-client-core/pull/169 to prevent conflict

PR #176, opened 2016-03-15
fix(travis): stop tests after server failure

petrjasek:
it won't work anyhow and it keeps tests running for long time


petrjasek:
added one more change - only using single build job for tests, not 3. this should speed up testing across repos because we can only run 5 builds in the same time for all superdesk org repositories, so every client core repo generates 3 and then another 3 when merged.
anyhow would be good if everyone sets up their forks on travis too, so they don't need to create prs just to run the tests..

petrjasek:
ping @superdesk/developers any comments?

ioanpocol:
:+1:

PR #178, opened 2016-03-15
Html tags are stripped from image captions.

eozan:
SDPA-294

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -92,11 +92,6 @@
         return angular.extend(dest, _.pick(src, _.keys(CONTENT_FIELDS_DEFAULTS)));
     }
 
-    var htmlRegex = /(<([^>]+)>)/ig;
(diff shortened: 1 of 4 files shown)


ioanpocol:
please rebase

scripts/superdesk-authoring/authoring.js
```diff
@@ -92,13 +92,25 @@
         return angular.extend(dest, _.pick(src, _.keys(CONTENT_FIELDS_DEFAULTS)));
     }
 
+    var htmlRegex = /(<([^>]+)>)/ig;
```
vied12: can be move one row below to respect the scope

scripts/superdesk-authoring/authoring.js
```diff
@@ -92,13 +92,25 @@
...
+    function removeHtmlRaw(content) {
```
vied12: This can be moved into `RemoveTagsDirective`'s `link()` method

naomiaro:
:+1:

PR #179, opened 2016-03-16
(fix)[SD-4215] Fix the user disabled e2e test.

mdhaman:
Depends on [PR-237](https://github.com/superdesk/superdesk-core/pull/237)


ioanpocol:
:+1:

PR #184, opened 2016-03-16
feat(monitoring): Add support for scheduled but not yet published gro…

ioanpocol:
…up in monitoring (SDNTB-88)

### DIFF
--- a/spec/monitoring_spec.js
+++ b/spec/monitoring_spec.js
@@ -25,17 +25,18 @@ describe('monitoring', function() {
     });
 
     it('can configure desk output as default when user switches desks and show it on monitoring view', function() {
-        expect(monitoring.getGroups().count()).toBe(6);
+        expect(monitoring.getGroups().count()).toBe(7);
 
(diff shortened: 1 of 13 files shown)


petrjasek:
imo these needs some refactoring, having to add all those ifs seems wrong

spec/monitoring_spec.js
```diff
@@ -6,7 +6,7 @@ var authoring = require('./helpers/authoring'),
     dashboard = require('./helpers/dashboard'),
     desks = require('./helpers/desks');
 
-describe('monitoring', function() {
+fdescribe('monitoring', function() {
```
mdhaman: it should be `describe`.

petrjasek: right, I was wondering why it was so fast on travis :smile:

ioanpocol:
@petrjasek A small refactoring is done.

petrjasek:
:+1:

pavlovicnemanja:
:+1:

PR #185, opened 2016-03-16
fix(time): make timestamps in list show non relative time

petrjasek:
it does only time for current day, day + time for current week,
month + day otherwise. + is configurable.

SD-4098


ioanpocol:
:+1:

naomiaro:
I think we need some kind of documentation for vars like config.longDateFormat or to put them all in template.js or index.html (for each client) just somewhere so we can quickly scan the list, see defaults (for tests & configuration). Otherwise it's sort of some wasted time trying to find what's currently done/available.

akintolga:
:+1:

petrjasek:
added docs to readme, @naomiaro wasn't sure about this model date/time format, doesn't seem like something that should be changed right?

naomiaro:
:+1:  Thanks for the docs!

PR #186, opened 2016-03-16
Byline field is autopopulated with user's display name, if autopopula…

eozan:
…teByline is on.

SDPA-285

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -2239,7 +2239,19 @@
         'gettext',
         '$interpolate'
     ];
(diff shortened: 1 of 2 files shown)


mdhaman:
:+1:

akintolga:
:+1:

scripts/superdesk-authoring/authoring.js
```diff
@@ -2268,6 +2268,9 @@
                                     });
                             }
                         }
+                        if (config.features.autopopulateByline && !item.byline) {
+                            item.byline = gettext('by') + ' ' + session.identity.display_name;
```
petrjasek: it would be better for translators to have it like
```
byline = $interpolate(gettext('by {{ display_name }}')({display_name: session.identity.display_name});
```

ioanpocol:
:+1:

PR #187, opened 2016-03-16
fix(layout): [SD-3956] [SD-3619] Issues with template names and avatar positions

pavlovicnemanja:
- [x] SD-3956 - Limit template name since afterwards the save button becomes not visible
- [x] SD-3619 - Wrong positioning for the user avatar in the comments section

### DIFF
--- a/scripts/superdesk-templates/styles/templates.less
+++ b/scripts/superdesk-templates/styles/templates.less
@@ -1,5 +1,15 @@
 .template-card {
-	.card-box__content {
-		min-height: 120px; /* should be big enough to fit in dropdown menu */
-	}
+    .card-box__content {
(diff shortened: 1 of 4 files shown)


scripts/superdesk-templates/styles/templates.less
```diff
@@ -1,5 +1,15 @@
...
+
+.template-header {
+    h3 span {
+        max-width: calc(~'100vw - 290px');
+        float: left;
+        text-overflow: ellipsis;
+        overflow: hidden;
+        &:first-of-type { margin-right: 3px; }
+    }
+}
```
ahilles107: missing new line at end of file.

petrjasek:
:+1:

PR #191, opened 2016-03-17
[SD-3951] Change Keyword label to template name in more templates popup

marwoodandrew:
The label was perhaps a bit misleaing


petrjasek:
:+1:

ioanpocol:
:+1:

PR #192, opened 2016-03-17
fix(react): only handle keyboard on focus

petrjasek:
instead of listening to body events it listens on its own element
which has to get focus in order to work. + user can use pageup/pagedown
to toggle between groups in monitoring
cc @vied12


naomiaro:
:+1:

PR #193, opened 2016-03-17
Fixed topics in preview.

eozan:


ahilles107:
Why PA stuff goes to client core? Or what means `patopic` filter name?

eozan:
thats right, my mistake, closing.

PR #200, opened 2016-03-21
[sd-4129]-Pressing 'enter' in the slugline of advanced search jumps t…

sjunaid:
…he cursor

### DIFF
--- a/scripts/superdesk-search/search.js
+++ b/scripts/superdesk-search/search.js
@@ -1517,6 +1517,16 @@
 
                         var input = elem.find('#search-input');
 
+                        var ENTER = 13;
+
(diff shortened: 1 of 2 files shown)


ioanpocol:
:+1:

scripts/superdesk-search/search.js
```diff
@@ -1517,6 +1517,16 @@
 
                         var input = elem.find('#search-input');
 
+                        var ENTER = 13;
+
+                        var inputField = elem.find('input[type="text"]');
+
+                        inputField.on('keydown', function(event) {
```
petrjasek: this will need unregistering otherwise there will be memory leak

ioanpocol:
:+1:

PR #202, opened 2016-03-21
fix(templates): fix templates saving

petrjasek:
there was an error in case headline or body was empty.

SD-4172


akintolga:
:+1:

PR #203, opened 2016-03-21
fix(scheduled stage): Show scheduled items only on scheduled stage (SDNTB-129)

ioanpocol:


ioanpocol:
@akintolga Please review this change

scripts/superdesk-monitoring/monitoring.js
```diff
@@ -122,13 +122,13 @@
                             {term: {'task.last_authoring_desk': desk_id}},
                             {and: [
                                 {term: {'task.desk': desk_id}},
-                                {terms: {state: ['scheduled', 'published', 'corrected', 'killed']}}
```
akintolga: Hi @ioanpocol, I checked with Lasky and we need to keep `scheduled` state in desk output. There's no problem for us in having a separate section called `scheduledDeskOutput` for `scheduled` items though. So maybe we can remove `scheduled` items from desk output only if `scheduledDeskOutput` view is enabled. If you think it is hard and complicated any other solution from you is welcome as well.

ioanpocol: I think that in order to don't confuse the users I will make scheduled stage configurable. By default it is disabled(not visible at all in the app) and out stage show also scheduled items, but if it is enabled the scheduled items will not be showed only on scheduled stage.

akintolga: That sounds good.

Aca-jov:
:+1:

PR #204, opened 2016-03-21
SDPA-343 : No source in search

naomiaro:
merging search providers and ingest providers for source lookup.


naomiaro:
https://github.com/superdesk/superdesk/pull/1856

small change related to adding a name, or defaulting to source.

petrjasek:
:+1:

PR #205, opened 2016-03-21
SDPA-317 :

naomiaro:
When publishing a NinJS content item both the description_html fields populated with text

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -102,7 +102,7 @@
     }
 
     function stripHtml(item) {
-        var fields = ['headline', 'abstract'];
+        var fields = ['headline'];
         _.each(fields, function(key) {
             if (angular.isDefined(item[key])) {
                 item[key] = stripHtmlRaw(item[key]);


scripts/superdesk-authoring/authoring.js
```diff
@@ -102,7 +102,7 @@
     }
 
     function stripHtml(item) {
-        var fields = ['headline', 'abstract'];
```
akintolga: The ticket talks about the description_text and description_html fields. Why are we removing abstract field here?

naomiaro: the description comes from the abstract.

akintolga: Not in our case. Abstract field in front end is mapped to abstract field in the back end. If the html is not stripped, backend needs to change to strip it there.

petrjasek: so seems like this change is fine, but requires a backend fix toom, right? any specific format where it should be stripped @akintolga ?

akintolga: It looks like it needs to be stripped from all outputs except ninjs when published. I created a ticket for this and assigned to our sprint: [SD-4247](https://dev.sourcefabric.org/browse/SD-4247)

akintolga: @naomiaro  I think this change can go ahead, we'll handle the non-ninjs formatters.

akintolga:
:+1:

naomiaro:
OK great! Sorry, been sick for a couple days.

PR #212, opened 2016-03-22
fix(react): scroll to active item when using keyboard

petrjasek:
make sure selected item is visible when using keyboard to
navigate within items list.

SD-4148


ioanpocol:
:+1:

PR #213, opened 2016-03-22
fix(sign-off): Fix error when sign_off_mapping is not set

ioanpocol:


petrjasek:
good fix, what about some test?

ioanpocol:
I have to do more changes on sign-off and I will add tests on next PR.

PR #214, opened 2016-03-22
SDPA-345 : Drag rather than click the POI circle

naomiaro:
POI initially updates on mousedown, you can drag the circle to adjust, then it will update again on mouse up/leave.

### DIFF
--- a/scripts/superdesk/upload/image-crop-directive.js
+++ b/scripts/superdesk/upload/image-crop-directive.js
@@ -189,6 +189,7 @@
                         CropBottom: Math.round(center.y + selectionHeight / 2)
                     };
                     angular.extend(cropData, crop);
+
                     refreshImage(img.src, [
                         cropData.CropLeft,
                         cropData.CropTop,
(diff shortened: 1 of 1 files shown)


pavlovicnemanja:
:+1:

scripts/superdesk/upload/image-crop-directive.js
```diff
@@ -293,27 +295,62 @@
...
+                    function leaveOverlay(e) {
+                        e.preventDefault();
+                        updatePOI(e.offsetX, e.offsetY);
                         $rootScope.$broadcast('poiUpdate', scope.point);
+                        overlay.onmousemove = null;
+                        overlay.onmouseup = null;
+                        overlay.onmouseleave = null;
+                    }
+
+                    overlay.addEventListener('mousedown', function(e) {
```
petrjasek: this listener should be removed as well, otherwise there is a memory leak

naomiaro: ah yes, forgot about scope destroy

petrjasek:
:+1:

PR #216, opened 2016-03-22
SDNTB-97 1st versio of Noewegian translation

vladnicoara:


app/scripts/translations.js
```diff
@@ -14,7 +14,7 @@ angular.module('gettext').run(['gettextCatalog', function (gettextCatalog) {
...
     gettextCatalog.setStrings('nl', {});
```
akintolga: There are few translations with angular binding such as`"{{ errorMessage }}":"{{ Feilmelding }}"` . There are few with <b> tags as well. I wanted to double check if they are ok?

vladnicoara: I don't know of it to cause harm. This is a temp translation anyway as it will get improved and it will change soon. I'll investigate further by then

petrjasek: it won't work, ppl shouldn't translate things in `{{ }}`

vladnicoara: @petrjasek you are right of course. I corrected what I could find and I will pass on the info to the translators.

petrjasek:
:+1:

ioanpocol:
:+1:

PR #221, opened 2016-03-23
Aca jov authoring toolbar

ioanpocol:


scripts/superdesk-authoring/views/authoring-topbar.html
```diff
@@ -15,128 +15,152 @@
...
+      <div class="subnav__button-stack subnav__button-stack--custom-buttons" ng-show="_editable && itemActions.save && action === 'edit'">
+        <button class="btn btn-custom"
+                ng-click="sendToNextStage()"
+                title="{{ :: 'To Desk' | translate }}">
+                <span class="btn__text" translate>To Desk</span>
+                <span class="btn__text--short">T D</span>
+        </button>
+        <button class="btn btn-custom"
+                type="submit"
+                ng-click="publish()"
```
mdhaman: It quite possible that the user won't have `publish` privileges  but the user has create/edit `privileges`. For example, a reporter does not have `publish` privileges but the report can create/edit content. Hence, checking `save` and `action === 'edit'` is not sufficient.

Aca-jov: > It quite possible that the user won't have `publish` privileges  but the user will have create/edit `privileges`.

PR #223, opened 2016-03-24
Cropping feature must respect the given sizes [SDPA-350]

vied12:
In `ChangeImageController`, we now save the renditions sizes in pixel to add them in the result.


mugurrus:
:+1:

ioanpocol:
:+1:

PR #224, opened 2016-03-24
fix(templates): avoid updating item when using helpline

petrjasek:
instead of putting tmp value on item, put it in extra fields,
so there is no need to delete it later on when saving.

also when adding a helpline info, put it together in a `<p></p>`
so it's not mixed when using multiple messages.

SD-3980

to actually save the template it requires https://github.com/superdesk/superdesk-core/pull/250


ioanpocol:
:+1:

PR #228, opened 2016-03-29
[SD-3712] - Mark the stories to send an sms once published

akintolga:


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/251

scripts/superdesk-authoring/authoring.js
```diff
@@ -72,6 +73,7 @@
         company_codes: {},
         ednote: {},
         headline: {maxlength: 64},
+        sms: {maxlength: 64},
```
mdhaman: i think this should be 160 or whatever is max sms message length.

marwoodandrew:
:+1:

ioanpocol:
:+1:

PR #243, opened 2016-03-31
Upload modal - field labels UI

darconny:


naomiaro:
:+1:

scripts/superdesk/upload/styles/upload.less
```diff
@@ -485,13 +486,24 @@ media uploader
 
 					input[type="text"] {
 						border: 0 !important;
-						height: 34px !important;
+						height: 26px !important;
```
vied12: Using `!important` is often a bad idea. It's better to play with selectors precision.
In my side, I remove them every time I see one :)

darconny: Yes I know, but it was set before like this to overwrite common input style. And because of the urgency of this PR (upload modal is broken), I don't have time to make it pretty :)

eozan:
:+1:

scripts/superdesk/upload/styles/upload.less
vied12: It's just the first one, you also added some rules after with `!imprortant` ;)

PR #245, opened 2016-03-31
SDPA-358 : Display the image size on the upload view

naomiaro:


scripts/superdesk-archive/views/upload.html
```diff
@@ -23,7 +23,7 @@ <h3 class="pull-left">{{ activity.label|translate }}</h3>
         </div>
         <ul class="upload-thumbs">
             <li ng-repeat="item in items">
-                <div class="thumb" sd-image-preview="item.url" data-file="item.file">
+                <div class="thumb" sd-image-preview="item.url" data-file="item.file" data-width="item.width" data-height="item.height">
```
vied12: why do you need this ?
Can't you directly use the values from the scope since you compute them is the same directive ?

naomiaro: the directive creates a new scope and also the values are needed later on in a different part of the template.

naomiaro: right... I think I partially forgot how angular js works when being sick

naomiaro: The sd-image-preview directive creates its own scope. This is also a template piece further down and outside of there this directive was called from.

vied12: I see, so why don't you load this `img` in order to get the sizes in the directive where you need it ?

naomiaro: (As we mentioned in Slack) Img srcs here are created from local file Blobs turned into a dataUrl. Might as well set the height/width of the image when we've created this dataURL here.

eozan:
:+1:

PR #248, opened 2016-03-31
Character count fixed for body in editor

Aca-jov:


ioanpocol:
:+1:

pavlovicnemanja:
:+1:

scripts/superdesk-authoring/authoring.js
```diff
@@ -1646,7 +1646,9 @@
                 limit: '=',
                 html: '@'
             },
-            template: '<span class="char-count" ng-class="{error: numChars > limit}" translate>{{numChars}}/{{ limit }} characters </span>',
+            template: '<span class="char-count" ng-class="{error: limit && numChars > limit}" translate> characters</span>' +
+                    '<span class="char-count" ng-class="{error: limit && numChars > limit}">{{numChars}}' +
+                    '<span ng-if="limit" ng-class="{error: limit && numChars > limit}">/{{ limit }}</span></span>',
```
vied12: at this point I would create a template file.
This is what I do when I have more than 1 line.

Aca-jov: Yes, your right, but I think in this case, change is too small to create new template file. Also this is not expected to be changed (and if it grows larger it would go in separate file, but i doubt it).

PR #254, opened 2016-04-01
[SD-4270] - Auto publish should not be visible on frontend

akintolga:
1. Macro api accepts a new parameter `includeBackend`
  1. If `false` then api returns only frontend macros
  2. If `true` then api returns both frontend and backend macros
2. When macros are listed for routing and desk/stage settings only the relevant ones will be displayed:
  1. Stage settings won't list `interactive` macros
  2. Routing settingswon't list `interactive` or `background-thread` macros

### DIFF
--- a/scripts/superdesk-authoring/macros/macros.js
+++ b/scripts/superdesk-authoring/macros/macros.js
@@ -13,8 +13,8 @@ function MacrosService(api, notify) {
             }));
     };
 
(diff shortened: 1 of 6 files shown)


mdhaman:
:+1:

marwoodandrew:
:+1:

ioanpocol:
:+1:

PR #258, opened 2016-04-01
Fixed ranking.

eozan:


naomiaro:
I think we should fix the way the CSS works instead... going to end up with too much garbage

naomiaro:
https://github.com/superdesk/superdesk-client-core/pull/260/files

PR #262, opened 2016-04-04
Ranking field added.

eozan:


eozan:
requires https://github.com/superdesk/superdesk/pull/1893

PR #267, opened 2016-04-05
fix(editor): avoid updating an item after autosave

petrjasek:
it might override changes done meanwhile so avoid
updating the item with response.

also fixing undo/redo not working for newly created items.

SD-4264 SD-4265


akintolga:
:+1:

sjunaid:
:+1:

PR #268, opened 2016-04-05
Fix for PA dynamic metadata

naomiaro:
https://github.com/superdesk/superdesk-client-core/pull/260/files

after this pull request, need a way to update schema for dynamic metadata based on services.


naomiaro:
this was becase of being confused by content_types.json

PR #270, opened 2016-04-05
SDPA-346

darconny:
- UI improvements for image edit/crop modal
- css refactoring

### DIFF
--- a/scripts/superdesk/upload/styles/upload.less
+++ b/scripts/superdesk/upload/styles/upload.less
@@ -580,87 +580,93 @@ media uploader
 edit image modal
 ======================================
 */
-.change-image {
-    &__metadata {
-        position: absolute;
-        top: 0;
-        right: 0;
-        bottom: 225px;
-        width: 335px;
-        overflow-y: auto;
-        background-color: #f0f0f0;
-        border-left: 1px solid #dfdfdf;
-        .box-sizing(border-box);
-        padding: 15px;
-        > div {
-            margin-bottom: 20px;
-        }
-        label {
-            font-size: 11px;
-            text-transform: uppercase;
-            color: @grayLight;
-        }
-    }
+.modal-body__flex {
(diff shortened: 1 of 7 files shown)


pavlovicnemanja:
:+1:

scripts/superdesk/upload/styles/upload.less
```diff
@@ -580,87 +580,93 @@ media uploader
...
-            font-size: 11px;
-            text-transform: uppercase;
-            color: @grayLight;
-        }
-    }
+.modal-body__flex {
+	display: flex;
+	flex-direction: column;
+	padding: 0;
+	overflow-y: hidden !important;
```
vied12: please avoid `!important`

styles/less/layouts.less
```diff
@@ -6,29 +6,27 @@
...
 body {
-    margin: 0;
-    font-family: @baseFontFamily;
-    font-size: @baseFontSize;
-    line-height: @baseLineHeight;
-    color: @sd-text;
-    background-color: @sd-background;
-    overflow: hidden;
-    height:100%;
+	margin: 0;
```
vied12: we indent with spaces :)
please configure your editor

scripts/superdesk/upload/styles/upload.less
darconny: this is here to overwrite overflow:auto; that comes from the standard body, and we don't need it here for flex implementation

vied12: then as I told you in another PR, it's better to play with selector precision than `important!`

darconny: In this case I do not agree.

vied12: https://css-tricks.com/when-using-important-is-the-right-choice/ ;)

PR #273, opened 2016-04-05
Render crops for featured images

vied12:
depends on https://github.com/superdesk/superdesk/pull/1903

* Perform renditions when featured image has been crop
* Add a `authoring/services.js` file for new services. (original file is now too long)

### DIFF
--- a/scripts/superdesk-authoring/services.js
+++ b/scripts/superdesk-authoring/services.js
@@ -1,0 +1,59 @@
+(function() {
+'use strict';
+
+angular.module('superdesk.authoring').service('cropPicture', CropPictureService);
+CropPictureService.$inject = ['$q', 'renditions', 'api', 'superdesk'];
+
+function CropPictureService($q, renditionsService, api, superdesk) {
+    this.crop = function(picture) {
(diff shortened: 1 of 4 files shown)


scripts/superdesk-authoring/services.js
```diff
@@ -0,0 +1,59 @@
...
+            .then(function(result) {
+                var renditionNames = [];
+                var savingImagePromises = [];
+                angular.forEach(result.cropData, function(croppingData, renditionName) {
+                    // if croppingData are defined
+                    if (angular.isDefined(croppingData.CropLeft) && !isNaN(croppingData.CropLeft)) {
+                        renditionNames.push(renditionName);
+                    }
+                });
+                console.log('renditionNames', renditionNames, 'result', result);
```
akintolga: console.log is left here...

akintolga:
:+1: Looks good. Maybe one simple test for the new service would be useful.

mdhaman:
:+1:

vied12:
@akintolga I won't write a test now because we have to change this workflow in the next task.
Let's do it when it will be done.

PR #302, opened 2016-04-12
fix(highlights): fix marking for highlights for uploaded images

petrjasek:
api is expecting to get guid, not _id, refactored this into highlight service.

SD-4125


akintolga:
:+1:

mdhaman:
:+1:

PR #305, opened 2016-04-12
fix spellcheck for editor2

petrjasek:
test a different rendering strategy, instead of updating the text
with markup for hiliting create a clone, do the hiliting there and
render it below the actual text. advantage is that there is no
cursor manipulation involved, it doesn't change markup so it doesn't
need to store and restore cursor position.

known issues:
- find and replace - before it was using hilite node to replace the
  word, it will require some changes to manipulate the text instead
- corrections dropdown - because the errors are rendered below the text,
  there would be no click event on this.

### DIFF
--- a/scripts/superdesk/editor2/editor.js
+++ b/scripts/superdesk/editor2/editor.js
@@ -345,9 +345,9 @@ function EditorService(spellcheck, $rootScope, $timeout, $q, _, renditionsServic
 
         // remove old hilite nodes if any
         var clones = parentNode.getElementsByClassName(CLONE_CLASS);
-        Array.prototype.forEach.call(clones, function(clone) {
-            parentNode.removeChild(clone);
-        });
+        if (clones.length) {
+            parentNode.removeChild(clones.item(0));
+        }
 
         // create a clone
         var hiliteNode = node.cloneNode(true);
@@ -1154,7 +1154,7 @@ angular.module('superdesk.editor2', [
                         var err, pos;
                         var point = {x: event.clientX, y: event.clientY};
                         var errors = elem[0].parentNode.getElementsByClassName('sderror');
-                        for (var i = 0; i < errors.length; i++) {
+                        for (var i = 0, l = errors.length; i < l; i++) {
                             err = errors.item(i);
                             pos = err.getBoundingClientRect();
                             if (isPointInRect(point, pos)) {
--- a/scripts/superdesk/editor2/editor.js
+++ b/scripts/superdesk/editor2/editor.js
@@ -12,15 +12,6 @@
 'use strict';
 
 /**
- * Generate click event on given target node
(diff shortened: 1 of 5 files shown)


scripts/superdesk/editor2/editor.js
```diff
@@ -325,19 +325,26 @@ function EditorService(spellcheck, $rootScope, $timeout, $q, _, renditionsServic
...
-        hiliteToken(node, token, className);
-        $timeout(function() {
-            hilite(node, tokens, className, true);
-        }, 0, false);
+        // remove old hilite nodes if any
+        var clones = parentNode.getElementsByClassName(CLONE_CLASS);
+        Array.prototype.forEach.call(clones, function(clone) {
+            parentNode.removeChild(clone);
+        });
+
```
naomiaro: getElementsByClassName returns a live collection, would be better to use slice to make an actual array from this DOM  collection first.

scripts/superdesk/editor2/editor.js
```diff
@@ -1113,7 +1119,26 @@ angular.module('superdesk.editor2', [
...
+                     * Test if given point {x, y} is in given bouding rectangle.
+                     */
+                    function isPointInRect(point, rect) {
+                        return rect.left < point.x && rect.right > point.x && rect.top < point.y && rect.bottom > point.y;
+                    }
+
                     editorElem.on('contextmenu', function(event) {
+                        var err, pos;
+                        var point = {x: event.clientX, y: event.clientY};
+                        var errors = elem[0].parentNode.getElementsByClassName('sderror');
```
naomiaro: This is also a live array, the length can't really be trusted/much slower to use than using slice to create a real array.

scripts/superdesk/editor2/editor.js
```diff
@@ -1113,7 +1119,26 @@ angular.module('superdesk.editor2', [
...
+                            pos = err.getBoundingClientRect();
+                            if (isPointInRect(point, pos)) {
+                                event.preventDefault();
+                                return;
+                            }
+                        }
+
                         if (editor.isErrorNode(event.target)) {
                             event.preventDefault();
                             var menu = elem[0].getElementsByClassName('dropdown-menu')[0],
```
naomiaro: hmm this is an old one but still...

petrjasek:
@naomiaro using slice was slower than using htmlcollection, was this the concern or something else?

ioanpocol:
:+1:

naomiaro:
@petrjasek concern was mostly it's a live collection, so the length is not fixed. It will update when you remove items so it's not the safest to use them in for loops an add/delete items from the collection.

petrjasek:
I see thx for info

naomiaro:
:+1:

PR #308, opened 2016-04-12
Ranking field is set to be displayed only on selected services.

eozan:
SDPA-338

### DIFF
--- a/scripts/superdesk-authoring/views/authoring-header.html
+++ b/scripts/superdesk-authoring/views/authoring-header.html
@@ -142,7 +142,7 @@
         <div class="authoring-header__item"
              sd-validation-error="error.urgency"
              data-required="schema.urgency.required"
-             ng-if="schema.urgency"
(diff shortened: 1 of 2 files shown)


scripts/superdesk-authoring/views/authoring-header.html
```diff
@@ -141,7 +141,7 @@
         <div class="authoring-header__item"
              sd-validation-error="error.urgency"
              data-required="schema.urgency.required"
-             ng-if="schema.urgency"
+             ng-if="schema.urgency && (!editor.urgency.service || editor.urgency.service[item.anpa_category[0].qcode])"
```
vied12: what is `anpa` ?
are we sure that `item.anpa_category[0].qcode` always exists ?
This condition isn't really clear to me. A function here could be easier to read

eozan: right, I will fix.

eozan:
Should be merged with https://github.com/superdesk/superdesk/pull/1939

PR #312, opened 2016-04-13
[SD 4030] Selected item count does not update properly after a selected item is killed

sjunaid:


scripts/superdesk-monitoring/monitoring.js
```diff
@@ -535,7 +542,12 @@
                             if (scope.total !== items._meta.total) {
                                 scope.total = items._meta.total;
                             }
-
+                            // For highlight page return only highlight items, i.e, include only last version if item type is published
+                            if (scope.group.type === 'highlights') {
```
akintolga: This section looks identical to one on line:529 so it'd be better to factor both to its own function.

akintolga:
I suggest to squash the commits...

scripts/superdesk-search/react.js
```diff
@@ -166,6 +172,33 @@
                     }
 
                     /**
+                     * Test if an multi-selection allowed
+                     *
+                     * @return {Boolean}
+                     */
+                    function isCheckAllowed(item) {
+                        if (item.state === 'killed' || (item._type === 'published' && !item.last_published_version)) {
```
mdhaman: duplicate code maybe just have `isCheckedAllowed`

ioanpocol:
:+1:

scripts/superdesk-archive/module.js
```diff
@@ -61,9 +63,21 @@
...
+         */
+        this.updateCount = function(itemId) {
+            _.remove(items, {_id: itemId});
+            this.count = items.length;
+        };
+
         // main
         this.reset();
         $rootScope.$on('$routeChangeStart', angular.bind(this, this.reset));
+        $rootScope.$on('multi:updateCount', function(_e, itemId) {
```
petrjasek: imo it should be called `multi:remove` , it's not obvious if on `updateCount` message it should remove given item or add it..

scripts/superdesk-search/react.js
```diff
@@ -390,6 +416,9 @@
                             return function(event) {
                                 event.stopPropagation();
                                 highlightsService.markItem(highlight._id, this.props.item);
+                                if (getViewType() === 'highlights') {
```
petrjasek: what is the idea here?

scripts/superdesk-search/react.js
```diff
@@ -466,7 +495,17 @@
 
                         render: function() {
                             var item = this.props.item.archive_item || this.props.item;
-                            var highlights = item.highlights || [];
+
+                            var highlights = [];
+                            if (isCheckAllowed(this.props.item)) {
+                                if (this.props.item.archive_item && this.props.item.archive_item.highlights) {
```
petrjasek: seems like check for `this.props.item.archive_item.highlights.length` should be also part of this, otherwise it's a bit confusing with `? :` right after

scripts/superdesk-search/react.js
```diff
@@ -862,6 +901,11 @@
                             scope.$apply(function() {
                                 activityService.start(this.props.activity, {data: {item: this.props.item}});
                             }.bind(this));
+
+                            if (_.contains(['spike', 'unspike'], this.props.activity.when)) {
+                                $rootScope.$broadcast('multi:updateCount', this.props.item._id);
```
petrjasek: imo this would be better in some `item:spike` and `item:unspike` event handler, so it works also for other users via websockets and not only current user

scripts/superdesk-search/react.js
```diff
@@ -1317,6 +1361,9 @@
                                 var itemId = search.generateTrackByIdentifier(item);
                                 var oldItem = itemsById[itemId] || null;
                                 if (!oldItem || !isSameVersion(oldItem, item) || !isArchiveItemSameVersion(oldItem, item)) {
+                                    if (oldItem && oldItem.selected != null) {
+                                        item.selected = oldItem.selected;
+                                    }
                                     itemsById[itemId] = item;
```
petrjasek: might be better to do `itemsById[itemId] = angular.extend({}, oldItem, item);` so we don't have to modify this in case there is another flag added

scripts/superdesk-search/react.js
petrjasek: this could be just `return !(item.state === 'killed' || ...);`

scripts/superdesk-search/react.js
sjunaid: In highlights only list view, removing the highlight of an item(single highlighted) removes that item from the list (unlike monitoring view item where just star removes). so only for viewType highlights we need to update the count if it's might be in multi-selection. however this should happen only if item is single highlighted. this check is also included in updated PR.

akintolga:
:+1:

mdhaman:
:+1:

petrjasek:
:+1:

PR #313, opened 2016-04-13
[SD-4263] Fixing the issue where double click on ingested item and publishing failed.

mdhaman:
- `action` was getting set to `true` because of the lock parameter. `lock` parameter is not needed.


akintolga:
:+1:

mugurrus:
:+1:

ioanpocol:
:+1:

PR #317, opened 2016-04-13
[SD-4293] Cannot upload image to avatar

jerome-poisson:
Fixed avatar upload, the issue was due to the use of ng-file-select
instead of ngf-select, probably an API change in ng-file-upload

SD-4293

### DIFF
--- a/scripts/superdesk-users/views/change-avatar.html
+++ b/scripts/superdesk-users/views/change-avatar.html
@@ -23,10 +23,10 @@
         </div>
 
         <div class="computer" ng-if="active.id == 'upload' && !preview.url" ng-hide="progress.width">
-            <div class="dropzone" ng-file-drop="preview.img = $files[0]">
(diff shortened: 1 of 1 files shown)


sjunaid:
:+1: (can you check the same for `ng-file-drop` as well)

scripts/superdesk-users/views/change-avatar.html
```diff
@@ -26,7 +26,7 @@ <h3 class="pull-left">{{ activity.label|translate }}</h3>
             <div class="dropzone" ng-file-drop="preview.img = $files[0]">
```
petrjasek: seems like here it will need the same change, right @sjunaid ?

jerome-poisson: yes indeed the drop was also bugged, fixed it

ioanpocol:
:+1:

petrjasek:
:+1:

PR #319, opened 2016-04-13
feat(cvs): Add support for custom controled vocabularies for non subj…

ioanpocol:
…ect fields (genre, place, ...)

### DIFF
--- a/scripts/superdesk-authoring/metadata/metadata.js
+++ b/scripts/superdesk-authoring/metadata/metadata.js
@@ -488,7 +488,7 @@ function MetaTermsDirective(metadata, $filter, $timeout) {
         scope: {
             item: '=',
             field: '@',
-            dependent: '@',
+            resetDependent: '&',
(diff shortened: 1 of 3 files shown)


scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -616,15 +624,15 @@ function MetaTermsDirective(metadata, $filter, $timeout) {
                     var o = {};
 
                     // dependent is set only for category
-                    if (scope.dependent) {
+                    if (scope.resetDependent) {
```
akintolga: I don't think this will ever resolve to `true`. Even if the function is not passed it will have a wrapper. You may need to use `iAttrs` attribute to check if it is there.

Aca-jov:
:+1:

pavlovicnemanja:
:+1:

petrjasek:
:+1:

PR #321, opened 2016-04-14
[sd-4255]-(fix) - Layout issue once multiple flags are set for an item

sjunaid:


mdhaman:
:+1:

akintolga:
:+1:

scripts/superdesk-search/react.js
```diff
@@ -653,12 +644,12 @@
                                     React.createElement('span', {className: 'broadcast-status', tooltip: broadcast.status}, '!') :
                                     null,
                                 flags.marked_for_not_publication ?
-                                    React.createElement('div', {className: 'state-label not-for-publication'},
-                                        gettext('Not For Publication')) :
+                                    React.createElement('div', {className: 'state-label not-for-publication',
+                                        title: gettext('not for publication')}, gettext('Not For Publications')) :
```
petrjasek: would be better to have same text for both text and title, so there is only 1 string to translate, not 2

scripts/superdesk-search/react.js
```diff
@@ -653,12 +644,12 @@
...
                                     null,
                                 flags.marked_for_legal ?
-                                    React.createElement('div', {className: 'state-label legal'}, gettext('Legal')) :
-                                    null,
+                                    React.createElement('div', {className: 'state-label legal', title: gettext('legal')},
```
petrjasek: same thing here, `'legal'` should be `'Legal'`

mdhaman:
squash commits.

petrjasek:
:+1:

PR #322, opened 2016-04-14
feat[shortcuts]: SD-3819 Keyboard shortcuts should be shown in tool tips

pavlovicnemanja:
SD-3819

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -389,10 +389,6 @@
 
             if (this.isPublished(original)) {
                 delete updates.dateline;
-
-                if (updates.embargo) {
-                    delete updates.embargo;
-                }
             }
 
             //check if rendition is dirty for real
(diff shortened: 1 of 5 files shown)


scripts/superdesk/keyboard/keyboard.js
```diff
@@ -330,6 +330,37 @@
                 elt['on' + type] = false;
             }
         };
+
+        this.registerShortcut = function (hotkey, callback, options, elem) {
+            this.register(hotkey[0], gettext(hotkey[2]), hotkey[1]);
+            /* hotkey[0] => Category, hotkey[1] => Description, hotkey[2] => Shortcut, hotkey[3] => Options */
+
+            this.bind(hotkey[2], function (e) {
```
petrjasek: does it ever unbind those shortcuts? because if you trigger it multiple times (like going from workspace to settings and back) it might do it again and it would leak memory

scripts/superdesk/keyboard/keyboard.js
```diff
@@ -330,6 +330,37 @@
                 elt['on' + type] = false;
             }
         };
+
+        this.registerShortcut = function (hotkey, callback, options, elem) {
+            this.register(hotkey[0], gettext(hotkey[2]), hotkey[1]);
```
petrjasek: does it populate this help screen too?

pavlovicnemanja: Yep, this is populating the help screen. My idea was to call shortcut registration just on one place, and best on the button, because it will trigger click, and we don't need to define double functions for them. If there is need for some additional functions, I created callback for it.

petrjasek: thing is it will only populate that help if button is visible on screen. so if you go away from workspace it won't be there, not sure if it's an issue or not, but I think the idea is that this help won't change

pavlovicnemanja: Hmmm, if button is visible, then action is available. Not sure if it is good case that we have all shortcuts visible on help screen (available and not available one) because I think it will be confusing. Someone will try to start some action that is visible, but not available, and he will think that it is a bug. So maybe is good case to have only currently available one.

If there is need to display all shortcuts, then I will find some solution for this. But I think we will need to specified them double. And maybe is better to define them only on one place, maybe on this service for displaying them.

scripts/superdesk/keyboard/keyboard.js
pavlovicnemanja: Maybe I can add check if these shortcut is already defined, and if it is, just ignore function. Is this case have a problem with leaking memory?

scripts/superdesk/keyboard/keyboard.js
petrjasek: yep I think that help screen should show you all of them, otherwise it would be done as you did it at first place. @eozan am I right?

pavlovicnemanja: ok so, do you have some idea how to initialize all shortcuts on app load? Or we need to define them separately? Is it better to define all of them in one place, like in one service, or as it was now, on every page were you call shortcut you also call a registration of that shortcut with `.run`?

petrjasek: I would use `.config` instead of `.run`. and do it per module, not in one place.

PR #325, opened 2016-04-14
fix(layout): SD-4144, SD-4213 Fix for couple of layout issues

pavlovicnemanja:
- [x] SD-4144 Layout issue in highlights configuration window
- [x] SD-4213 UI overlapping issue with a  'Content Filter' related control
- [x] SD-2636 Add Transifex link near Superdesk language selector


ioanpocol:
:+1:

PR #336, opened 2016-04-18
fix(archive): fix resend action after canceling

petrjasek:
due to `ngIf` it created child scope so on cancel setting it to null
made this part of child scope so it wouldn't pick changes from parent

SD-4345

added also one more related fix for resend subscribers dropdown:
once subscribers got loaded it didn't trigger a watch in meta terms
directive so it wouldn't populate the list. so now when assigning
it first creates a new instance of the list. plus it will hide that
subscribers dropdown before it gets popuplated.

SD-4346


akintolga:
:+1:

mdhaman:
:+1:

PR #337, opened 2016-04-18
feat(workspace): add config for ingest/content view

petrjasek:
those links can be switched off via config (and are off by default).

SD-3816


ioanpocol:
:+1:

petrjasek:
@vied12 @ioanpocol not sure how this should work for ntb/pa, so you might need those changes too if you want to keep old ingest/content views in workspace

petrjasek:
and by "those changes" I mean https://github.com/superdesk/superdesk/pull/1965, added this to wrong PR

PR #340, opened 2016-04-18
SDPA-385

darconny:
- input tag replaced with contenteditable div
- css updated


vied12:
Did you check if the form is still saving the data ?
Just asking, because `ng-model` doesn't work natively with `contenteditable` if I remember

darconny:
@vied12 Well for me its working fine, but that's why I asked for review to check if ng-model working properly.

petrjasek:
@vied12 it works because of this [ment.io](https://github.com/jeff-collins/ment.io/blob/master/ment.io/scripts.js#L140) we use
:+1:

PR #344, opened 2016-04-19
fix(preview): use single component for full screen preview

petrjasek:
it was using different component for items in package which was
not rendering full screen due to dom location

SD-4123


ioanpocol:
:+1:

akintolga:
:+1:

PR #347, opened 2016-04-19
Dynamic image size in cropping view

vied12:
Refactor the edit crop view, in order to support dynamic size for images later


scripts/superdesk/upload/image-crop-directive.js
```diff
@@ -279,136 +278,107 @@
...
+                        var overlay = element.find('.image-point__poi__overlay');
+                        var mousedown = false;
+                        overlay.on('mousedown', function(e) {
+                            // enable drag mode
+                            mousedown = true;
+                            updatePOI(e);
+                        });
+                        overlay.on('mousemove', function(e) {
+                            if (mousedown) {
+                                updatePOI(e);
```
naomiaro: Is it needed to update on mousemove? It seems a bit more efficient to update on mouseleave/mouseup. I'm not sure if the final mousemove will totally match the mouseup event.

scripts/superdesk/upload/image-crop-directive.js
```diff
@@ -279,136 +278,107 @@
...
+                    (function($img) {
+                        function updatePOI(e) {
+                            vm.updatePOI({
+                                x: Math.round(e.offsetX * 100 / $img.width) / 100,
+                                y: Math.round(e.offsetY * 100 / $img.height) / 100
+                            });
+                        }
+                        var overlay = element.find('.image-point__poi__overlay');
+                        var mousedown = false;
+                        overlay.on('mousedown', function(e) {
```
naomiaro: These on events need to be removed via scope destroy

naomiaro: since the overlay is not created dynamically now, you could add the on event via $scope to avoid having to remove these in scope destroy

naomiaro: http://stackoverflow.com/questions/26983696/angularjs-does-destroy-remove-event-listeners Was googling to double check :D

scripts/superdesk/upload/image-crop-directive.js
```diff
@@ -279,136 +278,107 @@
...
+                img.onload = function() {
+                    var $img = element.find('.image-point__image').get(0);
+                    drawPoint($img);
+                    // draw when needed
+                    scope.$on('poiUpdate', function() {
+                        drawPoint($img);
+                    });
+                    angular.element($window).on('resize', function() {
+                        drawPoint($img);
+                    });
```
naomiaro: memory removal.

scripts/superdesk/upload/image-crop-directive.js
```diff
@@ -279,136 +278,132 @@
...
                 }
+                // init directive element style
+                element.css({
+                    position: 'relative',
+                    display: 'block'
+                });
+                // load the image in order to know the size
+                var img = new Image();
+                img.onload = function() {
+                    function drawPointsFromModel() {
```
naomiaro: img onload function will be referenceing the img itself.
http://javascript.info/tutorial/memory-leaks#memory-leaks
should clean up the leak like here.

naomiaro: there's a circular reference here since the DOM function will reference img. I'm not sure if V8 can garbage collect this.

scripts/superdesk/upload/image-crop-directive.js
```diff
@@ -279,136 +278,132 @@
...
-
-                function drawPoint() {
-                    pointElem.css({
-                        left: (scope.point.x * img.width) - 15,
-                        top: (scope.point.y * img.height) - 15
+                    var topOffset = (poi.y * img.height) - circleRadius;
+                    var leftOffset = (poi.x * img.width) - circleRadius;
+                    var verticalLeftOffset = leftOffset + circleRadius - (lineThickness / 2);
+                    var horizontalTopffset = topOffset + circleRadius - (lineThickness / 2);
+                    element.find('.image-point__poi').css({
```
naomiaro: It's better to cache these objects in variables so the DOM doesn't have to be queried repeatedly.

scripts/superdesk/upload/image-crop-directive.js
vied12: fix here https://github.com/superdesk/superdesk-client-core/pull/347/commits/1532e9e23c5fec89ba501526dc706230c6024413

scripts/superdesk/upload/image-crop-directive.js
```diff
@@ -279,136 +278,140 @@
...
-                    container.append(crossTop);
-                    container.append(crossBottom);
-
-                }
-
-                function drawPoint() {
-                    pointElem.css({
-                        left: (scope.point.x * img.width) - 15,
-                        top: (scope.point.y * img.height) - 15
+                    var topOffset = (poi.y * img.height) - circleRadius;
```
naomiaro: These vars can also be declared at the top of the link function to avoid redeclaring a var each time which is much more expensive.

PR #348, opened 2016-04-19
fix(datetime): make sure mergeDateTime returns local time

petrjasek:
and not utc.

also moved `datetimeHelper` service to `superdesk.datetime` module from `superdesk.ui`

### DIFF
--- a/scripts/superdesk/datetime/datetime.js
+++ b/scripts/superdesk/datetime/datetime.js
@@ -73,6 +73,42 @@
         }
     }
 
+    DateTimeHelperService.$inject = ['moment', 'config'];
+    function DateTimeHelperService(moment, config) {
+
+        /*
+        * @param timestring 2016-03-01T04:45:00+0000
+        * @param timezone Europe/London
+        */
+        this.splitDateTime = function(timestring, timezone) {
+            var momentTS = moment.tz(timestring, timezone);
+
+            return {
+                'date': momentTS.format(config.model.dateformat),
+                'time': momentTS.format(config.model.timeformat)
+            };
+        };
+
+        this.isValidTime = function(value, format) {
+            var timeFormat = format || config.model.timeformat;
+            return moment(value, timeFormat, true).isValid();
+        };
+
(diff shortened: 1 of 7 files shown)


ioanpocol:
:+1:

scripts/superdesk/datetime/datetime.js
```diff
@@ -73,6 +73,42 @@
...
+            return moment(value, dateFormat, true).isValid();
+        };
+
+        this.mergeDateTime = function(date_str, time_str, timezone) {
+            var tz = timezone || config.defaultTimezone;
+            var merge_str = date_str + ' ' + time_str;
+            var formatter = config.model.dateformat + ' ' + config.model.timeformat;
+
+            // return without timezone information, which is stored separately
+            return moment.tz(merge_str, formatter, tz).format('YYYY-MM-DD[T]HH:mm:ss');
```
petrjasek: fyi calling `.format`  here is the only change, rest is just moving the code around

akintolga:
:+1:

mdhaman:
:+1:

scripts/superdesk/datetime/datetime.js
```diff
@@ -73,6 +73,42 @@
...
+
+        /*
+        * @param timestring 2016-03-01T04:45:00+0000
+        * @param timezone Europe/London
+        */
+        this.splitDateTime = function(timestring, timezone) {
+            var momentTS = moment.tz(timestring, timezone);
+
+            return {
+                'date': momentTS.format(config.model.dateformat),
```
mdhaman: @petrjasek from [Line 190](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/superdesk-authoring/metadata/metadata.js#L190) the `timestring` passed is `2016-04-20T09:45:00+0000` i.e. it contains utcoffset. This causes moment to calculate the time part wrong.

petrjasek: that's what is changed in PR, the time won't be utc anymore, but it will be local according to timezone

PR #352, opened 2016-04-21
[sd-4173]-(Imp)- Icons needed for highlight and take packages

sjunaid:


petrjasek:
can we rather store 'highlights' or 'takes' into package 'profile' field? and make use of that value for icon. because this will need yet another set of ifs when we introduce another package type..

akintolga:
I agree with you @petrjasek. If it is possible we want to merge like this for the time being but create a new ticket to:
- Refactor the backend to use either `profile` or `package_type` across the board to identify the type of the (composite) item
- Refactor client to handle server side changes

petrjasek:
:+1: sounds like a plan

mdhaman:
:+1:

PR #354, opened 2016-04-21
SDPA-418 : Can no longer embed images from API

naomiaro:


naomiaro:
also requires https://github.com/superdesk/superdesk-core/pull/312 to make sure stages are populated before desks.

petrjasek:
:+1:

PR #356, opened 2016-04-21
fix(editor): avoid undo/redo operations after load

petrjasek:
it might try to replace node content with same value (with different
whitespace) but then there is no change so it won't trigger highlights rendering.

SD-4065


akintolga:
:+1:

mdhaman:
:+1:

ioanpocol:
:+1:

PR #358, opened 2016-04-21
fix(autosave): Fix autosave issue

ioanpocol:


petrjasek:
:+1:

naomiaro:
👍  this was prob from those reversed booleans earlier.

PR #359, opened 2016-04-21
(fix) Fixing issues related to schedule and embargo dates.

mdhaman:
- For `publish_schedule` and `embargo` validation is done using utc dates.
- `Embargo` can be modified while correcting the story.
- Remove `embargo` setting if embargo as lapsed or killed.

### DIFF
--- a/scripts/superdesk/datetime/datetime.js
+++ b/scripts/superdesk/datetime/datetime.js
@@ -110,7 +110,6 @@
 
         /*
         * @param timestring 2016-03-01T04:45:00+0000.
-        * @param timezone Europe/London
         */
         this.greaterThanUTC = function(timestring) {
             return moment(timestring, 'YYYY-MM-DDTHH:mm:ssZZ') > moment.utc();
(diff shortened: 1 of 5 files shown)


akintolga:
:+1:

ioanpocol:
:+1:

scripts/superdesk/datetime/datetime.js
```diff
@@ -107,6 +107,14 @@
             // return without timezone information, which is stored separately
             return moment.tz(merge_str, formatter, tz).format('YYYY-MM-DD[T]HH:mm:ss');
         };
+
+        /*
+        * @param timestring 2016-03-01T04:45:00+0000.
+        * @param timezone Europe/London
```
petrjasek: seems like there is no such param

petrjasek:
:+1:

scripts/superdesk/datetime/datetime.js
mdhaman: yep. fixed.

PR #364, opened 2016-04-22
feat[shortcuts]: SD-3819 Keyboard shortcuts should be shown in tool tips

pavlovicnemanja:
SD-3819

Previous PR: #322 Closed because of some stupid git conflict :)

### DIFF
--- a/scripts/superdesk/keyboard/keyboard.js
+++ b/scripts/superdesk/keyboard/keyboard.js
@@ -332,6 +332,35 @@
         };
     }])
 
+    .directive('sdHotkey', ['keyboardManager', '$timeout', function (keyboardManager, $timeout) {
+            return {
+                link: function (scope, elem, attrs, ctrl) {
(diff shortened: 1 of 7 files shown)


scripts/superdesk/keyboard/keyboard.js
```diff
@@ -332,6 +332,35 @@
...
+                        options = scope.$eval(attrs.sdHotkeyOptions);
+
+                    keyboardManager.bind(hotkey, function (e) {
+                        e.preventDefault();
+                        if (callback) {
+                            callback();
+                        } else {
+                            elem.click();
+                        }
+                    }, options);
```
petrjasek: when it's global (`options.global == true`), there should be `unbind` on `scope.$destroy`, otherwise there is memory leak when navigating away from workspace and back

petrjasek:
:+1:

Aca-jov:
:+1:

PR #366, opened 2016-04-22
SDPA-406 : AD user might have different phone format. Validation not …

naomiaro:
…too important.

### DIFF
--- a/scripts/superdesk-users/views/edit-form.html
+++ b/scripts/superdesk-users/views/edit-form.html
@@ -133,11 +133,7 @@
 
                 <div sd-info-item>
                   <label for="phone" translate>phone number</label>
-                  <input type="text" name="phone" id="phone" ng-model="user.phone" ng-pattern="phonePattern">
-                  <div ng-show="userForm.phone.$error.pattern || error.phone" sd-valid-info>
-                    <i class="icon-info-sign-warning"></i>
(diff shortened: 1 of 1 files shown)


naomiaro:
Also depends on https://github.com/superdesk/superdesk-core/pull/317

petrjasek:
:+1:

scripts/superdesk-users/views/edit-form.html
```diff
@@ -133,11 +133,7 @@
 
                 <div sd-info-item>
                   <label for="phone" translate>phone number</label>
-                  <input type="text" name="phone" id="phone" ng-model="user.phone" ng-pattern="phonePattern">
-                  <div ng-show="userForm.phone.$error.pattern || error.phone" sd-valid-info>
-                    <i class="icon-info-sign-warning"></i>
-                    <span translate>please provide number in international format</span>
-                  </div>
+                  <input type="text" name="phone" id="phone" ng-model="user.phone">
```
gbbr: Is it not a better approach to change the pattern validator to a more appropriate form, so that it matches the needed phone format instead of removing it completely?

naomiaro: we can't predict the format though, depends on the LDAP/AD server.

PR #368, opened 2016-04-22
Crop Editor: validate data before to allow exit SDPA-420

vied12:
@eozan can you please check latest commit ?
Before it was keeping old values for some renditions because of https://github.com/superdesk/superdesk-client-core/blob/bbdeb26588934205dc3f008978b8e1d650ed69b7/scripts/superdesk/upload/image-crop-directive.js#L107
Weird... I would like to be sure I'm not making a mistake there. Thanks

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -919,8 +919,8 @@
         };
     }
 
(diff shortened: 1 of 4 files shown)


eozan:
@vied12, that line you linked to was because embed renditions were coming with different set of data, and I was using that to apply some of the logic for only some items. (I'm not too sure about the specifics, I can check if needed). If everything works ok with all kinds of renditions, it seems good to me.

PR #370, opened 2016-04-25
Header custom buttons now save and continue action

Aca-jov:


scripts/superdesk-authoring/authoring.js
```diff
@@ -303,6 +305,23 @@
             }
         };
 
+        this.rewrite = function (item, authoringWorkspace) {
```
petrjasek: any reason for having `authoringWorkspace` a param here?

Aca-jov: I need it in authoring service (rewrite), and since authoringWorkspace uses authoring service, I cannot add it in authoring, so I did it this way..

petrjasek: I see, you can avoid it using `$injector` service, so in `rewrite` you can do
```js
var authoringWorkspace = $injector.get('authoringWorkspace');`
```

Aca-jov: Fixed.

pavlovicnemanja:
:+1:

petrjasek:
:+1:

PR #377, opened 2016-04-26
fix(editor2): fix spellcheck in new editor

petrjasek:
- fix manual mode
- fix spellcheck after doing undo/redo

SD-4305
SD-4065


Aca-jov:
:+1:

PR #380, opened 2016-04-27
fix(metadata): show all categories for users without prefs

petrjasek:
before users set their preference regarding categories, they have
empty dropdown in authoring which prevents them from publishing.

so in case the preferences are empty list - show all categories.


akintolga:
:+1:

mdhaman:
:+1:

PR #382, opened 2016-04-27
Fix editor2 for aap

petrjasek:
- fix find&replace
- add option to disable toolbar
- add option to disable embeds

also fixed a memory leak in add-content

~~initially I've put some common parts for editor1 and editor2 into editor/utils.js, but not sure if it makes sense if we dump editor1~~


app/index.html
```diff
@@ -123,6 +123,7 @@
     <script src="scripts/superdesk/directives/sdWithParams.js"></script>
 
     <script src="scripts/superdesk/editor/editor.js"></script>
+    <script src="scripts/superdesk/editor/utils.js"></script>
```
akintolga: Is `utils.js` missing from the PR?

petrjasek: I didn't use it at the end, will remove this

petrjasek:
ping @superdesk/developers

mdhaman:
:+1:

akintolga:
:+1:

PR #384, opened 2016-04-28
[SD-4391] - Content filters fixes 

akintolga:
- Filters will display the name of the value rather than the qcode or _id in the preview
- Unique field was creating problem for _id field in desk selections


petrjasek:
:+1:

marwoodandrew:
:+1:

PR #386, opened 2016-04-28
SD-4394 -- Remove afterClose function for macros

vied12:
It was causing an unwanted redo in the editor

It reverts partially #363, (@eozan if you can check please)

cc @ioanpocol

### DIFF
--- a/scripts/superdesk-authoring/macros/macros.js
+++ b/scripts/superdesk-authoring/macros/macros.js
@@ -155,12 +155,7 @@ angular.module('superdesk.authoring.macros', [
                 order: 6,
                 needEditable: true,
                 side: 'right',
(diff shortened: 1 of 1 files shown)


eozan:
this breaks other fixes for both pa and aap :-1:

vied12:
Maybe @mdhaman has an idea about how we can solve this issue

mdhaman:
@vied12 , @eozan there were 2 problems:
1) If you open versions widget and go to previous version it shows previous version in readonly mode in the editor. If you close the versions widget you still see previous version of the item in readonly mode and only option left was to close editor.
2) Close the macro widget tab after macro is applied. 

What is the issue over here?

mdhaman:
@vied12 do we need this still?

vied12:
it seems that the initial problem has been solved. Thanks

PR #388, opened 2016-04-28
Editor2 embed: support SAMDesk SDPA-412

vied12:


scripts/superdesk/editor2/editor.js
```diff
@@ -1334,26 +1334,47 @@ angular.module('superdesk.editor2', [
         };
     }])
     .run(['embedService', 'iframelyService', function(embedService, iframelyService) {
-        var pattern = 'https?:\/\/(?:www)\.playbuzz\.com(.*)$';
-        var loader = '//snappa.embed.pressassociation.io/playbuzz.js';
-        var html = [
+        var playBuzzPattern = 'https?:\/\/(?:www)\.playbuzz\.com(.*)$';
+        var playBuzzlLoader = '//snappa.embed.pressassociation.io/playbuzz.js';
```
actionless: is that ok for such thing to be hardcoded on that master branch?

scripts/superdesk/editor2/editor.js
```diff
@@ -1334,26 +1334,47 @@ angular.module('superdesk.editor2', [
...
+                });
             }
         });
+        var samdeskEmbed = [
+            '<script>(function(d, s, id) { var fjs = d.getElementsByTagName(s)[0];',
+            'var js = d.createElement(s); js.id = id; js.src = \'https://embed.samdesk.io/js/2/embed.js\';',
+            'fjs.parentNode.insertBefore(js, fjs); }(document, \'script\', \'sam-embed-js\'));</script>',
+            '<div class="sam-embed" data-href="embed.samdesk.io/embed/$_ID"></div>'
+        ].join('');
+        var sameDeskPattern = 'https?://embed.samdesk.io/embed/(.+)';
```
naomiaro: being picky here, I'd drop the 'e' from this var name.

scripts/superdesk/editor2/editor.js
vied12: I believe it does. This file is just a wrapper to support playbuzz embeds.  This can be useful for every instances

actionless: my point what this variable have URL of PA's instance, not some public service url

vied12: So should we copy this file and host it ourself ?

actionless: why not include it to ./client/app/scripts ?

vied12: The main advantage to let PA to host it, is that they will update the script if needed (there are the author of this script and they use it often)
But if you think it's important to fork it in SD, I could do it. Just confirm it :)

actionless: the fact what that script could be updated independently from code actually sounds more like a con to me

vied12: Ok I will do it in another pr  (this is out of this scope) .  But in
general we can't maintain in all the embed scripts that we support. It
would be an endless job. We must rely on external sources as much as
possible.

PR #390, opened 2016-04-29
Img drag

naomiaro:
Chrome default list drag is broken in version 50. Nicer to just drag thumbnail around into editor.


petrjasek:
:+1:

PR #393, opened 2016-05-02
[SD-4295] - Spell check highlighting capitalised words

sjunaid:


scripts/superdesk/editor/spellcheck/spellcheck.js
```diff
@@ -90,17 +185,25 @@ function SpellcheckService($q, api, dictionaries) {
                 currentOffset = 0,
                 tree = document.createTreeWalker(node, NodeFilter.SHOW_TEXT);
 
+            // Abbreviations found in dictionary.
+            var re = /\w*(?:\.\w*)+/g;
```
petrjasek: after first `\w` there should be `+` imo, otherwise `...` would match too

scripts/superdesk/editor/spellcheck/spellcheck.spec.js
```diff
@@ -73,6 +74,22 @@ describe('spellcheck', function() {
...
 
+    it('can report error if paragraph starts with small letter', inject(function(spellcheck, api, $rootScope) {
+        var p = createParagraph('test');
+
+        spyOn(api, 'save');
+        spellcheck.errors(p).then(assignErrors);
+        $rootScope.$digest();
+        expect(errors.length).toBe(1);
+
+        spellcheck.addWordToUserDictionary('test');
```
petrjasek: not sure this test works, because it's mocking dictionaries, so would be better to test with some word that is already there

scripts/superdesk/editor/spellcheck/spellcheck.js
```diff
@@ -77,6 +77,101 @@ function SpellcheckService($q, api, dictionaries) {
...
+                } else {
+                    return false;
+                }
+            }
+            return dict.content[word];
+        }
+    }
+
+    function isFirstLetterCapital(word) {
+        var re = /^[A-Z][a-z0-9_-]*$/;
```
petrjasek: this should be probably only `[A-Z]`, given that it only checks first character.
it could be also `return word[0] === word[0].toUpperCase()` which should be faster

scripts/superdesk/editor/spellcheck/spellcheck.js
```diff
@@ -90,17 +185,25 @@ function SpellcheckService($q, api, dictionaries) {
                 currentOffset = 0,
                 tree = document.createTreeWalker(node, NodeFilter.SHOW_TEXT);
 
+            // Abbreviations found in dictionary.
+            var re = /\w*(?:\.\w*)+/g;
+            var abbrevList = _.words(Object.keys(d), re);
```
petrjasek: this should only change on dict updates, so there is no need to run this for every check

petrjasek: also not sure it does what you want, seems like it would return csv, not list

scripts/superdesk/editor/spellcheck/spellcheck.js
```diff
@@ -77,6 +83,100 @@ function SpellcheckService($q, api, dictionaries) {
...
+            nonSentenceWords.push({
+                word: _.trim(match[1]),
+                index: currentOffset + wordIndex
+            });
+        }
+
+        return nonSentenceWords;
+    }
+
+    function testExpression(textContent, currentOffset) {
```
petrjasek: pls add some comment to those functions, not exactly clear what it does by its name

scripts/superdesk/editor/spellcheck/spellcheck.js
```diff
@@ -91,16 +191,20 @@ function SpellcheckService($q, api, dictionaries) {
                 tree = document.createTreeWalker(node, NodeFilter.SHOW_TEXT);
```
naomiaro: Not sure if we care, but filter doesn't seem to be option in IE
https://developer.mozilla.org/en-US/docs/Web/API/Document/createTreeWalker

petrjasek: we don't care yet :wink:

petrjasek:
:+1:

naomiaro:
does jira need a capital SD in the commit message to properly link this PR to the ticket?

mdhaman:
@sjunaid  jira link is not working. 
squash the commits.

akintolga:
:+1:

PR #394, opened 2016-05-02
Fixing the issues where closepreview called by authoring widgets causing autosave to loose changes.

mdhaman:
Problem was `closepreview` called by the `versioning` widget caused  autosave to loose changes. Made change so that you can preview versions only after used has saved changes.

@vied12 , @petrjasek  can you review the changes.

### DIFF
--- a/scripts/superdesk-authoring/versioning/versions/views/versions.html
+++ b/scripts/superdesk-authoring/versioning/versions/views/versions.html
@@ -1,16 +1,6 @@
 <div class="article-versions" ng-controller="VersioningWidgetCtrl">
     <ul class="versions-list card-list">
-        <li class="item draft" ng-class="{selected: selected === item._autosave}" ng-if="item._autosave" ng-click="openAutosave()">
-            <div class="header">
-                <time title="{{ item._autosave._updated | date:'medium' }}">{{ item._autosave._updated | reldate }}</time>
-            </div>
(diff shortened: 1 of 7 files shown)


scripts/superdesk-authoring/versioning/versions/views/versions.html
```diff
@@ -1,16 +1,6 @@
 <div class="article-versions" ng-controller="VersioningWidgetCtrl">
     <ul class="versions-list card-list">
-        <li class="item draft" ng-class="{selected: selected === item._autosave}" ng-if="item._autosave" ng-click="openAutosave()">
```
petrjasek: so there won't be any indication that you have made some changes? or am I missing something

mdhaman: Issue here is if the autosave is in progress and somebody has version widget open and click on the previous version then you loose the latest changes from autosave version.  Hence on the closing the version widget you don't get the autosave changes.

Made changes such that version history won't show the autosave changes. You have indication that save button is enabled and you cannot preview previous version. You need to save current version to start previewing the all the previous versions.

petrjasek: I see

petrjasek:
:+1:

akintolga:
:+1:

PR #395, opened 2016-05-02
fix(highlights): append highlights popup to body

petrjasek:
it was hidden when part of an item elem because of overflow setting
we use for scrolling. so it has to be appended to some other element.
so reusing code for actions menu to render this same way.

SD-4449


akintolga:
:+1:

mdhaman:
:+1:

PR #399, opened 2016-05-02
Changed "Source" to "Originator" string

tictactatic:
To differentiate between original source and the default Source attached by Superdesk (as configured in Desks Settings).


ioanpocol:
:+1:

PR #400, opened 2016-05-02
Fix picture order on drag on drop SDPA-432

vied12:
in js, number++ doesn't return the incremented value


petrjasek:
:+1: better than using `++number` :wink:

actionless:
LGTM

PR #403, opened 2016-05-03
Editor2 embed field: Trigger onClose only if it was open before

vied12:


actionless:
:+1:

petrjasek:
:+1:

scripts/superdesk/editor2/add-embed.ctrl.js
```diff
@@ -157,15 +157,15 @@ function SdAddEmbedController (embedService, $element, $timeout, $q, _, EMBED_PR
     // toggle when the `extended` directive attribute changes
     $scope.$watch(function() {
         return vm.extended;
-    }, function(extended) {
+    }, function(extended, wasExtended) {
```
gbbr: What is `extended` and `wasExtended`? Some documentation would be nice. It might that I'm just out of context but it's hard for me to figure out what this stuff is.

gbbr: Whoops... This was already merged. Ignore my comment :)

vied12: `extented` is a visual state of the "add an embed" field bar

petrjasek: @gbbr fyi it's just current and previous value returned from that watched function, so value of `vm.extended`. you can find more info about `$watch` in [angular docs](https://docs.angularjs.org/api/ng/type/$rootScope.Scope)

gbbr: right, makes sense. thanks!

PR #404, opened 2016-05-03
Editor2: when a picture is inserted, show the following text during loading

vied12:


scripts/superdesk/editor2/editor.js
```diff
@@ -834,19 +834,25 @@ angular.module('superdesk.editor2', [
                         }
                     });
                 }
+                var debouncedInit = _.debounce(init, 250);
```
petrjasek: this doesn't work well with protractor, it should be using `$timeout`

naomiaro: The image cropping view is now using debounce, does it cause protractor to randomly fail @petrjasek ?

petrjasek: it could, protractor can wait for all $timeouts and $http requests to finish, but it has no idea about lodash methods

vied12:
@petrjasek: I replaced `_.debounce` by `$timeout`, it seems to work fine.
Thanks for the advice

petrjasek:
:+1:

mugurrus:
:+1:

PR #405, opened 2016-05-03
fix(schedule): use timezone when checking publish schedule

petrjasek:
it was ignoring timezone so when using timezones with negative
offset and time within that offset it was failing.

SD-4377


naomiaro:
Cool, thought there might be something like this hanging around after Nareg's experiences.

akintolga:
:+1:

PR #407, opened 2016-05-03
SDPA-429 : Spell-checker does not work well with capitalized words

naomiaro:


naomiaro:
https://github.com/superdesk/superdesk-client-core/pull/393/files this is already in progress working with grammar checks.

PR #415, opened 2016-05-04
ui: fix styling issue with sdTimezone component navigation.

gbbr:
Addresses [SD-4145](https://dev.sourcefabric.org/browse/SD-4145).

- Navigating with arrow components is now visible in the dropdown:
screen shot 2016-05-04 at 12 24 50 pm

- Timezone drop-down is on top of Publish/Send button

### DIFF
--- a/scripts/superdesk/directives/sdTypeahead.js
+++ b/scripts/superdesk/directives/sdTypeahead.js
@@ -6,13 +6,11 @@
          * Typeahead direcitve
          *
          * Usage:
-         *  <div sd-typeahead items="subjects" term="subjectTerm" search="searchSubjects(term)" select="selectSubject(item)">
-         *      <ul>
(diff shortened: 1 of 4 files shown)


petrjasek:
:+1:

akintolga:
:+1:  Looks good. 

Afaik `sd-typeahead` is used within `MetaTermsDirective` which is one gigantic directive used in few places. That would be great if you could check if the arrow keys are still functioning in Subject (multu-level) and Category fields in editor.

gbbr:
@akintolga not sure where `Category` is (couldn't find it), but `Subject` seems to be working fine (there is extra padding now but I think that makes it better visually), considering this is what you were seeing before:
screen shot 2016-05-05 at 10 31 03 am

akintolga:
@gbbr That looks good. Thanks a lot for checking this. Cheers.

petrjasek:
:+1:

PR #416, opened 2016-05-04
fix(templates): use profile scheme/editor config for templates

petrjasek:
moved this to a service where both authoring and templates can use it.
also added sms message handling into common article-edit directive.

SD-4433


akintolga:
:+1:

mdhaman:
:+1:

PR #402, opened 2016-05-03
[SD-4397]-(fix) - Action menu should refresh after a highlight is deleted from highlights section

sjunaid:

PR #419, opened 2016-05-05
Fix e2e

petrjasek:
fix authoring spec error


mdhaman:
:+1:

petrjasek:
damn, fixed 1, but probably introduced another 4 :smile:

petrjasek:
fix is in already in https://github.com/superdesk/superdesk-client-core/pull/402

PR #422, opened 2016-05-05
fix(editor2): fix editor config defaults for toolbar

petrjasek:
and avoid some extra dict reseting

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ For installation instructions, please refer to the README file at: [https://gith
 
 ##### Editor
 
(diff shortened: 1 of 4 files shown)


naomiaro:
👍  thanks for this!

actionless:
but it's still not reading it from `process.env`, right?

or we won't need to change that option on deployment?

naomiaro:
yeah the template.js should be still updated to be `true` by default really

petrjasek:
it shouldn't be changed on deployment

actionless:
LGTM then

PR #427, opened 2016-05-06
PR #382 introduced a bug with "add an embed" bar

vied12:
The (+) button was not behaving right and was always showing


petrjasek:
:+1: thx for fixing it :wink:

PR #428, opened 2016-05-06
[SD-4485][SD-4486] [SD-4463] Global Search Fixes.

mdhaman:
Fixes Following issues:
- Aggregations on  global search never updates.
Current aggregations don't change count at all i.e. aggregations count is same even if the user   switches between repos. For example, if user select only ingest repo then the user sees the aggregates of desk which imo is incorrect. If you search on aap multimedia repo then the user see aggregates of multimedia + superdesk.

- Remove `Keywords` from aggregates and allow to be search as text filter.
- Date range filter was not displayed when aggregates tab is open. It only displays when parameters  toggle is used.

This PR is dependent on [#349](https://github.com/superdesk/superdesk-core/pull/349)

### DIFF
--- a/scripts/superdesk-archive/views/metadata-view.html
+++ b/scripts/superdesk-archive/views/metadata-view.html
@@ -194,7 +194,7 @@
         <dd>{{ item.guid }}</dd>
     </dl>
     <dl ng-if="item.unique_name">
-        <dt translate>Unique ID</dt>
+        <dt translate>Unique Name</dt>
         <dd>{{ item.unique_name }}</dd>
     </dl>
     <dl>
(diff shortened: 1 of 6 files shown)


akintolga:
:+1:

petrjasek:
looks good, but why removing `keywords` from filter?

mdhaman:
Keywords could have lot of unique values and having aggregates on it does not make sense. We have ticket to remove aggregates and have it as text search https://dev.sourcefabric.org/browse/SD-4463.

petrjasek:
probably not, just wasn't sure given the previous title :wink:

mdhaman:
title is confusing as i merged 3 tickets into one.

sjunaid:
:+1:

PR #429, opened 2016-05-06
fix(editor2): port spellcheck updates to editor2

petrjasek:
it checks first words in sentences to start with uppercase.

@sjunaid what do you think? had to make some changes because `match.index == 0` is only beginning of text if there is no markup used, with markup it doesn't work


sjunaid:
@petrjasek , True that, thanks for the corrections. however, the another change at line 150 breaks the
excluding non sentence words code block at line 156. So, I taken your fixes for spellcheck.js and adjusted the required changes in PR https://github.com/superdesk/superdesk-client-core/pull/430.

pavlovicnemanja:
:+1:

Aca-jov:
:+1:

PR #434, opened 2016-05-09
make abstract use italics via config

petrjasek:
+ added `defaultConfig` provider to be able to set default config values in `.config` callback


akintolga:
:+1:

mdhaman:
:+1:

PR #435, opened 2016-05-09
[SD-4122]-(fix) - Users should not be able to highlight items of a no…

sjunaid:
…n highlight desk

### DIFF
--- a/scripts/superdesk-highlights/highlights.js
+++ b/scripts/superdesk-highlights/highlights.js
@@ -56,6 +56,7 @@
 
                 promise[key] = api('highlights').query(criteria)
                     .then(function(result) {
+                        setLabel(result._items);
                         cache.put(key, result);
                         promise[key] = null;
                         return $q.when(result);
@@ -64,6 +65,12 @@
                 return promise[key];
             }
         };
+
+        function setLabel(objItems) {
+            _.forEach(objItems, function (item) {
+                item.label = item.desks.length ? item.name : item.name + ' ' + gettext('(Global)');
+            });
+        }
 
         /**
          * Clear user cache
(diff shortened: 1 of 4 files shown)


scripts/superdesk-highlights/highlights.js
```diff
@@ -35,6 +35,7 @@
          * Fetches and caches highlights, or returns from the cache.
          */
         service.get = function(desk) {
+            service.clearCache();
```
petrjasek: don't think it should be here, would never use the cache..

scripts/superdesk-highlights/highlights.js
```diff
@@ -560,7 +563,7 @@
                             'button',
                             {disabled: isMarked, onClick: this.markHighlight},
                             React.createElement('i', {className: 'icon-star'}),
-                            highlight.name
+                            highlight.desks.length ?  highlight.name : highlight.name + gettext(' (Global)')
```
akintolga: I think this logic can be moved to highlight service instead of repeating in few places...

scripts/superdesk-search/react.js
```diff
@@ -836,9 +836,12 @@
                     var ActionsMenu = React.createClass({
                         toggle: function(event) {
                             this.stopEvent(event);
+                            highlightsService.clearCache();
```
petrjasek: why clearing cache here?

sjunaid: @petrjasek , no need as discussed, so removed. thanks.

mugurrus:
:+1:

scripts/superdesk-highlights/highlights.js
```diff
@@ -65,6 +66,12 @@
             }
         };
 
+        function setLabel(objItems) {
+            _.forEach(objItems, function (item) {
+                item.label = item.desks.length ? item.name : item.name + gettext(' (Global)');
```
petrjasek: for the sake of translation, might be better to have it like: `item.name + ' ' + gettext('(Global)')` so that if someone omits the space while translating this will still work

ioanpocol:
:+1:

PR #438, opened 2016-05-10
[SD-4197] - Target subscribers for publish action in routing

akintolga:


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/350

sjunaid:
:+1:

scripts/superdesk-ingest/module.js
```diff
@@ -1077,6 +1078,27 @@
                     });
                 });
 
+                adminPublishSettingsService.fetchSubscribers().then(function(items) {
```
mdhaman: look like this might fetch only top N subscribers not all.

mdhaman:
:+1:

PR #441, opened 2016-05-10
ui: restrict dashboard configuration to administrator only (SD-4362)

gbbr:
Fixes [SD-4362](https://dev.sourcefabric.org/browse/SD-4362).

### DIFF
--- a/scripts/superdesk-dashboard/module.js
+++ b/scripts/superdesk-dashboard/module.js
@@ -1,8 +1,8 @@
 (function() {
     'use strict';
 
-    DashboardController.$inject = ['$scope', 'desks', 'dashboardWidgets', 'api', 'session', 'workspaces', 'modal', 'gettext'];
-    function DashboardController($scope, desks, dashboardWidgets, api, session, workspaces, modal, gettext) {
+    DashboardController.$inject = ['$scope', 'desks', 'dashboardWidgets', 'api', 'session', 'workspaces', 'modal', 'gettext', 'privileges'];
+    function DashboardController($scope, desks, dashboardWidgets, api, session, workspaces, modal, gettext, privileges) {
         var vm = this;
 
         $scope.edited = null;
@@ -13,6 +13,8 @@
         function setupWorkspace(workspace) {
             vm.current = null;
             if (workspace) {
+                // is the user allowed to configure widgets on this desk?
+                $scope.configurable = workspaces.isCustom() || privileges.userHasPrivileges({desks: 1});
                 // do this async so that it can clean up previous grid
                 $scope.$applyAsync(function() {
                     vm.current = workspace;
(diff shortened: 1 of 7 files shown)


akintolga:
Is there a ticket for this PR?

gbbr:
I've updated the issue description with a link to the ticket.

akintolga:
I think non-admin users should be able to update their own widgets in their personal workspace, does your change handle this?

mdhaman:
instead of checking for `session.identity.user_type === 'administrator'`, it would be better to check for desk privilege for `Desk Dashboard`.

gbbr:
The above (regarding desk privilege) makes sense. I will look into it and
submit more changes for you guys to review.

gbbr:
@mdhaman which privileges should be allowed (or disallowed) to edit/modify widgets on the desk's dashboard?

gbbr:
PTAL - now the workspace is configurable if it is custom or if the user has the "_Desk Management_" privilege enabled

scripts/superdesk-dashboard/module.js
```diff
@@ -1,13 +1,21 @@
...
         var vm = this;
 
         $scope.edited = null;
         $scope.workspaces = workspaces;
         $scope.$watch('workspaces.active', setupWorkspace);
+
+        // canConfigure returns true if the current user is allowed to configure
+        // the active workspace. The active workspace may be configured if it is
+        // a custom workspace or if the user has desk management privileges.
+        $scope.canConfigure = function() {
```
akintolga: Maybe this function can be moved to workspaces service and can be used here and from the widget directive.

akintolga:
:+1: Looks good.

scripts/superdesk-dashboard/module.js
gbbr: I have thought about how to make this function less repetitive and have not came up with a better option. I think it is simple enough to have it as is, although I would also prefer it to be in one place, instead of two.

However, I don't think having it in the `workspaces` service is appropriate. I don't think it would be appropriate for the service to interact with session based data, but if you think otherwise, please explain.

gbbr:
Hey @akintolga - PTAL - I've submitted a new commit with a different approach where the configuration is inherited, rather than duplicated.

gbbr:
If you prefer [this version](https://github.com/superdesk/superdesk-client-core/pull/441/commits/14e678f76b43d0a8afc7b24532157df3abee2131), let me know and I'll revert my change.

scripts/superdesk-dashboard/module.js
petrjasek: imo this could be set in `setupWorkspace`, it won't change, so instead of function it could be just a value

petrjasek:
:+1:

gbbr:
I've addressed all comments and squashed into 1 commit.

PR #447, opened 2016-05-11
feat(search): Add the option to search by cvs (SDNTB-59)

ioanpocol:
By default only the subject search is present so all search related tests will run on it.


vied12:
:+1: (I would just add more comments when it becomes complex like some parts of this PR)

Aca-jov:
:+1:

PR #452, opened 2016-05-11
fix(authoring): fix publish schedule time check

petrjasek:
it wasn't using timezone if you first saved and then published,
because after save it returned timestamp with `+0000` and this
prevented it from using preset timezone.

also made some changes to make those error messages translatable

SD-4377


akintolga:
:+1:

mdhaman:
:+1:

PR #459, opened 2016-05-12
e2e tests for embeds

vied12:


vied12:
Will attempt again when we drop editor1

PR #462, opened 2016-05-12
feat(profiles): initial list implementation

petrjasek:
kind of hard to test, there is no way to create profiles yet..
but one can use initialize data..

SD-3872


Aca-jov:
:+1:

pavlovicnemanja:
:+1:

PR #464, opened 2016-05-13
fix(metadata): fix item source display

petrjasek:
it wasn't consistent, sometimes you would get ingest provider name
instead of source, sometimes just source. now there is mostly source
used, only in metadata view there can be also ingest provider, with
specific label.

SD-4163


Aca-jov:
:+1:

mdhaman:
:+1:

PR #465, opened 2016-05-13
fix(images): don't display image overlay if there is no action

petrjasek:
it seems like there should be a preview with this overlay,
so now it will only display it in case there is some action defined.

SD-4396

### DIFF
--- a/scripts/superdesk-archive/directives.js
+++ b/scripts/superdesk-archive/directives.js
@@ -417,52 +417,40 @@
                 templateUrl: 'scripts/superdesk-archive/views/item-rendition.html',
                 scope: {
                     item: '=',
-                    rendition: '@',
-                    ratio: '=?'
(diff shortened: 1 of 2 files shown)


scripts/superdesk-archive/directives.js
```diff
@@ -417,52 +417,40 @@
...
                         if (href) {
                             var img = new Image();
+
                             img.onload = function() {
                                 if (oldImg.length) {
                                     oldImg.replaceWith(img);
                                 } else {
                                     figure.html(img);
-                                    figure.append(previewHover);
+                                    if (attrs.ngClick) {
```
petrjasek: this is the actual fix, the rest is just refactoring

akintolga:
:+1:

pavlovicnemanja:
:+1:

PR #467, opened 2016-05-13
SDNTB-164 Label confusion

vladnicoara:


ioanpocol:
:+1:

petrjasek:
this might be confusing for other clients.. maybe would be better to handle this with translation?

mdhaman:
needs to be handle via translation.

PR #468, opened 2016-05-13
[SD-4492]-(Imp) - Render timestamps in search and monitoring views us…

sjunaid:
…ing 24hr clock

### DIFF
--- a/scripts/superdesk/datetime/datetime_spec.js
+++ b/scripts/superdesk/datetime/datetime_spec.js
@@ -25,11 +25,11 @@
                 var now = moment();
                 var day = now.isoWeekday();
 
-                expect(now.format('hh:mm')).toBe(datetime.shortFormat(now.format()));
+                expect(now.format('HH:mm')).toBe(datetime.shortFormat(now.format()));
 
                 // on monday do tuesday, otherwise day before
                 var week = day === 1 ? now.add(1, 'days') : now.subtract(1, 'days');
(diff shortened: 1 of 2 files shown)


scripts/superdesk/ui/ui.js
```diff
@@ -1190,6 +1190,7 @@
 
         .config(['defaultConfigProvider', function(defaultConfig) {
             defaultConfig.set('ui.italicAbstract', true);
+            defaultConfig.set('shortTimeFormat', 'HH:mm'); // 24h format
```
petrjasek: this setting should probably go into datetime module

scripts/superdesk/datetime/datetime_spec.js
```diff
@@ -21,15 +22,18 @@
...
 
         describe('datetime service', function() {
-            it('can short format datetime', inject(function(datetime, moment) {
+            it('can short format datetime', inject(function(datetime, moment, config) {
                 var now = moment();
                 var day = now.isoWeekday();
 
-                expect(now.format('hh:mm')).toBe(datetime.shortFormat(now.format()));
+                var TIME_FORMAT = config.shortTimeFormat || 'hh:mm';
+                var WEEK_FORMAT = config.shortWeekFormat || 'dddd, ' + TIME_FORMAT;
```
petrjasek: imo it shouldn't use the config, it's testing the config, so you should set the value explicitly

pavlovicnemanja:
:+1:

ioanpocol:
:+1:

PR #478, opened 2016-05-17
Editor2: Apply cycle fixed for embed metadata edition SDPA-465

vied12:


scripts/superdesk/editor2/editor.js
```diff
@@ -485,7 +485,9 @@ function SdTextEditorBlockEmbedController($timeout, editor, renditions) {
             // update the caption in the view
             vm.caption = caption;
             // on change callback
-            vm.onBlockChange();
+            $timeout(function() {
```
petrjasek: would be good to have some specific scope in place, not to trigger full page $digest, not sure if there is any

petrjasek:
:+1:

PR #479, opened 2016-05-17
fix(workspace): fix memory leaks

petrjasek:
there were keyboard bindings without unbinding, also some
$document listeners witout de-registration.
also updating some libs for testing

SD-4519


akintolga:
:+1:

mdhaman:
:+1:

PR #487, opened 2016-05-19
[SD-4477]-(fix) - Confirmation window not always appearing when closi…

sjunaid:
…ng an item from bottom panel

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -814,6 +814,9 @@
             };
         };
 
+        // Flag to set for instant checking if $scope is dirty for current item
+        this.dirty = null;
+
         /**
          * Called from workqueue in case of unsaved changes.
          */
@@ -1502,6 +1505,7 @@
                 };
 
                 $scope.save_enabled = function() {
+                    confirm.dirty = $scope.dirty;
                     return $scope.dirty || $scope.item._autosave;
                 };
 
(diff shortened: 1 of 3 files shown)


scripts/superdesk-authoring/workqueue/workqueue.js
```diff
@@ -127,16 +127,28 @@ function WorkqueueCtrl($scope, $rootScope, $route, workqueue, authoringWorkspace
...
+                        return confirm.reopen().then(function(reopen) {
+                            _reOpen(item);
+                        });
+                    }
+                }
                 _closeItem(item);
             });
     };
 
+    function _reOpen(item) {
```
mugurrus: For clarity I would rename this function to _reOpenItem

mugurrus:
:+1:

petrjasek:
:+1: but would be good to have some test for this

scripts/superdesk-authoring/authoring.js
```diff
@@ -1502,7 +1505,7 @@
                 };
 
                 $scope.save_enabled = function() {
-                    return $scope.dirty || $scope.item._autosave;
+                    return (confirm.dirty = $scope.dirty || $scope.item._autosave);
```
akintolga: Did you mean an assignment or a Boolean operation here?

sjunaid: both actually, the assignment just before returning the evaluated boolean

petrjasek: it's both, returns boolean and assigns it to confirm service. maybe it could done in 2 steps to make it more obvious

sjunaid: @petrjasek , slightly changed it as interested only in scope.dirty() or keep both?

petrjasek: seems ok now

ioanpocol:
:+1:

PR #491, opened 2016-05-19
Subject and category in metadata fixed

Aca-jov:
SDNTB-208

### DIFF
--- a/scripts/superdesk-archive/views/metadata-view.html
+++ b/scripts/superdesk-archive/views/metadata-view.html
@@ -17,7 +17,7 @@
             <dt translate>Word Count</dt>
             <dd>{{ item.word_count }}</dd>
         </dl>
-        <dl ng-if="item.keywords && item.keywords.length > 0">
+        <dl ng-if="item.keywords && item.keywords.length">
             <dt translate>Keywords</dt>
             <dd>{{ item.keywords | mergeWords}}</dd>
         </dl>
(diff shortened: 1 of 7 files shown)


scripts/superdesk-archive/views/metadata-view.html
```diff
@@ -47,11 +47,15 @@
...
         <dl>
             <dt translate>ANPA Category</dt>
-            <dd ng-if="item['anpa_category']">{{ item['anpa_category'] | mergeWords:'name' }}</dd>
+            <dd ng-if="item.anpa_category">{{ item.anpa_category | mergeWords:'name' }}</dd>
         </dl>
-        <dl >
+        <dl>
             <dt translate>Subject</dt>
-            <dd ng-if="item.subject && item.subject.length > 0">{{ item.subject | mergeWords:'name'}}</dd>
+            <dd>{{ item.subject | mergeWords:'name':'subject_custom'}}</dd>
```
ioanpocol: This subject_custom is just a cv name, so it should not be hardcoded here.

scripts/superdesk-archive/views/metadata-view.html
```diff
@@ -47,11 +47,15 @@
...
+        </dl>
+        <dl>
+            <dt translate>Category</dt>
```
ioanpocol: Category is a custom cv for NTB, so it should not appear here. The solution is to read the list of cv for current item's profile and show them dynamically in the way that are shown on item header.

petrjasek:
:+1:

ioanpocol:
:+1:

PR #496, opened 2016-05-20
feat(profile): Add the option to change the current profile for a content item

ioanpocol:
By default this option is disabled so all tests should pass.


Aca-jov:
:+1:

mugurrus:
:+1:

PR #497, opened 2016-05-20
Editor2: never use actual associations reference inside the editor

vied12:
But copied object only. This will keep origItem consistent.

SDPA-451

Depends on https://github.com/superdesk/superdesk-core/pull/370


mugurrus:
:+1:

actionless:
:+1:

PR #504, opened 2016-05-23
[SD-4586] - Cleaned Sms message from corrected stories

akintolga:


petrjasek:
:+1:
so this should keep the previous sms when doing correction right? couldn't find the info in jira

akintolga:
You are right @petrjasek the original ticket didn't have enough information. We have created a separate ticket for this requirement: https://dev.sourcefabric.org/browse/SD-4586

mdhaman:
:+1:

PR #510, opened 2016-05-24
ui(fix SD-4553): highlight star would not indicate correct status in monitoring view

gbbr:
**When on "Monitoring" page:**
The star will be red when the item is active within _at least one_ highlight package.

**When on "Highlights" page:**
The star will only be red if the item is active within the highlight package that is currently viewed.


akintolga:
:+1:

mdhaman:
:+1:

petrjasek:
:+1:

PR #512, opened 2016-05-24
chore(spec): add scripts to mark tests as forced (s/it/fit/)

actionless:


actionless:
that's needed in order to run smoke e2e testsuite during CI

petrjasek:
:+1:

PR #513, opened 2016-05-24
[SD-4242]-(fix) - Selecting "finance" category opens up company tag w…

sjunaid:
…hich in its turn is very slow to populate
- Implemented vs-repeat in meta-terms directive template.
- plus some improvements related to should display and data-list for company_codes dropdown.


naomiaro:
We already have this used in the project

https://github.com/kamilkp/angular-vs-repeat

I think this might be better if it's a dom rendering long list problem. It's already used in the dictionary screen when searching words.

mugurrus:
:+1:

mdhaman:
:+1:

PR #518, opened 2016-05-26
ui(fix SD-4552): add sd-password-strength indicator

gbbr:
Contains a new directive called `sd-password-strength` that enhances password inputs with strength validation. This directive will fail the validity of the model that is bound to the form if the minimum strength is not met (as per [requirements](https://dev.sourcefabric.org/browse/SD-4552)).

When less than 8 characters are entered, the strength indicator shows the text "_Short_":
screen shot 2016-05-26 at 1 39 47 pm

After more than 8 characters have been entered, strength will be computed on a scale from 1 to 4 based on how many points from below are found:
* a lower case letter (a-z)
* an upper case letter (A-Z)
* a number (0-9)
* a special characters (e.g. !@#$%^&*()_+|~-=\‘{}[]:";’<>?,./)

**Strength 1/4**:
screen shot 2016-05-26 at 1 39 57 pm

**Strength 2/4**:
screen shot 2016-05-26 at 1 40 03 pm

**Strength 3/4 (valid password)**:
screen shot 2016-05-26 at 1 40 09 pm

**Strength 4/4 (strong password)**:
screen shot 2016-05-26 at 1 40 17 pm

Additionally, hovering the question mark will display information about the password requirements:
screen shot 2016-05-26 at 1 40 22 pm

### DIFF
--- a/scripts/superdesk/directives/sdPasswordStrength.js
+++ b/scripts/superdesk/directives/sdPasswordStrength.js
@@ -1,0 +1,111 @@
+(function() {
+    'use strict';
+
+    // config holds the default configuration for the password strength calculator.
+    var config = {
+        MIN_LENGTH: 8,
+        ONE_LOWER: /^(?=.*[a-z])/g,
+        ONE_UPPER: /^(?=.*[A-Z])/g,
+        ONE_NUMBER: /^(?=.*[0-9])/g,
+        ONE_OTHER: /^(?=.*[^0-9^a-z^A-Z])/g,
+        MIN_STRENGTH: 3
+    };
+
+    /**
+     * sdPasswordStrength appends a strength indicator to the input that it is
+     * added to. Strength is computed in the following way:
+     *
+     *   - If the length of the password is less than 8, strength will be set
+     *     to 'Short'.
+     *
+     *   - Strength is increased whenever one of the following is found:
+     *        - a lower-case letter
+     *        - an upper-case letter
+     *        - a number
+     *        - another character
+     *
+     * Usage:
+     * <input type="password" sd-password-strength ng-model="pass">
+     *
+     * Params:
+     * @scope {Object} ngModel - model that the input is bound to
+     */
+    PasswordStrength.$inject = ['gettext', '$interpolate'];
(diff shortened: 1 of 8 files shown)


pavlovicnemanja:
:+1: Nice one! Great job!

Aca-jov:
:+1:  Well done

scripts/superdesk/directives/sdPasswordStrength.js
```diff
@@ -0,0 +1,111 @@
...
+        {txt: 'Short', cls: 'red'},
+        {txt: 'Weak', cls: 'red'},
+        {txt: 'Better', cls: 'yellow'},
+        {txt: 'OK', cls: 'green'},
+        {txt: 'Strong', cls: 'green'}
+    ];
+
+    // helpText holds the text that will be shown when the user hovers over the
+    // informational icon.
+    var helpText = 'Must be ' + config.MIN_LENGTH + ' characters long and ' +
```
petrjasek: would be good to translate this, ideally with `$interpolate` like [here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/superdesk-authoring/authoring.js#L2477)

scripts/superdesk/directives/sdPasswordStrength.js
```diff
@@ -0,0 +1,111 @@
...
+        ONE_UPPER: /^(?=.*[A-Z])/g,
+        ONE_NUMBER: /^(?=.*[0-9])/g,
+        ONE_OTHER: /^(?=.*[^0-9^a-z^A-Z])/g,
+        MIN_STRENGTH: 3
+    };
+
+    // styles holds each of the strength labels by index along with the class
+    // to be added to the indicator.
+    var styles = [
+        {txt: 'Short', cls: 'red'},
```
petrjasek: here it should use `gettext` for those txt values - it's used later so it would translate it, but without it being here it won't be picked when generating strings to translate

gbbr:
Thanks @petrjasek - I've amended the commit.

petrjasek:
:+1:

PR #522, opened 2016-05-27
feat(spellcheck): let user ignore word for edit session

petrjasek:
it keeps a list in memory per item and resets it when an item is closed.

SD-2151


akintolga:
:+1:

mdhaman:
:+1:

PR #523, opened 2016-05-30
content profiles (SD-3871)

gbbr:
**Demo**:
See [video](https://youtu.be/pKhN8vYiAgs) for details.

**Fixes issues**:
[SD-3871](https://dev.sourcefabric.org/browse/SD-3871) - Create a content profile
[SD-3894](https://dev.sourcefabric.org/browse/SD-3894) - Edit a content profile

It also fixes the front-end part of [SD-3895](https://dev.sourcefabric.org/browse/SD-3895) (Delete a content profile) and part of [SD-3896](https://dev.sourcefabric.org/browse/SD-3896) (Manage content profile fields).

### DIFF
--- a/scripts/superdesk-workspace/content/views/profile-settings.html
+++ b/scripts/superdesk-workspace/content/views/profile-settings.html
@@ -3,21 +3,23 @@
         <div class="header">
             <h2 translate>Content Profiles</h2>
 
-            <button id="add-new-content-profile" class="btn btn-info pull-right"
-                ng-click="ctrl.create()"
-                ng-disabled="true">
+            <button id="add-new-content-profile" class="btn btn-info pull-right" ng-click="ctrl.toggleCreate()">
                 <i class="icon-plus-sign icon-white"></i> {{ :: 'Add New Content Profile' | translate }}
             </button>
         </div>
 
-        <div class="content desk-settings">
+        <div class="content profile-settings desk-settings">
             <div class="flex-grid box wrap-items small-2 medium-4 large-6">
                 <div class="flex-item card-box"
-                    ng-repeat="type in ctrl.types | orderBy: 'label' track by type._id"
+                    ng-click="ctrl.toggleEdit(type)"
+                    ng-repeat="type in ctrl.items | orderBy: 'label' track by type._id"
                     ng-class="{inactive: !type.enabled}">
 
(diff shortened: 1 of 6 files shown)


scripts/superdesk-workspace/content/views/profile-settings.html
```diff
@@ -1,40 +1,103 @@
 <div sd-settings-view>
-    <div class="split-content">
-        <div class="header">
-            <h2 translate>Content Profiles</h2>
+	<div class="split-content">
```
vied12: we used to indent with spaces

ahilles107:
I like PR's with video! One question - why we can't edit profile properties when we create new one (another click on edit is require)? It's not so intuitive for user.

scripts/superdesk-workspace/content/views/schema-editor.html
```diff
@@ -0,0 +1,24 @@
+<div class="schema-editor">
```
vied12: Same here, it would be better to use spaces for indentation

gbbr:
@ahilles107 Yes, this was the same point brought up by @fritzSF - we would have to do this as a follow up and re-think the specs and workflow a bit. Currently the user must first create a profile before he is able to edit the fields. This is because (according to the issue description) when a user wants to create a new profile he should specify _Title_ and _Description_ only. Afterwards, the server decides what fields to pre-populate the profile with as a default, which are then returned to the client. Basically the client is not aware of the fields on a content profile before it is created.

One suggestion @fritzSF had was to replace the "Save" button on the creation form with a "Next", which will automatically open the editing form after creation and providing a better user experience which is closer to yours and his suggestion.

I hope this clarifies your question a bit. I recommend that we do this as a follow up though, because it is a bit out of scope for this sprint.

gbbr:
@vied12 addressed

scripts/superdesk-workspace/content/views/schema-editor.html
gbbr: I replaced all tabs with 4 spaces.

ahilles107:
I totally agree with @fritzSF idea about "next" button. It would solve that issue. Follow up sounds good - but please make sure that it will not end in backlog and live there till end of jiira.

petrjasek:
:+1:

PR #524, opened 2016-05-30
fix(dictionaries): add loading indicator when editing dict

petrjasek:
it can take some time when dictionary gets bigger, so adding
loading indicator. plus optimizing search trie generator.

SD-4548


akintolga:
:+1:

sjunaid:
:+1:

PR #528, opened 2016-05-31
fix(search): search provider radio button disappearing

petrjasek:
it was reseting list of search providers without fetching it again,
now it's set only once.

SD-4496


akintolga:
:+1:

sjunaid:
:+1:

PR #530, opened 2016-05-31
feat(desk): add content profiles setting to desk

petrjasek:
you can pick what content profiles will be available on desk
and what will be set as profile when an item is ingested

SD-4569


petrjasek:
requires https://github.com/superdesk/superdesk-core/pull/394

akintolga:
:+1:

PR #532, opened 2016-05-31
Fixed metadata dropdown

Aca-jov:


ioanpocol:
:+1:

scripts/superdesk/directives/sdTypeahead.js
```diff
@@ -67,6 +67,11 @@
                             $scope.hide = true;
                             $scope.focused = false;
                             $scope.select({item: item});
+                            $scope.active = null;
+                            $scope.term = null;
+                            if (item) {
+                                $document.triggerHandler('click');
```
petrjasek: what does it do?

Aca-jov: Previous when you click on some item, or press enter, active element remained there (wasn't reset) so if we remove that item and then press enter to open dropdown it will add previously selected item although we just pressed enter to open dropdown.

`$scope.term=null` resets search, so on next open of dropdown it will be emtpy.

And third part when you open dropdown, type something  and press enter, it wasn't closing dropdown. I tried, but couldn't resolve it different, it's a bit tricky..

petrjasek: yep it is, maybe can you add a comment there?

akintolga:
:+1:

scripts/superdesk/ui/ui.js
```diff
@@ -423,6 +423,13 @@
                                 }
                             });
 
+                            inputField.on('keydown', function(event) {
```
petrjasek: you will have to unregister this handler manually, otherwise it will leak memory.
with `scope.$on('$destroy', ...)`

Aca-jov: Y, totally forgot... I added it.

pavlovicnemanja:
:+1:

petrjasek:
:+1:

PR #536, opened 2016-06-01
fix(multiedit): use item schema for editor

petrjasek:
it was missing the info about schema and editor for each item.
plus adding a locked indicator - red border like in lists

SD-4560 SD-4562


akintolga:
:+1:

mdhaman:
:+1:

PR #539, opened 2016-06-02
[SD-4565] - Filtering subscribers as 'targetable by users'

akintolga:
Too many subscribers (as we have in pre-production) make it difficult to use resend or target subscriber collections:

- Subscribers `is_targetable` field set as `false` won't be listed in the target subscribers and resend selections.
- If the field is not set or true then the subscriber will be listed.


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/402

mdhaman:
:+1:

petrjasek:
:+1:

PR #544, opened 2016-06-06
SD-3487 As a user I want to be able to use a template in several desks

mugurrus:


mugurrus:
Created another one from a different branch

PR #547, opened 2016-06-06
fix(desks): only show profiles settings if there are any

petrjasek:
in case there are no active profiles hide the labels

SD-4569


ioanpocol:
:+1:

akintolga:
:+1:

PR #548, opened 2016-06-06
fix(profiles): use label from profile instead of its id

petrjasek:
sometimes those ObjectIds can be hard to remember.


akintolga:
:+1:

ioanpocol:
:+1:

PR #549, opened 2016-06-06
fix(desks): save changed done to content_profiles

petrjasek:
it was only working for the first time, but later
because it was only making changes in existing dict
it wasn't picked by update

SD-4660


ioanpocol:
:+1:

akintolga:
:+1:

PR #551, opened 2016-06-06
feat(users): add filters to users list

petrjasek:
by default it will only show active, with options to show
online, pending, inactive, disabled, and all

SD-4649


akintolga:
:+1:

sjunaid:
:+1:

PR #556, opened 2016-06-07
fix(notifications): deliver comments notifications to admin

petrjasek:
there was a filter in place which was confusing it seems. so now
every user mentioned in the comment will get it, no matter if admin
or not.

SD-1406


ioanpocol:
:+1:

akintolga:
:+1:

mdhaman:
:+1:

PR #560, opened 2016-06-07
[SD-4182] - Enabling highlight e2e tests

akintolga:
- This also addresses the problem of the highlight list popup is not appearing when the star clicked...
- Wrong highlight list is loaded when the user navigates to monitoring via address bar


mdhaman:
:+1:

petrjasek:
:+1:

PR #563, opened 2016-06-08
fix(widgets) - Disable widget buttons for readonly stories

akintolga:
- The problem was the isLocked function in authoring.js was hiding the isLocked function in the widget.js


petrjasek:
:+1:

sjunaid:
:+1:

PR #571, opened 2016-06-09
[search] show thumbnails for videos

jerome-poisson:
thumbnails are not restricted to pictures anymore, and show headline on
hover as for picture instead of showing it under the thumbnail.

SDNTB-210

### DIFF
--- a/scripts/superdesk-search/react.js
+++ b/scripts/superdesk-search/react.js
@@ -288,7 +288,7 @@
                      * @return {Boolean}
                      */
                     function hasThumbnail(item) {
-                        return item.type === 'picture' && item.renditions.thumbnail;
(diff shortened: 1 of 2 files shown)


scripts/superdesk-search/react.js
```diff
@@ -288,7 +288,7 @@
                      * @return {Boolean}
                      */
                     function hasThumbnail(item) {
-                        return item.type === 'picture' && item.renditions.thumbnail;
+                        return item.renditions.thumbnail;
```
petrjasek: maybe it throws an error if `item.renditions` is undefined, so should be probably `item.renditions && item.renditions.thumbnail`
otherwise not sure why it would fail so much on travis

ioanpocol:
:+1:

PR #580, opened 2016-06-10
[SDNTB-215] Change "subscriptions" in Scanpix search provider

jerome-poisson:
moved subscriptions options to search.js. Scanpix options are put
directly here for now as it was done in a similar way for PA, but
search providers handling should be refactored and providers
list/params should be updated by backend.

### DIFF
--- a/scripts/superdesk-search/search.js
+++ b/scripts/superdesk-search/search.js
@@ -1587,6 +1587,14 @@
                             scope.searchProviderTypes = searchProviderService.getProviderTypes();
                             scope.cvs = metadata.search_cvs;
                             scope.search_config = metadata.search_config;
+                            scope.scanpix_subscriptions = [{
+                                name: 'subscription',
+                                label: 'inside subscription',
+                            }, {
+                                name: 'all',
+                                label: 'all photos',
+                            }];
+                            scope.meta.scanpix_subscription = scope.scanpix_subscriptions[0].name;
                             scope.lookupCvs = {};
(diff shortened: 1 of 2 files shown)


scripts/superdesk-search/search.js
```diff
@@ -1783,12 +1791,19 @@
...
+                                        }
                                         if (typeof(val) === 'string'){
                                             if (val) {
                                                 metas.push(key + ':(' + val + ')');
                                             }
                                         } else if (angular.isArray(val)) {
                                             angular.forEach(val, function(value) {
+                                                if (typeof value.name !== 'undefined') {
+                                                    // if we have an object with a name, we must use it
+                                                    value = name;
```
petrjasek: imo `name` is undefined here

jerome-poisson: right, this part is useless, good catch

scripts/superdesk-search/search.js
```diff
@@ -1586,6 +1586,14 @@
                             scope.searchProviderTypes = searchProviderService.getProviderTypes();
                             scope.cvs = metadata.search_cvs;
                             scope.search_config = metadata.search_config;
+                            scope.scanpix_subscriptions = [{
+                                name: 'subscription',
+                                label: 'inside subscription',
+                            }, {
+                                name: 'all',
+                                label: 'all photos',
```
petrjasek: translation using `gettext` would be better here, because from here it can extract those strings.

petrjasek:
:+1:

PR #588, opened 2016-06-14
feat(monitoring): optimize updates

petrjasek:
- when an items is expired/moved away from stage - set it as gone with specific css.
- when there is an update to fetch new items and there is an item selected in stage - keep its position

### DIFF
--- a/scripts/superdesk-monitoring/monitoring.js
+++ b/scripts/superdesk-monitoring/monitoring.js
@@ -402,18 +402,16 @@
                     queryItems();
                 });
 
-                scope.$on('task:stage', queryItems);
-                scope.$on('ingest:update', queryItems);
-                scope.$on('item:spike', queryItems);
-                scope.$on('item:duplicate', queryItems);
-                scope.$on('item:copy', queryItems);
+                scope.$on('task:stage', scheduleQuery);
+                scope.$on('item:spike', scheduleQuery);
+                scope.$on('item:duplicate', scheduleQuery);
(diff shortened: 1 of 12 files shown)


akintolga:
:+1:

scripts/superdesk-monitoring/monitoring.js
```diff
@@ -423,18 +422,44 @@
...
+                    } else if (data && cards.shouldUpdate(scope.group, data)) {
                         scheduleQuery();
                     }
                 }
 
-                scope.$on('item:fetch', scheduleIfShouldUpdate);
-                scope.$on('content:update', scheduleIfShouldUpdate);
+                function extendItem(itemId, updates) {
+                    scope.$apply(function() {
+                        scope.items._items = scope.items._items.map(function(item) {
```
mdhaman: Can you please check if the `extendItem` works for the `published` items i.e. there will be multiple versions of the same item in the published collection?

mdhaman:
:+1:  looks except for the comment.

scripts/superdesk-monitoring/monitoring.js
petrjasek: this will only run on stage (when `from_stage === group._id`) so published items won't be affected by this

petrjasek: is this ok @mdhaman ?

mdhaman: no problems.

PR #601, opened 2016-06-15
fix(server): tweak travis config

petrjasek:
- set number of celery processes
- avoid errors when newrelic key is not set

### DIFF
--- a/test-server/app.py
+++ b/test-server/app.py
@@ -1,5 +1,17 @@
 
+import os
+
+
+if os.environ.get('NEW_RELIC_LICENSE_KEY'):
+    try:
+        import newrelic.agent
+        newrelic.agent.initialize(os.path.abspath(os.path.join(os.path.dirname(__file__), 'newrelic.ini')))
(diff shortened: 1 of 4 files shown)


test-server/app.py
```diff
@@ -1,6 +1,18 @@
...
+import os
+
+
+if os.environ.get('NEW_RELIC_LICENSE_KEY'):
+    try:
+        import newrelic.agent
+        newrelic.agent.initialize(os.path.abspath(os.path.join(os.path.dirname(__file__), 'newrelic.ini')))
+    except ImportError:
+        pass
+
```
jerome-poisson: a warning log message would be nice here: if somebody set the license key, she will probably expect it to be activated

jerome-poisson:
:+1:  (with the small remark about a log message)

test-server/app.py
petrjasek: yep this should be probably in the other place where I've added this - here it's only for e2e test server

PR #607, opened 2016-06-16
fix(publish) - Check the publish privilege for correct and kill

akintolga:
- If users don't have publish privilege, even if they have correct and kill, they won't be able to publish corrections or kills.


petrjasek:
:+1:

sjunaid:
:+1:

PR #608, opened 2016-06-16
[SD-3896]: Added field management on content profiles

gbbr:
screen shot 2016-06-16 at 4 12 33 pm

### DIFF
--- a/scripts/superdesk-workspace/content/views/schema-editor.html
+++ b/scripts/superdesk-workspace/content/views/schema-editor.html
@@ -2,7 +2,7 @@
     <h4 translate>Schema Configuration</h4>
     <ul class="pills-list">
         <li class="schema-item"
-            ng-repeat="(id, props) in schema track by $index"
+            ng-repeat="(id, props) in model.schema track by $index"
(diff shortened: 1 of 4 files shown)


scripts/superdesk-workspace/content/views/schema-editor.html
```diff
@@ -17,7 +17,33 @@ <h4 translate>Schema Configuration</h4>
             </div>
 
             <div class="body" ng-if="!!props">
-                Required: <div sd-check ng-model="props.required"></div>
+                <form>
+                    <fieldset>
+                       <div class="field">
+                           Required: <div sd-check ng-model="props.required"></div>
```
petrjasek: all those labels should be translated

petrjasek:
looks good, but there is missing translate annotation for labels

gbbr:
@petrjasek I've added `span[translate]`'s around those labels - is that what you were thinking? Also, there are the placeholders - do you reckon we should improve/extend (not sure if possible) the `translate` directive within this PR to include those?

petrjasek:
:+1: yep that's it

PR #609, opened 2016-06-16
feat(cvs): make it possible to set schema for cv

petrjasek:
and use it when editing. used to implement config for priority/urgency.

SD-4616


mdhaman:
:+1:

akintolga:
:+1:

PR #610, opened 2016-06-16
Editor embed code parser: allow to define custom handler in options SDPA-454

vied12:


petrjasek:
:+1:

vied12:
@petrjasek I finally moved the Vidible handler in https://github.com/superdesk/superdesk/pull/2117 using `config`.

because I think it shouldn't belong to core.
What do you think ?

scripts/superdesk/editor2/add-embed.ctrl.js
```diff
@@ -59,58 +61,31 @@ function SdAddEmbedController (embedService, $element, $timeout, $q, _, EMBED_PR
...
-                                        }
-                                    });
-                                })
-                            );
-                        }
                     }
                 ];
+                // prepend with custom handlers from config
+                if (config.editorEmbedCodeParsers) {
+                    knownPorviders = $injector.invoke(config.editorEmbedCodeParsers).concat(knownPorviders);
```
naomiaro: I would switch this to knownProviders, could be confusing spelling wise.

vied12: Hahaha right

petrjasek:
:+1:

PR #613, opened 2016-06-17
(fix)[SD-4768] Queue items should only refresh if there change in the items.

mdhaman:
Problems Fixed:
- Workqueue items were fetched on every `item:lock`,  `item:unlock`, `locationChangeSuccess` and `content:update` event.
- All queries were fired twice.

I have removed `locationChangeSuccess`. imo, 'content:update', 'item:lock' and 'item:unlock' are sufficient.


marwoodandrew:
:+1:

akintolga:
:+1:

PR #614, opened 2016-06-17
[SD-4681] ui(fix/content-profiles): reject server requests that return errors.

gbbr:
This was causing a bug when creating a content profile with a name that already exists. The "Edit" modal was being brought up, even though a content profile wasn't created.


petrjasek:
:+1:

akintolga:
:+1:

PR #615, opened 2016-06-19
SD-4744 Add proper indication on missing fields upon template creation

mugurrus:


scripts/superdesk-templates/templates.js
```diff
@@ -321,17 +322,30 @@
...
 
+                function validate(orig, item) {
+                    $scope.error = {};
+                    if (!item.template_name) {
+                        $scope.error.template_name = true;
+                    }
+                    if (!item.template_type) {
+                        $scope.error.template_type = true;
+                    }
+                    return Boolean($scope.error);
```
akintolga: It looks to me that this line always returns true.

petrjasek: :+1: right

akintolga:
:+1:

sjunaid:
:+1:

PR #618, opened 2016-06-20
fix(monitoring): stop preview on doubleclick

petrjasek:
it was only waiting for 200ms for second click, now it will
wait for 500ms before starting the preview so it will match
windows defaults

SD-4694


akintolga:
:+1:

sjunaid:
:+1:

PR #620, opened 2016-06-20
Added a method to set default value on init and search provider change

jerome-poisson:
Default value (only used for Scanpix so far) was only set on init, so
it was not working when opening advanced search with default to
inside superdesk, and then clicking on Scanpix.
This commit add setDefaultValues function which is called both on init
and search provider change making initial values consistant.

SDNTB-215


petrjasek:
:+1:

PR #624, opened 2016-06-21
[SD-4745]: fix timezone behavior

gbbr:
- Replaced pencil icon with remove (x) icon to correctly reflect the
  action that happens when it is clicked.

- Focus the timezone input once the field is cleared to encourage the
  user to type a new value.

Also fixes [SD-4791](https://dev.sourcefabric.org/browse/SD-4791).

### DIFF
--- a/scripts/superdesk/ui/ui.js
+++ b/scripts/superdesk/ui/ui.js
@@ -857,15 +857,15 @@
         };
     }
 
-    TimezoneDirective.$inject = ['tzdata', 'config'];
-    function TimezoneDirective(tzdata, config) {
+    TimezoneDirective.$inject = ['tzdata', 'config', '$timeout'];
+    function TimezoneDirective(tzdata, config, $timeout) {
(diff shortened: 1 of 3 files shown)


scripts/superdesk/ui/ui.js
```diff
@@ -926,6 +926,9 @@
                  * @method clearSelectedTimeZone
                  */
                 scope.clearSelectedTimeZone = function () {
+                    $timeout(function() {
+                        el.find('input')[0].focus();
+                    }, 0);
```
petrjasek: you can add `false` param not to trigger scope digest

petrjasek:
:+1:

scripts/superdesk/ui/ui.js
gbbr: Wonderful, I did not know this. I've amended.

akintolga:
:+1:

PR #626, opened 2016-06-21
fix(schedule): fix time picker time when utc time is missing

petrjasek:
after descheduling there is no utc schedule value, so it's using
`publish_schedule` field which has timezone info added by server (+0000)
which is wrong and should be ignored

SD-4759


mdhaman:
:+1:

akintolga:
:+1:

PR #628, opened 2016-06-21
SD-3487 As a user I want to be able to use a template in several desks

mugurrus:


akintolga:
:+1:

akintolga:
@mugurrus We have been enabling the e2e tests recently and I noticed that we don't have any e2e tests for templates. Would it be possible for you two write 1-2 two cases to cover the basics?

mdhaman:
:+1:

PR #632, opened 2016-06-22
[SD-4729] [SD-4730](fix) Adding target fields in template and routing.

mdhaman:
*Fixes Include*
- The broken query for templates in save as templates
- Filter not working if desk is select in the template management
- Change the query so that user will see the `personal template, public template assigned to the desk and public template assigned to no desk` in the modal for `create item from template`. 

@petrjasek @mugurrus @ioanpocol 

Depends on [Core PR 437](https://github.com/superdesk/superdesk-core/pull/437)


sjunaid:
:+1: LGTM

akintolga:
:+1:

petrjasek:
:+1:

PR #636, opened 2016-06-22
feat(associations): Customize related items to disable upload&editing…

ioanpocol:
… and allow only drag and drop (SDNTB-232)


jerome-poisson:
:+1:

PR #641, opened 2016-06-23
Editor: cut and paste a block SD-4673

vied12:
- Allow to cut a block selection_094

- Allow to paste a cut block by clicking on the new `Paste` button or by pressing `ctrl`+`v`
selection_095

### DIFF
--- a/scripts/superdesk/editor2/add-content.js
+++ b/scripts/superdesk/editor2/add-content.js
@@ -124,7 +124,7 @@ function AddContentCtrl (scope, element, superdesk, editor, $timeout, config, $q
         },
         actions: {
             addEmbed: function() {
-                vm.sdEditorCtrl.splitCurrentTextBlockAndInsertBetween(vm.textBlockCtrl).then(function() {
+                vm.sdEditorCtrl.splitAndInsert(vm.textBlockCtrl).then(function() {
                     // show the add-embed form
                     vm.textBlockCtrl.block.showAndFocusLowerAddAnEmbedBox();
                 });
@@ -142,7 +142,7 @@ function AddContentCtrl (scope, element, superdesk, editor, $timeout, config, $q
                             };
                         });
                     })).then(function(renderedImages) {
-                        vm.sdEditorCtrl.splitCurrentTextBlockAndInsertBetween(vm.textBlockCtrl, renderedImages);
+                        vm.sdEditorCtrl.splitAndInsert(vm.textBlockCtrl, renderedImages);
                     });
                 }, function() {
                     scope.node.focus();
@@ -153,7 +153,7 @@ function AddContentCtrl (scope, element, superdesk, editor, $timeout, config, $q
                 if (!vm.sdEditorCtrl.getCutBlock()) {
                     return false;
                 }
-                vm.sdEditorCtrl.splitCurrentTextBlockAndInsertBetween(vm.textBlockCtrl, vm.sdEditorCtrl.getCutBlock(true));
+                vm.sdEditorCtrl.splitAndInsert(vm.textBlockCtrl, vm.sdEditorCtrl.getCutBlock(true));
             },
         }
     });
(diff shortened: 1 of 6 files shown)


scripts/superdesk/editor2/add-content.js
```diff
@@ -162,7 +165,24 @@ function AddContentCtrl (scope, element, superdesk, editor, $timeout, config) {
...
-            }
+            },
+            pasteBlock: function() {
+                if (!vm.sdEditorCtrl.getCutBlock()) {
+                    return false;
+                }
+                var indexWhereToAddNewBlock = vm.sdEditorCtrl.getBlockPosition(vm.textBlockCtrl.block) + 1;
+                // cut the text that is after the caret in the block and save it in order to add it after the embed later
+                var textThatWasAfterCaret = vm.textBlockCtrl.extractEndOfBlock().innerHTML;
+                if (textThatWasAfterCaret && textThatWasAfterCaret !== '') {
```
gbbr: You can just say `if (textThatWasAfterCaret)` and it will cover the second assertion too. An empty string is a falsy value.

scripts/superdesk/editor2/editor.ctrl.js
```diff
@@ -247,15 +247,22 @@ function SdTextEditorController(_, EMBED_PROVIDERS, $timeout, $element, editor,
...
         insertNewBlock: function(position, attrs, doNotRenderBlocks) {
             var new_block = new Block(attrs);
-
-            $timeout(function() {
-                vm.blocks.splice(position, 0, new_block);
-                $timeout(vm.commitChanges, 0, false);
-                if (!doNotRenderBlocks) {
-                    $timeout(vm.renderBlocks, 0, false);
-                }
+            return $q(function(resolve, reject) {
```
gbbr: You may remove the second parameter if it's not used anywhere.

petrjasek:
:+1:

scripts/superdesk/editor2/editor.ctrl.js
```diff
@@ -237,6 +237,48 @@ function SdTextEditorController(_, EMBED_PROVIDERS, $timeout, $element, editor,
         getBlockPosition: function(block) {
             return _.indexOf(vm.blocks, block);
         },
+        splitCurrentTextBlockAndInsertBetween: function(textBlockCtrl, blockToInserts) {
+            var indexWhereToAddNewBlock = vm.getBlockPosition(textBlockCtrl.block) + 1;
```
gbbr: In my experience it is a better practice to use variable names as short and succinct as possible, without becoming to ambiguous. It makes it easier for programmers that are unfamiliar with your code to understand it. The reason for this being that long identifier names take a long time to read and a lot of 'brain space' to keep in mind. Whereas in something like literature it is nice for the words to be as expressive as possible, in programming I find it is better to make it easy for others to read the logic and flow of execution, as opposed to the clarity of identifier meanings.

For example, this whole component is mostly about 'block text', so it is a bit redundant to add 'Block' into every word.

In my opinion, it would look nicer as:

* `splitAndInsert` as opposed to `splitCurrentTextBlockAndInsertBetween`
* `idx` or `index` instead of `indexWhereToAddNewBlock`
* `after` instead of `textThatWasAfterCaret`
* `end()` instead of `extractEndOfBlock()`
* etc....

If any of the above prove to be slightly hard to follow, then an inline comment above the declaration can clarify their meaning.

This is just my personal view on declaring identifiers and I am happy for you to leave it as is if you disagree, but I encourage you to consider it. Readability is much more important than [brevity](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=define%20brevity) when it comes to programming.

petrjasek:
:+1:

scripts/superdesk/editor2/editor.ctrl.js
```diff
@@ -237,6 +237,49 @@ function SdTextEditorController(_, EMBED_PROVIDERS, $timeout, $element, editor,
         getBlockPosition: function(block) {
             return _.indexOf(vm.blocks, block);
         },
+        splitAndInsert: function(textBlockCtrl, blocksToInsert) {
```
gbbr: Super! :-) 👍

PR #650, opened 2016-06-27
[SD-4815]: missing labels appearing as empty on content schema editor

gbbr:
Whenever a translation is missing for an item in the schema of a content
profile, the key for the item is now displayed and a warning to check
the error console for more details.

Do note that whenever adding new items to the schema, subsequent
translations needs to be added in the front-end in
`(superdesk-workspace/content/content.js).ContentProfileSchemaEditor/labelMap`

### DIFF
--- a/scripts/superdesk-workspace/content/content.js
+++ b/scripts/superdesk-workspace/content/content.js
@@ -545,7 +545,12 @@
             'byline': gettext('By'),
             'dateline': gettext('Date'),
             'located': gettext('Located'),
-            'sign_off': gettext('Sign Off')
+            'sign_off': gettext('Sign Off'),
+            'sms': gettext('SMS'),
+            'body_footer': gettext('Body footer'),
+            'footer': gettext('Footer'),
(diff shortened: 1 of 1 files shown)


scripts/superdesk-workspace/content/content.js
```diff
@@ -560,7 +565,13 @@
                  * @description label returns the display name for a key.
                  */
                 scope.label = function(id) {
-                    return labelMap[id];
+                    if (labelMap.hasOwnProperty(id)) {
+                        return labelMap[id];
+                    }
+
+                    console.error('could not find label for ' + id +
```
petrjasek: imo would be better to use `console.warn` here, and it shouldn't show the warning to users - they might not be able to do much about it.

gbbr:
@petrjasek agreed & changed to warning + removed user visible warning

mdhaman:
:+1:

petrjasek:
:+1:

PR #654, opened 2016-06-27
feat(vocabulary): Limit list name only for 2 characters

pavlovicnemanja:
Because of rounded frames for News Value and Priority, we limited list name only for 2 characters. Approved by @fritzSF


akintolga:
:+1:

sjunaid:
:+1:

PR #655, opened 2016-06-27
[search providers] create 4 types of Scanpix search provider

jerome-poisson:
This commit creates 4 types of Scanpix search provider (ntbtema, ntbkultur, desk and npk).

the hardcoded tests are not good for maintaince, this should be
refactored as specified in comments, it's a temporaty fix.

fix SDNTB-217


petrjasek:
:+1:

PR #662, opened 2016-06-28
fix(profiles): disable remove button when profile is used

petrjasek:
works together with https://github.com/superdesk/superdesk-core/pull/448

SD-3895


akintolga:
:+1:

sjunaid:
:+1:

PR #664, opened 2016-06-28
fix(about): fix version info

petrjasek:
read it from client/package.json file instead of git,
because there is no git in place within node modules

SD-4770


akintolga:
:+1:

marwoodandrew:
:+1:

petrjasek:
I was trying to get revision id of the component, but it should be the main repo

PR #670, opened 2016-06-29
SDNTB-205 translation issues

vladnicoara:


scripts/superdesk-search/react.js
```diff
@@ -1019,7 +1019,8 @@
 
                         takekey: function(props) {
                             if (props.item.anpa_take_key) {
-                                return React.createElement('div', {className: 'takekey', key: 'takekey'}, props.item.anpa_take_key);
+                                return React.createElement('div', {className: 'takekey', key: 'takekey'},
+                                    gettextCatalog.getString(props.item.anpa_take_key));
```
petrjasek: just wondering - when you use `gettextCatalog` instead of `gettext` - is it to only translate but avoid another entry in po files?

petrjasek:
:+1:

gbbr:
👍

vladnicoara:
hi @petrjasek @gbbr because of https://github.com/superdesk/superdesk-client-core/blob/master/scripts/superdesk/gettext.js#L4 gettext only makes an entry in the po file and it has to be used with a translate filter/directive in the template. gettextCatalog.getString also pulls the translation( besides the po entry) and I am using it when the combination of gettext/translate is not available

gbbr:
Thanks for clarifying. LGTM

PR #671, opened 2016-06-29
Keep history of blocks manipulations

vied12:
Undo/Redo operation on block with `ctrl`+`z` and `ctrl`+`y`

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -48,6 +48,7 @@
     <script src="bower_components/ng-file-upload/ng-file-upload.js"></script>
     <script src="bower_components/ng-file-upload/ng-file-upload-shim.js"></script>
     <script src="bower_components/raven-js/dist/raven.js"></script>
+    <script src="bower_components/angular-history/history.js"></script>
 
     <script src="bower_components/angular-embed/dist/angular-embed.js"></script>
     <script src="bower_components/angular-contenteditable/angular-contenteditable.js"></script>
--- a/bower.json
+++ b/bower.json
@@ -31,8 +31,7 @@
     "angular-vs-repeat": "~1.1.4",
     "medium-editor-tables": "~0.5.0",
     "exif-js": "^2.1.1",
-    "angular-i18n": "^1.5.3",
-    "angular-history": "^0.8.1"
+    "angular-i18n": "^1.5.3"
   },
   "ignore": [
     "**/.*",
--- a/bower.json
+++ b/bower.json
@@ -31,7 +31,8 @@
     "angular-vs-repeat": "~1.1.4",
     "medium-editor-tables": "~0.5.0",
     "exif-js": "^2.1.1",
-    "angular-i18n": "^1.5.3"
+    "angular-i18n": "^1.5.3",
+    "angular-history": "^0.8.1"
   },
   "ignore": [
     "**/.*",
--- a/package.json
+++ b/package.json
@@ -33,10 +33,10 @@
     "load-grunt-config": "0.17.2",
     "load-grunt-tasks": "3.2.0",
     "request": "2.61.0",
-    "selenium-webdriver": "2.45.1"
+    "selenium-webdriver": "2.45.1",
+    "angular-history": "git+https://github.com/decipherinc/angular-history.git#v0.8.0"
   },
   "devDependencies": {
-    "angular-history": "git+https://github.com/decipherinc/angular-history.git#v0.8.0",
     "babel-core": "^6.10.4",
     "babel-loader": "^6.2.4",
     "babel-preset-es2015": "^6.9.0",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "selenium-webdriver": "2.45.1"
   },
   "devDependencies": {
+    "angular-history": "git+https://github.com/decipherinc/angular-history.git#v0.8.0",
     "babel-core": "^6.10.4",
     "babel-loader": "^6.2.4",
     "babel-preset-es2015": "^6.9.0",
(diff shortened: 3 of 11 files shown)


scripts/superdesk-authoring/services.js
```diff
@@ -1,7 +1,55 @@
 (function() {
 'use strict';
 
-angular.module('superdesk.authoring').service('renditions', RenditionsService);
+angular.module('superdesk.authoring')
+.service('renditions', RenditionsService)
```
gbbr: I think these two lines should be indented 4 spaces to the right based on most of our codebase (as far as I've seen).

vied12: right, I will change to stay consistent

scripts/superdesk/editor2/editor.spec.js
```diff
@@ -138,42 +138,6 @@ describe('text editor', function() {
 
     }));
 
-    it('can save model value', inject(function(editor, $rootScope) {
```
mdhaman: why this test is removed?

sjunaid:
Hi @vied12 , found few issues while quick testing;
- undo/redo not working with other input fields including slugline, headline, abstract, byline etc.
- when creating new item and directly paste some text in body (without typing anything) then try undo - notice undo not works.
- when creating new item type one two three in body field (but wait a bit before typing each word to allow rendering). now try undo one by one notice first word (one) not gets undo

scripts/superdesk/editor2/editor.spec.js
vied12: Because the tested methods have been removed by this PR.
Now the history is kept at the global editor level, not anymore at the text block level

vied12:
@sjunaid Now I save all fields and allow to return to the empty state.
I also wrote e2e tests

scripts/superdesk-authoring/authoring.js
```diff
@@ -2601,6 +2603,13 @@
 
                 /* End: Dateline related properties */
 
+                // watch item and save every change in history in order to perform undo/redo later
+                // ONLY for editor2 (with blocks)
+                try {
+                    angular.module('superdesk.editor2');
+                    history.watch('item', scope);
```
petrjasek: is this doing some recursive wathing on the item? so it would also track changes done in sidebars etc?

petrjasek:
:+1:

app/index.html
```diff
@@ -48,6 +48,7 @@
     <script src="bower_components/ng-file-upload/ng-file-upload.js"></script>
     <script src="bower_components/ng-file-upload/ng-file-upload-shim.js"></script>
     <script src="bower_components/raven-js/dist/raven.js"></script>
+    <script src="bower_components/angular-history/history.js"></script>
```
gbbr: Should we start to slowly move away from bower and use npm instead? Might be nicer on the long run and would completely remove index.html clutter. I was thinking to maybe just `npm install angular-history` and `import 'angular-history';` in `index.js`, or better yet, importing it only locally in your own module.

What do you think?

bower.json
```diff
@@ -31,7 +31,8 @@
     "angular-vs-repeat": "~1.1.4",
     "medium-editor-tables": "~0.5.0",
     "exif-js": "^2.1.1",
-    "angular-i18n": "^1.5.3"
+    "angular-i18n": "^1.5.3",
+    "angular-history": "^0.8.1"
```
gbbr: If we do the above suggestion, then this can go away, and so can the addition to `karma.conf.js` below.

scripts/superdesk-authoring/services.js
```diff
@@ -1,7 +1,62 @@
...
+    .service('renditions', RenditionsService)
+    .factory('history', HistoryFactory);
+
+/**
+ * Watches an expression to keep history of its states
+ * and binds ctrl-z and ctrl-y to undo/redo its states
+ */
+HistoryFactory.$inject = ['History', '$window', '$timeout'];
+function HistoryFactory(History, $window, $timeout) {
+    var KeyOperations = {};
```
gbbr: This isn't necessary, but I think it will work now with ES6:
```
var KeyOperations = {
    'Z'.charCodeAt(0): History.undo,
    'Y'.charCodeAt(0): History.redo
};
```

gbbr:
I would get rid of the bower dependency, otherwise LGTM 👍

scripts/superdesk-authoring/services.js
vied12: Actualy it's more
```js
var KeyOperations = {
    ['Z'.charCodeAt(0)]: History.undo,
    ['Y'.charCodeAt(0)]: History.redo
};
```
Thanks for hint, I didn't know !

app/index.html
vied12: Ok I will import it with
```
import 'angular-history/history.js';
```
in authoring.js, where I use it

gbbr:
👍

package.json
```diff
@@ -36,6 +36,7 @@
     "selenium-webdriver": "2.45.1"
   },
   "devDependencies": {
+    "angular-history": "git+https://github.com/decipherinc/angular-history.git#v0.8.0",
```
petrjasek: not 100% sure, but imo this should be part of dependencies, because you need it installed also when using this repo as library.

gbbr: Based on this logic, if we consider our scenario where the main repo `superdesk` embeds this one as a module, then all of the `devDependencies` should be in `dependencies`. I am also not sure of the implications of that. But at least it would remove the duplication of `package.json`

petrjasek:
:+1:

app/index.html
gbbr: @vied12 you forgot to remove this line before merging

vied12: oups... will do it in my next PR

vied12: Actually this file is ignored by git now, by you ;)

PR #674, opened 2016-06-30
[SD-4856] - Unable to unlock other users' draft

akintolga:
- `draft` is the "version 0" state of the story and currently every users draft is visible to everyone.
- Ideally `draft` stories should have been limited to the owners only. 
- But that change may affect other areas so we're limiting the actions on `draft` stories so that a user cannot do anything on another users draft story

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -531,6 +531,13 @@
             var user_privileges = privileges.privileges;
             var action = angular.extend({}, DEFAULT_ACTIONS);
             var itemOnReadOnlyStage = item && item.task && item.task.stage && desks.isReadOnlyStage(item.task.stage);
+            var someoneElsesDraft = item.state === 'draft' && lock.isLocked(item) && !lock.isLockedByMe(item);
+
+            if (someoneElsesDraft) {
+                // No action is allowed if the story is someone else's draft
+                action.view = false;
+                return action;
+            }
 
             // takes packages are readonly.
             // killed item and item that have last publish action are readonly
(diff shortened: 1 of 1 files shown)


mdhaman:
:+1:

scripts/superdesk-authoring/authoring.js
```diff
@@ -3206,7 +3213,18 @@
          */
         this.edit = function(item, action) {
             if (item) {
-                authoringOpen(item._id, action || 'edit', item._type || null);
+                var actions = authoring.itemActions(item);
+                if (action) {
```
gbbr: You can use `if (action && actions[action])` here.

scripts/superdesk-authoring/authoring.js
```diff
@@ -3206,7 +3213,18 @@
...
+                    // action is provided, check if it is permitted
+                    if (actions[action]) {
+                        authoringOpen(item._id, action, item._type || null);
+                    }
+                } else {
```
gbbr: Here you could do `else if` instead of nesting.

scripts/superdesk-authoring/authoring.js
```diff
@@ -3206,7 +3213,18 @@
          */
         this.edit = function(item, action) {
             if (item) {
```
gbbr: Is there ever a case when no `item` is provided for editing? If no, then the condition is not needed imo.

gbbr:
Looks good besides minor style comments.

akintolga:
Closing this PR as there's a new PR to solve this problem by not displaying other users' draft stories...

PR #677, opened 2016-06-30
fix(ftp): set passive mode by default

petrjasek:
when configuring ingest or publish via ftp set passive mode on
by default, to match server config

SDNTB-17


gbbr:
👍

pavlovicnemanja:
:+1:

PR #678, opened 2016-06-30
ui(fix): when slug line is too long it breaks layout in preview (SD-4775)

gbbr:
Additionally, when hovering over the slugline, the full text can be seen in the tooltip.


pavlovicnemanja:
:+1:

petrjasek:
:+1:

PR #679, opened 2016-06-30
fix(splitter): SD-4735 Fixed position of splitter after changing view

pavlovicnemanja:
SD-4735 - When authoring is expanded to the left the borders should remain unchanged if users navigate thru the left panel 

SD-4851 - Three dot menu of items of a package is not working in firefox


petrjasek:
:+1:

Aca-jov:
:+1:

PR #680, opened 2016-06-30
Updating search results after multi send fixed.

eozan:
SD-4807

### DIFF
--- a/scripts/superdesk-ingest/module.js
+++ b/scripts/superdesk-ingest/module.js
@@ -1582,8 +1582,8 @@
         });
     }]);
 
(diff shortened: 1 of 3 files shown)


petrjasek:
imo would be better to fetch single item instead of reloading the whole thing, also this change might not be available yet due to elastic caching, but will be there if you get item by id. but probably you get the updated item from server already, so it's just a matter of updating that item in list

eozan:
This is done the same way authoring>sendTo works. Imo, let's add a refactoring task to fix all of them together, if this is the wrong way.

eozan:
Now changed to update only sent item.

mugurrus:
:+1:

PR #682, opened 2016-07-01
ui(fix): disable default browser behavior when dropping onto editor

gbbr:
When dragging an image onto the body field in the editor, by default,
the browser will open the image. After this change, this will no longer
happen and the user will stay with the editor.

Addresses [latest comment](https://dev.sourcefabric.org/browse/SD-4648?focusedCommentId=84964&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-84964) in [SD-4648](https://dev.sourcefabric.org/browse/SD-4648).

The line removed in the change is no longer needed because the default 
behavior has now changed via #644.

### DIFF
--- a/scripts/superdesk/editor2/editor.js
+++ b/scripts/superdesk/editor2/editor.js
@@ -962,8 +962,6 @@ angular.module('superdesk.editor2', [
                         setEditorFormatOptions(editorConfig, sdTextEditor.editorformat, scope);
                         // if config.multiBlockEdition is true, add Embed and Image button to the toolbar
                         if (scope.config.multiBlockEdition) {
-                            // this dummy imageDragging stop preventing drag & drop events
-                            editorConfig.extensions = {'imageDragging': {}};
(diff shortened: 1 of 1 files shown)


scripts/superdesk/editor2/editor.js
```diff
@@ -963,7 +963,6 @@ angular.module('superdesk.editor2', [
                         // if config.multiBlockEdition is true, add Embed and Image button to the toolbar
                         if (scope.config.multiBlockEdition) {
                             // this dummy imageDragging stop preventing drag & drop events
```
vied12: Then this comment can be removed. And does everything works fine with image drag&drop now ?

gbbr: You're right! Sorry about that. Well, according to the [SD-4648](https://dev.sourcefabric.org/browse/SD-4648) we shouldn't be able to drag images. Additionally, if someone requires images to be dragged into a field it can now be enabled via config `imageDragging: true`. This was changed in #644

gbbr:
I'm closing this at the change can not be committed.

PR #688, opened 2016-07-04
fix(users): SD-4722 Fix user filtering

pavlovicnemanja:
SD-4722 - User management issues with filtering by disabled, pending and inactive flags

Done in same way as adding status labels.


mdhaman:
:+1:

akintolga:
:+1:

PR #697, opened 2016-07-07
Extend the mixin which generate field orders for flex

vied12:
And use the same default order in Preview as in Editor

SD-4932


gbbr:
👍

PR #698, opened 2016-07-07
SDNTB-218 Drop a video in a story body

vied12:


scripts/superdesk/editor2/add-content.js
```diff
@@ -132,7 +132,7 @@ function AddContentCtrl (scope, element, superdesk, editor, $timeout, config, $q
             addPicture: function() {
                 superdesk.intent('upload', 'media').then(function(images) {
                     $q.all(images.map(function(image) {
```
gbbr: Since these can now be both images and videos, maybe it's best we rename `images` and `image` to `items` or `item` to avoid confusion.

scripts/superdesk/editor2/add-content.js
```diff
@@ -132,7 +132,7 @@ function AddContentCtrl (scope, element, superdesk, editor, $timeout, config, $q
...
-                        return editor.generateImageTag(image).then(function(imgTag) {
+                        return editor.generateMediaTag(image).then(function(imgTag) {
```
gbbr: Same here, s/`imgTag`/`tag`

scripts/superdesk/editor2/add-content.js
vied12: I don't think so, because we are in `addPicture`. We are adding only picture there

gbbr: Ah, you're right. My bad.

gbbr:
LGTM 👍

PR #484, opened 2016-05-18
[SD-4494] Modified editor to allow for automatic text expansion.

mdhaman:
This PR is dependent on [#362](https://github.com/superdesk/superdesk-core/pull/362) 

- [x] Confirmation for overwriting the existing abbreviation
- [x] Some event to refresh the cache for abbreviation.

@petrjasek can you review this?

PR #699, opened 2016-07-08
ui/fix: fixes a problem where dictionary contents were not visible [fixes SD-4817]

gbbr:
I believe that initially the intention here was to duplicate the response object. In that case `_.create` is not the best option and I assume `_.cloneDeep` is more appropriate and closer to what was planned.

Reverses some changes made in #484 /cc @mdhaman


mdhaman:
:+1:

ioanpocol:
:+1:

PR #709, opened 2016-07-11
fix(templates): SD-4852 Fix list of available templates

pavlovicnemanja:
SD-4852

### DIFF
--- a/scripts/superdesk-templates/templates.js
+++ b/scripts/superdesk-templates/templates.js
@@ -151,8 +151,6 @@
                 criteria.template_type = type;
             }
 
-            criteria.schedule_desk = null;
-
             if (templateName) {
                 criteria.template_name = {'$regex': templateName, '$options': '-i'};
             }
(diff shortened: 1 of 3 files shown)


scripts/superdesk-templates/templates.js
```diff
@@ -151,8 +151,6 @@
                 criteria.template_type = type;
             }
 
-            criteria.schedule_desk = null;
```
pavlovicnemanja: @mdhaman can you confirm that this is not breaking something? This was preventing templates with `schedule_desk` to be listed in list.

mdhaman: Original thinking was not to show 'schedule' templates. It should not cause any problems if we enable it. 
Please enable it as it is also a "template" but it would be good to have indicator in available templates list that it is a scheduled template as well.

mdhaman:
:+1:

akintolga:
:+1:

PR #710, opened 2016-07-12
Monitoring widget not updating after item send fixed.

eozan:
SD-4847

### DIFF
--- a/scripts/superdesk-monitoring/aggregate.js
+++ b/scripts/superdesk-monitoring/aggregate.js
@@ -12,9 +12,9 @@
     'use strict';
 
     AggregateCtrl.$inject = ['$scope', 'api', 'desks', 'workspaces', 'preferencesService', 'storage',
-                             'gettext', 'multi', 'config'];
+                             'gettext', 'multi', 'config', '$timeout'];
     function AggregateCtrl($scope, api, desks, workspaces, preferencesService, storage,
-            gettext, multi, config) {
+            gettext, multi, config, $timeout) {
         var PREFERENCES_KEY = 'agg:view';
         var defaultMaxItems = 10;
         var self = this;
@@ -381,6 +381,13 @@
             });
         };
 
+        this.reload = function() {
+            $timeout(function() {
+                self.search(self.searchQuery);
+            }, 1, false);
+            self.search();
+        };
+
         this.state = storage.getItem('agg:state') || {};
         this.state.expanded = this.state.expanded || {};
 
@@ -409,6 +416,10 @@
 
         $scope.$on('open:resend', function(evt, item) {
             $scope.resend = item;
+        });
+
+        $scope.$on('item:move', function(evt, item) {
+            self.reload();
         });
     }
 


scripts/superdesk-authoring/authoring.js
```diff
@@ -2352,6 +2352,7 @@
                         })
                         .then(function(value) {
                             notify.success(gettext('Item sent.'));
+                            $rootScope.$broadcast('item:fetch');
```
gbbr: I'm not 100% familiar with the functionality here, but wouldn't `item:move` be more appropriate? It seems like the item is being sent / moved, not fetched. `item:move` is a new event that has been added by the backend and there could already be handlers in place.

scripts/superdesk-monitoring/aggregate.js
```diff
@@ -381,6 +381,13 @@
             });
         };
 
+        this.reload = function() {
+            $timeout(function() {
+                self.search(self.searchQuery);
+            }, 1, false);
+            self.search();
```
gbbr: Why is `self.search` run twice?

gbbr:
LGTM 👍  besides questions

scripts/superdesk-authoring/authoring.js
ioanpocol: Fetch is an operation to add on archive an item from ingest list. The operation needs more changes compared with a simple move.

gbbr: :+1: thanks for clarifying

scripts/superdesk-monitoring/aggregate.js
eozan: @gbbr, ideally it should just refresh current search by only calling it once. But because of the way we have set it up, search sets watched values which eventually update actual view. In its current state it's much more easier to have it working this way than refactoring, though alternative solutions are welcome.

gbbr: Thanks for explaining that. I've just checked out your branch and I've created an item in the personal space, but I can see it in the "Activity Stream" widget using the `admin` user. Am I doing something wrong?

gbbr: Please consider this solution instead: https://github.com/eozan/superdesk-client-core/pull/3

scripts/superdesk-monitoring/aggregate.js
```diff
@@ -12,9 +12,9 @@
     'use strict';
 
     AggregateCtrl.$inject = ['$scope', 'api', 'desks', 'workspaces', 'preferencesService', 'storage',
-                             'gettext', 'multi', 'config'];
+                             'gettext', 'multi', 'config', '$timeout'];
     function AggregateCtrl($scope, api, desks, workspaces, preferencesService, storage,
-            gettext, multi, config) {
+            gettext, multi, config, $timeout) {
```
gbbr: Forgot this one, please remove it. It's no longer needed.

gbbr:
Please remove the dependency and then feel free to merge :-) Thanks for the patience.

mugurrus:
:+1:

gbbr:
👍

petrjasek:
seems like it broke some test, or am I missing something @eozan ?

PR #718, opened 2016-07-13
ui/fix: hide headline if disabled in profile (fixes SD-4865)

gbbr:
This is a requirement from SD-4865 but as it seems in the code the headline was never be meant to be disabled.


mdhaman:
:+1:

sjunaid:
:+1:

PR #719, opened 2016-07-13
ui/fix: added sd-width property to editor (fixes SD-4867)

gbbr:
This will allow users to interact with the layout of the editor, which
might sometime become misaligned if the configuration is an odd one. For
example, adding a component such as an abstract (which has a character
counter) with a narrow width such as `'quarter'` will definitely break
layout.

The default for all will be `'full'`.


mugurrus:
:+1:

ioanpocol:
:+1:

PR #720, opened 2016-07-13
Confirm delete added for personal workspace

Aca-jov:
SDNTB-248

### DIFF
--- a/scripts/superdesk-archive/module.js
+++ b/scripts/superdesk-archive/module.js
@@ -540,12 +540,19 @@
                     label: gettext('Spike Item'),
                     icon: 'trash',
                     monitor: true,
-                    controller: ['spike', 'data', '$rootScope', function spikeActivity(spike, data, $rootScope) {
-                        return spike.spike(data.item).then(function(item) {
-                            $rootScope.$broadcast('item:spike');
-                            return item;
-                        });
-                    }],
+                    controller: ['spike', 'data', '$rootScope', 'modal', '$location', '$q',
+                        function spikeActivity(spike, data, $rootScope, modal, $location, $q) {
+                            var txt = gettext('Do you want to delete the item permanently?');
+                            var isPersonal = $location.path() === '/workspace/personal';
+
+                            return $q.when(isPersonal ? modal.confirm(txt) : 0)
+                                .then(function() {
+                                    return spike.spike(data.item).then(function(item) {
+                                        $rootScope.$broadcast('item:spike');
+                                        return item;
+                                    });
+                                });
+                        }],
                     filters: [{action: 'list', type: 'archive'}],
                     action: 'spike',
                     keyboardShortcut: 'ctrl+x',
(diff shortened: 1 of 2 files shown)


pavlovicnemanja:
:+1:

scripts/superdesk-archive/module.js
```diff
@@ -540,12 +540,23 @@
                     label: gettext('Spike Item'),
                     icon: 'trash',
                     monitor: true,
-                    controller: ['spike', 'data', '$rootScope', function spikeActivity(spike, data, $rootScope) {
-                        return spike.spike(data.item).then(function(item) {
-                            $rootScope.$broadcast('item:spike');
-                            return item;
-                        });
-                    }],
+                    controller: ['spike', 'data', '$rootScope', 'modal', '$location',
```
gbbr: Please simplify this as:
```js
controller: ['spike', 'data', '$rootScope', 'modal', '$location', '$q',
    function spikeActivity(spike, data, $rootScope, modal, $location, $q) {
        var txt = gettext('Do you want to delete the item permanently? ');
        var isPersonal = $location.path() === '/workspace/personal';
        return $q.when(isPersonal ? modal.confirm(txt) : 0)
            .then(function() {
                return spike.spike(data.item).then(function(item) {
                    $rootScope.$broadcast('item:spike');
                    return item;
                });
            });
    }],
```
It will contain less repetition and will be easier to follow imo.

scripts/superdesk-search/search.js
```diff
@@ -2248,9 +2248,18 @@
          * Multiple item spike
          */
         this.spikeItems = function() {
-            spike.spikeMultiple(multi.getItems());
-            $rootScope.$broadcast('item:spike');
-            multi.reset();
+            if ($location.path() === '/workspace/personal') {
```
gbbr: You could do the same here:
```js
var txt = gettext('Do you want to delete the item permanently? ');
var isPersonal = $location.path() === '/workspace/personal';
return $q.when(isPersonal ? modal.confirm(txt) : 0)
    .then(function() {
        spike.spikeMultiple(multi.getItems());
        $rootScope.$broadcast('item:spike');
        multi.reset();
    });
```

scripts/superdesk-search/search.js
```diff
@@ -2248,9 +2248,18 @@
...
+                modal.confirm(gettext('Do you want to delete the item permanently? '))
```
gbbr: Since these are multiple items, the more appropriate text would be _"Do you want to delete these items permanently?"_

Also, I might be wrong, but as far as I know these items can be unspiked, meaning that the action is not permanent as the confirmation dialog says.

Aca-jov: For personal workspace there isn't unspike, so action is permanent. 

For other comments, I will fix it as you suggested ;)

gbbr:
👍 LGTM

PR #725, opened 2016-07-15
ui/fix: make spikeMultiple return a promise

gbbr:
`spike` returns a promise, so users of the API should expect `spikeMultiple` to returns the same.


akintolga:
:+1:

mdhaman:
:+1:

PR #731, opened 2016-07-18
Webpack Build System

gbbr:
This PR adds the Webpack build system and introduces ES6 features to our front-end codebase. If you'd like to read more about how to use ES6 with Angular 1.x, [this](https://medium.com/@daviddentoom/switching-to-es6-with-angular-1-x-is-easy-a08c40c2fc72#.wnnltjvm8) blog article is a basic introduction.

### Plan

- [x] Development Environment
- [x] Karma Unit Tests
- [x] Protractor
- [x] Linters / Hinters
- [x] Travis
- [x] Deploy to [`webpack.dev.superdesk.org`](https://webpack.dev.superdesk.org/)

Some changes to the current workflow can be seen below. These could be fixed in subsequent PRs.

### To do

- Enable and fix one [broken test](https://github.com/superdesk/superdesk-client-core/pull/731/files#diff-07ae57eff5b73ba33e360d1b5ff32fa7R70) in `sdImageCrop.spec.js`.
- Enable and fix [broken assertions](https://github.com/superdesk/superdesk-client-core/pull/731/files#diff-80b9c51c28525a6a8183e94e3920dcc9R161) in `takes_spec.js`.

### Pitfalls

- When changing HTML files, manual browser refresh is necessary for now. Changing `index.html` requires server restart or re-running `grunt template:dev` task.
- Can not use ES6 in test / spec files (yet).
- Bower still present for now.


scripts/superdesk-archive/controllers/baseList.js
```diff
@@ -67,3 +67,6 @@ function BaseListController($scope, $location, superdesk, api, search, desks, pr
         self.fetchItems({source: query});
     };
 }
+
+//TODO(x) This is nasty
+window.BaseListController = BaseListController;
```
petrjasek: can this be handled via import? as a proof of concept? :wink:

gbbr: It definitely can. I could do this in the immediate next PR so we can get
the ball moving. Unless you feel strongly about doing it now, in which case
I don't have a problem with that, but it might postpone things a bit.
Ideally I would keep PRs separate but you decide :)

tasks/options/template.js
```diff
@@ -67,13 +67,17 @@ module.exports = function(grunt) {
             options: data('http://localhost:5000/api', true),
             files: files
         },
+        dev: {
+            options: data('http://localhost:5000/api', true),
+            files: {'index.html': '<%= appDir %>/index.html'}
```
petrjasek: shouldn't this index go into dist folder?

petrjasek:
:+1:

tasks/options/template.js
gbbr: Yes, this is a bit odd based on our setup, but I couldn't figure out a way to set a `base` in WebPack like we did with `grunt-connect`. The only way I can think of is to set the root in the `dist` folder but that would make all of the scripts unavailable because they would become located at `../script`.

There might be a way though but I didn't want to let this stall me from finishing the rest of the work...

vied12:
A new step to the future ! :clap::clap::clap:

PR #734, opened 2016-07-19
fix(templates): fix abbreviations expantion when there is no language

petrjasek:
for empty item language `spellcheck.getAbbreviationsDict` would stop
promise chain which would stop `editor.commitScope`. thus all editor
directives wouldn't work at all.

so instead it will resolve with empty dict.

SD-4934


pavlovicnemanja:
:+1:

sjunaid:
:+1:

mdhaman:
:+1:

PR #742, opened 2016-07-20
fix(desks): fix desk re-editing

petrjasek:
update desk reference with latest etag obtained from server

SD-4700


akintolga:
:+1:

sjunaid:
:+1:

PR #751, opened 2016-07-25
Optional test environment indicator on header

akintolga:
We have Test, UAT and PROD environments deployed locally. Especially the UAT and PROD has identical data so without a visual indicator it is easy to confuse PROD as UAT which happened few times.

So an optional boolean parameter `isTestEnvironment` is created to indicate a deployment is a "Test" instance in the top banner. If the parameter is not used then there'll be no change in UI.

### DIFF
--- a/scripts/superdesk/menu/menu.js
+++ b/scripts/superdesk/menu/menu.js
@@ -43,8 +43,9 @@
             'userNotifications',
             'asset',
             'privileges',
+            'config',
             'lodash',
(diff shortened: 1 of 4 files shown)


petrjasek:
:+1: but maybe this could be some string config rather than boolean so you could set it to `TEST`, `UAT`, etc.

mdhaman:
:+1:

marwoodandrew:
:+1:

PR #752, opened 2016-07-25
ui/fix: remove references to BaseListController via window object

gbbr:
Addresses https://github.com/superdesk/superdesk-client-core/pull/731#discussion_r71855622 /cc @petrjasek 
<sub>(_might be best viewed with [split view](https://github.com/superdesk/superdesk-client-core/pull/752/files?diff=split)_)</sub>


petrjasek:
:+1:

PR #755, opened 2016-07-25
Make better contained modules and clean index.js

gbbr:
Made `index.js` more compact and better outlined and contained modules.


petrjasek:
:+1:

PR #756, opened 2016-07-25
SDPA-502 If a subservice is required then the parent service should automatically be set

mugurrus:
Sets the anpa category corresponding to the required subservice:
if a subservice field (defined in vocabularies) was declared as
required in a content profilebthen make sure that the corresponding
anpa category was added to the anpa_category field in the newly
created item. Without this value the anpa category field is not
displayed.

### DIFF
--- a/scripts/superdesk-authoring/authoring.spec.js
+++ b/scripts/superdesk-authoring/authoring.spec.js
@@ -1,0 +1,51 @@
+fdescribe('authoring', function() {
+    'use strict';
+
+    beforeEach(module('superdesk.vocabularies'));
+    beforeEach(module('superdesk.archive'));
+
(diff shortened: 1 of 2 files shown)


petrjasek:
:+1:

gbbr:
The logic seems a bit complex, so a test would be nice if you have the time :-)

gbbr:
Thanks for adding the test. Looks good 👍

scripts/superdesk-authoring/authoring.spec.js
```diff
@@ -0,0 +1,58 @@
+fdescribe('authoring', function() {
```
petrjasek: seems like forgotten `fdescribe`

PR #758, opened 2016-07-26
fix(templates): let templates handle target_subscriber

petrjasek:
works best with https://github.com/superdesk/superdesk-core/pull/492

SD-4875


mugurrus:
:+1:

akintolga:
:+1:

PR #759, opened 2016-07-26
Remove duplication of index.html and index.less (fixes SD-5119)

gbbr:
This will cause the client-core to always use its own `index.html` file, even when embedded into the main `superdesk` repo, allowing the removal of `index.html` and `index.less` in the main repo.

Corresponds to https://github.com/superdesk/superdesk/pull/2189

### DIFF
--- a/tasks/options/template.js
+++ b/tasks/options/template.js
@@ -1,3 +1,4 @@
+var path = require('path');
 
 module.exports = function(grunt) {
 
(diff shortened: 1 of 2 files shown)


petrjasek:
we do some client specific configuration in index.html, so imo it can't be removed without introducing other means of configuration

gbbr:
Yes, I was just about to close this because of that. We need a different solution.

PR #761, opened 2016-07-26
SD-4992 Editor: prevent chrome to add extra markup with style

vied12:
It appears that Chrome adds <span> with style in contenteditable
element if the style is not explicitly set in CSS.

SD-4992


gbbr:
👍

petrjasek:
does chrome stop to add any span altogether?

vied12:
@petrjasek yes it does, from my observations.
Please give a try on https://cceditorpreventextramarkup.dev.superdesk.org

petrjasek:
:+1:

PR #764, opened 2016-07-26
fix(profiles): handle defaults for cvs

petrjasek:
- add select box instead of string input so you can pick some of those structured values.
- add min/max length validation
- remove negative numbers from order/min/max inputs

SD-5032 SD-5020 SD-5019 SD-5035


akintolga:
:+1:

mdhaman:
:+1:

PR #767, opened 2016-07-27
[SD-4956] - Formatted story preview

akintolga:
With this change users can preview the published format of a story:
- An optional preview formatted button to display the story the same format that it would be sent to subscribers
- To enable the functionality a config change required to define which formats are allowed for the user


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/497

petrjasek:
:+1:

mdhaman:
:+1:

PR #768, opened 2016-07-27
Separate config from index.html (SD-5127)

gbbr:
- [x] Separate config data from codebase into distinct file.
- [x] Allow default route route to be configurable in the build (for NTB)
- [x] Karma unit tests
- [x] Transfer `tansa` spellchecker functionality into client core (for NTB) ([from here](https://github.com/superdesk/superdesk/blob/ntb/client/app/index.html#L170-L192))
- [x] Transfer embed provider Vidible feature to client core (for PA-POC) ([from here](https://github.com/superdesk/superdesk/blob/pa-poc/client/app/index.html#L100-L113))
- [x] Find a solution to translations
- [x] Move docs to client core
- [x] Test build and dev on all branches: master, NTB, PA-POC, AAP

This PR also disables the failing spec which is failing all PR builds, so we can see green until it's fixed.


superdesk.config.js
```diff
@@ -0,0 +1,61 @@
+/**
+ *
+ * This is the default configuration file for the Superdesk application.
```
petrjasek: imo would be better to by default get that file from `cwd`, if it's not there get it from core. so we don't have to use that --config flag all the time when working on superdesk repo

gbbr: I like that idea! Will do so :)

scripts/superdesk/editor2/add-embed.ctrl.js
```diff
@@ -71,9 +71,13 @@ EMBED_PROVIDERS, $scope, editor, config, $injector) {
                         name: EMBED_PROVIDERS.youtube
                     }
                 ];
-                // prepend with custom handlers from config
-                if (config.editorEmbedCodeParsers) {
-                    knownProviders = $injector.invoke(config.editorEmbedCodeParsers).concat(knownProviders);
```
gbbr: /cc @vied12 please review this part - it involves editor and pa-poc

scripts/superdesk-tansa/index.js
```diff
@@ -0,0 +1,47 @@
+/*
```
gbbr: /cc @ioanpocol @Aca-jov

scripts/superdesk/editor2/add-embed.ctrl.js
vied12: I need then to move https://github.com/superdesk/superdesk/blob/pa-poc/server/pa/vidible.py to `superdesk-core`

scripts/index.js
```diff
@@ -56,6 +60,12 @@ import 'superdesk-settings';
 import 'superdesk-ingest';
 import 'superdesk-search-providers';
 
-// App
-import 'app/scripts/translations';
-import 'app/main';
+// load spellcheckers
+if (appConfig.spellcheckers.indexOf('tansa') > -1) {
```
gbbr: Comment from @Aca-jov - use `features.tansaProofing` instead of `config.spellcheckers`

superdesk.config.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * This is the default configuration file for the Superdesk application. By default,
+ * the app will use the file with the name "superdesk.config.js" found in the current
+ * working directory, but other files may also be specified using relative paths with
+ * the SUPERDESK_CONFIG environment variable or the grunt --config flag.
+ */
+
+module.exports = function(grunt) {
```
petrjasek: not sure if it needs grunt here, can it just return a dict?
imo ideally this file would be loaded into index.html rather then during build, so you can change something there without the need to rebuild, so we can build the app once and use it with different configurations.

scripts/index.js
Aca-jov: `features.useTansaProofing` to be precise ;)

superdesk.config.js
gbbr: I agree that a simple JSON object would be better here. You've said during one of our discussions that you'd like this to be a function so we can make use of environment variables. But a function gives too much freedom to do "hacks" to a third party. I will make this into a simple JSON object and name it `superdesk.config.json`.

As for using it live, I think it is better as it is because when we disable a feature, that specific module will not be included into the bundle so when you create a build you will create it containing only the files that are actually used, no extra luggage. This will move us closer to having a core and other pluggable components in my opinion.

scripts/superdesk/editor2/add-embed.ctrl.js
vied12: I just merged https://github.com/superdesk/superdesk-core/pull/508, so you can now add `apps.vidible` to the installed app in this branch

gbbr: :+1: thanks Edouard

petrjasek:
:+1:

vied12:
:+1:

PR #773, opened 2016-07-28
fix(content_profile): use meta-* directives for defaults

petrjasek:
select elem doesn't handle lists surprisingly..

SD-5035

### DIFF
--- a/scripts/superdesk-workspace/content/content.js
+++ b/scripts/superdesk-workspace/content/content.js
@@ -55,8 +55,7 @@
     });
 
     var CV_ALIAS = Object.freeze({
-        locators: 'place',
-        categories: 'anpa_category'
+        locators: 'place'
     });
 
     angular.module('superdesk.workspace.content', [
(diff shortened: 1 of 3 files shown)


scripts/superdesk-workspace/content/content.js
```diff
@@ -619,6 +622,19 @@
...
+
+                /**
+                 * Copy value from `id` key into `default` within schema
+                 *
+                 * it's a workaround for meta-* directives, those have specific logic based on `id`
+                 *
+                 * @param {String} id
+                 */
+                scope.setdefault = function(id) {
+                    console.log('default', id, scope.model.schema[id][id]);
```
akintolga: console.log is left here...

akintolga:
:+1:

scripts/superdesk-workspace/content/content.js
petrjasek: ups

PR #776, opened 2016-07-29
feat(archive): let user pick desk/stage when unspiking

petrjasek:
SD-4706

### DIFF
--- a/scripts/superdesk-archive/module.js
+++ b/scripts/superdesk-archive/module.js
@@ -97,8 +97,8 @@ import BaseListController from './controllers/baseList';
         //ToDo: Track upcoming item:move event as well for updateCount.
     }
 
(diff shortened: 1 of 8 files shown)


gbbr:
LGTM

scripts/superdesk-archive/module.js
```diff
@@ -140,7 +140,24 @@ import BaseListController from './controllers/baseList';
          * @param {Object} item
          */
         this.unspike = function(item) {
```
gbbr: This is perfect as is, but just for fun, it could be:
```js
this.unspike = item => getUnspikeDestination().then(config => unspike(item, config));
```

petrjasek: that's too funky for me. yet :wink:

akintolga:
:+1:

mdhaman:
:+1:

PR #777, opened 2016-07-29
fix(editor): fix preformatted spellcheck highlights alignment

petrjasek:
there was font/line size set for spans within editor to match
its defined size which avoids inserting those styles by chrome.
but size was defined on `p` tag and not on `pre` tag so then
it would be different for preformatted text and spans used
for spellcheck highlights which would have different size than the other text.

SD-5155


gbbr:
LGTM

akintolga:
:+1:

PR #782, opened 2016-08-01
 SD-4992 Editor: prevent chrome from adding an empty <div>

vied12:
Prevent chrome from adding an empty `<div>` beginning of a new item.

This will fix the problem for every new created item.

This idea comes from the [medium-editor tests](https://github.com/yabwe/medium-editor/blob/98ebd576bc8d29e2382d5941f92932770da60730/spec/content.spec.js#L283)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ templates-cache.js
 test-server/env
 test-server/__pycache__
 e2e-test-results
+.env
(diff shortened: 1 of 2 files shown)


petrjasek:
:+1:

sjunaid:
:+1:

PR #783, opened 2016-08-01
fix(content_profiles): add warning for maxlength=0

petrjasek:
this would make the field useless, so let user know this does't work.

SD-5020


sjunaid:
:+1:

PR #786, opened 2016-08-02
[SD-4907] - Differentiate send and publish buttons

akintolga:
- Publish and Send-To panels are separated, user clicks the header to activate the panel
- Initially Send-To panel is shown then when changed it is remembered throughout the session
- Fetch, Fetch-To panels behave like Send-To so no change required for that section
- Multi-send action displays only the send to panel

@fritzSF Would you mind please pulling and playing these changes to let me know what you think?

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -2009,11 +2009,11 @@ import 'angular-history/history.js';
     SendItem.$inject = ['$q', 'api', 'desks', 'notify', 'authoringWorkspace',
         'superdeskFlags', '$location', 'macros', '$rootScope',
         'authoring', 'send', 'editor', 'confirm', 'archiveService',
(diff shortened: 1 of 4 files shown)


petrjasek:
:+1:

fritzSF:
@akintolga How about trying to change the color of the publish button as a first step? :) Maybe this will be enough. This is adding one extra click for publishing which is not ideal. Another option (or addition to the first one) would be to add a prompt dialog for publishing (with the option to disable the prompt).

akintolga:
@fritzSF I think the main problem is the close proximity of send and publish buttons. Even if we change the colors they will still click the publish button mistakenly. If we add a prompt then this will add a extra step to the deliberate publish and I am not sure the users will like it. I think the buttons need to be separated.

akintolga:
I am closing this PR for the time being. I will revisit this when "Send & Publish" story [SD-3567](https://dev.sourcefabric.org/browse/SD-5367) is underway...

PR #787, opened 2016-08-02
fix(search_providers): use allowed api to filter out search providers

petrjasek:
and in case there is no search provider type configured on server
disable add new search provider button

SD-5118


akintolga:
:+1:

PR #790, opened 2016-08-02
Monitoring widget not updating after item send fixed.

eozan:
SD-4847

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -2415,6 +2415,7 @@ import 'angular-history/history.js';
                         })
                         .then(function(value) {
                             notify.success(gettext('Item sent.'));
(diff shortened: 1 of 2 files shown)


eozan:
This includes previously discarded solution from https://github.com/superdesk/superdesk-client-core/pull/710 as replacing solution does not work. This PR is not ideal, but in current scope it should be sufficient.

sjunaid:
Hi @eozan, however LGTM, but did a quick testing, appears that if its first time on workspace screen then item sent and widget reloads, but then trying another send not reloads widget, i also not figured out the resolution, just wanted to let you know if you can double check.

petrjasek:
lets not do this again, it reloads even if there should be no reload for given group, thus creating extra load, and it needs to wait for certain time before reloading, otherwise it will get same old data from elastic. so imo should be handled within specific group using existing update scheduling

PR #792, opened 2016-08-03
fix(authoring): fix autosaving not working

petrjasek:
it was due to wrong `_locked` param handling, it should indicate
that item is locked by current user in current session, thus editable.

SD-5197


mdhaman:
@petrjasek couple months ago `_locked` handling was changed and following are the PRs related to it.
 - https://github.com/superdesk/superdesk-client-core/pull/280
 - https://github.com/superdesk/superdesk-client-core/pull/358
 - https://github.com/superdesk/superdesk-client-core/pull/394

I am not sure that the BUG was introduced because of the above PRs. But, can you please verify the following scenarios for `autosave`:

- Create a story, add content but do not save. Minimize the story, or double click on another story. Same for existing story.
- With above conditions, refresh the browser.
- `Reopen` the minimized not-saved story.
- On the existing story make some changes and do not save the changes. Open the version history tab and the user should be able to navigate all versions without loosing `autosave` version.

Plus we are seeing some issues with `undo` and `redo` operations. Not sure if it is related to `autosave` or recent `history` changes. It looks like the model is not updated after `undo/redo` operations. When you save the item, it is different from the what you see it on the editor.

petrjasek:
@mdhaman minimizing and refreshing seems to work ok, but this versioning tab only works after refresh, but once you do some changes and open it you can't go to older versions before you hit save, but that seems to be the case now as well. not related to locking, it's disabled when scope is dirty.

mdhaman:
:+1:  LGTM

petrjasek:
@mdhaman added a fix here for this undo/redo issue

mdhaman:
@petrjasek still it looses the content on undo.

petrjasek:
@mdhaman can you check once more pls? should work now

mdhaman:
I have done changes https://github.com/petrjasek/superdesk-client-core/pull/1 which fixes couple of scenarios.

- Setting the etag to original etag if the undo goes beyond the last saved version.
- Setting autosave = null in case of there  no changes to the item.

We still have a problem when the user start with a blank item, then adds slugline, headline, abstract and body_html and save the changes. Now, perform the undo operation till the populated fields are all gone or one of the fields is gone. Then save the changes you will see that the changes are not saved. I think history needs to compare the `OrigItem` object and the set the field to empty values in the current `item` object if the field does not exist the current history.

In terms of user experience, the undo operation will undo other fields that the user does not have focus on and the user won't notice the changes. imo, we should move angular-history changes to separate branch for further testing and POs approval for the change in behaviour.

scripts/superdesk-authoring/authoring.js
```diff
@@ -236,13 +235,23 @@ import 'angular-history/history.js';
             } else {
                 return api.find('archive', _id, {embedded: {lock_user: 1}})
                 .then(function _lock(item) {
-                    if (!read_only && lock.isLocked(item)) {  // Check if item is still editable
-                        item._locked = true;
+                    if (read_only) {
+                        item._locked = lock.isLockedInCurrentSession(item);
```
naomiaro: 👍  for better naming of this function

scripts/superdesk-authoring/authoring.js
```diff
@@ -236,13 +235,23 @@ import 'angular-history/history.js';
...
                         item._editable = false;
                         return $q.when(item);
-                    } else {
-                        item._editable = !read_only;
-                        return lock.lock(item);
+                    } else { // we want to lock
+                        if (lock.isLocked(item)) { // is locked by someone else
```
naomiaro: actually still just saying isLocked, but not have it include the current session is too confusing for later changes in the code I think.

mdhaman:
:+1:

scripts/superdesk-authoring/authoring.js
petrjasek: yep, this needs more work, but don't want to change it now just before next release

PR #795, opened 2016-08-04
SD-5178 : Change in dateline value doesn't trigger save

naomiaro:


akintolga:
:+1:

mdhaman:
:+1: @naomiaro User profile also uses Dateline. Can please you verify if it is still working properly?

naomiaro:
@mdhaman yep user profile is fine, it was using totally different code from this.

naomiaro:
Looks like this still needs something else, save is only appearing sometimes.

naomiaro:
The problems have to do with autosaving https://github.com/superdesk/superdesk-client-core/pull/792 this should fix it I think.

PR #796, opened 2016-08-04
fix(dictionaries): use inheritance to avoid sending all dict

petrjasek:
send only changes again. but send everything for abbreviations dict.

SD-4986


akintolga:
:+1:

mdhaman:
:+1:

PR #799, opened 2016-08-05
[SD-4839] - Panel width change shouldn't affect other dropdowns

akintolga:
- My earlier PR https://github.com/superdesk/superdesk-client-core/pull/781 affected all dropdowns so this will target only the one in subscriber settings pop-up


petrjasek:
:+1:

mdhaman:
:+1:

PR #804, opened 2016-08-05
Arrow navigation fixed for ntb custom monitoring widget

Aca-jov:
SDNTB-238 - fixed scrolling in monitoring widget
SD-4695

### DIFF
--- a/scripts/superdesk-desks/desks.js
+++ b/scripts/superdesk-desks/desks.js
@@ -230,15 +230,16 @@
 
                 var container = elem[0];
                 var offsetY = 0;
+                var itemHeight = 0;
                 elem.bind('scroll', function() {
                     scope.$apply(function() {
                         if (container.scrollTop + container.offsetHeight >= container.scrollHeight - 3) {
-                            container.scrollTop = container.scrollTop - 3;
+                            container.scrollTop -= 3;
                             scope.fetchNext();
                         }
                         if (container.scrollTop <= 2) {
                             offsetY = 2 - container.scrollTop;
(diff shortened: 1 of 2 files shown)


scripts/superdesk-desks/desks.js
```diff
@@ -280,31 +279,26 @@
...
                     if (!scope.fetching && scope.page > 2) {
                         scope.fetching = true;
-                        scope.page = scope.page - 1;
-                        if (scope.page > 2) {
-                            criteria.source.from = (scope.page - 3) * criteria.source.size;
-                        } else {
-                            criteria.source.from = 0;
-                        }
+                        scope.page -= 1;
+                        criteria.source.from = scope.page > 3 ? (scope.page - 3) * criteria.source.size : 0;
```
gbbr: I think it would be very helpful to add an inline comment and describe what happens here, when navigating backwards from page numbers > 3.

gbbr:
It's a bit hard to review as I'm not sure what this is fixing. I suspect the issue ID in your description could be wrong. Can you please check and potentially add the correct issue that this addresses? The one you linked is [SD-4695](https://dev.sourcefabric.org/browse/SD-4695)

Aca-jov:
I updated description.

scripts/superdesk-desks/desks.js
Aca-jov: Well, I'm not using this `criteria.source.from`, I suppose it's used somewhere else, I just changed code to be better written, that's all.

gbbr:
Very nice, thank you! LGTM

pavlovicnemanja:
:+1

PR #806, opened 2016-08-08
fix(headline) - Headline style fix while editing

akintolga:
- When a new story is edited, headline style doesn't get applied due to a p tag
- If sms message is enabled, it had the markup from headline


petrjasek:
:+1:

sjunaid:
:+1:

PR #807, opened 2016-08-08
[SD-4988]- When a user has them scrolled, views should not change the…

sjunaid:
…ir scroll position on update events

- This change is about maintaining the user view, in case of items added/removed in the current scope of items, and provides the refresh button instead when more  updates comes in, i-e, when items added or removed from list as a result of any action/activity; provided that if scroll position is not on top of list or user is currently previewing any item.
- Content updates, locking/ unlocking remain unaffected, and reflects the change on item in current view.
- Hitting refresh button or manually scroll to top of list forces the list to refresh and scroll position initialized to its top positions.
- Applies to monitoring (groups & single view), highlights, spiked and search view
- covers SD-5253

### DIFF
--- a/scripts/superdesk-items-common/styles/media-archive.less
+++ b/scripts/superdesk-items-common/styles/media-archive.less
@@ -928,7 +928,6 @@
     }
 }
 
-
 /*modal window*/
 @modalMaxImageWidth : 379px;
 @modalMaxImageHeight : 250px;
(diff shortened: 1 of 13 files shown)


petrjasek:
there is some code duplicated, like this items merging and canShowRefresh logic, would be good to have it in one place only, with some tests for it ideally.

akintolga:
Pinging @petrjasek

petrjasek:
:+1:

akintolga:
:+1:

mdhaman:
:+1:

PR #809, opened 2016-08-08
[SD-5139]: Move from Bower to NPM

gbbr:
The purpose of this task is to further simplify deployment and dependency management by removing Bower from the build process.

### Changes in this PR: 

- Removes `bower` from build process and replaces it with `npm`.
- Updates `lodash` to latest version and replaces deprecated methods with new ones.
- Tests now allow ES6. Some changes had to be made to acommodate this, such as replacing `module` calls with `window.module` calls as per docs.
- Reduces Travis build failures by using `npm@latest` and introducing retry attempts on failed `npm install`'s
- Improved Protractor debug output (now shows failing suite _and_ test name)
- Removed function (`(function() { ... }())`) encapsulation on some files, as this is no longer needed with Webpack. **Note for reviewers**: This was a minor change but the diffs look terrible on `world-clock.js` and `activity.js`
- Removed use of `angular.bind` and polyfills in `menu/notifications/notifications.js` and replaced with arrow functions for the same effect.
- Embedded `langmap` package into codebase because it was not ES6 compatible as-is.


vied12:
:+1: Well done !

pavlovicnemanja:
:+1: Great Job! :)

petrjasek:
would it be possible define that `module` global for specs?
otherwise :+1:

gbbr:
I've not found a way to do so. All the Google searches I've made regarding ES6 and testing were using it in this form: `window.module`

PR #811, opened 2016-08-08
Sd 4983: Image metadata validation upon save and publish

thnkloud9:
1. Added media metadata field validation to authoring module when editing media directly from article edit.
2. Made media metadata required fields configurable via superdesk.config.js and mocks.js


vied12:
:+1:

PR #818, opened 2016-08-09
SD-4564: remove "add to package" for items already added to highlight package

mscam:
In order to remove the "add to package" menu item for already added packages I was forced to use _PackagesService_ to store package group items, as it was not possible to access menu controller _$scope.list_ variable from activity's _additionalCondition_, I also added a test for this issue, that is passing successfully. If you see some test failing, it's related to a different issue (selection of multiple items, that is failing also on upstream/master on my installation)


scripts/superdesk-packaging/packaging.js
```diff
@@ -764,11 +791,11 @@
             .activity('addtopackage', {
                 label: gettext('Add to current'),
                 priority: 5,
-                dropdown: ['item', 'className', 'authoringWorkspace', 'packages', 'api', '$rootScope',
-                function(item, className, authoringWorkspace, packages, api, $rootScope) {
+                dropdown: ['item', 'className', 'authoringWorkspace', 'packages', 'api', 'translatedLabel', '$rootScope',
+                function(item, className, authoringWorkspace, packages, api, translatedLabel, $rootScope) {
```
vied12: Do you really need `translatedLabel` here ?

vied12:
:+1: and welcome to the contributors list !

scripts/superdesk-packaging/packaging.js
```diff
@@ -12,11 +12,12 @@
...
     'use strict';
 
-    PackagesService.$inject = ['api', '$q', 'archiveService', 'lock', 'autosave', 'authoring', 'desks'];
-    function PackagesService(api, $q, archiveService, lock, autosave, authoring, desks) {
+    PackagesService.$inject = ['api', '$q', 'archiveService', 'lock', 'autosave', 'authoring', 'authoringWorkspace', 'desks', '$rootScope'];
+    function PackagesService(api, $q, archiveService, lock, autosave, authoring, authoringWorkspace, desks, $rootScope) {
         var self = this;
 
         this.groupList = ['main', 'story', 'sidebars', 'fact box'];
+        this.packageGroupItems = {};
```
petrjasek: imo it would be better to somehow reset this on that directive `$destroy` event or when it's opening, it could have old data, like if you would add an item, go somewhere else, somebody would remove it, and you would start editing that highlight again, it would return true that an item is added using this imo

scripts/superdesk-packaging/packaging.js
mscam: No I don't need it. Sorry I was missing that. I can remove it from there.

scripts/superdesk-packaging/packaging.js
mscam: Ok, let me make the changes. I also have to fix the test as well, then I'll send you back the updated pull request for a review

spec/helpers/monitoring.js
```diff
@@ -268,7 +268,7 @@ function Monitoring() {
 
     this.selectGivenItem = function(item) {
         var itemTypeIcon = item.element(by.css('.type-icon'));
-        browser.actions().mouseMove(itemTypeIcon, {x: -50, y: -50}).mouseMove(itemTypeIcon).perform();
+        browser.actions().mouseMove(itemTypeIcon, {x: -45, y: -45}).mouseMove(itemTypeIcon).perform();
```
mscam: @petrjasek @vied12  I was forced to reduce the value of coordinates passed to _mouseMove_. With the previous value, the test was failing as the browser click action was targeting a different element (highlight main menu instead of item checkbox)

mscam:
For some reasons mouseover on checkboxes is not working well after testing multiple package operations in a single spec. I have cleaned the packaging specs without handling browser size issues. We now have separate specs for packaging features, instead of having only one generic "perform package operations"

petrjasek:
:+1:

PR #820, opened 2016-08-10
fix(workspace): fix uploaded images are not visible

petrjasek:
use `original_creator` field for filtering items in draft
state instead of `task.user` which is only set on next save

SD-5217


akintolga:
:+1:

vied12:
:+1:

PR #822, opened 2016-08-10
feat(multiselect): SD-5114 Added ability to select items with shortcut

pavlovicnemanja:
SD-5114 - Select multiple items with fewest actions

### DIFF
--- a/scripts/superdesk-search/react.js
+++ b/scripts/superdesk-search/react.js
@@ -1574,13 +1574,13 @@ angular.module('superdesk.search.react', [
                     },
 
                     select: function(item, event) {
-                        if (event.shiftKey) {
+                        if (event && event.shiftKey) {
                             return this.selectMultipleItems(item);
                         }
 
                         this.setSelectedItem(item);
 
-                        if (event.ctrlKey) {
+                        if (event && event.ctrlKey) {
                             return this.selectItem(item);
                         }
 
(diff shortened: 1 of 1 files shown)


scripts/superdesk-search/react.js
```diff
@@ -1551,17 +1551,37 @@
                         },
 
                         multiSelect: function(item, selected) {
```
akintolga: Would you please rename the parameter `item` to `items` please as now we are sending a list I guess...

scripts/superdesk-search/react.js
```diff
@@ -1576,6 +1588,35 @@ import classNames from 'classnames';
...
+                            }
+
+                            for (var i = positionStart; i <= positionEnd; i++) {
+                                var item = this.state.itemsById[this.state.itemsList[i]];
+                                if (isCheckAllowed(item)) {
+                                    selectedItems.push(item);
+                                }
+                            }
+
+                            multi.reset();
```
akintolga: If user selects the first item by single select then selects items 3,4,5 with multiselect, will this line clear the previously selected first item?

akintolga:
:+1:

scripts/superdesk-search/react.js
pavlovicnemanja: It works same as in ubuntu, it will select from last item you picked. For example, with ctrl you select item 1, 2 and 3, then press shift and select 6th item. Selection will be from 3 to 6 (as I say, same as in windows/ubuntu file system)

akintolga: :+1:

petrjasek:
:+1:

mdhaman:
:+1:

PR #824, opened 2016-08-11
Validation is done before tansa check for ntb

Aca-jov:
SDNTB-256

### DIFF
--- a/scripts/superdesk-authoring/authoring.js
+++ b/scripts/superdesk-authoring/authoring.js
@@ -384,6 +384,14 @@ import 'angular-history/history.js';
             });
         };
 
+        this.validateBeforeTansa = function (orig, diff) {
+            diff = extendItem({}, diff);
+
+            this.cleanUpdatesBeforePublishing(orig, diff);
+
+            return api.save('validate', {'act': 'publish', 'type': orig.type, 'validate': diff});
+        };
+
         this.saveWorkConfirmation = function saveWorkAuthoring(orig, diff, isDirty, message) {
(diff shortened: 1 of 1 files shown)


pavlovicnemanja:
:+1:

scripts/superdesk-authoring/authoring.js
```diff
@@ -384,6 +384,17 @@ import 'angular-history/history.js';
             });
         };
 
+        this.validateBeforeTansa = function (orig, diff) {
+            diff = extendItem({}, diff);
+
+            this.cleanUpdatesBeforePublishing(orig, diff);
+
+            return api.save('validate', {'act': 'publish', 'type': orig.type, 'validate': diff})
+            .then(function(result) {
```
gbbr: This `.then` statement is redundant, you can remove it. The same result will be returned.

scripts/superdesk-authoring/authoring.js
```diff
@@ -384,6 +384,17 @@ import 'angular-history/history.js';
             });
         };
 
+        this.validateBeforeTansa = function (orig, diff) {
```
gbbr: This function, on its own, seems unrelated to tansa. Would it not be better suited to call it simply `validate` ? Correct me if I'm wrong.

gbbr: @Aca-jov replied on Slack: the `validate` API endpoint is apparently tansa specific and we already have a private validate function, so we can keep as is.

gbbr:
LGTM

PR #825, opened 2016-08-11
fix(authoring): fix undo/redo related issues

petrjasek:
there was an issue when reverting after hitting save due to inheritance,
so changed the code to not update origItem with content changes after
saving.

also fixing some components (dateline) not reseting properly on redo
when value becomes null again.

SD-5279 SD-5280 SD-5281

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Fixed
 
 - Fix undo/redo changes not being saved or autosaved
+- Fix undo/redo for new item not reverting to empty body after save
+- Fix undo/redo not working properly with complex fields (dateline, category, subject)
 
 ## [1.0.0-beta1] 2016-04-26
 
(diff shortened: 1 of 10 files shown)


akintolga:
:+1:  I pulled and tested the tickets, this resolves all the tickets. I saw only one problem and this is how it happens:
- In body type '1' & save, '2' & save, '3' & save
- Start performing UNDO: you'll see 123 > 12 > 1 > Empty in the body field which is good. 
- Once you see body is gone back to empty save
- Now close and re-open the story, body value after the last save comes back. 

I think the diff doesn't include the field if there's no value after undo which then fails to update the value in db. This applies to all fields in editor.

mdhaman:
:+1:

petrjasek:
thanks for testing, will merge this and work on that scenario in a new PR

PR #827, opened 2016-08-12
fix (macros, abbreviations) [SD-5289][SD-5291] Fixes for macros and abbreviations.

mdhaman:
- If you apply macros then incorrect _etag get passed while publishing the item.


petrjasek:
:+1:

vied12:
:+1:

PR #830, opened 2016-08-12
fix(authoring): fix reverting to an empty item after save

petrjasek:
populate item with default values when editing starts, so that it
can be reverted to initial state, even if origItem changed after saving.

so this is actually a workaround for js inheritance, ideally we wouldn't
use it then, but this seems to be quite some work due to those `_locked`
and `_editable` and other flags.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
       export CHROME_BIN=`pwd`/chrome-linux/chrome && $CHROME_BIN --version ;
 
 install:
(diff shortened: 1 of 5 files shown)


petrjasek:
this should only work with https://github.com/superdesk/superdesk-core/pull/519

akintolga:
:+1:  I have tested these and it looks good. 

There are two tickets SD-5294 and SD-5302 that we observed, It looks like this PR can address those two tickets as well.

PR #836, opened 2016-08-16
fix(authoring): fix drag and drop image into body

petrjasek:
it was removing all event handlers, also these defined in
drop zone directive. so rather remove only specific handler
defined later in `$render`.

SD-5310


petrjasek:
+ added fix for images after upload not visible from master

thnkloud9:
:thumbsup:

PR #842, opened 2016-08-16
SD-4631: Replace "keywords required" error message with "slugline required"

mscam:
Simple text replacement. No tests required.


petrjasek:
:+1: but pls squash those commits into 1

mdhaman:
:+1:

PR #849, opened 2016-08-17
Removes encapsulation of modules

gbbr:
Removes function encapsulation (`(function() { ... }())`) that was used
in script files, as this is no longer needed in a modular system.
Pollution of the global scope is no longer a concern with the new set-up.

#### Note to reviewers:
The diff looks ugly but all that happened is that the top of each module:
```js
(function() {
    'use strict';
```
and the bottom:
```js
})();
```
was removed and indentation re-aligned.


petrjasek:
:+1:

thnkloud9:
:thumbsup:

PR #850, opened 2016-08-17
SD-5075: Unable to follow up the warning message in Edit crops

thnkloud9:
edit crops no longer allows saving invalid crops already, but this change fixes a bug preventing media saving if metadata fields were not displayed


petrjasek:
:+1:

PR #851, opened 2016-08-17
fix(urls): show warning in console when using unknown resource

petrjasek:
this implies that some module is missing on server, so make it visible


akintolga:
:+1:

mdhaman:
:+1:

PR #853, opened 2016-08-17
SD-4980 : Clean pasted html.

naomiaro:


akintolga:
@naomiaro Although this looks like a straight-forward change, I suggest this change to be merged when there's enough regression testing in place for QA. I know that this is targeted for master not 1.0 but soon there'll be a 1.1 from master and without proper regression this change may have a potential to break few things for us. I don't think that QA is aware of all the scenarios and test cases, I think they will only check if a copy-paste to description field stays italic.

@petrjasek May be we need a new label to indicate regression testing is needed for a PR. This will tell us:
- that if there's no regression test, that PR cannot be merged.
- If there's regression tests in place a clear notice from developer or reviewers that any additional case is needed.

I will inform Andrew Lasky to provide for a set of test cases for QA...

mdhaman:
@naomiaro settings can be changed via config https://github.com/superdesk/superdesk-client-core/blob/master/scripts/superdesk/editor2/editor.js#L761

Examples are https://github.com/superdesk/superdesk-aap/blob/master/client/superdesk.config.js#L11 and 
https://github.com/superdesk/superdesk/blob/pa-poc/client/superdesk.config.js#L26

naomiaro:
Aha, OK @mdhaman I can move this into a config file.

petrjasek:
@akintolga I agree we need such testing, but not sure if QAs are checking these pull requests. probably we should put such info in changelog, what do you think?

akintolga:
@petrjasek That sounds good. Actually regression testing needs to be done even if there's no flag.

naomiaro:
Replaced by https://github.com/superdesk/superdesk-aap/pull/99

PR #855, opened 2016-08-18
[SD-5001] - Separation of parameters, filters and saved panels in global search

akintolga:
- [x] Separate parameters section to its own tab in global search
- [x] Queries with aggregations will run only when the filters panel is opened
- [x] Reduce the query timeout from 3 sec to 300 msec
- [x] Re-arrange controls in parameters tab
- [x] Add Company Codes to the parameters
- [x] Change category and company code from `q` to `params` in query
- [x] Search from parameters will run only when `Search` button is clicked
- [x] Fix e2e tests


akintolga:
@petrjasek This change has also potential to create problems so I'll provide regression test details to QA for this change. Does this require a separate feature branch?

petrjasek:
:+1: and yes feature branch would make sense here

PR #856, opened 2016-08-18
Show the right icons for takes-pack and highlight-pack in related items

naspeh:
[SD-5214](https://dev.sourcefabric.org/browse/SD-5214)


petrjasek:
:+1: good refactoring, but imo would be even better to make it directive, something like:
```
<i sd-file-type-icon data-item="item"></i>
```

naspeh:
@petrjasek, thanks for comment :). I'm still working on this fix further, I will do this derective, just I'm not familiar with Angular and how it is better doing things :).

naspeh:
@petrjasek: It needs one more portion of review

gbbr:
Nice one! LGTM

scripts/superdesk/directives/sdFiletypeIcon.js
```diff
@@ -0,0 +1,25 @@
+export default angular.module('superdesk.directives.filetypeIcon', [])
```
gbbr: Can we add some documentation here to say what this directive does?

scripts/superdesk/itemList/itemList.js
```diff
@@ -5,6 +5,66 @@ var DEFAULT_OPTIONS = {
     sort: [{_updated: 'desc'}]
 };
 
+function extendScope(scope, itemList, itemPinService) {
```
gbbr: Also here some documentation would be very welcome :-)

naspeh:
Yep, I'll add documentation

naspeh:
@petrjasek, @gbbr please review again.

I've changed binding type in this directive from `=` to `@`, seems it is better here.

scripts/superdesk/directives/sdFiletypeIcon.js
```diff
@@ -1,7 +1,16 @@
 export default angular.module('superdesk.directives.filetypeIcon', [])
+/**
+ * sdFiletypeIcon adds proper "filetype-icon-*" class based item.
```
gbbr: Maybe it would be better to say:
```
sdFileTypeIcon adds the "filetype-icon-*" class based on the item type.
```
(imo, not necessary)

gbbr:
LGTM, one small comment, but I'm ok with the current state too 👍

scripts/superdesk/directives/sdFiletypeIcon.js
naspeh: I wanted to write item type, but there other checks like `item.highlight`, etc. But I'll add "the", I'm not perfect in grammar :).

naspeh:
@gbbr: done

gbbr:
Thank you @naspeh

scripts/superdesk/directives/sdFiletypeIcon.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+.directive('sdFiletypeIcon', function() {
+    return {
+        scope: {item: '='},
+        link: function(scope, element, attrs) {
+            scope.$watch('item', function(item) {
+                if (item) {
+                    initIcon(item);
+                }
+            }, true);
```
petrjasek: I don't think it should watch item recursively, that's performance issue

petrjasek: also it doesn't change for an item, so once it's set you can remove the watch:
```
var stopWatch = scope.$watch('item', function(item) {
    if (item) {
        stopWatch();
    ...
```

naspeh: Thanks for the comments, I supposed that I could be wrong here. I'll fix that.

scripts/superdesk/itemList/itemList.js
```diff
@@ -5,6 +5,74 @@ var DEFAULT_OPTIONS = {
...
+
+    scope.pin = function(item) {
+        itemPinService.add(scope.options.pinMode, _.clone(item));
+    };
+
+    scope.unpin = function(item) {
+        itemPinService.remove(item);
+    };
+
+    scope.isPinned = function(item) {
```
gbbr: You don't have to change this now, but it's worth noting for the future that functions like this serve no purpose being on the `scope` and generally the scope should be limited to items that you wish to (two-way) bind to, such as a model or in rare cases functions. If my understanding of the way Angular works is correct, this affects performance seriously and all these scope items become processed during the digest cycle.

This is one of the main reasons people run into performance issues with Angular. A better approach is to create controllers and have these functions there.

It's ok to leave this as is for now, but I'm just saying this as a remark since you were saying that you are learning Angular.

You don't need to take any action.

naspeh:
@petrjasek, @gbbr done

petrjasek:
:+1:

gbbr:
:+1:

naspeh:
I can't merge it! Who can?

PR #858, opened 2016-08-18
fix(editor): Fixed issue when toolbar editor goes over header

pavlovicnemanja:
This issue can't be fixed just with z-index. Because toolbar is set on body it always goes over authoring header.

Also, I tried to move toolbar inside `.main-article` but medium-editor is having some issues with position on `relativeContainer`.

If someone has another idea how to solve this, be free to suggest!

### DIFF
--- a/scripts/superdesk/editor2/editor.js
+++ b/scripts/superdesk/editor2/editor.js
@@ -734,12 +734,14 @@ angular.module('superdesk.editor2', [
     }])
     .directive('sdTextEditorBlockText', ['editor', 'spellcheck', '$timeout', 'superdesk', '$q', 'gettextCatalog', 'config',
     function (editor, spellcheck, $timeout, superdesk, $q, gettextCatalog, config) {
+        var TOP_OFFSET = 134; // header height
+
         var EDITOR_CONFIG = {
             toolbar: {
                 static: true,
                 align: 'left',
                 sticky: true,
-                stickyTopOffset: 134, // header height
+                stickyTopOffset: TOP_OFFSET,
                 updateOnEmptySelection: true
             },
             paste: {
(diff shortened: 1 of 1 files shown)


scripts/superdesk/editor2/editor.js
```diff
@@ -734,12 +734,14 @@ angular.module('superdesk.editor2', [
     }])
     .directive('sdTextEditorBlockText', ['editor', 'spellcheck', '$timeout', 'superdesk', '$q', 'gettextCatalog', 'config',
     function (editor, spellcheck, $timeout, superdesk, $q, gettextCatalog, config) {
+        var TOP_OFFSET = 134; // header height
```
gbbr: Is there any way we can retrieve the height programatically? I'm thinking if we change the CSS or skin in the future, this could cause issues. If that's too difficult or doesn't work, you can leave it as is.

P.S. If you prefer, we can now use `const` instead of `var` for constants.

gbbr:
LGTM besides minor comment / question

scripts/superdesk/editor2/editor.js
```diff
@@ -1033,6 +1035,15 @@ angular.module('superdesk.editor2', [
                         }
                     });
 
+                    // hide toolbar if element is under header
+                    scope.medium.subscribe('positionedToolbar', function(e, elem) {
+                        var toolbar = scope.medium.getExtensionByName('toolbar'),
+                            elemPosition = elem.getBoundingClientRect();
+                        if (toolbar) {
+                            toolbar.toolbar.hidden = elemPosition.top < TOP_OFFSET;
```
petrjasek: what does it do if you set it to `false`?

pavlovicnemanja: It show's toolbar :)

petrjasek: I see, thx

petrjasek:
:+1:

scripts/superdesk/editor2/editor.js
pavlovicnemanja: I will need to get height of three different elements, so I think this is unnecessary use of resources.

PR #861, opened 2016-08-18
SD-4847: Update monitoring items when an item is sent to another stage/desk.

mscam:
* Update monitoring items when an item is sent to another stage/desk.


scripts/superdesk-desks/desks.js
```diff
@@ -216,15 +216,18 @@
                  *
                  * In case it gets called multiple times it will query only once
                  */
-                function scheduleQuery() {
+                function scheduleQuery(delay) {
```
gbbr: You can do:
```js
function scheduleQuery(delay = 5000) {
```
to obtain default parameter values, and remove the `if` statement below.

scripts/superdesk-desks/desks.js
```diff
@@ -172,7 +172,7 @@
                 }
 
                 scope.$watch('filter', queryItems);
-                scope.$on('task:stage', function(event, data) {
+                scope.$on('task:stage', function($event, data) {
```
gbbr: Why this change? The parameter is not used.

gbbr:
Can you please squash all the commits into one?

gbbr:
Many times broadcasting on `$rootScope` is not necessary - do you think this might be one of those cases? I think the server already sends `item:move` via WebSockets, which is automatically broadcasted already. Maybe it's not picked up?

screen shot 2016-08-18 at 17 58 50

scripts/superdesk-desks/desks.js
mscam: Consistency with the other event listeners that are using $event as argument name, no particular reason.

mscam:
I can squash commits with github with the merge button, but I can also rebase my feature branches in future.

mscam:
I am not against broadcasting events to $rootScope, sometimes is useful. But it's also good to discover why that event is not picked up. What do you think? Let's discuss about the WebSockets  issue tomorrow.

gbbr:
Sure, let's do a hangout

scripts/superdesk-desks/desks.js
gbbr: The event listeners that you see using `$event` are using Angular's [`$event`](https://docs.angularjs.org/guide/expression#-event-), which is fine. In this case this is a custom event so `e` or `event` should be better to avoid confusing the two concepts. Just like above where you've used `e` yourself.

scripts/superdesk-archive/module.js
```diff
@@ -91,8 +91,9 @@ function MultiService($rootScope) {
     $rootScope.$on('item:publish', function(e, data) {
         self.remove(data.item);
     });
-
-    //ToDo: Track upcoming item:move event as well for updateCount.
```
gbbr: 👍 nice one

scripts/superdesk-desks/desks.js
mscam: ok great, I will change all the relevant unused argument name respecting that naming convention.

gbbr:
Much better, thank you @mscam - now if we could make it all into one commit, then we can merge. I assume this will give many conflicts when attempting to squash so it might be best to just create a new branch and copy over the changes to get a clean commit.

scripts/superdesk-desks/desks.js
```diff
@@ -172,22 +172,22 @@
...
                 });
 
-                scope.$on('item:move', function($event, data) {
-                    if ((data.to_stage && data.to_stage === scope.stage) ||
-                        (data.from_stage && data.from_stage === scope.stage)) {
-                        scheduleQuery();
+                scope.$on('item:move', function(_e, data) {
+                    if ((data.to_desk && data.from_desk !== data.to_desk) ||
+                        (data.to_stage && data.from_stage !== data.to_stage))  {
+                        scheduleQuery(1000); // smaller delay.
```
petrjasek: it should be a bit more, like 2s, it takes some time to elastic after a change to show the updates

gbbr:
Closing in favour of https://github.com/superdesk/superdesk-client-core/pull/870

scripts/superdesk-desks/desks.js
mscam: Ok, thanks.

PR #865, opened 2016-08-19
SD-5082 : "failed to get lock on item" error when opening read only i…

naomiaro:
…tems in monitoring

### DIFF
--- a/scripts/superdesk-search/react.js
+++ b/scripts/superdesk-search/react.js
@@ -1420,6 +1420,12 @@ angular.module('superdesk.search.react', [
                         }
                     },
 
+                    dbClick: function(event) {
+                        if (!this.props.item.gone) {
+                            this.props.onDbClick(this.props.item);
+                        }
+                    },
+
                     getInitialState: function() {
                         return {hover: false};
                     },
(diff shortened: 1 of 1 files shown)


scripts/superdesk-search/react.js
```diff
@@ -1576,6 +1582,27 @@ angular.module('superdesk.search.react', [
                         }, 500, false);
                     },
 
+                    dbClick: function(item) {
+                        var activities = superdesk.findActivities({action: 'list', type: 'archive'}, item);
+
+                        var canEdit = _.reduce(activities, function(result, value, key) {
```
gbbr: `key` is not used here

gbbr:
LGTM, besides minor unused arg comment

scripts/superdesk-search/react.js
gbbr: No need to change this, but just FYI, it can also be:
```js
var canEdit = _.reduce(activities, (result, value) => result || value._id === 'edit.item', false);
```
Also, your build might fail because of using double quotes `"`

naomiaro: aha yeah I always forget some expected styling, jshint caught this though

naomiaro: right I guess we can use ES6 now,

petrjasek:
:+1:

PR #867, opened 2016-08-19
SD-5078 : Display a proper error message in case a wrong GUID is ente…

naomiaro:
…red in product test section

### DIFF
--- a/scripts/superdesk-products/products.js
+++ b/scripts/superdesk-products/products.js
@@ -215,7 +215,8 @@ function ProductsConfigController ($scope, gettext, notify, api, products, modal
                 }
             });
         }, function(response) {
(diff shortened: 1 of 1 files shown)


gbbr:
This is good. I'm only wondering if all the other stuff in the `response` object is irrelevant to debugging and if `_message` property will always be there. Otherwise, LGTM

naomiaro:
I think it's always there, I can add a check for if message I guess. The message is the only thing that needs to be flashed to the user.

gbbr:
As you think, thanks for explaining 👍

petrjasek:
:+1:

gbbr:
👍

mdhaman:
:+1:

PR #873, opened 2016-08-22
SD-5075: Unable to follow up the warning message in Edit crops and save changes in image item

thnkloud9:
updated authoring to not extend renditions if the original image size is too small for the rendition.
also fixed an issue with incorrectly setting isDirty flag on Edit/Apply crop area


petrjasek:
:+1:

PR #876, opened 2016-08-23
[SD-5359]-(fix) - Global Search -> Blue frame does not appear when an…

sjunaid:
… item is deleted by another user


thnkloud9:
:thumbsup:

petrjasek:
:+1:

PR #878, opened 2016-08-23
fix(templates): when you save item as template, it should be create type

petrjasek:
so there won't be any dropdown

SD-5231


thnkloud9:
:thumbsup:

PR #881, opened 2016-08-23
ui(hotfix): fixes a problem with a regression in webpack-dev-server

gbbr:
There was a regression in the webpack-dev-server that introduced a lot of problems with the proxy, making it impossible for some people to run the app locally. This freezes the version to avoid these issues in the future.


pavlovicnemanja:
:+1:

PR #889, opened 2016-08-24
fix(DeskMentions): Fix the problem where desk mentioned query was called recursively.

mdhaman:


scripts/superdesk-monitoring/monitoring.js
```diff
@@ -749,7 +752,7 @@
                  */
                 scope.acknowledge = function(notification) {
                     deskNotifications.markAsRead(notification, scope.desk);
-                    $timeout(init, 5000);
+                    $timeout(reload(), 5000);
```
petrjasek: I guess it should be only `reload` here, without `()`

mdhaman: yes.

petrjasek:
this PR should fix travis https://github.com/superdesk/superdesk-client-core/pull/891

mdhaman:
thanks @petrjasek

akintolga:
:+1:

petrjasek:
:+1:

PR #895, opened 2016-08-24
[SD-5364] - Hide publish button for highlight packages

akintolga:
- Highlights will be published when they exported as flat file (text)


mdhaman:
:+1:

petrjasek:
:+1:

PR #900, opened 2016-08-26
fix(multiedit): SD-4659 Fixed sms switch in multiedit mode

pavlovicnemanja:
SD-4659 - switching ON or OFF does not work for SMS control in multiedit mode


Aca-jov:
:+1:

petrjasek:
:+1:

PR #909, opened 2016-08-30
SD-4564: An item added to the highlight package is still available to be added to the highlight package

mscam:
This is a small bugfix for highlight package items, that were showing "Add to current"  even if already present in the newly created item.


ioanpocol:
:+1:

PR #920, opened 2016-08-31
fix(monitoring): set max height to 10 if not set

petrjasek:
and detect spike as not limited group

SD-5412 SD-5383


akintolga:
:+1:

sjunaid:
:+1:

PR #887, opened 2016-08-24
[SD-5360]-(fix) - Global Search -> item deleted from top of screen di…

sjunaid:
…sappears instantaneously w/o being grayed out
- included item:deleted for gray out functionality when ingested item removed/deleted

PR #932, opened 2016-09-02
fix(search) - Referesh frame and button not appearing on updates

sjunaid:
- fix bug introduced in recent refact(search) 2nd last changes in SearchResults.js, that was preventing to canShowRefresh() to compare and decide if refresh frame/button should appear.
- also added the removed item:deleted, work done in https://github.com/superdesk/superdesk-client-core/pull/887


petrjasek:
:+1:

mdhaman:
:+1:

PR #934, opened 2016-09-02
Fix webpack-dev-server for Vagrant environment

lnogues:
* Added some VIM specific extension in .gitignore
* Change webpack-dev-server URL from `localhost` to `0.0.0.0` in order for it to be accessible from the host machine.


thnkloud9:
:thumbsup:

vied12:
:+1:

PR #937, opened 2016-09-02
ui(refactor): split 'superdesk-legal-archive' module.

gbbr:
In cases like this one, it could be considered a bit overkill
to create folders and `index.js` files given that it's such a small
module and could instead consider a version where there are no folders
and instead the module only contains files (1 html, 1 service, 1
directive, 1 controller). A counter-argument would be that this is a
more scalable form. I will leave it as is for the sake of consistency
and simplicity and we can think about the final form after this
intermediate refactoring step is completed.

`superdesk-legal-archive` becomes:
```
├── controllers
│   ├── LegalArchiveController.js
│   └── index.js
├── directives
│   ├── LegalItemSortbar.js
│   └── index.js
├── services
│   ├── LegalArchiveService.js
│   └── index.js
├── tests
│   └── legal.spec.js
├── views
│   └── legal_archive.html
└── index.js

5 directories, 9 files
```


vied12:
:+1:

petrjasek:
:+1:

PR #940, opened 2016-09-05
fix(relatedItem, search)[SD-5454][SD-5453] Modified related item widget and search for dates

mdhaman:
- Related Item widget is configurable to all slugline search using (match any word, exact match and start with) 
- Indication on the authoring screen if the item is related or not
- Modified search for dates (versioncreated, firstcreated, etc) to use default time zone. To use default timezone for search is configurable.


marwoodandrew:
:+1:

petrjasek:
:+1:

ioanpocol:
please rebase the PR

PR #941, opened 2016-09-05
SD-4983: Image metadata validation upon save and publish

thnkloud9:
added case to check for empty <br> tags in validateMediaFields()

### DIFF
--- a/scripts/superdesk-authoring/authoring/controllers/ChangeImageController.js
+++ b/scripts/superdesk-authoring/authoring/controllers/ChangeImageController.js
@@ -51,7 +51,8 @@ export function ChangeImageController($scope, gettext, notify, modal, $q, _, api
         function validateMediaFields() {
             _.each($scope.data.requiredFields, function (key) {
                 var value = $scope.data.metadata[key];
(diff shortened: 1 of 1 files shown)


scripts/superdesk-authoring/authoring/controllers/ChangeImageController.js
```diff
@@ -50,7 +50,8 @@ export function ChangeImageController($scope, gettext, notify, modal, $q, _, api
         /* Throw an exception if a required metadata field is missing */
         function validateMediaFields() {
             _.each($scope.data.requiredFields, function (key) {
-                if ($scope.data.metadata[key] == null || _.isEmpty($scope.data.metadata[key])) {
+                var value = $scope.data.metadata[key];
+                if (value == null || _.isEmpty(value) || value === '<br>') {
```
gbbr: Would simply `if (!!!value && value === '<br>')` not do? It should already catch all falsy values (both _null_ and _empty_ string).
Also, what about the other forms of `br`, such as `<br/>`, `<br >`, `<br />`, etc ?

vied12:
:+1:

gbbr:
👍

PR #942, opened 2016-09-05
SD-5002 - highlight multiple spaces in spell check

lnogues:
Hightlight multiple spaces in spellcheck and suggest a single space instead.

### DIFF
--- a/scripts/superdesk/editor/spellcheck/spellcheck.js
+++ b/scripts/superdesk/editor/spellcheck/spellcheck.js
@@ -297,9 +297,10 @@ function SpellcheckService($q, api, dictionaries, $rootScope, $location) {
      * @param {string} word
      */
     this.suggest = function suggest(word) {
-        if (word.match(/^\s+$/i))
+        if (word.match(/^\s+$/i)) {
             return Promise.resolve([{key: ' ', value: 'Add single space'}]);
-        else
+        }
+        else {
             return api.save('spellcheck', {
                 word: word,
                 language_id: lang
@@ -317,6 +318,7 @@ function SpellcheckService($q, api, dictionaries, $rootScope, $location) {
                     return {key: key, value: key};
                 });
             });
+        }
     };
 
     /**
--- a/scripts/superdesk/editor/spellcheck/spellcheck.js
+++ b/scripts/superdesk/editor/spellcheck/spellcheck.js
@@ -253,13 +253,27 @@ function SpellcheckService($q, api, dictionaries, $rootScope, $location) {
         return getDict().then(function(d) {
             var errors = [],
                 regexp = /[0-9a-zA-Z\u00C0-\u1FFF\u2C00-\uD7FF]+/g,
+                dblSpacesRegExp = /\S(\s{2,})\S/g,
                 match,
+                dblSpacesMatch,
                 currentOffset = 0,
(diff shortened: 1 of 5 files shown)


scripts/superdesk/editor/spellcheck/spellcheck.js
```diff
@@ -285,19 +297,26 @@ function SpellcheckService($q, api, dictionaries, $rootScope, $location) {
...
+            }).then(function(result) {
+                var allDict = getDict();
+                var wordFoundInDict = _.pick(allDict.content, function(value, key) {
+                    if (key.toLowerCase() === word.toLowerCase()) {
+                        return key;
+                    }
+                });
+                angular.extend(result.corrections, Object.keys(wordFoundInDict));
+
+                return result.corrections.map(function(key) {
```
gbbr: You may use ES6 syntax here if you'd like `return result.corrections.map(key => ({key: key, value: key}));`

lnogues: Well since I didn't see any arrow functions in the file, I kept the old syntax

gbbr: I know, it's fine. We've just recently started supporting ES6 and I'm trying to encourage it more.

scripts/superdesk/editor2/views/block-text.html
```diff
@@ -20,7 +20,7 @@
     </code>
 
     <ul class="dropdown-menu">
-        <li ng-repeat="word in suggestions"><button ng-click="replace(word)">{{ :: word }}</button></li>
+        <li ng-repeat="word in suggestions"><button ng-click="replace(word.key)">{{ :: word.value }}</button></li>
```
sjunaid: Hi @lnogues , there is a change impact in superdesk\editor2\editor.js, 
- the below code need to be adjusted for capital word suggestion,

spellcheck.suggest(node.textContent).then(function(suggestions) {
                            if (scope.sentenceWord) {
                                ```suggestions.push(scope.replaceWord[0].toUpperCase() + scope.replaceWord.slice(1));```
                                scope.suggestions = _.unique(suggestions);
                            } else {

- use _.uniq instead, as _.unique alias is removed in favor of _.uniq in ver. 4. https://github.com/lodash/lodash/wiki/Changelog

scripts/superdesk/editor2/editor.js
```diff
@@ -1151,8 +1151,14 @@ angular.module('superdesk.editor2', [
...
                             if (scope.sentenceWord) {
-                                suggestions.push(scope.replaceWord[0].toUpperCase() + scope.replaceWord.slice(1));
-                                scope.suggestions = _.unique(suggestions);
+                                suggestions.push({
+                                    key: scope.replaceWord[0].toUpperCase() + scope.replaceWord.slice(1),
+                                    value: scope.replaceWord[0].toUpperCase() + scope.replaceWord.slice(1)
+                                });
+
+                                scope.suggestions = suggestions.filter(suggestion => {
+                                    return suggestion.key !== scope.replaceWord;
```
gbbr: You don't necesarilly have to change this, as it is already correct, but just FYI, `return` can be omitted here with this form:
```js
scope.suggestions = suggestions.filter(suggestion => suggestion.key !== scope.replaceWord);
```

vied12:
:+1:

gbbr:
👍

PR #916, opened 2016-08-31
feat(content type): Change default content type to support feature me…

ioanpocol:
…dia (support image and video)

PR #944, opened 2016-09-05
Revert "feat(content type): Change default content type to support feature me…"

ioanpocol:
Reverts superdesk/superdesk-client-core#916


petrjasek:
:+1:

PR #945, opened 2016-09-05
fix(editor): add dragdrop and stopPropigate to sdTextEditorDropZone directive

thnkloud9:
not sure why this change was needed now as the feature worked fine last week, but this PR should fix the issues with the current chrome browser.


ioanpocol:
:+1:

ioanpocol:
This issue fixed the issue to drag&drop on body field but it still an issue to drag&drop on featured image/media.

PR #946, opened 2016-09-05
SDNTB-205 Translation issues

vladnicoara:


scripts/superdesk/datetime/datetime.js
```diff
@@ -150,11 +150,12 @@ export default angular.module('superdesk.datetime', [
 
     .directive('sdDatetime', DateTimeDirective)
 
-    .filter('reldate', function reldateFactory() {
+    .filter('reldate', ['gettextCatalog', function reldateFactory(gettextCatalog) {
         return function reldate(date) {
+            moment.locale(gettextCatalog.currentLanguage);
```
gbbr: I'm not an expert with moment, but should this not be a project-wide setting? Does the locale only apply to this filter? Won't changing the locale here change it everywhere?

If so, maybe this would be best suited in `scripts/bootstrap.js` inside a `run` clause, such as:
```js
.run(['gettextCatalog', gtc => moment.locale(gtc.currentLanguage)]);
```

ioanpocol:
:+1:  (except comment)

scripts/superdesk/datetime/datetime.js
vladnicoara: you are right, moved to
https://github.com/superdesk/superdesk-client-core/pull/946/files#diff-72e8716fc4deadd7a7a6fa4e986cf46eL22

so it reacts as soon as the language is loaded/changed

gbbr: Nice. Thanks Vlad!

gbbr:
:+1:

PR #947, opened 2016-09-06
[SD-5403]-(fix)- Highlighted items stay highlighted after search term…

sjunaid:
… is removed
included fix for [SD-5399](https://dev.sourcefabric.org/browse/SD-5399) as well

### DIFF
--- a/scripts/superdesk-search/services/SearchService.js
+++ b/scripts/superdesk-search/services/SearchService.js
@@ -448,8 +448,9 @@ export function SearchService($location, gettext, config, session) {
                     _.forEach(_.keys(item.es_highlight), function (key) {
                         item[key] = item.es_highlight[key][0];
                     });
+                } else {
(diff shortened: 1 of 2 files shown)


mdhaman:
:+1:

ioanpocol:
:+1:

scripts/superdesk-search/services/SearchService.js
```diff
@@ -448,6 +448,7 @@ export function SearchService($location, gettext, config, session) {
                     _.forEach(_.keys(item.es_highlight), function (key) {
                         item[key] = item.es_highlight[key][0];
                     });
+                    item.reRender = true; // used to make it re-render.
```
petrjasek: imo would be better to create a new item here with `return angular.extend({}, item);`, rather then adding here a flag and create this new item later.

petrjasek:
:+1:

PR #951, opened 2016-09-06
refactor(search-providers): split 'superdesk-search-providers' module.

gbbr:
This could be another way to split smaller modules (didn't create
folders).

`superdesk-search-providers` becomes:
```
.
├── views
│   ├── search-provider-config.html
│   └── settings.html
├── constants.js
├── directive.js
├── index.js
└── service.js

1 directory, 6 files
```


petrjasek:
:+1:

PR #954, opened 2016-09-06
SD-5462: Image drag and drop to editor feature media field has no effect in chrome

thnkloud9:
also fixes an issue when clicking the feature media field when it is empty that prevented the upload dialog from being shown.


ioanpocol:
:+1:

pavlovicnemanja:
:+1:

PR #956, opened 2016-09-07
fix(featured media): Disable crop editing for video

ioanpocol:


scripts/superdesk-authoring/authoring/directives/ItemAssociationDirective.js
```diff
@@ -72,20 +72,29 @@ export function ItemAssociationDirective(superdesk, renditions, $timeout, api, $
             renditions.get();
 
             scope.edit = function(item) {
-                scope.loading = true;
-                return renditions.crop(item).then(function(updatedItem) {
-                    var data = updateItemAssociation(updatedItem);
+                if (item.renditions && item.renditions.original && scope.isVideo(item.renditions.original)) {
```
petrjasek: imo would be better to change the logic - if picture do crop otherwise just attach. because there is also audio without cropping, not sure about graphics atm

ioanpocol: Yes, true, I will change the code

petrjasek:
:+1:

pavlovicnemanja:
👍

PR #957, opened 2016-09-08
SDPA-498 "Slugline" should be replaced by "Keyword"

mugurrus:
translation updates

### DIFF
--- a/po/en_GB.po
+++ b/po/en_GB.po
@@ -11,6 +11,10 @@ msgstr ""
 "Language: en_GB\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
+#: scripts/superdesk/lang/missing-translations-strings.js:19
+msgid "'SLUGLINE is a required field'"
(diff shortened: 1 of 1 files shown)


petrjasek:
:+1:

po/en_GB.po
```diff
@@ -11,6 +11,10 @@ msgstr ""
 "Language: en_GB\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
+#: scripts/superdesk/lang/missing-translations-strings.js:19
+msgid "'SLUGLINE is a required field'"
+msgstr "'KEYWORD is a required field'"
```
petrjasek: this probably should be empty like the others

mugurrus: No, it's needed for PA

petrjasek: that's strange, it's not there for `Slugline` below

gbbr:
:+1:

mugurrus:
This is not the proper solution to the issue, will make another PR.

PR #961, opened 2016-09-08
Sd 5318 - Solving the issue with Cursor jumping to the beginning of pasted text in the caption of an image

lnogues:
Solving the issue with Cursor jumping to the beginning of pasted text in the caption of an image


scripts/superdesk/editor2/editor.js
```diff
@@ -566,11 +566,23 @@ function SdTextEditorBlockEmbedController($timeout, editor, renditions, config)
             vm.model.caption = caption;
             // update the caption in the view
             vm.caption = caption;
+
             // on change callback
             $timeout(function() {
                 vm.onBlockChange();
             });
         },
+        handlePasta: function(e) {
```
petrjasek: I guess it should be `handlePaste`

petrjasek:
:+1: looks good

scripts/superdesk/editor2/editor.js
pavlovicnemanja: Mmmm finally, handling food with Superdesk :) I prefer meat, but pasta also works for me :)

pavlovicnemanja:
👍

mdhaman:
@petrjasek @lnogues change looks good. I hope this change does not break our formatters. I am just being cautious. Would it be possible to move this change to separate branch and test it?

PR #963, opened 2016-09-08
SD-5316 Pressing enter in the body changes the font for the text after the cursor

lnogues:
This issue seems to be directly related to contenteditable. So for now, the only way I found to modify it, is to manually remove any styling attributes in any span elements.

### DIFF
--- a/scripts/superdesk/editor2/editor.js
+++ b/scripts/superdesk/editor2/editor.js
@@ -1020,7 +1020,10 @@ angular.module('superdesk.editor2', [
                         scope.medium.subscribe(eventName, updateAddContentButton);
                     });
                     // listen updates by medium editor to update the model
-                    scope.medium.subscribe('editableInput', function() {
+                    scope.medium.subscribe('editableInput', function(e, elem) {
+                        elem.querySelectorAll('span[style]')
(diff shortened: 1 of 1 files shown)


scripts/superdesk/editor2/editor.js
```diff
@@ -1020,7 +1020,10 @@ angular.module('superdesk.editor2', [
                         scope.medium.subscribe(eventName, updateAddContentButton);
                     });
                     // listen updates by medium editor to update the model
-                    scope.medium.subscribe('editableInput', function() {
+                    scope.medium.subscribe('editableInput', function(e, elem) {
+                        elem.querySelectorAll('span[style]')
+                            .forEach(span => span.removeAttribute('style'));
```
petrjasek: would it be possible to remove those spans altogether?

lnogues: Sure, unless there are some specific usage for them?

petrjasek: I don't think so, it's just added by chrome, in firefox it shoudn't be there at all

lnogues: I can remove it

petrjasek:
:+1:

PR #965, opened 2016-09-08
fix(monitoring) SD-5288 On closing preview, remove selection of item

pavlovicnemanja:
SD-5288 - un-highlight an item in list view
SD-5317 - Make whole Superdesk drop point for drag&drop images
SD-5444 - Hide the task option on application right menu (monitoring, personal, ...)

### DIFF
--- a/scripts/superdesk-search/directives/ItemList.js
+++ b/scripts/superdesk-search/directives/ItemList.js
@@ -1526,6 +1526,10 @@ export function ItemList(
                     }
                 },
 
+                deselectAll: function() {
+                    this.setState({selected: null});
+                },
+
                 updateAllItems: function(itemId, changes) {
                     var itemsById = angular.extend({}, this.state.itemsById);
                     _.forOwn(itemsById, function(value, key) {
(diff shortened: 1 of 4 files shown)


petrjasek:
:+1:

scripts/superdesk-search/directives/ItemList.js
```diff
@@ -1550,6 +1550,13 @@ export function ItemList(
                 },
 
                 setSelectedItem: function(item) {
+                    if (monitoringState.selectedGroup !== scope.$id) {
+                        //If selected item is from another group, deselect all
+                        $rootScope.$broadcast('item:unselect');
+                    } else {
```
gbbr: This `else` clause is not needed. `monitoringState.selectedGroup` already has the valud `$scope.$id` here.

scripts/superdesk-search/directives/ItemList.js
```diff
@@ -1550,6 +1550,13 @@ export function ItemList(
                 },
```
gbbr: I would make a function here:
```js
deselectAll: () => listComponent.setState({selected: null}),
```

And use that below in both places. Instead of `$rootScope.$broadcast('item:unselect');` and further down:
```js
scope.$on('item:unselect', this.deselectAll);
```
That broadcast can involve many side-effects later, and this way you can use a local function instead of broadcasting again via root.

pavlovicnemanja: Hey, I need to use `$broadcast` because I need to affect all scopes of same directive where I will remove selection. `this` will affect only current scope. I will create this `deselectAll` function.

scripts/superdesk-search/directives/ItemList.js
pavlovicnemanja: my mistake, this should be fired only on selection items between groups... changed...

gbbr:
👍

PR #966, opened 2016-09-08
Added a local .vimrc

lnogues:
For whoever is using vim, it's possible to create a project specific vimrc.
To enable this functionality, you need to add `set exrc` in your vim config.

This current .vimrc is doing two things:

* Forcing a 4 spaces indentation
* Changing the file saving strategy and thus greatly improving the webpack hot reloading.

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -1,0 +1,10 @@
+filetype plugin indent on
+" show existing tab with 4 spaces width
+set tabstop=4
+" when indenting with '>', use 4 spaces width
+set shiftwidth=4
+" On pressing tab, insert 4 spaces
+set expandtab
+
+" Otherwise webpack doesn't detect your changes
+set backupcopy=yes


naspeh:
I'm not sure it should be in the repo

You could add `.vimrc` to [global .gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore), so you'll have this functionality locally.

Who uses vim, probably could setup it properly without outside help ;).

We could create vim tips page somewhere in docs for mentioned technics.

gbbr:
This is cool, I didn't know you can do this! Still, I don't think enough people use vim that have had issues setting these options up, to make it worthwhile including in the project. I would rather be ok with adding it to `.gitignore` to allow you to use it, as suggested above.

naspeh:
I said about **global .gitignore**, you don't need to put it to repo's .gitignore.

https://thomashunter.name/blog/global-gitignore-vs-repository-gitignore/

petrjasek:
I see this is project specific config (unlike this gitingore thing), so I wouldn't mind it, lets see how others will vote
:+1:

.vimrc
```diff
@@ -0,0 +1,10 @@
+filetype plugin indent on
```
naspeh: I'm not using this setting, I don't like it. Please remove it. Other settings may sense.

naspeh:
BTW: From: http://stackoverflow.com/questions/456792/vim-apply-settings-on-files-in-directory

> I'd strongly suggest not using set exrc
> 
> Even with set secure, under *nix, vim will still run autocommands, shell, et al, if you own the file. So if you happend to edit a file in that tarball I sent you with a .vimrc containing:
> 
> autocmd BufEnter * :silent! !echo rm -rf ~/
> you'll probably be less amused than I will.

Also about `set secure` after `set exrc`: http://www.ilker.de/specific-vim-settings-per-project.html

gbbr:
Not the best solution (because it applies to all JS), but what I have is:
```
au FileType javascript setlocal ts=4 sw=4 expandtab
```

naspeh:
This is why I told that this file is not for the repo. We are different and have different preferences. In Vim we could do the same thing with many ways...

naspeh:
@petrjasek do you use VIM?

petrjasek:
yep

gbbr:
Seems this isn't going in, so I am closing it for now. Please re-open otherwise...

PR #967, opened 2016-09-09
fix(relatedItem) Modified related item widget to indicate possible update items or take items.

mdhaman:
- Adding an indicator to related items for 'associate as update' or 'associated as take'

### DIFF
--- a/scripts/superdesk-archive/index.js
+++ b/scripts/superdesk-archive/index.js
@@ -1,3 +1,7 @@
+//styles
+import './styles/related-item.less';
+
+//scripts
 import './related-item-widget/relatedItem';
 
 import * as directive from './directives';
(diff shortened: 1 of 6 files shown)


petrjasek:
:+1:

scripts/superdesk-archive/index.js
```diff
@@ -1,3 +1,7 @@
+//styles
+import './styles/related-item.less';
```
gbbr: Would it serve better de-coupling if `relative-item-widget` would import its own styles?

gbbr:
LGTM besides minor structural comment

scripts/superdesk-archive/index.js
mdhaman: sure will do that.

PR #978, opened 2016-09-12
fix(search): keep field name untranslated in query

petrjasek:
otherwise it will just fail while it won't match elastic schema

SD-5484

### DIFF
--- a/scripts/superdesk-authoring/metadata/metadata.js
+++ b/scripts/superdesk-authoring/metadata/metadata.js
@@ -645,7 +645,7 @@ function MetaTermsDirective(metadata, $filter, $timeout) {
                 scope.terms = filterSelected(scope.list);
                 if (scope.cv) { // filter out items from current cv
                     scope.selectedItems = _.filter(selected, function(term) {
-                        return term.scheme === scope.cv._id;
+                        return term.scheme === (scope.cv._id || scope.cv.id);
                     });
                 } else {
                     scope.selectedItems = selected;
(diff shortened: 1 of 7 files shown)


petrjasek:
@vladnicoara what do you think?

vladnicoara:
@petrjasek  LGTM

scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -700,17 +700,25 @@ function MetaTermsDirective(metadata, $filter, $timeout) {
...
             }
 
             function addTerm(term) {
+                if (!term) {
+                    return;
+                }
+
                 // Only select terms that are not already selected
-                if (!_.find(scope.item[scope.field], function(i) {return i[scope.uniqueField] === term[scope.uniqueField];})) {
+                if (!_.find(scope.item[scope.field], function(i) {
```
gbbr: It would be easier to figure this logic out, imo, if we would separate it into a function. One possible example could be:
```js
let uf = scope.uniqueField;
let isSelected = _.find(scope.item[scope.field], i => i && i[uf] === term[uf];

if (isSelected) {
    return;
}

// ...
```
This is just an alternative. It also reduced indentation. But, if you feel strongly about yours, I'm happy to keep it.

scripts/superdesk-search/services/TagService.js
```diff
@@ -23,6 +23,13 @@ export function TagService($location, desks, userList, metadata, search) {
...
 
+    function tag(label, value) {
+        return {
+            label: label,
+            value: value || label
+        };
+    }
+
     function initSelectedParameters (parameters) {
         tags.selectedParameters = [];
```
gbbr: If the `tag` function is only used in the scope of this function, should it not be here:
```js
let tag = (l, v) => ({label: l, value: v||l});
```
I feel like the ES6 police :-)

gbbr:
LGTM (besides minor styling comments)

scripts/superdesk-search/services/TagService.js
petrjasek: it's not only there, but all over the service. might give es6 a go, but is there any advantage to use a function there? just wondering..

gbbr: I think the only advantage is better encapsulation, since `tag` is only used in `initSelectedParameters`, defining it there makes this obvious. It's just styling really... Not necessarily something that needs changing, just a personal opinion.

petrjasek: it's not used only there, but also in other methods

gbbr: Sorry, didn't see that.

PR #979, opened 2016-09-12
Settings improvements 1

fritzSF:


scripts/superdesk/directives/sdCheck.js
```diff
@@ -30,10 +30,19 @@ export default angular.module('superdesk.check.directives', [])
                 element.on('click', function(e) {
                     $scope.$apply(function() {
                         ngModel.$setViewValue(!ngModel.$viewValue);
+                        console.log(attrs);
```
gbbr: Is this needed?

gbbr:
_as per Slack discussions with @pavlovicnemanja:_ 

We will try to use [`transclude`](https://docs.angularjs.org/api/ng/directive/ngTransclude) here as discussed on Slack. It will give developers more control and freedom over styling the checkbox.

Evaluating and appending user input to the DOM can many times be an open window to XSS attacks, so we should be a bit more weary when considering to use them. They are considered advanced operations that are for extremely rare use-cases. Most common operations can easily be resolved with standard angular functionalities.

pavlovicnemanja:
👍

Aca-jov:
:+1:

PR #985, opened 2016-09-14
fix(workspace): fix spike view for custom workspace

petrjasek:
it was overlapping when there were multiple desks selected in
custom workspace config.

SD-5445


pavlovicnemanja:
👍

PR #986, opened 2016-09-14
Added web publisher module for swp

Aca-jov:
SWP-244

This is just adding web-publisher module and some visual elements (side navigation and monitoring), it will be reworked in next sprints according to requirements.


pavlovicnemanja:
👍 LGTM

PR #987, opened 2016-09-14
fix(toolbar): SD-5480 Keep toolbar visible when scrolling

pavlovicnemanja:
SD-5480 - The formatting toolbar for the body field doesn't stay visible when scrolling down


Aca-jov:
:+1:

PR #989, opened 2016-09-15
fix(authoring): set scope dirty when something is dropped on featured

petrjasek:
it could trigger a new digest cycle when another was running already,
thus raising an error so onchange handler was never called.
now it will only call `$apply` when needed - on drop.

SD-5486

### DIFF
--- a/scripts/superdesk-authoring/authoring/directives/ItemAssociationDirective.js
+++ b/scripts/superdesk-authoring/authoring/directives/ItemAssociationDirective.js
@@ -1,5 +1,5 @@
-ItemAssociationDirective.$inject = ['superdesk', 'renditions', '$timeout', 'api', '$q', 'config'];
-export function ItemAssociationDirective(superdesk, renditions, $timeout, api, $q, config) {
+ItemAssociationDirective.$inject = ['superdesk', 'renditions', 'api', '$q', 'config'];
+export function ItemAssociationDirective(superdesk, renditions, api, $q, config) {
     return {
         scope: {
             rel: '=',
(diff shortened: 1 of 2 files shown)


naomiaro:
👍  nice tests

scripts/superdesk-authoring/authoring/directives/ItemAssociationDirective.js
```diff
@@ -47,21 +47,17 @@ export function ItemAssociationDirective(superdesk, renditions, $timeout, api, $
                         scope.loading = false;
                     });
                 } else {
-                    var data = updateItemAssociation(item);
-                    scope.onchange({item: scope.item, data: data});
+                    scope.$apply(function() {
```
gbbr: Please simplify as:
```js
scope.$apply(updateItemAssociation);
```

scripts/superdesk-authoring/authoring/directives/ItemAssociationDirective.spec.js
```diff
@@ -0,0 +1,36 @@
...
+        scope.item = item;
+        elem = $compile(`<div sd-item-association
+            data-item="item"
+            data-rel="rel"
+            data-onchange="onChange()"></div>`
+        )(scope);
+        $rootScope.$digest();
+    }));
+
+    it('can trigger onchange handler on drop', inject(function(renditions, $q, $rootScope) {
```
gbbr: You're not using `$q` and `renditions`, can they be removed?

petrjasek: sure

scripts/superdesk-authoring/authoring/directives/ItemAssociationDirective.js
petrjasek: it needs `item` arg

gbbr: Yes, sorry, my bad.

PR #991, opened 2016-09-16
Scroll list view when outside of the view port

lnogues:
Allow the view port to scroll up or down if the current selected element is outside of it

### DIFF
--- a/scripts/superdesk-search/directives/ItemList.js
+++ b/scripts/superdesk-search/directives/ItemList.js
@@ -1345,6 +1345,22 @@ export function ItemList(
                     dragitem.start(event, this.props.item);
                 },
 
+                componentDidUpdate: function() {
+                    // If content-list is present in the dom, we can update its scroll position
+                    // in case the current element is outside of the viewport.
+                    if (this.props.flags.selected && document.getElementsByClassName('content-list')) {
+                        let itemRect = this.refs.item.getBoundingClientRect();
+                        let contentList = document.getElementsByClassName('content-list')[0];
+                        let contentListRect = contentList.getBoundingClientRect();
+
+                        if (itemRect.bottom > contentListRect.bottom) {
+                            contentList.scrollTop += itemRect.bottom - contentListRect.bottom;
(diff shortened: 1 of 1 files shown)


scripts/superdesk-search/directives/ItemList.js
```diff
@@ -1345,6 +1345,22 @@ export function ItemList(
                     dragitem.start(event, this.props.item);
                 },
 
+                componentDidUpdate: function() {
+                    // If content-list is present in the dom, we can update its scroll position
+                    // in case the current element is outside of the viewport.
+                    if (this.props.flags.selected && document.getElementsByClassName('content-list')) {
+                        let itemRect = this.refs.item.getBoundingClientRect();
```
gbbr: Can we move the contents of this `if` condition into a separate function? It will make `componentDidUpdate` a bit more cleaner, just in case more stuff goes in here later. If we call it something like `this.adjustScroll` it will also make a lot more sense imo.

It could be something like:
```js
// adjustScroll adjusts the content list's scroll position in case the current item is
// out of view
adjustScroll: function() {
    let el = document.getElementsByClassName('content-list');
    if (!el.length) {
        return;
    }
    let itemRect = this.refs.item.getBoundingClientRect();
    let contentList = el[0];
    let contentListRect = contentList.getBoundingClientRect();

    if (itemRect.bottom > contentListRect.bottom) {
        contentList.scrollTop += itemRect.bottom - contentListRect.bottom;
    } else if (itemRect.top < contentListRect.top) {
        contentList.scrollTop -= contentListRect.top - itemRect.top;
    }
},

componentDidUpdate: function() {
    if (this.props.flags.selected) {
        this.adjustScroll();
    }
},
// ...
```
The React component's life-cycle methods are where most people look to understand how a component works. We should keep those as simple as possible imo.

Sorry if I've made any mistakes, I was just trying to illustrate the point.

ioanpocol:
@lnogues  This PR is very old. Can you fix it or close it?

gbbr:
I'm closing this. It's way outdated. Re-open if needed.

PR #992, opened 2016-09-16
feat(send): add type ahead for desk select in send to

petrjasek:
SD-4947

### DIFF
--- a/scripts/superdesk-desks/directives/DeskSelect.js
+++ b/scripts/superdesk-desks/directives/DeskSelect.js
@@ -1,0 +1,158 @@
+DeskSelect.$inject = ['Keys', 'lodash'];
+export function DeskSelect(Keys, _) {
+    const UP = -1;
+    const DOWN = 1;
+
+    class DeskSelectController {
+        constructor() {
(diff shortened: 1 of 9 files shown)


scripts/superdesk-desks/index.js
```diff
@@ -13,6 +13,8 @@ import * as ctrl from './controllers';
 import * as directive from './directives';
 import { DesksFactory } from './services';
 
+console.assert(directive.DeskSelect, directive.DeskSelect);
```
gbbr: Is this forgotten or should we have it? I think Angular will fail anyway because of missing directives, so it's probably safe to remove.

scripts/superdesk-desks/directives/DeskSelect.js
```diff
@@ -0,0 +1,153 @@
+
```
gbbr: Extra space here.

scripts/superdesk-desks/index.js
petrjasek: that's forgotten

PR #993, opened 2016-09-16
Automatically load external node modules via config

gbbr:
_Work in progress..._

### Proposal

Enables telling superdesk to load external resources via configuration. In `superdesk.config.js`, adding:
```js
apps: ['superdesk-planning', 'my-app']
```
will cause the client to include `node_modules/superdesk-planning` and `node_modules/my-app` into the compilation. 

External applications that want to register Angular modules into Superdesk should do so by calling:
```js
window.RegisterSuperdeskApplication('angular.module.name');
```
This will cause `angular.module.name` to be included into the main app. This is basically the entry point to your module.

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,6 +9,7 @@
   </head>
   <body ng-class="config.bodyClass">
     <div sd-superdesk-view></div>
-    <script src="app.bundle.js" /></script>
+    <script src="core.bundle.js" /></script>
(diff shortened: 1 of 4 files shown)


gbbr:
Closing in favour of simplified version https://github.com/superdesk/superdesk-client-core/pull/995

PR #995, opened 2016-09-16
Auto-load external Angular modules using config

gbbr:
Enables telling superdesk to load external resources via configuration. In `superdesk.config.js`, adding:
```js
apps: ['superdesk-planning', 'my-app']
```
will cause the client to include `node_modules/superdesk-planning` and `node_modules/my-app` into the compilation, as well as load all exported Angular modules into superdesk. 

--
All that external apps need to do is export an Angular module that is the entry point as default. ie:

```js
export default angular.module('my.superdesk.app', []);
```
--
### TODO

- [x] Allow importing `npm` modules by config
- [x] Write documentation

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -50,6 +50,7 @@ module.exports = function(grunt) {
     grunt.registerTask('server', [
         'clean',
         'copy:index',
+        'ngtemplates:gen-importer',
         'ngtemplates:dev',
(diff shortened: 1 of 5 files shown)


gbbr:
Correct. You can name your module anything, as long as it doesn't clash
with other already existing ones.

PR #988, opened 2016-09-15
feat(search): Adding the remove aggregate functionality.

mdhaman:
- [x] allows user to exclude a particular facet.
- [x] Documentation for the new functionality
- [x] Unit test and e2e test

PR #996, opened 2016-09-22
authoring/autosave: improve save method

gbbr:
Noticed some improvements to `autosave` service during the hackathon and decide to apply them. Also rewrote with ES6 (probably best viewed in split diff).

Copied over some changes to `.travis.yml` from https://github.com/superdesk/superdesk-client-core/pull/988 to fix build.

### DIFF
--- a/scripts/superdesk-authoring/authoring/services/AutosaveService.js
+++ b/scripts/superdesk-authoring/authoring/services/AutosaveService.js
@@ -1,71 +1,81 @@
 import * as helpers from 'superdesk-authoring/authoring/helpers';
 
-AutosaveService.$inject = ['$q', '$timeout', 'api'];
-export function AutosaveService($q, $timeout, api) {
-    var RESOURCE = 'archive_autosave',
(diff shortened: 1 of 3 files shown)


scripts/superdesk-authoring/authoring/services/AutosaveService.js
```diff
@@ -1,77 +1,89 @@
...
-AutosaveService.$inject = ['$q', '$timeout', 'api'];
-export function AutosaveService($q, $timeout, api) {
-    var RESOURCE = 'archive_autosave',
-        AUTOSAVE_TIMEOUT = 3000,
-        timeouts = {},
-        self = this;
+const RESOURCE = 'archive_autosave',
+    AUTOSAVE_TIMEOUT = 3000;
+
+let $q, $timeout, api;
```
petrjasek: was wondering if this could be somehow avoided, but I guess not

gbbr:
Also not my favourite solution to injectables. Should be fine though, since services are only instantiated once.

PR #1002, opened 2016-09-28
Tansa lincese

Aca-jov:


scripts/superdesk-tansa/index.js
```diff
@@ -14,8 +14,8 @@ document.write(`
                 // need to populate dynamically
                 tansaUserId: 'tansa',
                 checkboxPreference: false,
-                parentAppId: ‘02d65e3e-f62d-41f9-92c6-ffaf16bf209e’,
-                licenseKey: ‘10747a8a5016-95b1-4d4c-846b-f34347e8f288’
+                parentAppId: '02d65e3e-f62d-41f9-92c6-ffaf16bf209e',
+                licenseKey: '10747a8a5016-95b1-4d4c-846b-f34347e8f288'
```
gbbr: Maybe this should be private - wouldn't other people be able to use it otherwise?

We could have it as an ENV var and inject it via webpack....

PR #1008, opened 2016-09-29
WIP: Enable snapshot functionality for e2e tests

naspeh:
Regarding: [SDESK-129](https://dev.sourcefabric.org/browse/SDESK-129)

Backend part: [superdesk-core/pull/592](https://github.com/superdesk/superdesk-core/pull/592)

PR #1012, opened 2016-09-30
WIP: Tests check

naspeh:
The same as #1008


naspeh:
New clean PR: https://github.com/superdesk/superdesk-client-core/pull/1014

PR #1013, opened 2016-10-03
Fix travis build

gbbr:
Update Python to fix `bcrypt` error that was crashing build and removed `-q` flag from `pip` for easier debugging. Also removed `npm` restriction of Python version since this is no longer causing issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 
-python: 3.4
+python: 3.5
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


ahilles107:
You know that You can start travis build on Your fork branch without PR's like those?

gbbr:
@ahilles107 sorry for the spam, no, I didn't.

gbbr:
Hopefully this will eventually turn into the PR that fixes the actual build, so it should be ok.

PR #1018, opened 2016-10-04
[SD-5508] Allow ingest providers to be flagged as pass throught

marwoodandrew:


marwoodandrew:
Requires superdesk/superdesk-core#601

scripts/apps/ingest/views/settings/ingest-sources-content.html
```diff
@@ -123,6 +123,12 @@ <h3 translate ng-hide="provider._id" translate>Add New Source</h3>
                                 <option value=""></option>
                             </select>
                         </div>
+                        <div class="field">
+                        <label class="right">
```
gbbr: The indentation for `div.field`'s children is a bit off...

marwoodandrew: Thanks, I'd made a mess of that!

PR #1020, opened 2016-10-05
[SD-5512] - Embargo datetime fields made dependent on embargo  privilege

akintolga:


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/602

PR #1021, opened 2016-10-05
Fix publish

Aca-jov:
SDNTB-306

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.js
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.js
@@ -418,9 +418,13 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
 
             };
 
+            /**
+             * continueAfterPublish is passed only from $scope.publishAndContinue as bool,
+             * in other cases this is object (don't use parameter in those cases)
+             */
             $rootScope.publishAfterTansa = function () {
                 if (!onlyTansaProof) {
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -420,7 +420,7 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
 
             $rootScope.publishAfterTansa = function () {
                 if (!onlyTansaProof) {
-                    $scope.saveTopbar().then(continueAfterPublish ? $scope.publishAndContinue : $scope.publish);
+                    $scope.saveTopbar().then(continueAfterPublish === true ? $scope.publishAndContinue : $scope.publish);
```
gbbr: I'm just curious: why does this fix the problem? What other truthy values can `continueAfterPublish` have that cause issues here?

Aca-jov: This is because of `$scope.publish = function(continueOnPublish) {`.

Here I pass true: 
```$scope.publishAndContinue = function() {
                $scope.publish(true).then(function(published) {```

And on other occasions, on publish - item is passed, so there is object, so it is needed to check only for true.

gbbr: Thanks for the explanation.

This is quite confusing and unexpected. `continueAfterPublish` indicates that it's a bool value. Can we please add a comment here to at least specify what you just said above?

It would be nice to fix this so that bool is bool and item is item. Not either bool or item.

PR #886, opened 2016-08-24
[SD-5368]- Update page title with the Url and Desk/Workspace

akintolga:
- A new service is created to update the page title
- The title will be 'Superdesk {- Url} {-Desk/Workspace}
- Url can be:
 - Dashboard, Monitoring, Spike, Highlights, Search, Settings
- If Url is search then there won't be desk

PR #1024, opened 2016-10-05
[SD-5368]- Update page title with the Url and Desk/Workspace

akintolga:
The original PR was https://github.com/superdesk/superdesk-client-core/pull/886 and for some reason the changes are lost during rebase so creating the PR again...

- A new service is created to update the page title
- The title will be 'Superdesk {- Url} {-Desk/Workspace}
- Url can be:
 - Dashboard, Monitoring, Spike, Highlights, Search, Settings
- If Url is search then there won't be desk

### DIFF
--- a/scripts/core/services/pageTitle.js
+++ b/scripts/core/services/pageTitle.js
@@ -1,0 +1,33 @@
+/**
+ * PageTitle service
+ *
+ * PageTitle service is used to save and update the title of the page.
+ * The title will be 'Superdesk {- Url} {-Desk/Workspace}
+ * Url can be: Dashboard, Monitoring, Spike, Highlights, Search, Settings
+ * If Url is search then there won't be desk
+ *
+ */
+export default angular.module('superdesk.services.pageTitle', [])
+    .service('pageTitle', function() {
+        this.title = 'Superdesk';
+        this.url = '';
+        this.workspace = '';
+
+        this.setPageUrl = function(url) {
+            this.url = url;
+            setPageTitle(this.url, this.workspace);
+        };
+
+        this.setPageWorkspace = function(activeWorkspace) {
+            this.workspace = activeWorkspace;
+            setPageTitle(this.url, this.workspace);
+        };
+
+        function setPageTitle(url, activeWorkspace) {
+            document.title = 'Superdesk' + (url?' - ' + url:'') + (activeWorkspace?' - ' + activeWorkspace:'');
+        }
+
+        this.clearPageTitle = function() {
+            document.title = this.title ;
+        };
+    });
--- a/scripts/core/services/pageTitle.js
+++ b/scripts/core/services/pageTitle.js
@@ -7,27 +7,27 @@
  * If Url is search then there won't be desk
  *
  */
-export default angular.module('superdesk.services.pageTitle', [])
+export default angular.module('superdesk.core.services.pageTitle', [])
     .service('pageTitle', function() {
         this.title = 'Superdesk';
         this.url = '';
         this.workspace = '';
 
-        this.setPageUrl = function(url) {
+        this.setUrl = function(url) {
             this.url = url;
-            setPageTitle(this.url, this.workspace);
+            setTitle(this.url, this.workspace);
         };
 
-        this.setPageWorkspace = function(activeWorkspace) {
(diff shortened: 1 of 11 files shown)


akintolga:
@petrjasek For some reason the Travis doesn't kick in for this PR ...

petrjasek:
will close and reopen, that should trigger travis, in theory

scripts/core/services/pageTitle.js
```diff
@@ -0,0 +1,33 @@
...
+ * If Url is search then there won't be desk
+ *
+ */
+export default angular.module('superdesk.services.pageTitle', [])
+    .service('pageTitle', function() {
+        this.title = 'Superdesk';
+        this.url = '';
+        this.workspace = '';
+
+        this.setPageUrl = function(url) {
```
gbbr: I think `pageTitle.setPage*` is a bit stuttering. In my opinion it would be better to just call these methods `pageTitle.set*`.

Also, this is not really a URL, but more like a section. So would it not be better called simply `set`? Since this is a service, let's make it versatile and more global, so we can set any title we like.

scripts/core/services/pageTitle.js
```diff
@@ -0,0 +1,33 @@
...
+            this.url = url;
+            setPageTitle(this.url, this.workspace);
+        };
+
+        this.setPageWorkspace = function(activeWorkspace) {
```
gbbr: Should we call this `setWorkspace` instead?

If it would be a global function, then `setPageWorkspace` would make sense, but when you take into account that it is called in the context of the name of this service (`pageTitle.setPageWorkspace`), then it's a bit duplicated.

scripts/core/services/pageTitle.js
```diff
@@ -0,0 +1,33 @@
...
+            this.workspace = activeWorkspace;
+            setPageTitle(this.url, this.workspace);
+        };
+
+        function setPageTitle(url, activeWorkspace) {
+            document.title = 'Superdesk' + (url?' - ' + url:'') + (activeWorkspace?' - ' + activeWorkspace:'');
+        }
+
+        this.clearPageTitle = function() {
```
gbbr: Here too, simply `clear` is better imo, instead of `pageTitle.clearPageTitle`.

scripts/core/services/pageTitle.js
```diff
@@ -0,0 +1,33 @@
...
+            setPageTitle(this.url, this.workspace);
+        };
+
+        this.setPageWorkspace = function(activeWorkspace) {
+            this.workspace = activeWorkspace;
+            setPageTitle(this.url, this.workspace);
+        };
+
+        function setPageTitle(url, activeWorkspace) {
+            document.title = 'Superdesk' + (url?' - ' + url:'') + (activeWorkspace?' - ' + activeWorkspace:'');
```
gbbr: I'm just wondering, would it make more sense for the title to be of the form: `Superdesk - <Workspace> - <Title>`? I'm fine with both, whatever makes more sense.

akintolga: I'm fine with both too so I'll keep this like this for the time being...

scripts/core/services/pageTitle.js
akintolga: I have changed it to `clear`

scripts/core/services/pageTitle.js
```diff
@@ -0,0 +1,33 @@
...
+ * If Url is search then there won't be desk
+ *
+ */
+export default angular.module('superdesk.core.services.pageTitle', [])
+    .service('pageTitle', function() {
+        this.title = 'Superdesk';
+        this.url = '';
+        this.workspace = '';
+
+        this.setUrl = function(url) {
```
gbbr: I would still just call this `setName` or something else, since it's not really a URL in the context of the entire application. If it were to be a URL, we wouldn't have to set it and we could make it work automatically by hooking it up to the router. What do you think? Amend as you decide and feel free to merge imo.

akintolga: I agree with you that setUrl sounds more like a navigation action if you see the method name in isolation. But since the context of the function is pageTitle service then it won't be much of an issue. Nonetheless, I'm changing the name to something less confusing.

PR #1026, opened 2016-10-06
[SD-5515] Changes to add Item History to Item Preview pane

nrvikas:


scripts/apps/authoring/versioning/history/history.js
```diff
@@ -58,7 +60,10 @@ function HistoryController($scope, authoring, api, notify, desks, archiveService
 VersioningHistoryDirective.$inject = [];
 function VersioningHistoryDirective() {
     return {
-        templateUrl: 'scripts/apps/authoring/versioning/history/views/history.html'
+        templateUrl: 'scripts/apps/authoring/versioning/history/views/history.html',
+	scope: {
+            item: '='
```
akintolga: Do we need to isolate the scope here?

scripts/apps/authoring/versioning/history/views/history.html
```diff
@@ -2,7 +2,7 @@
     <ul class="history-list">
         <li class="item"
             ng-class="{selected: selected === version}"
-            ng-repeat="version in versions track by version._current_version"
+            ng-repeat="version in versions"
```
akintolga: If I am not wrong the _current_version should be unique for versions of the same item. So if track by fails here, it maybe because of some other problem.

scripts/apps/authoring/versioning/history/history.js
```diff
@@ -30,7 +30,9 @@ function HistoryController($scope, authoring, api, notify, desks, archiveService
                             $scope.openVersion($scope.last);
                         }
                     }
-                });
+                }, function(err) {
```
akintolga: Under which condition do we get an error here?

nrvikas: When we have an 'ingested' item being previewed, it has no versions and the error occurs. If we do not set $scope.versions to null in this case, the previous item's history is not removed from the display.
Also, currently the change affects items with all states ( not only the published ones ). Kindly let me know if this is OK or we need to have this work done strictly for published items only.

scripts/apps/authoring/versioning/history/history.js
nrvikas: This is needed when an Item is previewed from Legal Archive page ( followed design similar to sd-media-metadata directive ).

scripts/apps/authoring/versioning/history/views/history.html
nrvikas: When we select different items to be previewed, some version revisions ( having same _current_version already on display) are not replaced. As we are now dealing with different items being selected for display, 'track by' was removed so that new item's versions are rendered on the DOM. Kindly suggest/correct me as if I'm missing anything.

scripts/apps/authoring/versioning/history/history.js
mdhaman: Ingested items don't have history. Remove the Item History tab for ingested item.

scripts/apps/monitoring/controllers/MonitoringController.js
```diff
@@ -37,6 +37,7 @@ export function MonitoringController($rootScope, $location, desks) {
     function preview(item) {
         vm.previewItem = item;
         vm.state['with-preview'] = !!item;
+        vm.previewItem.state === 'ingested' ? vm.showHistoryTab = false : vm.showHistoryTab = true;
```
petrjasek: it seems a bit confusing that there is assignment after `?`, can you put it like:
```
vm.showHistoryTab = vm.previewItem.state !== 'ingested';
```

scripts/apps/search/directives/SearchResults.js
```diff
@@ -318,6 +318,8 @@ export function SearchResults(
                 }
                 scope.selected.preview = item;
                 scope.shouldRefresh = false; // prevents $routeUpdate to refresh, just on preview changes.
+                scope.selected.preview.state === 'ingested' ? scope.showHistoryTab = false : scope.showHistoryTab = true;
```
petrjasek: same here

scripts/apps/monitoring/controllers/MonitoringController.js
nrvikas: Hi Peter, I have incorporated these changes in the later commits of the file.

scripts/apps/search/directives/SearchResults.js
nrvikas: Hi Peter, I have incorporated these changes in the later commits of the file.

spec/helpers/authoring.js
```diff
@@ -270,7 +270,7 @@ function Authoring() {
 
     this.showHistory = function() {
         this.showVersions();
-        return element(by.css('[ng-click="tab = \'history\'"]')).click();
+        return  (element(by.id('authoring-container'))).element(by.css('[ng-click="tab = \'history\'"]')).click();
```
petrjasek: seems like there is an extra space after return, thus lint fails

nrvikas: Hi Petr, this is taken care of with this morning's commit.

gbbr:
Please squash all of these commits as you deem logical before merging. There is a lot of duplication and unneeded merge commits that will make the main branch's history hard to follow.

PR #1027, opened 2016-10-06
[SD-5028] - Report original changed errors in desk settings

akintolga:


akintolga:
I moved 412 check to `desks.save`.   Actually this check could be moved further up to http-enpoint-factory.js to enable it for all resources at a later time.

petrjasek:
yep, could be quite generic, just we have to make sure that on reload users won't get lost somehow. so for now I would keep it in desks

PR #1028, opened 2016-10-06
feat(swimlane/column view) - Swimlane view or Column view functionali…

sjunaid:
…ty for monitoring view

- Swimlane view available via switch view button (next to + button) in monitoring view, enabled when configured via config.features with defined columnsLimit.
- Swimlane view stuff is controlled via sd-columns.swimlane, sd-column.swimlane and swimlane css classes 
- Number of columns (i-e groups) displayed is limitTo columnsLimit.
- When "authoring is opened" and an item is selected for preview then only selected item's column would be visible with its preview, on close of preview, all defined number of columns displayed back.
- Columns order can be Re-Arraged via another available button "Re-Arrange Groups" in monitoring, which opens existing monitoring settings modal with only Re-Order tab visible,
all viewable columns are highlighted/greyed in reorder screen.
- Re-Arranged columns/groups (in order) are stored in user preferences, which allows users to have their own prefered display order of columns/groups in monitoring view. 
- when changes like stage activated/deactivated in desk monitoring settings, then column order in monitoring view will reset back to what is defined in desk monitoring setting, until the user Re-Arrange it in monitoring view.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ To configure the build, the `superdesk.config.js` file must export a function th
 - `features.alchemy`: `false` - allow alchemy widget for keywords
 - `features.elasticHighlight`: `false` - allow highlighting of search terms by elasticsearch
 - `features.noTakes`: `false` - disable takes related functionality
+- `features.swimlane`: `null` - enables switch view button in monitoring view, which allows to switch between list view or swimlane view. Example: `features: {swimlane: {columnsLimit: 4}}` will enable switch view button and displays 4 columns when turned ON, set null or keep undefined to disable
 
 ##### Workspace
 - `workspace.content`: `false` - enable content view in workspace (obsolete)
--- a/scripts/apps/monitoring/controllers/AggregateCtrl.js
+++ b/scripts/apps/monitoring/controllers/AggregateCtrl.js
@@ -11,6 +11,9 @@ export function AggregateCtrl($scope, api, desks, workspaces, preferencesService
     this.groups = [];
     this.spikeGroups = [];
     this.modalActive = false;
+    this.displayOnlyCurrentStep = false;
+    this.columnsLimit = null;
+    this.currentStep = 'desks';
     this.searchLookup = {};
     this.deskLookup = {};
     this.stageLookup = {};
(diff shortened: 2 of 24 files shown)


scripts/apps/monitoring/controllers/AggregateCtrl.js
```diff
@@ -99,10 +102,42 @@ export function AggregateCtrl($scope, api, desks, workspaces, preferencesService
...
-                        }
+                        return preferencesService.get(PREFERENCES_KEY)
+                            .then(function(preference) {
+                                var desk = self.deskLookup[activeWorkspace.id];
+
+                                // Read available groups from user preferences first
+                                if (preference && preference[activeWorkspace.id] && preference[activeWorkspace.id].groups) {
+                                    if (desk.monitoring_settings !== undefined) {
+                                        // difference between desk.monitoring_settings and preference[activeWorkspace.id].groups
+                                        var diff = _.xor(_.map(desk.monitoring_settings, '_id'),
```
gbbr: I think you can also do:
```js
let monitoringSettings = desk.monitoring_settings || [];
let activePrefGroups = preference[activeWorkspace.id].groups || [];
let ids = _.uniqBy(monitoringSettings.concat(activePrefGroups), '_id');
```
It should have the same outcome and is more obvious imo.

scripts/apps/monitoring/controllers/AggregateCtrl.js
```diff
@@ -99,10 +102,42 @@ export function AggregateCtrl($scope, api, desks, workspaces, preferencesService
...
+                                             _.map(preference[activeWorkspace.id].groups, '_id'));
+
+                                        if (_.isEmpty(diff)) {
```
gbbr: `if (ids)` should be good enough.

scripts/apps/monitoring/controllers/AggregateCtrl.js
```diff
@@ -99,10 +102,42 @@ export function AggregateCtrl($scope, api, desks, workspaces, preferencesService
...
+                                            // update groups in preferences with any changes in desk's monitoring settings groups.
+                                            _.map(preference[activeWorkspace.id].groups, function(group) {
```
gbbr: Generally `map` is used when we need the return value, which is not the case here, so please use native `forEach`. It's best to avoid using `lodash` since modern JS pretty much offers all you need.

scripts/apps/monitoring/controllers/AggregateCtrl.js
```diff
@@ -99,10 +102,42 @@ export function AggregateCtrl($scope, api, desks, workspaces, preferencesService
...
-                        var desk = self.deskLookup[activeWorkspace.id];
-                        if (desk && desk.monitoring_settings) {
-                            return {'type': 'desk', 'groups': desk.monitoring_settings};
-                        }
+                        return preferencesService.get(PREFERENCES_KEY)
+                            .then(function(preference) {
+                                var desk = self.deskLookup[activeWorkspace.id];
+
+                                // Read available groups from user preferences first
+                                if (preference && preference[activeWorkspace.id] && preference[activeWorkspace.id].groups) {
```
gbbr: Please simplify this condition and try to use less indentation if possible. It would be very helpful. It's a bit difficult to follow.

scripts/core/ui/ui.js
```diff
@@ -1122,16 +1125,59 @@ function splitterWidget(superdesk, superdeskFlags, $timeout) {
...
+                        scope.$applyAsync(function() {
+                            headerContainer.width(stageContainer.width());
+                        });
+                    }
+                }
+            }
+
+            scope.$watch(function() {
+                if (scope.viewColumn) {
+                    resize();
```
petrjasek: this should be probably like `scope.$watch('viewColumn', resize);` otherwise it might call resize in every `$digest`

scripts/apps/monitoring/directives/MonitoringView.js
```diff
@@ -14,20 +14,65 @@ export function MonitoringView($rootScope, authoringWorkspace) {
...
+                return containerElem.scrollTop + containerElem.offsetHeight + 200 >= containerElem.scrollHeight;
+            }
+
+            /**
+             * Trigger render in case user scrolls to the very end of list
+             */
+            function renderIfNeeded($event) {
+                if (isListEnd($event.currentTarget)) {
+                    scope.rendering = scope.loading = true;
+                    $rootScope.$broadcast('render:next');
```
petrjasek: imo this could be `scope.$broadcast` to make it more obvious that it targets some component inside

README.md
```diff
@@ -29,6 +29,7 @@ To configure the build, the `superdesk.config.js` file must export a function th
 - `features.alchemy`: `false` - allow alchemy widget for keywords
 - `features.elasticHighlight`: `false` - allow highlighting of search terms by elasticsearch
 - `features.noTakes`: `false` - disable takes related functionality
+- `features.swimlane`: `object{columnsLimit: +ve_Integer} | null` - enables switch view button in monitoring view, which allows to switch between list view or swimlane view. Example: `features: {swimlane: {columnsLimit: 4}}` will enable switch view button and displays 4 columns when turned ON, set null or keep undefined to disable
```
sjunaid: done

scripts/apps/monitoring/controllers/AggregateCtrl.js
sjunaid: done

scripts/core/ui/ui.js
sjunaid: @petrjasek , done little differently.

sjunaid:
Keeping the scope of this PR limited, the major refactoring can be done later imo.

scripts/apps/monitoring/controllers/AggregateCtrl.js
```diff
@@ -87,25 +90,56 @@ export function AggregateCtrl($scope, api, desks, workspaces, preferencesService
...
             return workspaces.getActiveId()
-                .then(function(activeWorkspace) {
-                    if (self.settings != null && self.settings.desk) {
-                        return deskMonitoringConfig(self.settings.desk);
-                    } else if (activeWorkspace.type === 'workspace') {
-                        return preferencesService.get(PREFERENCES_KEY)
-                            .then(function(preference) {
-                                if (preference && preference[activeWorkspace.id] && preference[activeWorkspace.id].groups) {
-                                    return {'type': 'workspace', 'groups': preference[activeWorkspace.id].groups};
+            .then(function(activeWorkspace) {
```
gbbr: This will become unmaintainable in it's current state, please refactor this to contain less `if` statements and please split it into smaller chunks by creating functions and documenting the flow of execution.

scripts/apps/monitoring/controllers/AggregateCtrl.js
```diff
@@ -87,25 +90,56 @@ export function AggregateCtrl($scope, api, desks, workspaces, preferencesService
...
+
+                                    return {'type': 'desk', 'groups': activePrefGroups};
                                 }
-                                return {'type': 'workspace', 'groups': []};
-                            });
-                    } else if (activeWorkspace.type === 'desk') {
-                        var desk = self.deskLookup[activeWorkspace.id];
+                            }
+
+                            return {'type': 'desk', 'groups': activePrefGroups};
```
gbbr: This same `return` statement is repeated too many times. Can you just create a `groups` variable and assign whatever value is needed and instead only return once at the end? ie:
```js
return {'type': 'desk', 'groups': groups};
```

scripts/apps/monitoring/controllers/MonitoringController.js
```diff
@@ -34,24 +44,33 @@ export function MonitoringController($rootScope, $location, desks) {
...
+        if (!item) {
+            vm.selectedGroup = null;
+        }
     }
 
     function closePreview() {
         preview(null);
         $rootScope.$broadcast('item:unselect');
+        if (vm.viewColumn) {
+            $rootScope.$broadcast('resize:header');
```
gbbr: This doesn't seem like something that should be an _app-wide_ event. I would find another way to do this.

scripts/apps/monitoring/controllers/MonitoringController.js
```diff
@@ -61,4 +80,18 @@ export function MonitoringController($rootScope, $location, desks) {
         vm.singleGroup.singleViewType = null;
         vm.singleGroup = null;
     }
+
+    function switchView(viewColumn) {
```
gbbr: We need to document _all_ functions in the codebase. This will soon be in our contributions guidelines which are in the making

scripts/apps/monitoring/directives/AggregateSettings.js
```diff
@@ -66,6 +66,16 @@ export function AggregateSettings(desks, workspaces, session, preferencesService
                 WizardHandler.wizard('aggregatesettings').next();
             };
 
+            scope.shouldHideStep = function(code) {
```
gbbr: Same here, all functions need to be documented using Angular's `dgeni` format. See [here](https://github.com/wingedfox/dgeni-alive).

scripts/core/ui/ui.js
```diff
@@ -1122,23 +1125,77 @@ function splitterWidget(superdesk, superdeskFlags, $timeout) {
...
     };
 }
 
 /*
+ * Header Resize directive for swimlane view, applies the same width of
+ * stage column to its fixed positioned header in order to resize it.
+ * sd-header-resize used in monitoring-view.html with respective sd-monitoring-group
+ *
+ */
+headerResizeDirective.$inject = ['$rootScope', '$timeout', '$window', 'workspaces'];
```
gbbr: Capital `H` here.

scripts/core/ui/ui.js
```diff
@@ -1122,23 +1125,77 @@ function splitterWidget(superdesk, superdeskFlags, $timeout) {
...
+function headerResizeDirective($rootScope, $timeout, $window, workspaces) {
+    return {
+        link: function(scope, element) {
+            let window = angular.element($window);
+            let resize = _.debounce(calcSize, 250);
+            window.on('resize', resize);
+
+            function calcSize() {
```
gbbr: I'm pretty sure that using CSS correctly will eliminate the need for computations like this and the need for `$rootScope` broadcasting events related to resizing. I am aware that the current state of the layout might make this more difficult, but we should change that to accommodate this change if we want to implement new features, instead of further propagating issues like this.

scripts/apps/monitoring/controllers/MonitoringController.js
```diff
@@ -13,15 +13,25 @@ export function MonitoringController($rootScope, $location, desks) {
     this.viewSingleGroup = viewSingleGroup;
     this.viewMonitoringHome = viewMonitoringHome;
 
+    this.getSwimlaneView = config.features && config.features.swimlane ? 1 : 0;
```
gbbr: I would change this name, `getSwimlaneView` sounds like a function that is "_getting_" something. While this is more like a check, so something like `hasSwimlaneView` would be more appropriate.

scripts/core/ui/ui.js
petrjasek: not sure if this can be solved with css, @pavlovicnemanja @Aca-jov any idea?
there is fixed position header that should fit in parent, whose width might change (based on number of columns).

akintolga:
@petrjasek @gbbr Thanks very much for the reviews.  Syed created another PR for the superdesk repo to enable the swimlanes view where QA can get their hands on.

PR #1029, opened 2016-10-07
Enables JSX  and remove previous planning module

vied12:
Enables JSX

Remove previous planning module


ioanpocol:
This is more like a delete not an add :)

scripts/index.js
```diff
@@ -15,13 +15,21 @@ if (appConfig.features.useTansaProofing) {
 // since tests do not import this file.
 angular.module('superdesk.config').constant('config', appConfig);
 
+// Superdesk modules to be loaded in bootstrap process
```
gbbr: Please undo the changes in this file and remove the `features.planning` item from the configuration. It is no longer needed after https://github.com/superdesk/superdesk-client-core/pull/995.

To enable the `superdesk.planning` module, simply install it with `npm` and add this into `superdesk.config.js`:
```js
apps: ['superdesk-planning']
```

webpack.config.js
```diff
@@ -52,26 +52,31 @@ module.exports = function makeConfig(grunt) {
...
                 {
-                    test: /\.js$/,
+                    test: /\.jsx?$/,
                     exclude: function(p) {
                         'use strict';
                         // exclude parsing node modules, but allow the 'superdesk-core'
                         // node module, because it will be used when building in the
                         // main 'superdesk' repository.
-                        return p.indexOf('node_modules') > -1 && p.indexOf('superdesk-core') < 0;
+                        var allowing = ['superdesk-core', 'superdesk-planning'];
```
gbbr: You can remove these changes too after #995

webpack.config.js
```diff
@@ -164,6 +169,8 @@ function getDefaults(grunt) {
 
         // app features
         features: {
+            // superdesk-planning feature
```
gbbr: This too.

vied12:
I rebased and made the changes. well done, it works like a charm !

ioanpocol:
I think that we have to include & activate EsLint and Jest on package.json

gbbr:
@ioanpocol JEST is not a very good testing framework and Facebook doesn't give much support... We can use our current set-up to test React with some small adjustments. Shouldn't be an issue.

vied12:
Should we merge it or do we still "code freeze" :snowflake:  ?

package.json
```diff
@@ -29,6 +29,7 @@
   "main": "scripts/index.js",
   "dependencies": {
     "angular-history": "git+https://github.com/decipherinc/angular-history.git#v0.8.0",
+    "superdesk-planning": "git+https://github.com/superdesk/superdesk-planning.git#1f04215",
```
gbbr: I wonder if this should be here. If yes, then why not include it straight into client-core if it is already permanently part of it? Do we really need a separate node module given that it is not optional?

It seems like this kind of defeats the whole purpose of having external modules... What do the others thinks?

vied12: I added it here because last time I tried (before your #995) it was impossible to add it at the `superdesk/superdesk` level. Ideally, in my opinion, it would be loaded there in order to be included in integration tests. But the feature must stay option-able in term of a user's point of view.

gbbr: Makes sense. Well, now it can be added there, and into that `superdesk.config.js`. So maybe in this case we should not have it here, right?

vied12: Deal !

gbbr:
@vied12 not a big issue right now, but you might have forgotten to add `html-loader` to `package.json`

vied12:
@gbbr that's right ! It was working for me because I have it in the sd pluggin.
here is the fix: https://github.com/superdesk/superdesk-client-core/pull/1036

PR #1032, opened 2016-10-10
[SD-5547] - Improve performace after scroll lock

akintolga:
- With this change when the scroll position is locked, system will query only the updated item instead of all items if the single item is provided.


akintolga:
Created a function in search service to form a a query by ids.

petrjasek:
:+1:

PR #1033, opened 2016-10-10
[SEARCH] [SCANPIX] restored default subscription

jerome-poisson:
in search parameters, defaut subscription was removed by
commits 6d6e6d961d2d65a730f87a32972548cef61221d5 and 9ead052938a503bd39c428f1bb7454d312b5d3e8.

this commit restore it using angular.values because of the refactoring
done since the former version.

fix SDNTB-288


scripts/apps/search/directives/ItemRepo.js
```diff
@@ -14,6 +14,16 @@ export function ItemRepo(
         link: function(scope, elem) {
 
             /*
+             * function to initialize default values on init or search provider change
+             */
+            scope.setDefaultValues = function() {
```
gbbr: This does not need to be on the scope. Generally, we only put functions that need to be "_two-way bound_" between view and model on the scope as this increases Angular digest cycle processing and reduces app performance. In this case, this is not even used in the view, so please make just a regular function here.

scripts/apps/search/directives/ItemRepo.js
```diff
@@ -14,6 +14,16 @@ export function ItemRepo(
...
+                if (scope.repo && scope.repo.search && scope.repo.search.indexOf('scanpix') === 0) {
```
gbbr: Is `scanpix` always on position `0`?

scripts/apps/search/directives/ItemRepo.js
jerome-poisson: OK thanks. I'm beginner with angular, so thanks for your advice :), pushing an update soon

scripts/apps/search/directives/ItemRepo.js
jerome-poisson: yes, it was a quick workaround to have several Search Providers for the same instance with different login/password (this was not anticipated in the code).

I hope to remove all of this and put in plugins in the future, scanpix should not be hardcoded at all in the client, but this is the way it is done for the moment.

naspeh:
@jerome-poisson your build is green now, you should merge it as soon as possible :)

naspeh:
@gbbr it needs your approvement...

naspeh:
FYI: I just restarted it when it was red

jerome-poisson:
thanks @naspeh ,
help us @gbbr you are our only hope now, you are our only hope now :)

PR #1004, opened 2016-09-29
LBSD-1000 Blog formatting is not saved

vladnicoara:
when editing some contend and only changing markup( making things bold or italic for example), the changes can't be saved

PR #1034, opened 2016-10-11
Revert "LBSD-1000 Blog formatting is not saved"

petrjasek:
Reverts superdesk/superdesk-client-core#1004


petrjasek:
my bad, merged this by mistake and it would break some tests

PR #1039, opened 2016-10-12
[SD-5551] Allow item to be duplicated to user's context and default desks

nrvikas:
This Pull-Request depends on the it's counterpart superdesk-core's pull request:
https://github.com/superdesk/superdesk-core/pull/616 and https://github.com/superdesk/superdesk-core/pull/618

### DIFF
--- a/scripts/apps/archive/index.js
+++ b/scripts/apps/archive/index.js
@@ -128,13 +128,7 @@ angular.module('superdesk.apps.archive', [
                 label: gettext('Duplicate'),
                 icon: 'copy',
                 monitor: true,
-                controller: ['api', 'notify', '$rootScope', 'data', function(api, notify, $rootScope, data) {
-                    api.save('duplicate', {}, {desk: data.item.task.desk}, data.item)
-                        .then(function() {
-                            $rootScope.$broadcast('item:duplicate');
-                            notify.success(gettext('Item Duplicated'));
-                        });
-                }],
+                controller: ctrl.DuplicateController,
                 filters: [{action: 'list', type: 'archive'}],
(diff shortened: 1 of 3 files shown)


scripts/apps/archive/index.js
```diff
@@ -128,8 +128,15 @@ angular.module('superdesk.apps.archive', [
                 label: gettext('Duplicate'),
                 icon: 'copy',
                 monitor: true,
-                controller: ['api', 'notify', '$rootScope', 'data', function(api, notify, $rootScope, data) {
-                    api.save('duplicate', {}, {desk: data.item.task.desk}, data.item)
+                controller: ['api', 'notify', '$rootScope', 'data', 'desks', '$location', 'workspaces', 'session',
```
gbbr: I know you're just extending what has already been here, but since this function has now grown, would you mind adding it into a separate file? For example `apps/archive/controllers/DuplicateController.js` might work.

scripts/apps/archive/index.js
```diff
@@ -128,8 +128,15 @@ angular.module('superdesk.apps.archive', [
...
+                function(api, notify, $rootScope, data, desks, $location, workspaces, session) {
+                    var currentDeskId = null;
```
gbbr: Is `= null;` necessary?

scripts/apps/archive/index.js
nrvikas: Incorporated the changes.

scripts/apps/archive/index.js
nrvikas: Not really. I have removed it in the latest commit.

PR #1041, opened 2016-10-13
fix(templates): let user with templates privilege edit desk templates

petrjasek:
but only on desks where he's a member.

SDESK-203


petrjasek:
@akintolga I won't risk another travis build :wink:

PR #1042, opened 2016-10-14
fix(npm): fix package dependencies

petrjasek:
use dev dependencies only for testing related libs,
others put in dependencies so you get it installed
in superdesk repo. plus remove peerDependencies which
work different with recent npm.

main idea is that in superdesk repo there will be only `superdesk-core` in `package.json`.
so everything to do `grunt build` and `grunt server` must be in dependencies, things we use
here on travis are in devDependencies, and there are no peerDependencies anymore.


petrjasek:
using it at https://github.com/superdesk/superdesk/pull/2346

PR #1043, opened 2016-10-16
SDNTB-205 Translation issues

vladnicoara:


scripts/index.js
```diff
@@ -15,6 +15,11 @@ if (appConfig.features.useTansaProofing) {
     require('apps/tansa');
 }
 
+// load locale for angular
+if (appConfig.ddangular_i18n) {
```
gbbr: Please document this features inside `README.md`. Also, why `ddangular_i18n`?

scripts/index.js
```diff
@@ -15,6 +15,11 @@ if (appConfig.features.useTansaProofing) {
...
+	require('angular-i18n/' + appConfig.angular_i18n);
```
gbbr: Dynamic imports are not possible with webpack (or any other bundler as far as I know). If you do this, everything inside that folder will be imported at compile time and the correct import will be selected at runtime. It's a bit much to include this many files in the project. We have to find a better way to do this.

scripts/index.js
vladnicoara: sorry, I was doing some tests, it's supposed to be angular_i18n because it only work for some angular bootstrap components

PR #1046, opened 2016-10-17
cssRefactoring [toggle buttons]

fritzSF:


styles/sass/buttons/toggle-buttons.scss
```diff
@@ -0,0 +1,201 @@
...
+   .toggle-button, .btn {
+       float:left;
+       margin-right: 6px;
+   } 
+}
+.button-list--padded {
+   padding: 11px 0 11px 20px;
+}
+
+// .toggle-list {
```
gbbr: I think it would be a good idea to make an effort and not commit any commented out code (or styling) into the `master` branch...

fritzSF: You are right, forgot to delete it.

PR #1049, opened 2016-10-17
fix(validate): Prevent validate to modify orig item

ioanpocol:


scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -335,9 +335,11 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
             }
 
             function validate(orig, item) {
+            	var updated = {};
```
gbbr: Is this to prevent cases when `orig` is null? If so, we can assign a default value by changing the function signature as follows:
```js
function validate(orig = {}, item) {
```
This is an ES6 feature we can use.

Either way, the indentation is a bit off here, I think you are missing a space.

ioanpocol: I fixed the indentation.
On helpers.extendItem(orig, item); is changed the orig item
There are cases when the save will not work well because after a change on an item field we will get the same value on orig field.

petrjasek:
imo would be good to refactor it a bit and move that validate function into a service

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
gbbr: In that case please do:
```js
let copy = _.cloneDeep(orig);
 helpers.extendItem(copy, item);
```
instead of `var updated = {}`

ioanpocol: Done

PR #1050, opened 2016-10-17
[SEARCH PROVIDER] [SCANPIX] fixed default values

jerome-poisson:
fixed bad reset of meta in SearchParameters init.

SDNTB-288

### DIFF
--- a/scripts/apps/search/directives/SearchParameters.js
+++ b/scripts/apps/search/directives/SearchParameters.js
@@ -1,7 +1,7 @@
 SearchParameters.$inject = [
-    '$location', 'asset', 'tags', 'metadata', 'meta', 'searchCommon', 'desks', 'userList', 'gettext', 'gettextCatalog', 'ingestSources'
+    '$location', 'asset', 'tags', 'metadata', 'searchCommon', 'desks', 'userList', 'gettext', 'gettextCatalog', 'ingestSources'
 ];
-export function SearchParameters($location, asset, tags, metadata, meta, common, desks, userList, gettext, gettextCatalog, ingestSources) {
+export function SearchParameters($location, asset, tags, metadata, common, desks, userList, gettext, gettextCatalog, ingestSources) {
     return {
         scope: {
             repo: '=',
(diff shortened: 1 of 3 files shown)


scripts/apps/search/directives/SearchParameters.js
```diff
@@ -28,7 +28,12 @@ export function SearchParameters($location, asset, tags, metadata, meta, common,
                 scope.query = params.q;
                 scope.flags = false;
                 scope.common = common;
-                scope.meta = meta = {};
+                scope.meta = meta;
+                for (var member in meta) {
```
gbbr: What exactly are you trying to do here? Something doesn't seem right :-)

jerome-poisson: on the last P.R. (merged this morning), I did try to reset meta to see if it was fixing  the tests, it did indeed, but because I replaced the meta object, the values were not changed anymore (no more double binding), and it was merged like this.

This P.R. fix it by reseting meta instead of replacing the object, and the default values works again.

Actually all of this is because of some removed code a couple of weeks ago, resulting in default values not set anymore and https://dev.sourcefabric.org/browse/SDNTB-288 to fix .

Because code was then splitted in directives in a refactoring, I had to find a way to communicate between itemRepo and SearchParameter, which I did using angular.value, but that broke the tests and since then I want to curse everybody involved in javascript creation and ecosystem.

jerome-poisson: @gbbr checks are passing, so this P.R. actually fix the problem, dunno if there is a better option, but his one works :)

gbbr: I might be wrong though, either way, this is very weird. In case my suggestion above doesn't work, it might be more obvious to do:
```js
Object.keys(meta).forEach(key => delete meta[key]);
```

petrjasek: imo would be simpler if all that logic is in searchParameters, it could watch repo to trigger that setdefault func, so there is no need to split this logic it seems, or am I missing something?

jerome-poisson:
@petrjasek @gbbr I have updated the P.R. and moved meta to searchCommon, this is simplier and the feature still work, crossing finger to see if the tests are passing.

I think this option is acceptable and cleaned than the previous one.

jerome-poisson:
that's with a big emotion that I'm pushing the "merge pull request" button now, thank you javascript people of the world for making so nice and exciting to code something which was doable with a "selected" attribute in the (not so) old time.

PR #1051, opened 2016-10-18
Favicon

darconny:


gbbr:
Please also test this with the build:
```sh
grunt build
cd dist
npm install -g http-server && http-server # or any other HTTP server of choice
```
And see if it works in the browser. I know the behaviour is a build different on deploy.

darconny:
@gbbr Tested and it is working in the browser. Thanks.

PR #1052, opened 2016-10-18
SDESK-233: Date fields in search filters do not reset when value is cleared

thnkloud9:
allow null values in watch to update the field so values can be reset.


thnkloud9:
cool, cool, more for myself to remember why I removed the condition, but I get your point.  removing.

PR #1055, opened 2016-10-19
fix(publish): If item is already published set 'act' correct for vali…

ioanpocol:
…dation


scripts/apps/authoring/authoring/services/AuthoringService.js
```diff
@@ -202,11 +202,15 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
     };
 
     this.validateBeforeTansa = function (orig, diff) {
+        var act = 'publish';
+        if (orig && orig.state === 'published') {
```
mdhaman: After `publish` action, the item could be `corrected` or `killed`.  `AuthoringDirective` knows about the action. Can we pass the action to this method?

PR #1062, opened 2016-10-20
[SDESK-212] Live suggestions

gbbr:
Based on [design](https://wiki.sourcefabric.org/display/NR/Live+Suggestions) from Wiki. The button that enables the feature is found in the _triple-dot_ menu at the top right of any open authoring item.

In this PR:
- new module `apps.authoring.suggest`
- new directive `sd-suggest`
- new service `suggest` for controlling and interacting with live suggestions.
- unit tests
- e2e helper functions and initial test

Backend in progress at https://github.com/superdesk/superdesk-core/pull/609

### DIFF
--- a/.jscs.json
+++ b/.jscs.json
@@ -17,7 +17,6 @@
     "requireLineFeedAtFileEnd": true,
     "requireCommaBeforeLineBreak": true,
     "validateJSDoc": {
-        "checkParamNames": true,
         "requireParamTypes": true
     },
     "requireSpaceBeforeBinaryOperators": ["=", "+", "-", "*", "/", "==", "===", "!=", "!=="],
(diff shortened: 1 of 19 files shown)


gbbr:
@petrjasek I didn't know another way to make those functions available to the template (it didn't make any sense to put them on the scope and make them part of the digest cycle, so I chose a controller), but if you know a better pattern, do let me know.

petrjasek:
you can keep it on controller, so instead of eg `this.scope.ngShow = true` you would set `this.ngShow = true;` and in template you use it as `ctrl.ngShow`, more info [here](https://toddmotto.com/digging-into-angulars-controller-as-syntax/)

PR #1068, opened 2016-10-25
[SDESK-268] Allow complex boolean queries in search

marwoodandrew:


scripts/apps/search/services/SearchService.js
```diff
@@ -337,12 +337,24 @@ export function SearchService($location, gettext, config, session) {
                 criteria.post_filter = {'and': post_filters};
             }
 
-            if (search.q) {
+            var query_string = null;
+            if (search.q && !search.raw) {
+                query_string = search.q.replace(/\//g, '\\/');
+            } else {
```
petrjasek: this part might be a bit more readable as functional:
```
query_string = [search.q, search.raw].filter(q => q).map(q => '(' + q.replace(...) + ')').join(' AND ');
```
but maybe it's not :wink:

scripts/apps/search/directives/RawSearch.js
```diff
@@ -0,0 +1,25 @@
+RawSearch.$inject = ['asset', '$location'];
+export function RawSearch(asset, $location) {
```
gbbr: Please document what this is.

scripts/apps/search/directives/RawSearch.js
```diff
@@ -0,0 +1,25 @@
...
+    return {
+        templateUrl: asset.templateUrl('apps/search/views/raw-search.html'),
+        link: function(scope) {
+            var params = $location.search();
+            scope.rawquery = params.raw;
+
+            var ENTER = 13;
```
gbbr: This can be a const.

scripts/apps/search/directives/RawSearch.js
```diff
@@ -1,12 +1,22 @@
-RawSearch.$inject = ['asset', '$location'];
+/**
+ * @memberof superdesk.apps.search
```
gbbr: Here we need to use `@module` instead of `@memberof` for this to work.

scripts/apps/search/directives/RawSearch.js
```diff
@@ -1,12 +1,22 @@
...
+ * @ngdoc directive
+ * @name RawSearch
```
gbbr: I know this might be a bit counter-intuitive, but we need to use the name that this service is [registered as](https://github.com/marwoodandrew/superdesk-client-core/blob/5eaedb32f72f050d4d8cf9a888682625b1193724/scripts/apps/search/index.js#L46) for documentation to be efficient, in this case `sdRawSearch`. It makes more sense from a documentation perspective.

scripts/apps/search/directives/RawSearch.js
marwoodandrew: Thanks Gabriel, I have made the changes! 
(When looking for an examples of documenting directives I must have chosen badly!)

gbbr: It's ok, what you chose makes more sense, but it won't work unfortunately. I've made the same mistake. I am hoping to have a guide and the documentation website up by end of sprint.

PR #1069, opened 2016-10-25
fix(ingest): Disable open for external and non editable(no fetch available) ingest items

ioanpocol:


scripts/apps/authoring/authoring/services/AuthoringWorkspaceService.js
```diff
@@ -44,6 +44,11 @@ export function AuthoringWorkspaceService($location, superdeskFlags, authoring,
             }
         }.bind(this);
 
+        if (item._type === 'externalsource' && config.features && 'editFeaturedImage' in config.features &&
```
petrjasek: would be good to add a comment here or put this into a function to give it a name, it's clear from PR name what it does but not so much from the code

scripts/apps/authoring/authoring/services/AuthoringWorkspaceService.js
```diff
@@ -44,6 +44,12 @@ export function AuthoringWorkspaceService($location, superdeskFlags, authoring,
             }
         }.bind(this);
 
+        // disable open for external ingest sources that are not editable(fetch not available)
+        if (item._type === 'externalsource' && config.features && 'editFeaturedImage' in config.features &&
```
gbbr: This should also work as:
```js
(item._type === 'externalsource' && !config.features.editFeaturedImage)
```
`config.features` is always there as far as I know, it defaults as object. See [here](https://github.com/superdesk/superdesk-client-core/blob/master/webpack.config.js#L178). Much shorter and clearer imo.

ioanpocol: done

PR #1072, opened 2016-10-26
[SDESK-270] Changes to have Item-List row-height reduced (if configured).

nrvikas:


nrvikas:
Corresponding config changes: https://github.com/superdesk/superdesk-aap/pull/154

scripts/apps/search/directives/ItemList.js
```diff
@@ -808,9 +811,13 @@ export function ItemList(
             });
 
             var ListItemInfo = function(props) {
+                var itemInfoClassName = 'item-info';
+                if (config.reducedItemListRowHeight) {
+                    itemInfoClassName = 'item-info item-info-reduced-rowheight';
```
gbbr: Here, I would prefer:
```js
let cx = 'item-info' + config.list.narrow ? ' item-info-reduced-rowheight' : '';
```
or
```
if (config.list.narrow) {
    cx += ' item-info-reduced-rowheight';
}
```
`cx` is a common used name in React applications for class names. Or at most `className(s)`, imo `itemInfoClassName` is a bit repetitive.

scripts/apps/search/directives/ItemList.js
```diff
@@ -1653,6 +1669,10 @@ export function ItemList(
                     var createItem = function createItem(itemId) {
                         var item = this.state.itemsById[itemId];
                         var task = item.task || {desk: null};
+                        var versionCreatorName;
```
gbbr: You can start off with `null` here and remove `|| null` on line 1691. Alternatively, as a better approach, you could create a function on this component called `displayName` which will return the display_name for a user, or null otherwise, and then call it as `this.displayName(item.version_creator)`.

styles/sass/media-archive.scss
```diff
@@ -606,6 +606,9 @@ $rightfield-width:60px;
             .item-info {
             padding: 10px 30px 10px 10px;
             }
+            .item-info-reduced-rowheight {
+            padding: 5px 10px 5px 10px;
```
gbbr: Spacing issue here.

scripts/apps/search/directives/ItemList.js
```diff
@@ -808,9 +811,13 @@ export function ItemList(
...
 
             var ListItemInfo = function(props) {
+                var itemInfoClassName = 'item-info';
+                if (config.reducedItemListRowHeight) {
+                    itemInfoClassName = 'item-info item-info-reduced-rowheight';
+                }
                 return React.createElement(
                     'div',
-                    {className: 'item-info'},
+                    {className: itemInfoClassName},
```
petrjasek: you can use `classNames` to simplify this class names logic in a few places:
```
{className: classNames('item-info', {'item-info-reduced-rowheight': config.reducedItemListRowHeight})}
```

gbbr: Sorry, ignore my above comment, you'll have to use `classNames` library here, which is already included in the project and is a common React pattern.

At the top of the file `import cx from 'classnames';`

```js
{className: cx('item-info', {'item-info-reduced-rowheight': config.list.narrow}),
```

scripts/apps/search/directives/ItemList.js
```diff
@@ -493,6 +493,9 @@ export function ItemList(
 
             var ListPriority = function(props) {
                 var css = {className: 'list-field urgency'};
+                if (config.reducedItemListRowHeight) {
```
nrvikas: I really liked the idea you have suggested here -  I have renamed the config to config.list.thinRows ( I just thought 'narrow' would indicate narrowing the vertical dimension of the list more than reducing the row height ). Please feel free to comment if you think 'narrow' would still be a better option to go with - will be absolutely happy to change it.

scripts/apps/search/directives/ItemList.js
nrvikas: I noticed displayName as an attribute was being added by React for each class - hence named this to modifiedUserName() instead.

scripts/apps/search/directives/ItemList.js
nrvikas: Done, thanks for the suggestion.

nrvikas: Sure. There already was a classNames import in the file, so I have used the same. Thanks.

scripts/apps/search/directives/ItemList.js
nrvikas: I will come back on this tomorrow regarding the documentation in readme file.

gbbr: Thanks @nrvikas

PR #1079, opened 2016-10-28
SDNTB-205 Prepared Activity Message for translation & .pot updates

vladnicoara:


po/superdesk.pot
```diff
@@ -4697,6 +4753,10 @@ msgstr ""
 msgid "Role"
 msgstr ""
 
+#: scripts/core/lang/missing-translations-strings.js:36
+msgid "role {{role}} is granted new privileges: Please re-login."
```
gbbr: Minor comment, but I think "_has been_ granted" would be more correct.

po/superdesk.pot
```diff
@@ -6202,6 +6274,18 @@ msgstr ""
 msgid "User"
 msgstr ""
 
+#: scripts/core/lang/missing-translations-strings.js:32
+msgid "user {{user}} has been added to desk {{desk}}: Please re-login."
+msgstr ""
+
+#: scripts/core/lang/missing-translations-strings.js:36
+msgid "user {{user}} is granted new privileges: Please re-login."
```
vladnicoara: @gbbr same as above, I will talk to Nelu and see if he can help

po/superdesk.pot
gbbr: Oh, ok, never mind then. It's not that critical, nor is it that wrong. Should be fine :shipit:

PR #1084, opened 2016-11-01
fix(search-results) - Reverted the queryItems changes

akintolga:
Reverting the change I have done to improve the query behavior. We saw few problems during UAT testing that this change affected the results reflected on the list. For instance an item duplicate action took longer than the query so the results were not visible until the next query runs.


akintolga:
@sjunaid has created a separate PR https://github.com/superdesk/superdesk-client-core/pull/1085 for the master branch including these changes

PR #1088, opened 2016-11-01
Update react to latest version

vied12:
and prevent to load it more than once


webpack.config.js
```diff
@@ -52,7 +52,9 @@ module.exports = function makeConfig(grunt) {
                 'angular-embedly': 'angular-embedly/em-minified/angular-embedly.min',
                 'jquery-gridster': 'gridster/dist/jquery.gridster.min',
                 'external-apps': path.join(process.cwd(), 'dist', 'app-importer.generated.js'),
-                'i18n': path.join(process.cwd(), 'dist', 'locale.generated.js')
+                'i18n': path.join(process.cwd(), 'dist', 'locale.generated.js'),
+                // ensure that react is loaded only once (3rd party apps can load more...)
+                'react': path.resolve('./node_modules/react'),
```
gbbr: Could you please help me understand this? What problem are you solving?

vied12: Only after a `grunt build`, the component react-bootstrap I am using in the planning was failing with a ` traverseParentPath(...): Cannot traverse from and to the same ID`. From what I red, this is due to multiple react instances loaded in the app. This alias must ensure that all the imports refer to the same `react`. And now it works after a build :)

PR #1089, opened 2016-11-01
feature(planning): add google api key config

thnkloud9:
will be used by planning component

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ To configure the build, the `superdesk.config.js` file must export a function th
 ##### Services
 
 - `iframely.key` - iframely api key
+- `google.key` - google api key
 - `raven.dsn` - sentry api key
 - `analytics.ga` - google analytics id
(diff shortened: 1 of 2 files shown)


gbbr:
Which Google API is this for? I think they have more than one :-)

thnkloud9:
@gbbr , as far as I am aware google apis are now controlled via developer console and enabled / disabled per app there.  In this particular case, at this time, I am only using maps and places apis, but there is no reason that I could not enable other google apis later for the same key.  At least that has been my understanding of the current google apis paradigm.  However, please correct me if I'm wrong, as I would def. not claim to be an expert here.

gbbr:
Neither am I am expert. I was just wondering whether we could be more specific about what API this is, but since you say that the key is valid for all of their APIs, then this looks good to me.

thnkloud9:
yep, makes sense to add.  should be there now.  thanks!

PR #1095, opened 2016-11-03
[SD-303] - Desk and Fetched desk fields are separated

akintolga:
Separated Fetched Desk and Desk in content list..
- Fetched desk information is creating a new query per item which is affecting the performance.
- Desk information is grabbed from the item so that doesn't have performance impact. Also users require that information.

NTB or master instance shouldn't get affected by this.

### DIFF
--- a/scripts/apps/search/directives/ItemList.js
+++ b/scripts/apps/search/directives/ItemList.js
@@ -28,7 +28,8 @@ const DEFAULT_LIST_CONFIG = {
         'category',
         'provider',
         'expiry',
-        'desk'
+        'desk',
+        'fetchedDesk'
     ]
 };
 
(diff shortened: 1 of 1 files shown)


petrjasek:
@akintolga is this something short term? otherwise I would rather put it on master to avoid regressions, but not sure what you mean by:
> NTB or master instance shouldn't get affected by this.

akintolga:
@petrjasek No it is not short term but we needed a hot fix in production that's why I initially created the PR here in 1.3. I am planning to create a PR in master today as well. What I meant for NTB was I checked their preferred fields in their branch and they won't get affected when this eventually gets into master.

petrjasek:
I see, thx for info @akintolga

PR #1096, opened 2016-11-03
fix(editor): Turn off automatic spellchecking if there is no dictionary

pavlovicnemanja:
SDESK-56 - As a user I don’t want the automatic spell checker to be run if there is no dictionary available for the item’s language

Also, removed old editor, version 1.


vied12:
@gbbr absolutely. But take in consideration that editor3 is on the way :)

PR #1098, opened 2016-11-04
feature(editor3): Initial version of editor3

ioanpocol:
This Editor is based on example editor from https://facebook.github.io
/draft-js/

### DIFF
--- a/scripts/apps/authoring/views/article-edit.html
+++ b/scripts/apps/authoring/views/article-edit.html
@@ -1,4 +1,4 @@
-<div class="field" sd-width="{{editor.headline.sdWidth || 'full'}}" ng-if="schema.headline" ng-class="{'limit-error': item.headline.length > schema.headline.maxlength}" order="{{editor.headline.order}}" sd-validation-error="error.headline" data-required="schema.headline.required">
+<div class="field" sd-width="{{editor.headline.sdWidth || 'full'}}" ng-if="schema.headline && !editor.headline.editor3" ng-class="{'limit-error': item.headline.length > schema.headline.maxlength}" order="{{editor.headline.order}}" sd-validation-error="error.headline" data-required="schema.headline.required">
   <label translate>Headline</label>
     <span ng-if="schema.headline.maxlength" sd-character-count data-item="item.headline" data-html="true" data-limit="schema.headline.maxlength"></span>
     <div id="title" class="headline"
@@ -12,6 +12,19 @@
         ng-if="_editable"
         ng-trim="false"></div>
   <div class="headline" ng-if="!_editable">{{item.headline}}</div>
+</div>
+
+<div class="field" sd-width="{{editor.headline.sdWidth || 'full'}}" ng-if="schema.headline && editor.headline.editor3" ng-class="{'limit-error': item.headline.length > schema.headline.maxlength}" order="{{editor.headline.order}}" sd-validation-error="error.headline" data-required="schema.headline.required">
+  <label translate>Headline</label>
+    <span ng-if="schema.headline.maxlength" sd-character-count data-item="item.headline" data-html="true" data-limit="schema.headline.maxlength"></span>
+    <div id="title" class="headline"
+        sd-editor3
+        data-editor-format="editor.headline.formatOptions"
+        data-language="item.language"
+        data-value="item.headline"
+        data-on-change="autosave(item)"
+        data-read-only="!_editable"
+        data-trim="false"></div>
 </div>
 
 <div class="field" ng-if="schema.sms && item._type !== 'archived'" sd-width="{{editor.sms.sdWidth || 'full'}}" order="{{editor.sms.order}}">
@@ -59,7 +72,7 @@
     data-onchange="autosave(item)"></div>
 </div>
 
-<div class="field abstract" sd-width="{{editor.abstract.sdWidth || 'full'}}" ng-class="{'limit-error': item.abstract.length > schema.abstract.maxlength}" ng-if="schema.abstract" order="{{editor.abstract.order}}" sd-validation-error="error.abstract" data-required="schema.abstract.required">
+<div class="field abstract" sd-width="{{editor.abstract.sdWidth || 'full'}}" ng-class="{'limit-error': item.abstract.length > schema.abstract.maxlength}" ng-if="schema.abstract && !editor.abstract.editor3" order="{{editor.abstract.order}}" sd-validation-error="error.abstract" data-required="schema.abstract.required">
   <label translate>Abstract</label>
     <span sd-character-count
           data-item="item.abstract"
@@ -77,6 +90,23 @@
         ng-if="_editable"
         ng-trim="false"></div>
   <div class="abstract" ng-if="!_editable" sd-html-preview="item.abstract"></div>
+</div>
+
+<div class="field abstract" sd-width="{{editor.abstract.sdWidth || 'full'}}" ng-class="{'limit-error': item.abstract.length > schema.abstract.maxlength}" ng-if="schema.abstract && editor.abstract.editor3" order="{{editor.abstract.order}}" sd-validation-error="error.abstract" data-required="schema.abstract.required">
+  <label translate>Abstract</label>
+    <span sd-character-count
+          data-item="item.abstract"
+          data-html="true"
+          data-limit="schema.abstract.maxlength"></span>
+    <span sd-word-count data-item="item.abstract" data-html="true"></span>
+    <div id="abstract" class="abstract"
+        sd-editor3
+        data-editor-format="editor.abstract.formatOptions"
+        data-language="item.language"
+        data-value="item.abstract"
+        data-on-change="autosave(item)"
+        data-read-only="!_editable"
+        data-trim="false"></div>
 </div>
 
 <div class="field" ng-if="item._type !== 'archived' && schema.byline" sd-width="{{editor.byline.sdWidth || 'full'}}" order="{{editor.byline.order}}" sd-validation-error="error.byline" data-required="schema.byline.required">
@@ -148,7 +178,7 @@
     </div>
 </div>
 
-<div class="field body" ng-if="!isMediaType && schema.body_html" order="{{editor.body_html.order}}" sd-validation-error="error.body_html" data-required="schema.body_html.required" sd-width="{{editor.body_html.sdWidth || 'full'}}">
+<div class="field body" ng-if="!isMediaType && schema.body_html && !editor.body_html.editor3" order="{{editor.body_html.order}}" sd-validation-error="error.body_html" data-required="schema.body_html.required" sd-width="{{editor.body_html.sdWidth || 'full'}}">
   <label translate>Body</label>
     <span sd-character-count data-item="item.body_html" data-html="true"></span>
     <span sd-word-count data-item="item.body_html" data-html="true"></span>
@@ -165,6 +195,20 @@
          data-type="item.type">
     </div>
     <div class="text-editor" sd-html-preview="item.body_html" ng-if="!_editable"></div>
+</div>
+
+<div class="field body" ng-if="!isMediaType && schema.body_html && editor.body_html.editor3" order="{{editor.body_html.order}}" sd-validation-error="error.body_html" data-required="schema.body_html.required" sd-width="{{editor.body_html.sdWidth || 'full'}}">
+  <label translate>Body</label>
+    <span sd-character-count data-item="item.body_html" data-html="true"></span>
+    <span sd-word-count data-item="item.body_html" data-html="true"></span>
+    <div id="bodyhtml"
+         sd-editor3
+         data-editor-format="editor.body_html.formatOptions"
+         data-value="item.body_html"
+         data-read-only="!_editable"
+         data-on-change="autosave(item)"
+         data-language="item.language">
+    </div>
 </div>
 
 <div class="field" ng-if="isMediaType" order="{{editor.media.order}}" sd-width="{{editor.media.sdWidth || 'full'}}">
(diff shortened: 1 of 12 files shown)


scripts/apps/index.js
```diff
@@ -50,5 +50,6 @@ export default angular.module('superdesk.apps', [
     'superdesk.apps.publish',
     'superdesk.apps.templates',
     'superdesk.apps.monitoring',
-    'superdesk.apps.profiling'
+    'superdesk.apps.profiling',
+    'superdesk.apps.editor3',
```
gbbr: This is part of core, so `superdesk.core.editor3`, and load it in `core` please, not apps.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -0,0 +1,224 @@
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import {Editor, EditorState, RichUtils, ContentState} from 'draft-js';
+
+export class Editor3 extends React.Component {
```
gbbr: All of this needs to be documented, module, classes, methods, properties. Please see the contribution guidelines (documentation section)(https://github.com/superdesk/superdesk-client-core/blob/master/CONTRIBUTING.md) for info how to do that.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -0,0 +1,224 @@
...
+    this.onChange(
+      RichUtils.toggleInlineStyle(
+        this.state.editorState,
+        inlineStyle
+      )
+    );
+  }
+
+  render() {
+    const {editorState} = this.state;
```
gbbr: Have we suddetly switched the whole project from 2 space indentation to 4 spaces? We need to keep it consistent.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -0,0 +1,224 @@
...
+    this.readOnly = props.readOnly || false;
+    this.showToolbar = props.showToolbar || false;
+    this.editorFormat = props.editorFormat || [];
+    this.parentOnChange = props.onChange;
+
+    this.focus = () => this.refs.editor.focus();
+    this.onChange = (editorState) => this._onChange(editorState);
+
+    this.handleKeyCommand = (command) => this._handleKeyCommand(command);
+    this.onTab = (e) => this._onTab(e);
```
gbbr: I'm just curious - why are we not defining this methods as public? Why do you use this pattern where you declare a public variable that exposes a private method with no changes?

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -0,0 +1,224 @@
...
+    var initialContentState = ContentState.createFromText(props.value);
+    this.state = {editorState: EditorState.createWithContent(initialContentState)};
+
+    this.readOnly = props.readOnly || false;
+    this.showToolbar = props.showToolbar || false;
+    this.editorFormat = props.editorFormat || [];
+    this.parentOnChange = props.onChange;
+
+    this.focus = () => this.refs.editor.focus();
+    this.onChange = (editorState) => this._onChange(editorState);
```
gbbr: I was wondering why not declare this as a function using the nice ES6 syntax?
```js
onChange(state) {
    this._onChange(state);
}
```
Also, why not just declare `_onChange` as `onChange` directly? Why the wrappers?

akintolga:
@ioanpocol This looks good. You must have worked out the pros and cons of using this editor I believe. Is there any document or any correspondence (or a ticket) for that so we can have a look how it is going to be good for us too?

scripts/core/editor3/components/Editor3.jsx
ioanpocol: This was from Draft.js example and it is a kind on pattern in order to bind different methods to this.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -0,0 +1,224 @@
...
+
+    this.handleKeyCommand = (command) => this._handleKeyCommand(command);
+    this.onTab = (e) => this._onTab(e);
+    this.toggleBlockType = (type) => this._toggleBlockType(type);
+    this.toggleInlineStyle = (style) => this._toggleInlineStyle(style);
```
ioanpocol: see above comment

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -0,0 +1,224 @@
...
+        this.editorFormat = props.editorFormat || [];
+        this.parentOnChange = props.onChange;
+
+        this.focus = () => this.refs.editor.focus();
+        this.onChange = (editorState) => this._onChange(editorState);
+
+        this.handleKeyCommand = (command) => this._handleKeyCommand(command);
+        this.onTab = (e) => this._onTab(e);
+        this.toggleBlockType = (type) => this._toggleBlockType(type);
+        this.toggleInlineStyle = (style) => this._toggleInlineStyle(style);
```
gbbr: Hi @ioanpocol 

I'm not sure I'm convinced by this. It makes no sense. This is the same as writing `this.toggleInlineStyle = this._toggleInlineStyle;`. And if so, why would we do this? Why not just declare `_toggleInlineStyle` directly as `toggleInlineStyle`.

In JS, to my knowledge, methods starting with underscore are an attempt at private methods that should not be tested, which is fine, but in this case, the public method does nothing on top of the private one and this is just extra code which is unnecessary and confusing.

Where did you see this example? Having seen it as an example somewhere doesn't necessarily make it good, we need to know why we are using what we are using, and if there are good arguments, then sure.

Also, I don't quite understand why we would declare methods like this when we have the ES6 syntax available.

I'm ok with keeping these changes if there is a valid explanation, but just because we've seen it somewhere else, doesn't mean it's good. Is this an example from Facebook? If so, please share it.

ioanpocol: The arrow function binds all those functions to "this"
An alternative is to do:
this.toggleBlockType = this._toggleBlockType.bind(this);
or
this.toggleBlockType = this.toggleBlockType.bind(this);
With this pattern it looks a little nicer, depends by test of course.

gbbr: Maybe using `.bind` would be nicer, because then it's actually obvious why we are doing it. But you decide, I am fine with either.

thnkloud9: just FYI, we have been using the bind method in the planning component.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -0,0 +1,224 @@
...
+                            ref="editor"
+                        />
+                    </div>
+            );
+        }
+    }
+}
+
+// Custom overrides for "code" style.
+const styleMap = {
```
vied12: As you created a css file for the editor3, is it possible to put all the style there ?

scripts/apps/authoring/views/article-edit.html
```diff
@@ -167,6 +181,23 @@
     <div class="text-editor" sd-html-preview="item.body_html" ng-if="!_editable"></div>
 </div>
 
+<div class="field body" ng-if="!isMediaType && schema.body_html && editor.body_html.editor3" order="{{editor.body_html.order}}" sd-validation-error="error.body_html" data-required="schema.body_html.required" sd-width="{{editor.body_html.sdWidth || 'full'}}">
+  <label translate>Body</label>
+    <span sd-character-count data-item="item.body_html" data-html="true"></span>
+    <span sd-word-count data-item="item.body_html" data-html="true"></span>
+    <div id="bodyhtml"
+         sd-editor3
+         data-config="{disableEditing: isPublished(item), multiBlockEdition: true, tabindex: editor.body_html.order}"
```
vied12: Does `multiBlockEdition` still makes sense with editor3 ? Maybe this state depends more on the available editor features

ioanpocol: Probably no, I will delete all settings that are not used yet and I will add back if will be needed later.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: OK, so we will use bind version then.

ioanpocol:
@akintolga Internally there were a lot of discussions about creating a new editor based on Draft.js.
You can talk with @petrjasek and @vied12 about this subject.
In short the current editor is not very stable and seams that an editor based on Draft.js will add much stability and it will be much easy to control, customize and extend.
In order to prevent any broke in the current editor, the new editor will be used only if the content types are set to use the new editor. For the integration of the new editor we changed only an html file, so the new editor and old editor have completely separated code base.

petrjasek:
@akintolga it supports highlighting for spellchecking or find/replace, and handles better editing while there are any highlights. also when re-rendering it only renders modified parts so it's much faster. it can also render different highlights at the same time.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: Done

scripts/apps/authoring/views/article-edit.html
ioanpocol: Done

scripts/core/editor3/components/Editor3.jsx
ioanpocol: Used bind instead in order to be more clear

scripts/core/editor3/components/Editor3.jsx
ioanpocol: Used direct bind call instead in order to be more clear

scripts/core/editor3/components/Editor3.jsx
ioanpocol: The example is the one from main draft.js page:
https://facebook.github.io/draft-js/

scripts/core/editor3/index.js
```diff
@@ -0,0 +1,23 @@
+import './styles.scss';
+import * as ctrl from './controllers';
+
+export default angular.module('superdesk.core.editor3', [])
```
gbbr: The module is not documented.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -0,0 +1,226 @@
...
+     * @param {String} language the current language used for spellchecker
+     * @description Editor3 is a draft.js based editor that support customizable formatting (shortly more features will come ...)
+     */
+    constructor(props) {
+        super(props);
+        var initialContentState = ContentState.createFromText(props.value);
+        this.state = {editorState: EditorState.createWithContent(initialContentState)};
+
+        this.readOnly = props.readOnly || false;
+        this.showToolbar = props.showToolbar || false;
```
gbbr: Please use [React propTypes](https://facebook.github.io/react/docs/typechecking-with-proptypes.html) to document all props and their types. This is very helpful and offers type safety. They will also have to be documented.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -0,0 +1,226 @@
...
+    focus() {
+        this.refs.editor.focus();
+    }
+
+    onChange(editorState) {
+        this.setState({editorState});
+        this.parentOnChange(editorState.getCurrentContent().getPlainText());
+    }
+
+    handleKeyCommand(command) {
```
gbbr: Please document all public functions (in JS the pattern is that functions which are not public start with underscore `_`). In this case, all of the methods are public and they need to be documented.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -0,0 +1,226 @@
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import {Editor, EditorState, RichUtils, ContentState} from 'draft-js';
+
+export class Editor3 extends React.Component {
+    /**
+     * @ngdoc react component
```
gbbr: Will this work? `@ngdoc react component`? Need to test.

akintolga:
@ioanpocol @petrjasek Thanks for the explanation. That looks promising.

For us there are few key things:

- The mark-up the editor produces behind the scenes: our publishing has a coupling to the `<p>` tags the 2nd editor produces. We also strip out lots of other mark-up so the change in the mark-up structure may be significant. 
- Handling of plain text story editing is also important. The editor should resist any change in style of the story if we opt for a plain text mode. Especially when a piece of rich text is copied and pasted.
- If we move to this editor, can we still open and edit previously written stories in the system? Mind you we have not started to use content profiles yet.
- All the work done to make the editor-2 stable (the fixes for undo, redo, etc), as far as I can see they are the part of the editor-2 implementation so these features are need to be re-tested.

I'm sure you all aware of these points already. I'm looking forward to play with it to check these points out.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: @gbbr Do you have any example how to document React component
Seams that right now .jsx files are ignored at all when the docs are generated.
If the filter for dgeni-alive grunt task is changed in order to include jsx files I get the error:
error:   Error: No file reader found for /home/ioan/sd/superdesk-client-core/scripts/core/editor3/components/Editor3.jsx

ioanpocol:
@gbbr Do you have any example how to document React component in order to work with dgeni-alive.
Seams that right now .jsx files are ignored at all when the docs are generated.
If the filter for dgeni-alive grunt task is changed in order to include jsx files I get the error:
error: Error: No file reader found for /home/ioan/sd/superdesk-client-core/scripts/core/editor3/components/Editor3.jsx

gbbr:
We'll have to write some code and custom dgeni preprocessors to support this. For now, if you could just make sure everything is documented, including the methods, and adding [`propTypes`](https://facebook.github.io/react/docs/typechecking-with-proptypes.html) should suffice.

I will take care of converting whatever you write to the format that it needs to be.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: Done

scripts/core/editor3/components/Editor3.jsx
ioanpocol: Done

PR #1102, opened 2016-11-07
publisher: manage page layout

tomaszrondio:
that is only main layout for page manager view. Html was changed as well

### DIFF
--- a/scripts/apps/web-publisher/styles/_helpers_variables.sass
+++ b/scripts/apps/web-publisher/styles/_helpers_variables.sass
@@ -1,2 +1,13 @@
 .contentPadding
     padding: 20px
+.rightPadding
+    padding-right: 20px
+
+
+.publisherManager.modal label
+    min-width: auto
+
+.headerMargin
+    margin-top: 48px
+    &--two
+        margin-top: 96px
--- a/scripts/apps/web-publisher/styles/_helpers_variables.sass
+++ b/scripts/apps/web-publisher/styles/_helpers_variables.sass
@@ -1,0 +1,2 @@
+.contentPadding
+    padding: 20px
--- a/scripts/apps/web-publisher/views/manage-site.html
+++ b/scripts/apps/web-publisher/views/manage-site.html
@@ -6,30 +6,32 @@
             <button class="btn btn--primary" ng-disabled="createForm.$invalid || !createForm.$dirty" ng-click="webPublisher.save()" translate>Save</button>
         </div>
     </div>
-    <div class="modal-body manage-page">
-        <ul class="nav nav-tabs nav-stacked">
-            <li ng-class="{active: webPublisher.activeTab === 'general'}">
-                <button
-                    ng-click="webPublisher.changeTab('general')"
-                    translate>General</button>
-            </li>
+    <div class="modal-body sidebarLayout">
+        <div class="sidebarLayout__sidebar">
+            <ul class="nav nav--stackedPublisher">
+                <li ng-class="{active: webPublisher.activeTab === 'general'}">
+                    <button
+                        ng-click="webPublisher.changeTab('general')"
+                        translate>General</button>
+                </li>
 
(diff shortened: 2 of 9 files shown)


scripts/apps/web-publisher/styles/_helpers_variables.sass
```diff
@@ -0,0 +1,45 @@
+$superdesk-green:       rgba(49,172,102,1);
```
gbbr: Is web publisher branding different from Superdesk? If not, let's use the variables already available in [`~variables/_colors.scss`](https://github.com/superdesk/superdesk-client-core/blob/master/styles/sass/variables/_colors.scss). I can see that the `$superdesk-green` is there, although slightly different.

scripts/apps/web-publisher/styles/_helpers_variables.sass
```diff
@@ -0,0 +1,45 @@
...
+
+$sd-text-label:         #909090;
+
+
+$todo-color :           #b70000;
+$inprogress-color :     #cc7800;
+$done-color :           #288001;
+
+.contentPadding
+  padding: 20px
```
gbbr: This syntax without the curly braces is quite cool. We've not used it elsewhere in the project but I'd be happy to adopt it!

Although, we do use 4 space indentation instead of 2, so please amend that.

scripts/apps/web-publisher/views/manage-site.html
```diff
@@ -27,9 +28,10 @@ <h3 class="pull-left" translate>Manage {{new.name}}</h3>
                     translate>Navigation</button>
             </li>
         </ul>
-        <div class="manage-content" ng-switch="webPublisher.activeTab">
-            <form name="createForm">
-                <fieldset ng-switch-when="general">
+    </div>
+        <div class="sidebarLayout__content" ng-switch="webPublisher.activeTab">
+            <form name="createForm" class="contentPadding" ng-switch-when="general">
+                <fieldset >
```
gbbr: Extra space here, inside the tag name

scripts/apps/web-publisher/views/manage-site.html
```diff
@@ -6,8 +6,9 @@ <h3 class="pull-left" translate>Manage {{new.name}}</h3>
             <button class="btn btn--primary" ng-disabled="createForm.$invalid || !createForm.$dirty" ng-click="webPublisher.save()" translate>Save</button>
         </div>
     </div>
-    <div class="modal-body manage-page">
-        <ul class="nav nav-tabs nav-stacked">
+    <div class="modal-body sidebarLayout">
+        <div class="sidebarLayout__sidebar">
+        <ul class="nav nav--stackedPublisher">
```
gbbr: `<ul>` should be indented as a child of the `div`

scripts/apps/web-publisher/styles/_helpers_variables.sass
tomaszrondio: branding is the same but for some reason I can't use global variables defined in variables.scss so I just copied them. There are/will be new elements that might be usefull for other components in superdesk. We can talk about putting them in main styles folder later.

scripts/apps/web-publisher/views/manage-site.html
tomaszrondio: ok

tomaszrondio:
update ready

scripts/apps/web-publisher/styles/_helpers_variables.sass
gbbr: Why can you not get local variables?

gbbr: Superdesk already has its own styling, talk to @fritzSF or @pavlovicnemanja if you need help with branding, I think adding any theming at module level will create problems later.

You just need to `@import '~variables.scss'` - why is that not working?

PR #1105, opened 2016-11-08
[SDESK-290] Added user prefrence desks:preferred to send/publish item

nrvikas:
Depends on server side pull request: https://github.com/superdesk/superdesk-core/pull/645

### DIFF
--- a/scripts/apps/desks/directives/DeskSelect.js
+++ b/scripts/apps/desks/directives/DeskSelect.js
@@ -1,5 +1,12 @@
-DeskSelect.$inject = ['Keys', 'lodash'];
-export function DeskSelect(Keys, _) {
+/**
+ * @ngdoc directive
+ * @module superdesk.apps.desks
+ * @name sdDeskSelect
+ * @requires preferencesService
+ * @description Directive to handle desk selection functionality in drop-downs/menus
+ */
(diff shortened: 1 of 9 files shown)


scripts/apps/desks/directives/DeskSelect.js
```diff
@@ -29,6 +29,14 @@ export function DeskSelect(Keys, _) {
 
             scope.filter = '';
 
+            preferencesService.get('desks:preferred').then(function(result) {
+                scope.preferredDesks = result;
+            });
+
+            scope.deskPreferenceOrdering = function(desk) {
```
gbbr: Do you mind documenting this function please? As well as the `scope.desks` you've added further down in `UserPreferencesDirective`. You'll have to make sure they are displayed correctly in `grunt docs`. See our [contribution guidelines](https://github.com/superdesk/superdesk-client-core/blob/master/CONTRIBUTING.md) for more information.

For them to display on that page, you'll also have to document their host parents, which are `DeskSelect` (remember to use the name of the directive, and not _DeskSelect_, in the documentation), as well as `UserPreferencesDirective`.

I would really appreciate it if you could do this. We have to try and add more and more documentation and for this to happen we could really use some help :-)

nrvikas: Sure, I will. Thanks.

PR #1107, opened 2016-11-08
Fixed issue in filters when switching between presets and dates (SDESK-232)

pavlovicnemanja:
SDESK-232


scripts/apps/search/directives/SearchPanel.js
```diff
@@ -246,20 +246,37 @@ export function SearchPanel($location, desks, privileges, tags, asset, metadata,
                 }
             }
 
+            /*
```
gbbr: This is not the correct documentation syntax. You'll have to check with `grunt docs` if this displays. It might be necessary to document the parents too (SearchPanel). See the examples in the contribution guidelines. Or in the `TranslationService` you wrote 👍

pavlovicnemanja: :+1:

PR #1109, opened 2016-11-09
fix(monitoring) - Prevent monitoring widgets to grow in size

akintolga:
- This also has improvements for single item query for updates

### DIFF
--- a/scripts/apps/search/services/SearchService.js
+++ b/scripts/apps/search/services/SearchService.js
@@ -1,6 +1,18 @@
 import { PARAMETERS, EXCLUDE_FACETS } from 'apps/search/constants';
 
 SearchService.$inject = ['$location', 'gettext', 'config', 'session'];
+/**
+ * @ngdoc service
+ * @module superdesk.apps.search
+ * @name search
+ *
+ * @requires https://docs.angularjs.org/api/ng/service/$injector $location
+ * @requires gettext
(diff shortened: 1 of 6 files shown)


akintolga:
@petrjasek We're planning to not to release 1,4 but wait for 1.5, So can I do only one extra PR to 1.5?

petrjasek:
@akintolga yep should be fine

petrjasek:
those e2e tests are failing due to celery version in test-server, you can fix it via updating core version there

scripts/apps/search/services/SearchService.js
```diff
@@ -584,6 +584,19 @@ export function SearchService($location, gettext, config, session) {
     };
 
     /**
+     * @ngdoc method
+     * @name doesSearchAgainstRepo
```
gbbr: Hi @akintolga, please check with `grunt docs` that this displays properly. Here, I think you have to use `@name search#doesSearchAgainstRepo` to link it to the service.

I think you'll also have to document SearchService for this to show up, would you mind doing that? Otherwise there's really no point in documenting the methods. Remember that the "@name" for SearchService is "search" as it is defined.

akintolga:
@gbbr Merging this PR with this shape, I'll create another one for master branch where I'll add the fixes for the documentation.

PR #1111, opened 2016-11-11
fix(monitoring) - Prevent monitoring widgets to grow in size

akintolga:


akintolga:
@gbbr I checked docs, it looked good to me. Let me know if you see anything missing.

scripts/apps/search/services/SearchService.js
```diff
@@ -592,6 +604,19 @@ export function SearchService($location, gettext, config, session) {
     };
 
     /**
+     * @ngdoc method
+     * @name search#doesSearchAgainstRepo
+     * @public
+     * @return {bool}
```
gbbr: I generally used `Boolean`, but not sure which one is more correct. :-)

scripts/apps/search/services/SearchService.js
```diff
@@ -592,6 +604,19 @@ export function SearchService($location, gettext, config, session) {
...
+     * @description Checks if the given search object will do the search agains the given repo
+     * @param {Object} search search criteria
+     * @param {string} repo name of the repo: ingest, archive, published, archived
```
gbbr: Here, I used to capitalise `String`, but probably yours is more correct since it reflects the output of `typeof` in JS. I guess it's fine to keep it like this.

PR #1113, opened 2016-11-11
fix(build): put version hash into app bundle name

petrjasek:


petrjasek:
@akintolga this adds file hash to app bundle - eg `app.bundle.69bc603c.js`

akintolga:
💯

PR #1114, opened 2016-11-11
Allow React components to be documented (SDESK-331)

gbbr:
This was achieved by copying [dgeni-alive](https://github.com/wingedfox/dgeni-alive) into our repository in `./tasks/dgeni` and adding an extra file reader to allow JSX files, as well as a new type and template for React components.

React components can now be documented the same as directives (for now) using the `@ngdoc React` header.

Component view:

screen shot 2016-11-11 at 12 15 58

Module view:

screen shot 2016-11-11 at 12 15 51

Split into two commits for easy reviewing. First commit is the copy of `dgeni-alive` to `tasks/dgeni`, and the second commit is my changes to the configuration.


petrjasek:
shouldn't we fork dgeni-alive and make those changes there, probably submiting a PR to upstream?
I don't think we should have a copy in our repository

gbbr:
This is the same thing that Nelu was concerned about. I have considered this as well and decided not to. I think we should copy dgeni-alive as a means to have a starting point for a configuration that is specific to our project, and work with it to add more and more features that are appropriate for Superdesk only. 

Keep in mind that the dgeni framework is still an npm module, and this is only a configuration. There was no point to build from scratch so I figured this would be a good start. 

Besides this, a lot more work would be required to add full JSX support. This is just a small step forward.

@petrjasek what do you think?

gbbr:
P.S. We will have to change additional stuff soon to customise the looks of the documentation website, as well potentially add new documentation tags and structures to better reflect our project's module hierarchy (core, apps, sub-modules, etc). That won't be possible without making a 100% custom configuration. So I think copying it is a very good advantage, as opposed to removing it and writing another one from scratch. As it is, dgeni-alive is not customisable enough to add all of these perks.

gbbr:
In time, the plan is to remove a lot of the stuff in there, and eventually have a cleaner config as we progress forward.

Sorry for the many comments.

naspeh:
But you add **176!** files to the repo, it's too much as to me. I totally agreed with @petrjasek.

We could name this fork like `superdesk-cc-dgeni` or so to point that it related to only Superdesk

gbbr:
@naspeh we talked this through on Slack and we have to keep it in the repo. But I will add more commits to this PR and clean up, removing unused files (many of them are not necessary, such as some weird coverage reports). In the end I think 80% of it can be safely removed. So no worries. Yes, this is way too much at the moment.

PR #1117, opened 2016-11-15
Added web publisher module

Aca-jov:


scripts/apps/web-publisher/index.js
```diff
@@ -0,0 +1,35 @@
+import './styles/web-publisher.scss';
+
+import { WebPublisherManagerController } from './controllers';
+import * as services from './services';
+
+export default angular.module('superdesk.apps.web_publisher', [
+    'superdesk.core.activity',
+    'superdesk.config'
+    ])
```
gbbr: I think no spaces here is what we are aiming for.

scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
```diff
@@ -0,0 +1,68 @@
+WebPublisherManagerController.$inject = ['$scope', 'publisher', 'modal'];
+export function WebPublisherManagerController($scope, publisher, modal) {
```
gbbr: Checkout [`SuggestController`](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/suggest/SuggestController.js) for guidance on how to document controllers and their methods.

Please run `grunt docs` inside the client repository to make sure all of this documentation you've written displays correctly.

scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+    /**
+     * Deleting site
+     *
+     * @param {string} code
+     */
+    self.remove = (code) => {
+        modal.confirm(gettext('Please confirm you want to delete website.')).then(
+            () => {
```
gbbr: Just do:
```js
() => publisher.removeSite(code).then(refreshSites)
```
without the curly braces (`{}`), so that the promise gets chained.

scripts/apps/web-publisher/index.js
```diff
@@ -0,0 +1,35 @@
+import './styles/web-publisher.scss';
+
+import { WebPublisherManagerController } from './controllers';
+import * as services from './services';
+
+export default angular.module('superdesk.apps.web_publisher', [
```
gbbr: You need to document this module too, see the [`authoring`](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/authoring/index.js#L12-L18) module as an example on how to do it. If you don't document the module, it's components will not be displayed (controllers, directives, etc) on the documentation page. It works as a hierarchy.

Aca-jov:
Documentation fixed and checked.

scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
```diff
@@ -0,0 +1,79 @@
+/**
+ * @ngdoc controller
+ * @module superdesk.apps.web_publisher
+ * @name WebPublisherManagerController
+ * @requires api/superdesk.apps.web_publisher/service/PublisherFactory publisher 
+ * @requires superdesk.core.services.modal modal
```
gbbr:
```js
@requires modal
```
is enough here, since that is the name of the service. It is requiring a service, not a module.

scripts/apps/web-publisher/services/pub-api-factory.js
```diff
@@ -0,0 +1,120 @@
+/**
+ * @ngdoc service
+ * @module superdesk.apps.web_publisher
+ * @name PubAPIFactory
```
gbbr: I know this may be counter-intuitive, but we need to call this `pubabi` as it is defined in the module. So `@name pubabi`.

scripts/apps/web-publisher/services/pub-api-factory.js
```diff
@@ -0,0 +1,120 @@
...
+            let pubConfig = config.publisher || {};
+            this._base = pubConfig.base || '';
+            this._protocol = pubConfig.protocol || 'http';
+            this._tenant = pubConfig.tenant || 'default';
+            this._domain = pubConfig.domain || '';
+        }
+
+        /**
+         * @ngdoc method
+         * @name PubAPIFactory#setTenant
```
gbbr: Don't forget to update these here with `@name pubabi#setTenant` etc.

scripts/apps/web-publisher/services/publisher-factory.js
```diff
@@ -0,0 +1,72 @@
+/**
+ * @ngdoc service
+ * @module superdesk.apps.web_publisher
+ * @name PublisherFactory
```
gbbr: Same here, `@name publisher`

scripts/apps/web-publisher/services/publisher-factory.js
```diff
@@ -0,0 +1,72 @@
...
+ * @description Publisher service
+ */
+PublisherFactory.$inject = ['pubapi'];
+export function PublisherFactory(pubapi) {
+
+    class Publisher {
+
+        /**
+         * @ngdoc method
+         * @name PublisherFactory#setTenant
```
gbbr: And then don't forget to update here to `@name publisher#setTenant`

scripts/apps/web-publisher/services/publisher-factory.js
```diff
@@ -0,0 +1,72 @@
...
+         * @param {String} tenant
```
gbbr: We need to decide between `String` and `string`. Up to you.

scripts/apps/web-publisher/services/publisher-factory.js
```diff
@@ -0,0 +1,72 @@
...
+         * @return this
```
gbbr: It's `@returns`, not `@return`. And you need to specify a type, so maybe `@returns {Object} the instance for chaining`.

gbbr:
CI is failing because of linting. You may check in advance with `grunt hint`

superdesk.config.js
```diff
@@ -8,6 +8,15 @@
 module.exports = function(grunt) {
     return {
         defaultRoute: '/workspace',
-        requiredMediaMetadata: ['headline', 'description_text', 'alt_text']
+        requiredMediaMetadata: ['headline', 'description_text', 'alt_text'],
+        publisher: {
```
gbbr: It might make more sense to remove the `features.webPublisher` configuration key completely, and instead simply check that this key (`publisher`) is defined. Otherwise it seems a bit duplicated. I mean, if people define this object, then they obviously want to use web publisher. What do you think?

scripts/apps/index.js
```diff
@@ -22,6 +22,7 @@ import 'apps/monitoring';
 import 'apps/settings';
 import 'apps/ingest';
 import 'apps/search-providers';
+import 'apps/web-publisher';
```
gbbr: I think it might be best to remove the `if` condition in `scripts/apps/web-publisher/index.js` line 22 (about the config), and instead do it here, like:
```js
/* globals __SUPERDESK_CONFIG__: true */
const appConfig = __SUPERDESK_CONFIG__;
const withPublisher = typeof appConfig.publisher !== 'undefined';

if (withPublisher) {
    require('apps/web-publisher');
}

    require('apps/tansa');
}
```

And do you mind removing the tansa logic from [scripts/index.js#L16-L18](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/index.js#L16-L18) ? It makes more sense for it to be here.

scripts/apps/index.js
```diff
@@ -50,5 +51,6 @@ export default angular.module('superdesk.apps', [
     'superdesk.apps.publish',
     'superdesk.apps.templates',
     'superdesk.apps.monitoring',
-    'superdesk.apps.profiling'
+    'superdesk.apps.profiling',
+    'superdesk.apps.web_publisher'
 ]);
```
gbbr: And then here we would make this line:
```js
].concat(withPublisher ? 'superdesk.apps.web_publisher' : []));
```

scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
```diff
@@ -0,0 +1,79 @@
...
+            .queryMenus().then(menus => $scope.menus = menus);
+    };
+
+    /**
+    * @ngdoc method
+    * @name WebPublisherManagerController#save
+    * @description Saving site
+    */
+    self.save = () => {
+        publisher.manageSite({tenant: $scope.new}, siteCode).then(refreshSites);
```
gbbr: I think it would make sense to return this value, so that people can do stuff like `save.then(...`.... And so you can know when save completed.

In which case, also `@returns {Promise}`

scripts/apps/web-publisher/services/index.js
```diff
@@ -0,0 +1,2 @@
+export { PublisherFactory } from './publisher-factory';
```
gbbr: We generally use to name these files `PublisherFactory.js` and `PubAPIFactory.js` in the rest of the project. Your approach is not wrong, but this is mostly how `core` is built, and not `apps`. It would be nice if you could follow the same pattern.

PR #1119, opened 2016-11-16
feat(projections) - Projecting the search results

akintolga:
- Search results will return only the necessary fields for the list view to function
- There will be another query when the preview panel is opened to get the entire story

### DIFF
--- a/scripts/apps/search/services/SearchService.js
+++ b/scripts/apps/search/services/SearchService.js
@@ -516,6 +516,23 @@ export function SearchService($location, gettext, config, session) {
     };
 
     /**
+     * Merges the highlighted fields to the item
+     *
+     * @param {Object} item - story
+     * @return {Object} item
+     */
+    this.mergeHighlightFields = function(item) {
+        if (item.es_highlight) {
+            _.forEach(_.keys(item.es_highlight), function (key) {
+                item[key] = item.es_highlight[key][0];
+            });
+        } else {
+            item.es_highlight = [];
+        }
+        return item;
+    };
+
+    /**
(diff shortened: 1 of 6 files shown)


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/658

scripts/apps/search/services/SearchService.js
```diff
@@ -596,6 +604,31 @@ export function SearchService($location, gettext, config, session) {
     };
 
     /**
+     * Returns the query criteria for a single item while keeping
```
gbbr: I'm not sure that this documentation style is correct for the methods. Best way to make sure it's ok is to run `grunt docs` and see if it displays.  Let me know if you've already did and it's working.

I would think you need to take the approach like all the other methods, such as the one below, for example (`doesSearchAgainstRepo`)

scripts/apps/search/services/SearchService.js
```diff
@@ -516,6 +516,23 @@ export function SearchService($location, gettext, config, session) {
     };
 
     /**
+     * Merges the highlighted fields to the item
```
gbbr: I know it's now confusing with many types of documentation syntaxes throughout the project. I'm working on evening it all out and converting it, but in the meantime, we'll need help! Please check using `grunt docs`. I think you need to use the format that is used further down in this file on `this.doesSearchAgainstRepo`. Do you mind correcting?

akintolga:
Closing this PR as opening a replacement for 1.5 (https://github.com/superdesk/superdesk-client-core/pull/1127)

PR #1120, opened 2016-11-16
feat(publish): add content api token to subscriber modal

petrjasek:
initial version - you can only see generated token

SDESK-285

### DIFF
--- a/scripts/apps/publish/views/subscribers.html
+++ b/scripts/apps/publish/views/subscribers.html
@@ -142,6 +142,12 @@
                             </div>
                         </div>
                     </tab>
+
+                    <tab heading="{{ :: 'Content API Token' | translate }}" class="TabContent">
+                        <h3 translate>Authentication Token</h3>
+                        <p><small><code>{{ subscriber.content_api_token }}</code></small></p>
+                    </tab>
+
                 </tabset>
             </fieldset>
         </form>
--- a/test-server/requirements.txt
+++ b/test-server/requirements.txt
@@ -1,4 +1,4 @@
 gunicorn==19.6.0
 honcho==0.6.6
 newrelic>=2.66,<2.67
-git+git://github.com/superdesk/superdesk-core.git@0c0409f#egg=Superdesk-Core
+git+git://github.com/superdesk/superdesk-core.git@f5dbb8b#egg=Superdesk-Core


petrjasek:
yep it failed on travis too, I've restarted it there, but I can't restart the other one right @naspeh ?

naspeh:
you can ask me to restart for now, but I'll add this ability for sure (before disabling travis).

the error is persistent, on both containers, mine and travis:
the same here: https://github.com/superdesk/superdesk-client-core/pull/1116
- https://travis-ci.org/superdesk/superdesk-client-core/builds/176279573
- https://test.superdesk.org/push/sdcpr-1116/76eec7502e/20161116-055700-push.log


> @naspeh is there any way to have friendlier output than plain text? If not, how can one learn more about what's happening here?
@gbbr friendlier, you mean with colors? now it shows
```
Failures:
1) highlights add highlights configuration: highlights management
  Message:
[31m    Failed: unknown error: Element is not clickable at point (529, 143). Other element would receive the click: <input type="text" class="fullwidth-input ng-pristine ng-untouched ng-valid ng-not-empty ng-valid-required" ng-model="configEdit.name" ng-keyup="handleEdit($event);" required="">
      (Session info: chrome=54.0.2840.100)
...
```
so one e2e test is failing...

petrjasek:
@naspeh you could use `aha` on the log to convert it to html without those color sequences

naspeh:
yep, I saw `aha`, I need some simple UI for logs (now just nginx serves a file) but need more like `tail -f` with colors (probably with help of `aha` or https://github.com/ralphbean/ansi2html or https://github.com/drudru/ansi_up)

BTW: you can restart with push new commit :) @petrjasek just added something and they runs again.

naspeh:
@petrjasek have you change something? or travis just passed in the last run without special sauce...

petrjasek:
I've updated server core there.

naspeh:
colors: https://test.superdesk.org/push/sdcpr-1120/7065b6aace/20161117-015433/push.log.htm

gbbr:
Haha, nice one @naspeh ! For some reason this is much friendlier. 

Do you think it would be worthwhile writing something somewhere (in the README or CONTRIBUTING?) about what these mean? ie. what is `push`? what is `web`? I'm just thinking how we could debug potential issues.

Let me know if my suggestion is pointless or if the scope of these integrations should be obvious. It might be just me lacking knowledge.

naspeh:
@gbbr: I need to document all things if it needs we could add that to README too after we switch to new checks.

PR #1123, opened 2016-11-18
[SDESK-314] - UI issues in swimlane when an item is open in authoring

sjunaid:
- Covers SDESK-314, SDESK-318 and SDESK-320


fritzSF:
All the changes done here will be obsolete once I finish the UI changes for this view. I will close this PR as it will be overwritten anyway.

PR #1124, opened 2016-11-18
feat(desks): Added ability to add default language for desks

pavlovicnemanja:
SDESK-35 - As an admin I want to set a default language for all new desk content, this one will override the SD default if it is a different one

SDESK-47 - As a user I want to create new articles in a language set as default for my desk

### DIFF
--- a/scripts/apps/desks/views/desk-config-modal.html
+++ b/scripts/apps/desks/views/desk-config-modal.html
@@ -48,6 +48,17 @@
                                 </div>
                             </div>
                             <div class="field">
+                                <label translate>Desk language</label>
+                                <div class="line-select">
+                                    <select id="deskLanguage" ng-model="desk.edit.desk_language">
+                                        <option value=""></option>
+                                        <option ng-repeat="language in languages" value="{{language.language}}" ng-selected="desk.edit.desk_language === language.language">
+                                            {{ language.label | translate}}
+                                        </option>
+                                    </select>
+                                </div>
+                            </div>
+                            <div class="field">
                                 <label translate>SERVICE</label>
                                 <div sd-meta-terms
                                     class="data"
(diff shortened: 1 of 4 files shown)


scripts/apps/desks/views/desk-config-modal.html
```diff
@@ -48,6 +48,12 @@ <h3 ng-show="desk.edit._id" translate>Edit "{{desk.edit.name}}" Desk</h3>
                                 </div>
                             </div>
                             <div class="field">
+                                <label translate>Desk language</label>
+                                <input type="text" class="line-input fullwidth-input"
```
petrjasek: imo there should be a select box with some predefined values, we need those to match with dictionary language code, and also with language you can set in authoring, and this would be useful everywhere

pavlovicnemanja: ok, so, maybe I can add list of languages that we have set in dictionaries? I was thinking about this, but on some planning we had discussion about this and conclusion was to left input box...

What you think about setting language list from dictionaries?

petrjasek: what about using languages from [settings](https://github.com/superdesk/superdesk-ntb/blob/master/server/settings.py#L194)?
I wouldn't get it from dictionaries, those should use same component to select language imo

pavlovicnemanja: yep, maybe you are right. Should I display all languages selected there, or just languages that are source (have setted source as true)?

petrjasek: I would display them all for now

pavlovicnemanja: :+1: adding!

PR #1127, opened 2016-11-21
feat(projections) - Projecting the search results

akintolga:
- Search results will return only the necessary fields for the list view to function
- There will be another query when the preview panel is opened to get the entire story

### DIFF
--- a/scripts/apps/search/services/SearchService.js
+++ b/scripts/apps/search/services/SearchService.js
@@ -1,4 +1,9 @@
-import {PARAMETERS, EXCLUDE_FACETS} from 'apps/search/constants';
+import {
+    PARAMETERS,
+    EXCLUDE_FACETS,
+    CORE_PROJECTED_FIELDS,
+    UI_PROJECTED_FIELD_MAPPINGS,
+    DEFAULT_LIST_CONFIG} from 'apps/search/constants';
 /**
  * @ngdoc service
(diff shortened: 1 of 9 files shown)


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/663

akintolga:
@gbbr Thanks for the points in the previous PR. I have updated the comments. For some reason though when I run grunt docs I don't see the details of the three functions documented in search service. I compared to tags service where it has two functions documented and I can see the details. I'll keep looking.

scripts/apps/search/services/SearchService.js
```diff
@@ -524,6 +523,25 @@ export function SearchService($location, gettext, config, session) {
     };
 
     /**
+     * @ngdoc method
+     * @name search#mergeHighlightFields
+     * @public
+     * @description Merges the highlighted fields to the item
+     * @param {Object} item
+     * @return {Object}
```
gbbr: I think `@returns` is the correct form. It might be the reason it's not working? Not sure...

akintolga:
@petrjasek I have changed the list of fields for projection to infer them from the `config.list`. There is a list of `CORE_PROJECTED_FIELDS`  that is required regardless. The others will be inferred from `UI_PROJECTED_FIELD_MAPPINGS` . So we don't need a separate config field for the projected fields.

PR #1131, opened 2016-11-21
fix(publish): Fix translation issue for publish

ioanpocol:


vladnicoara:
It fixes the problem with the different messages

PR #1132, opened 2016-11-22
[SDESK-290] Added user help text regarding preferred desks selection.

nrvikas:


scripts/apps/users/views/user-preferences.html
```diff
@@ -89,6 +89,9 @@
                 <label>
                     {{ :: preferences['desks:preferred'].label | translate}}
                   </label>
+                  <p class="discreet" translate>
+                    By selecting the desks as your preferred desks, they appear first in the order when sending or publishing items.
+                  </p>git
```
gbbr: Is this `git` test intended? Looks out of place. Ignore my message if I'm wrong.

nrvikas: @gbbr you have sharp eyes. This text, I believe made its way because I was entering some git commands on another terminal on my machine. It's really in a wrong place. Sorry, I will fix that.

PR #1138, opened 2016-11-23
Added routes management

Aca-jov:
SWP-337, SWP-338, SWP-339, SWP-340


Aca-jov:
Oh, sorry, this is on publisher branch, only for web-publisher we're working on. I'll add tickets in descritpion.
e2e tests will be added later, for now it's fine.

scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
```diff
@@ -29,8 +41,8 @@ export function WebPublisherManagerController($scope, publisher, modal) {
      * @description Opens modal window for creating new site
      */
     self.toogleCreateSite = () => {
```
gbbr: I'm just wondering - why are some functions defined as arrow functions and others not? Is it to bind to `this`? You are already using self here ...

Could potentially be much cleaner as a class...

Aca-jov: Y, i missed that, i changed all functions to be arrow functions.

PR #1139, opened 2016-11-23
 fix(swimlane-update) - Design update for the swim lane view

fritzSF:


pavlovicnemanja:
Merged with @naspeh approval.

PR #1141, opened 2016-11-25
fix(swimlane-update) - Fix some remaining ui issues, SDESK-318 and SDESK-320

sjunaid:


scripts/apps/monitoring/styles/monitoring.scss
```diff
@@ -200,6 +206,30 @@
       }
     }
   }
+  &:not(.authoring) {
```
gbbr: Can we do something about this? A simpler version would be nicer.

fritzSF:
This will collide with the new design of the swimlane view. I'm closing this PR, if the problem still exists we will take care of this.

PR #1143, opened 2016-11-25
refactor search providers

petrjasek:
- read available types and labels from new api
- use search_provider id as param, thus allow multiple instances of same
  type

### DIFF
--- a/scripts/apps/search-providers/service.js
+++ b/scripts/apps/search-providers/service.js
@@ -1,13 +1,8 @@
 export default function SearchProviderService(providerTypes, $filter, api, allowed) {
     return {
-        getAllowedProviderTypes: function() {
-            return allowed.filterKeys(providerTypes, 'search_providers', 'search_provider');
-        },
-        getSearchProviders: function() {
-            return api.search_providers.query({}).then(
-                (result) => $filter('sortByName')(result._items, 'search_provider')
-            );
-        }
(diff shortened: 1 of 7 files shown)


petrjasek:
client changes for https://github.com/superdesk/superdesk-core/pull/674

scripts/apps/search/directives/SearchResults.js
```diff
@@ -11,6 +11,12 @@ SearchResults.$inject = [
     'config'
 ];
 
+let HEX_REG_EXP = /[a-f0-9]{24}/;
```
gbbr: Since it's caps, maybe a `const` is better. Also, it doesn't seem to be needed outside the `isObjectId` function, so maybe it would be better in there?

scripts/apps/search-providers/service.js
```diff
@@ -1,7 +1,10 @@
 export default function SearchProviderService(providerTypes, $filter, api, allowed) {
+
     return {
         getAllowedProviderTypes: function() {
-            return allowed.filterKeys(providerTypes, 'search_providers', 'search_provider');
+            return api.get('search_providers_allowed').then(function(allowedTypes) {
```
gbbr: You'll probably need to change to arrow here, now with the new linting rules.
```js
.then((allowedTypes) => allowedTypes._items);
```

PR #1144, opened 2016-11-28
ESLint: Variables

gbbr:
ESLint [variables section](http://eslint.org/docs/rules/#variables).

### DIFF
--- a/scripts/apps/archive/controllers/UploadController.js
+++ b/scripts/apps/archive/controllers/UploadController.js
@@ -153,7 +153,7 @@ export function UploadController($scope, $q, upload, api, archiveService, sessio
                 item.upload.abort();
             }
         }
-        if (index !== undefined) {
+        if (typeof index !== 'undefined') {
             $scope.items.splice(index, 1);
         }
         if (_.isEmpty($scope.items)) {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -108,7 +108,12 @@
         "radix": 2,
         "yoda": 2,
 
-        "strict": 0,
+        "no-shadow-restricted-names": 2,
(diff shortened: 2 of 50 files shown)


scripts/apps/archive/controllers/UploadController.js
```diff
@@ -153,7 +153,7 @@ export function UploadController($scope, $q, upload, api, archiveService, sessio
                 item.upload.abort();
             }
         }
-        if (index !== undefined) {
+        if (index) {
```
petrjasek: imo `index` can be `0`

gbbr:
@petrjasek the problem is that a key inside an object defined as `undefined`, is still a defined key. As in, it exists in the prototype. I'm ok with removing this rule if you are ok with taking this risk. What do you think?

gbbr:
You can read more [here](http://eslint.org/docs/rules/no-undefined).

petrjasek:
I see. I can imagine keeping this but then I wouldn't disable `== null`, because those 2 together make things quite complex. but lets wait what others think.

gbbr:
`=== null` is staying if that's what you mean. As for `== null` could be an issue because `undefined == null` :-)

petrjasek:
I mean `== null` to check for both `null` and `undefined`. this would become:
`x === null || typeof x === 'undefined'`

gbbr:
Yeah, but we have `_.isNil(v)` which does just that.

gbbr:
@petrjasek I'm closing this. I don't feel very strongly about it and it's already used too much in our codebase. It's hard to factor it out and make all the tests pass and if you don't fully agree with it then I am ok with dropping it.

PR #1149, opened 2016-11-29
ESLint: ECMAScript 6

gbbr:
Rules from [ES6 section](http://eslint.org/docs/rules/#ecmascript-6) of ESLint.


petrjasek:
personally I would only use `const` together with `UPPER_CASE`, otherwise you can't tell if something is a constant or not. code like that seems counter-intuitive to me.

gbbr:
I think the idea here is, to quote from the ESLint [page](http://eslint.org/docs/rules/prefer-const):
> If a variable is never reassigned, using the `const` declaration is better.
> 
> `const` declaration tells readers, “_this variable is never reassigned_”, reducing cognitive load and improving maintainability.

In my opinion this is very true and can deem helpful when reading the code of others. You know that an identifier was declared only as a simplification (and not as a moving part) when you see it as a `const`.

On the other hand, I do agree that app-specific configuration and more "_global_" constants should be upper-cased and that is a good practice, which unfortunately there is no rule for. But I think we can enforce that through pull requests. This rule and what you are saying are two different things in my opinion.

If you think this would be too confusing, it's easy for me to remove it, and I don't feel that strongly about it, but in my opinion it's a good thing to have.

gbbr:
P.S. This is now becoming a "_best practice_" even in new React applications and can be seen in many examples, including our code-base, check out some of the Editor3 [code](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/components/Editor3.jsx#L205).

petrjasek:
> const declaration tells readers, “this variable is never reassigned”, reducing cognitive load and improving maintainability.

that's my issue with it, you have to check declaration to see if something is const or not, but you can't tell if you only look at some part of it. like if you read diff on github

gbbr:
@petrjasek fine. Indeed it would be nice to enforce some naming convention for `const`'s. Personally I don't see a problem either way but I can understand how this could be too restrictive for some. I have removed that commit.

gbbr:
What do you think of the rest? /cc @vied12 maybe you'd like to give some input as well.

scripts/apps/archive/related-item-widget/relatedItem.js
```diff
@@ -71,10 +71,11 @@ function RelatedItemController(
...
     function today() {
         if (config.search && config.search.useDefaultTimezone) {
-            return moment().tz(config.defaultTimezone).format('YYYY-MM-DD') +
-                'T00:00:00' + moment.tz(config.defaultTimezone).format('ZZ');
+            return `${moment()
+                .tz(config.defaultTimezone)
+                .format('YYYY-MM-DD')}T00:00:00${moment.tz(config.defaultTimezone).format('ZZ')}`;
         }
-        return moment().format('YYYY-MM-DD') + 'T00:00:00' + moment().format('ZZ');
+        return `${moment().format('YYYY-MM-DD')}T00:00:00${moment().format('ZZ')}`;
```
gbbr: Removed

PR #1152, opened 2016-11-29
fix(api): detect duplicate requests

petrjasek:
and serve such requests from cache.

SDESK-361

### DIFF
--- a/scripts/core/api/api-service.js
+++ b/scripts/core/api/api-service.js
@@ -1,3 +1,5 @@
+var angular = require('angular');
+
 /**
  * @ngdoc provider
  * @name apiProvider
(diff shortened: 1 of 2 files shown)


scripts/core/api/api-service.js
```diff
@@ -52,13 +54,41 @@ function APIProvider() {
...
+
+                if (config.method !== 'GET') {
+                    return $http(config);
+                }
+
+                let now = Date.now();
+                let key = config.url + angular.toJson(config.params || {});
+                let last = cache[key] || null;
+                if (last && now - last.now < 100) {
+                    console.warn('duplicate request',
```
gbbr: Won't this create some amount of spam in the console? We have quite a lot of occurrences of duplicate requests.

scripts/core/api/api-service.js
```diff
@@ -52,13 +54,41 @@ function APIProvider() {
...
-                return $http(config);
+
+                if (config.method !== 'GET') {
+                    return $http(config);
+                }
+
+                let now = Date.now();
+                let key = config.url + angular.toJson(config.params || {});
+                let last = cache[key] || null;
+                if (last && now - last.now < 100) {
```
gbbr: In my opinion `100` should be a const at the top, so it's clear and easy to config. It would also give the value meaning.

scripts/core/api/api-service.js
petrjasek: actually not that many, but the idea is that we should avoid these, so imo it should be visible

PR #1154, opened 2016-11-30
fix(authoring): make min tansa urgency configurable

petrjasek:
and change default to be 5.

SDNTB-341

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.js
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.js
@@ -462,12 +462,14 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
                 }
             }
 
+            let minTansaUrgency = $rootScope.config.tansa && $rootScope.config.tansa.urgency || 5;
+
             /**
              * Depending on the item state one of the publish, correct, kill actions will be executed on the item
              * in $scope.
              */
             $scope.publish = function(continueOnPublish) {
-                if ($scope.useTansaProofing() && $scope.item.urgency > 3 && !isCheckedByTansa) {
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -462,12 +462,14 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
                 }
             }
 
+            let minTansaUrgency = $scope.config.tansa && $scope.config.tansa.urgency || 5;
```
gbbr: Should we put `5` at the top in `const DEFAULT_URGENCY = 5` or something? So it's easy to spot and config if needed.

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -462,12 +462,14 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
                 }
             }
 
+            let minTansaUrgency = $rootScope.config.tansa && $rootScope.config.tansa.urgency || 5;
```
gbbr: Another option is for 5 to live in the defaults, over at the default [config](https://github.com/superdesk/superdesk-client-core/blob/master/webpack.config.js#L119). Not sure if that's a better option, but it would completely remove this line and you could use `$rootScope.config.tansa.urgency` directly.

petrjasek: I will keep it close to that usage, imo those whole ansa specific part should be eventually moved away

gbbr:
Yeah. It would look weird there

PR #1133, opened 2016-11-23
SWP-447 : Merging and testing navigation

naomiaro:

PR #1158, opened 2016-12-01
Added listing routes on web site management

Aca-jov:
Also, fixed and added:
https://github.com/superdesk/superdesk-client-core/pull/1133

### DIFF
--- a/scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
+++ b/scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
@@ -11,6 +11,10 @@ WebPublisherManagerController.$inject = ['$scope', 'publisher', 'modal'];
 export function WebPublisherManagerController($scope, publisher, modal) {
     var self = this;
 
+    $scope.routes = [];
+    $scope.menu = undefined;
+    $scope.menus = [];
+    $scope.menuObj = {};
     self.TEMPLATES_DIR = 'scripts/apps/web-publisher/views';
 
     /**
@@ -40,9 +44,20 @@ export function WebPublisherManagerController($scope, publisher, modal) {
      * @description Opens modal window for creating new site
      */
     self.toogleCreateSite = () => {
+        self.openSiteModal = !self.openSiteModal;
+        if (!self.openSiteModal && self.selectedSite.subdomain) {
+            /**
+             * @ngdoc event
+             * @name WebPublisherManagerController#refreshRoutes
+             * @eventType broadcast on $scope
+             * @param {String} subdomain - subdomain for which to refresh routes
+             * @description event is thrown when modal window is closed and saved site is selected
+             */
+            $scope.$broadcast('refreshRoutes', self.selectedSite.subdomain);
+        }
+
         self.selectedSite = {};
         $scope.newSite = {};
-        self.openSiteModal = !self.openSiteModal;
         publisher.setTenant('default');
     };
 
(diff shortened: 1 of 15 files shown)


pavlovicnemanja:
:+1:

scripts/apps/web-publisher/services/PublisherFactory.js
```diff
@@ -81,11 +81,12 @@ export function PublisherFactory(pubapi) {
         /**
          * @ngdoc method
          * @name publisher#queryRoutes
+         * @param {Object} type - which routes to query
          * @returns {Promise}
          * @description Query routes
```
gbbr: This description doesn't really say anything about what this function does.

scripts/apps/web-publisher/directives/SiteRoutesDirective.js
```diff
@@ -0,0 +1,38 @@
...
+    class SiteRoutes {
+
+        constructor() {
+            this.scope = {site: '=site'};
+            this.templateUrl = 'scripts/apps/web-publisher/views/site-routes.html';
+        }
+
+        link(scope) {
+
+            scope.$on('refresh:routes', function(e, data) {
```
gbbr: Why not `scope.$on('refresh:routes', queryItems)`?

scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
```diff
@@ -44,6 +45,9 @@ export function WebPublisherManagerController($scope, publisher, modal) {
         $scope.newSite = {};
         self.openSiteModal = !self.openSiteModal;
         publisher.setTenant('default');
+        if (!self.openSiteModal) {
+            $rootScope.$broadcast('refresh:routes');
```
gbbr: imo this should not be `$rootScope`. I would limit it to the $scope of the `SiteRoutesDirective` or find another way to communicate this change. `$rootScope` events concern the whole application, while this only concerns a mere directive.

scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
```diff
@@ -11,6 +11,10 @@ WebPublisherManagerController.$inject = ['$scope', 'publisher', 'modal'];
 export function WebPublisherManagerController($scope, publisher, modal) {
     var self = this;
```
gbbr: I wouldn't use `self` if it isn't needed. But I am ok with leaving it. Just a comment.

scripts/apps/web-publisher/services/PubAPIFactory.js
```diff
@@ -52,6 +64,23 @@ export function PubAPIFactory(config, $http, $q) {
...
+         * @param {String} resource
+         * @param {Number} id
+         * @returns {Promise}
+         * @description GET a given resource by id.
+         */
+        get(resource, id) {
+            return this.req({
+                url: this.resourceURL(resource, id),
+                method: 'GET'
+            }).then(response => {
```
gbbr: This `.then` clause - why is it needed? Doesn't the promise already return it's response?

scripts/apps/web-publisher/directives/SiteRoutesDirective.js
```diff
@@ -0,0 +1,40 @@
...
```
gbbr: I would document this event. Take a look [here](https://github.com/wingedfox/angular-route-segment/blob/master/src/route-segment.js#L679-L687) to see how events can be documented. Please check with `grunt docs` that it shows up.

scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
Aca-jov: I'll rework controller as class, for now its like this.

scripts/apps/web-publisher/directives/SiteRoutesDirective.js
```diff
@@ -0,0 +1,47 @@
...
+             * @param {Object} data - subdomain for which to refresh routes
+             * @description event is thrown when modal window is closed and saved site is selected in WebPublisherManagerController
+             */
+            scope.$on('refreshRoutes', function(e, data) {
+                if (data === scope.site.subdomain) {
+                    queryItems();
+                }
+            });
+
+            function queryItems() {
```
gbbr: Since you've created an ES6 class (which is cool!) I think we can now take benefit of declaring these methods as part of the class, and even documenting them. A common pattern is naming it `_queryItems` and documenting it as `@private`.

scripts/apps/web-publisher/directives/SiteRoutesDirective.js
```diff
@@ -0,0 +1,47 @@
...
+                scope.loading = true;
+                publisher.setTenant(scope.site.subdomain);
+                publisher.queryRoutes({type: 'collection'}).then(routes => {
+                    scope.loading = false;
+                    scope.routes = routes;
+                });
+            }
+
+            queryItems();
```
gbbr: Then here, you would call `this._queryItems`

gbbr:
LGTM

PR #1160, opened 2016-12-01
[SDESK-264] - Marking stories for desks

akintolga:
- Made stories can be marked/unmarked for desks (_as they are marked/unmarked for highlights_)
- Changed 'output stage' query to include marked stories.
- Created a new search parameter to return marked stories for a desk
- Consolidated (most) common directives for highlights and marked_desks


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/681

scripts/apps/desks/directives/DesksReactDropdown.js
```diff
@@ -0,0 +1,64 @@
...
+ * @requires noDesksLabel
+ *
+ * @param {Object} [desks] collection of desks
+ *
+ * @description Creates dropdown react element with list of available desks
+ */
+
+import React from 'react';
+
+DesksReactDropdown.$inject = ['item', 'className', 'desks', 'gettext', 'noDesksLabel'];
```
gbbr: Can we turn this whole file into an actual React component implemented as an ES6 class that extends `React.Component`. We can document it using `@ngdoc React`. Also, you can safely use JSX, it works fine. No need to programatically create the elements.

Also, please use `propTypes` to document component properties.

scripts/apps/desks/services/DesksFactory.js
```diff
@@ -1,5 +1,5 @@
-DesksFactory.$inject = ['$q', 'api', 'preferencesService', 'userList', 'notify', 'session', '$filter'];
-export function DesksFactory($q, api, preferencesService, userList, notify, session, $filter) {
+DesksFactory.$inject = ['$q', 'api', 'preferencesService', 'userList', 'notify', 'session', '$filter', 'privileges'];
+export function DesksFactory($q, api, preferencesService, userList, notify, session, $filter, privileges) {
```
gbbr: Can we document this factory and at the very least the new methods?

scripts/apps/search/directives/ItemList.js
```diff
@@ -668,6 +668,147 @@ export function ItemList(
                 }
             });
 
+            var MarkedDesksList = React.createClass({
```
gbbr: Same here, can we use ES6 classes, JSX, and document all of this? Also, please put this in a separate file. `ItemList` is already one of our biggest files and it's begging for a refactor :)

scripts/apps/archive/controllers/ArchiveListController.js
```diff
@@ -125,6 +125,7 @@ export class ArchiveListController extends BaseListController {
         $scope.$on('content:update', refreshItems);
         $scope.$on('item:deleted', refreshItems);
         $scope.$on('item:highlight', refreshItems);
+        $scope.$on('item:marked_desks', refreshItems);
```
gbbr: Just FYI, not sure if you knew this, or if it applies here, but you can (and imo should) document events using `@ngdoc event` directive like [here](https://github.com/wingedfox/angular-route-segment/blob/master/src/route-segment.js#L679-L687). They should probably be documented at their origin though (where they are broadcasted).

scripts/apps/search/directives/ItemList.js
akintolga: @gbbr I agree with your point. But if it is possible I'd like look into the ItemList file in a separate PR after this. Please let me know if that is is ok with you.

scripts/apps/desks/directives/DesksReactDropdown.jsx
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @description Creates dropdown react element with list of available desks
+ */
+import React from 'react';
+
+DesksReactDropdown.$inject = ['item', 'className', 'desks', 'noDesksLabel'];
+export function DesksReactDropdown(item, className, desks, noDesksLabel) {
+    /** Dropdown component */
+    class MarkDesksDropdown extends React.Component {
+        render() {
```
gbbr: I would do:
```js
render() {
    const items = desks.desks._items;

    return (
        <ul className={this.props.className}>
            {items.length ?
                items.map((d) => (<MarkedDeskItem key={d._id} desk={d} item={item}/>))
                : <li><button disabled="true">{noDesksLabel}</button></li>}
        </ul>
    );
}
```
seems much simpler

scripts/apps/desks/directives/DesksReactDropdown.jsx
```diff
@@ -0,0 +1,79 @@
...
+            return (
+                <ul className={this.props.className}>
+                    {items}
+                </ul>
+            );
+        }
+    }
+
+    /** Button component for marking items for desks */
+    class MarkBtn extends React.Component {
```
gbbr: Each component should be in its own file. There is a pending PR which will be merged soon that will disallow this via linting anyway, so might as well do it now.

scripts/apps/desks/directives/DesksReactDropdown.jsx
```diff
@@ -0,0 +1,79 @@
...
+            return (
+                <button disabled= {this.isMarked} onClick={this.markDesk}>
+                    <i className="icon-bell" ></i>{this.desk.name}
+                </button>
+            );
+        }
+    }
+
+    /** List item created per desk */
+    class MarkedDeskItem extends React.Component {
```
gbbr: I think this can also be written as:
```js
const MarkedDeskItem = (props) =>
    <li key={'desk-' + this.props.desk._id}>
        <MarkBtn item={this.props.item} desk={this.props.desk} />
    </li>;
```

scripts/apps/desks/directives/DesksReactDropdown.jsx
```diff
@@ -0,0 +1,79 @@
...
+ *
+ *
+ * @description Creates dropdown react element with list of available desks
+ */
+import React from 'react';
+
+DesksReactDropdown.$inject = ['item', 'className', 'desks', 'noDesksLabel'];
+export function DesksReactDropdown(item, className, desks, noDesksLabel) {
+    /** Dropdown component */
+    class MarkDesksDropdown extends React.Component {
```
gbbr: This component should be in a different file. You can pass angular injectables via props for now.

scripts/apps/search/directives/ItemList.js
gbbr: I think we should do it within this PR because moving 3 classes into a separate file, converting them to JSX and simply importing them here to be used as normal shouldn't take more than 30-45 minutes imo. For Angular's injectables we can send them via props. If there are any serious issues with doing this that I've omitted, please let me know and in that case we can leave it for another PR.

scripts/apps/archive/directives/MarkedItemTitle.js
```diff
@@ -0,0 +1,105 @@
...
+MarkedItemTitle.$inject = ['desks', '$timeout', 'authoring', 'highlightsService'];
+export function MarkedItemTitle(desks, $timeout, authoring, highlightsService) {
+    return {
+        scope: {
+            item: '=item',
+            markField: '@field'
+        },
+        templateUrl: 'scripts/apps/archive/views/marked_item_title.html',
+        // todo(petr): refactor to use popover-template once angular-bootstrap 0.13 is out
+        link: function(scope, elem) {
```
gbbr: This link function is way too long imo and is hard to follow the logic. Can we maybe create a class and split all this logic into multiple smaller (documented) functions?

For directives you need to `return new ClassName()` and it will work fine. See [this example] from web publisher.

scripts/apps/archive/directives/MarkedItemTitle.js
```diff
@@ -0,0 +1,105 @@
...
+             * @param {string} isOpen
+             */
+            scope.toggleClass = function(isOpen) {
+                scope.open = isOpen;
+            };
+
+            scope.marks = scope.item[scope.markField];
+            scope.hasMarkItemPrivilege = authoring.itemActions(scope.item).mark_item_for_desks;
+
+            scope.$on('item:' + scope.markField, ($event, data) => {
```
gbbr: This, and the `if` below show that you are trying to implement something that is in fact two directives. So why not have two? This is very error prone.

I think that instead of using jQuery (which we shouldn't use in an Angular app) event listeners we can make a service to communicate to other parts of the application, instead of clicking around using the `.click()` method. I would not merge this PR in its current state.

petrjasek: I agree that this switching is error prone, but I'm also curious how to make this 2 without code duplication which is imo error prone too, any suggestions @gbbr ? if this would be class, maybe inheritance?

gbbr: I didn't look into it that deeply, but if it presents issues I'm happy to help. I think the critical part is breaking this down into smaller parts and documenting it so it's easy to follow. As well as finding an alternative way to introducing the functionality without use of jQuery (if possible) and use of `.click()`.

akintolga: Agree with @petrjasek that unless we use some sort of inheritance we'll either have duplication or one or two if statements. I'll leave it like this for now.

akintolga:
@gbbr Thanks a lot your review Gabriel, this PR helped me to understand the new structure in a better way. There's only one comment I haven't addressed, it's for `MarkedItemTitle` so I'll keep this for another day, if it is not merged (tests might fail as well) I'll look into that as well, but if you think this can be merged then I'll do another PR for that one...

scripts/apps/archive/directives/MarkedItemTitle.js
```diff
@@ -0,0 +1,105 @@
...
+
+            elem.on({
+                click: function(e) {
+                    e.preventDefault();
+                    e.stopPropagation();
+                },
+                mouseenter: function(e) {
+                    popup = $(this).find('.highlights-list');
+                    popup.not('.open').children('.dropdown-toggle')
+                    .click();
```
gbbr: I'm ok with keeping the link function like this. But can we find a better, more programatic way, of opening the dropdown? Without `click()`-ing around? Maybe a service? An event that the dropdown directive listens too (if there is one) ?

scripts/apps/desks/components/MarkBtn.jsx
```diff
@@ -0,0 +1,41 @@
...
+        super(props);
+        this.item = props.item;
+        this.desk = props.desk;
+        this.desks = props.desks;
+        this.isMarked = this.item.marked_desks &&
+            this.item.marked_desks.some((md) => md.desk_id === this.desk._id);
+        this.markDesk = this.markDesk.bind(this);
+    }
+
+    markDesk() {
```
gbbr: Can we document the methods here?

scripts/apps/desks/components/MarkBtn.jsx
```diff
@@ -0,0 +1,41 @@
...
+ * @name MarkBtn
+ * @param {Object} item the story to be marked
+ * @param {Object} desk desk for the item to be marked
+ * @param {Object} desk desksService
+ * @description Represents a button for given item to be marked for the given desk
+ */
+export class MarkBtn extends React.Component {
+    constructor(props) {
+        super(props);
+        this.item = props.item;
```
gbbr: These properties (`this.item`, `this.desk`, `this.desks`) they are not used anywhere. Am I missing something?

scripts/apps/desks/components/MarkBtn.jsx
```diff
@@ -0,0 +1,41 @@
...
+        return (
+            <button disabled= {this.isMarked} onClick={this.markDesk}>
+                <i className="icon-bell" />{this.desk.name}
+            </button>
+        );
+    }
+}
+
+MarkBtn.propTypes = {
+    desks: React.PropTypes.object.isRequired,
```
gbbr: Would be nice to document props too, otherwise it will be hard to reuse these components.

scripts/apps/desks/directives/DesksReactDropdown.jsx
```diff
@@ -0,0 +1,23 @@
...
+ * @ngdoc directive
+ * @module superdesk.apps.desks
+ * @name DesksReactDropdown
+ *
+ * @requires React
+ * @requires item
+ * @requires className
+ * @requires desks
+ * @requires noDesksLabel
+ *
```
gbbr: Extra blank line here.

scripts/apps/desks/directives/index.js
```diff
@@ -14,3 +14,5 @@ export {DeskeditMacros} from './DeskeditMacros';
 export {ActionPicker} from './ActionPicker';
 export {DeskSelect} from './DeskSelect';
 export {DeskSelectItem} from './DeskSelectItem';
+export {MarkDesksDropdown} from './MarkDesksDropdown';
+export {DesksReactDropdown} from './DesksReactDropdown.jsx';
```
gbbr: No need for the `.jsx` extension.

scripts/apps/search/components/MarkedDesksList.jsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import {
+    closeActionsMenu
+} from 'apps/search/helpers';
+
+export class MarkedDesksList extends React.Component {
```
gbbr: We should try and commit to documenting everything. This component has been omitted.

scripts/apps/search/components/MarkedDesksList.jsx
```diff
@@ -0,0 +1,102 @@
...
+                    desks.hasMarkItemPrivilege() ? React.createElement(
+                        'button',
+                        {className: 'btn btn--mini', onClick: this.removeMarkedDesk(desk)},
+                        gettext('REMOVE')
+                    ) : null
+                );
+            }
+        }.bind(this);
+
+        var items = [
```
gbbr: Why not JSX?

scripts/apps/search/services/TagService.js
```diff
@@ -111,63 +111,73 @@ export function TagService($location, desks, userList, metadata, search, ingestS
...
+            }
+
+            var codeList = metadata.values[cv.list];
+            var selecteditems = JSON.parse(index);
+
+            processSelectedItems(selecteditems, codeList);
+        });
+    }
+
+    var fieldProcessors = {
```
gbbr: We should try and document these...

scripts/apps/desks/components/MarkBtn.jsx
akintolga: They are used in line 25...

scripts/apps/desks/components/MarkBtn.jsx
akintolga: Any example of it somewhere?

scripts/apps/search/components/MarkedDesksList.jsx
akintolga: Isn't it the case for all the other jsx files you commited yesterday or am I missing something?

scripts/apps/desks/directives/DesksReactDropdown.jsx
akintolga: Done.

scripts/apps/desks/components/MarkBtn.jsx
gbbr: No examples yet, sorry. But anything you find reasonable, as long as it contains a description that makes it easier to understand what they stand for should do.

I still need to add functionality into dgeni to display these on the site, and once I do I will convert them to the correct format, as long as you add the description. Even just a simple inline comment is ok for now.

scripts/apps/search/components/MarkedDesksList.jsx
gbbr: The "all other JSX files I committed yesterday" are not new. They are coming from the "old" `ItemList.js` which I've split up into multiple files as part of a refactor.

New components that go into the codebase should be documented.

While I would love to spend time and document all of the old components too, that would probably take me many months of work. These requirements are part of our contribution guidelines and I don't understand what the problem is.

scripts/apps/search/components/MarkedDesksList.jsx
```diff
@@ -0,0 +1,90 @@
...
+                            {gettext('REMOVE')}</button> : null}
+                    </li>)
+            }
+                </ul>
+        );
+    }
+}
+
+/**
+ * svc: contains $timeout and desks and is required
```
gbbr: It's enough that you've documented these in the `params` section of the component. Sorry if I didn't notice this before. This doesn't hurt though! Thanks for adding!

PR #1161, opened 2016-12-02
fix(search): Keep repo filters when clearing searches

pavlovicnemanja:
SDESK-286 - Clicking the "Clear filters" button switches user to standard SD search from Search provider

### DIFF
--- a/scripts/apps/search/directives/SaveSearch.js
+++ b/scripts/apps/search/directives/SaveSearch.js
@@ -39,7 +39,11 @@ export function SaveSearch($location, asset, api, session, notify, gettext, $roo
             scope.clear = function() {
                 scope.editingSearch = false;
                 scope.edit = null;
-                $location.url($location.path());
+                _.each($location.search(), (item, key) => {
+                    if (key !== 'repo') {
+                        $location.search(key, null);
+                    }
+                });
             };
 
             scope.search = function() {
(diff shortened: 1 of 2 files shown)


naspeh:
eslint is failing :)
```
/opt/superdesk/client-core/scripts/apps/search/directives/SaveSearch.js
  42:44  error  Unexpected function expression                prefer-arrow-callback
  42:52  error  Unexpected space before function parentheses  space-before-function-paren
```
https://test.superdesk.org/push/sdcpr-1161/783035f59c/20161202-111901/check-npmtest.log

petrjasek:
it seems that it actually broke that e2e test

scripts/apps/search/directives/SaveSearch.js
```diff
@@ -39,7 +39,11 @@ export function SaveSearch($location, asset, api, session, notify, gettext, $roo
             scope.clear = function() {
                 scope.editingSearch = false;
                 scope.edit = null;
-                $location.url($location.path());
+                _.each($location.search(), (item, key) => {
+                    if (key !== 'repo') {
```
vied12: maybe a little comment there to know why you skip 'repo'

scripts/apps/search/directives/SaveSearch.js
```diff
@@ -39,7 +39,11 @@ export function SaveSearch($location, asset, api, session, notify, gettext, $roo
...
+                        $location.search(key, null);
```
vied12: instead of calling several times `$location.search`, I think you could set an object as parameter

petrjasek: imo calling it many times is not a problem, it will only change url once per digest, this way it seems quite readable

PR #1170, opened 2016-12-08
[SDESK-455] fix(search): Fixing aap multimedia filters for type and credit.

mdhaman:


scripts/apps/search/directives/SearchPanel.js
```diff
@@ -306,6 +306,11 @@ export function SearchPanel($location, desks, privileges, tags, asset, metadata,
                         _.find(scope.tags.selectedFacets[type], (facet) => facet.value === key);
                 }
 
+                if (type === 'credit') {
+                    return scope.tags && scope.tags.selectedFacets[type] &&
```
gbbr: Can we just put:
```js
if (!scope.tags || !scope.tags.selectedFacets[type]) {
    return false;
}
```
on line 304 and remove it from all the other conditions? It would make things a lot more readable and less repetitive.

gbbr: I would rewrite this function as:
```js
scope.hasFilter = function(type, key) {
    if (!scope.tags || !scope.tags.selectedFacets[type]) {
        return false;
    }

    const facet = scope.tags.selectedFacets[type];
    const isDesk = type === 'desk' && _.find(facet, (f) => f.value === key);
    const isCredit = type === 'credit' && facet.indexOf(scope.aggregations.credit[key].qcode) >= 0;
    
    return isDesk || isCredit || facet.indexOf(key) >= 0;
}
```
What do you think? You can keep whichever one you prefer, but I would remove the duplicate condition as suggested in the above comment either way.

mdhaman: Agreed. Done the change.

scripts/apps/search/directives/SearchPanel.js
```diff
@@ -300,14 +300,25 @@ export function SearchPanel($location, desks, privileges, tags, asset, metadata,
                 return date ? moment(date).format('YYYY-MM-DD') : null; // jshint ignore:line
             };
 
+            /**
+             * @ngdoc method
+             * @name hasFilter
```
mdhaman: cool. will change that.

mdhaman:
@petrjasek can't add this PR to 1.5 due to refactoring for search.

petrjasek:
@mdhaman if you need this on 1.5 we can backport it, let me know

PR #1174, opened 2016-12-08
content lists UI + css updates

tomaszrondio:


gbbr:
Please use "_Squash and merge_" when merging.

PR #1175, opened 2016-12-08
CI: fix angular version to 1.5.9

naspeh:
The new 1.6.0 is break "npm test" and "protractor"

`Angular` and `Protractor` what is wrong with you today?

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   ],
   "main": "scripts/index.js",
   "dependencies": {
-    "angular": "^1.5.5",
+    "angular": "1.5.9",
     "angular-animate": "^1.5.5",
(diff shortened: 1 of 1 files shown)


naspeh:
I also tried to change this:
```
     jasmineNodeOpts: {
         showColors: true,
-        defaultTimeoutInterval: 200000
+        defaultTimeoutInterval: 30000
     },
```
After that e2e tests were failing by this timeout, so I removed this change and keep only angular version, but seems it works, I checked already on #1170 (it was run as development build with this change for angular). Anyway, I'll wait for green.

gbbr:
Yeah, that's a long time, but there's probably a good reason for it ...

PR #1179, opened 2016-12-12
[SDESK-419] Support duplication of text archived items

nrvikas:
Depends on core's pull request: https://github.com/superdesk/superdesk-core/pull/690

### DIFF
--- a/scripts/apps/archive/controllers/DuplicateController.js
+++ b/scripts/apps/archive/controllers/DuplicateController.js
@@ -7,7 +7,7 @@ export function DuplicateController(api, notify, $rootScope, data, desks, $locat
     } else {
         currentDeskId = desks.getCurrentDeskId();
     }
-    api.save('duplicate', {}, {desk: currentDeskId}, data.item)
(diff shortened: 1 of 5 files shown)


scripts/apps/search/directives/ItemList.js
```diff
@@ -718,11 +718,11 @@ export function ItemList(
                                 ));
 
                     var elem = React.createElement('div',
-                        {
-                            className: 'dropdown__menu more-activity-menu',
-                            onMouseOver: this.toggle,
-                            onMouseLeave: this.close
-                        },
+                                                   {
```
gbbr: What happened here? 😄

nrvikas: Hi Gabriel, I got this error with lint. Hence added the indentations as lint expected.
ss

nrvikas: But, it fails for the exact opposite reason now. So, I'm going to remove it :)
ss_today

PR #1180, opened 2016-12-12
[SDESK-270] Narrow view when we have preview and authoring open

nrvikas:
Note: Works in conjunction with the thinRows configuration as per the documentation in README.md

### DIFF
--- a/scripts/apps/search/directives/ItemList.js
+++ b/scripts/apps/search/directives/ItemList.js
@@ -32,7 +32,8 @@ ItemList.$inject = [
     'config',
     '$interpolate',
     'metadata',
-    'storage'
+    'storage',
+    'superdeskFlags'
 ];
 
 export function ItemList(
@@ -61,7 +62,8 @@ export function ItemList(
     config,
     $interpolate,
     metadata,
-    storage
+    storage,
+    superdeskFlags
 ) {
     // contains all the injected services to be passed down to child
     // components via props
@@ -91,7 +93,8 @@ export function ItemList(
         config: config,
         $interpolate: $interpolate,
         metadata: metadata,
-        storage: storage
+        storage: storage,
+        superdeskFlags: superdeskFlags
     };
 
     return {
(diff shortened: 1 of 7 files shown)


petrjasek:
@gbbr can you elaborate on this?

> This doesn't seem like something that should live in global config, but more like something that should be configured within the UI.

I'm not sure what you mean.

gbbr:
@petrjasek these `thinRows` and `narrowView` settings seem like theme changes that should be enabled in the UI (Superdesk settings) and not as a build setting. If we do decide to include them as a build setting (superdesk.config.js), then we should maybe at least group them together into an object that is specific to the theme. Possibly called `theme` ?

gbbr:
Should we possibly add a section to the Settings page that addresses these theme settings on a user basis? I'm just thinking that the superdesk.config.js might become too complex if we keep adding minor tweaks like this.

petrjasek:
afaik this is not user specific, so imo makes sense as a global config.

gbbr:
Sure. If that's the case then maybe I'm just lacking knowledge about this aspect. Although I would still refrain from using `$rootScope` so freely, unless there's any good reason too.

scripts/apps/search/directives/ItemList.js
```diff
@@ -1950,6 +1967,10 @@ export function ItemList(
                     }
                 });
 
+                if (listConfig.thinRows) {
+                    scope.$on('item:previewed', listComponent.setNarrowView);
```
petrjasek: imo there is no need to put here this `if`, in case this is not configured it shouldn't fire that event, so it seems like duplication

nrvikas:
@gbbr , @petrjasek I appreciate your inputs. I will discuss major changes well ahead from next time so that we are all on the same page. I have responded to other review comments on this PR specifically to avoid confusion. Thanks gain.

scripts/apps/search/directives/ItemList.js
nrvikas: Yes, I have  made the required changes. Please let me know if it needs any corrections.

PR #1181, opened 2016-12-12
Split ItemList.js [SDESK-470]

gbbr:
Splits `ItemList.js` by putting all of the React components into separate files in the `components` folder. Angular injectables and scope are shared via props. Some common functions are in `helpers.js`.

React classes have also been converted to ES6 and propTypes were added to everything.

Folder structure of `apps/search/components` folder looks like this:
```
.
├── DesksDropdown.jsx
├── ErrorBox.jsx
├── FetchedDesksInfo.jsx
├── GridTypeIcon.jsx
├── HighlightsInfo.jsx
├── HighlightsList.jsx
├── Item.jsx
├── ItemContainer.jsx
├── ItemList.jsx
├── ItemPriority.jsx
├── ItemUrgency.jsx
├── ListItemInfo.jsx
├── ListPriority.jsx
├── ListTypeIcon.jsx
├── MediaInfo.jsx
├── MediaPreview.jsx
├── ProgressBar.jsx
├── SelectBox.jsx
├── TimeElem.jsx
├── TypeIcon.jsx
├── actions-menu
│   ├── ActionsMenu.jsx
│   ├── Divider.jsx
│   ├── Item.jsx
│   ├── Label.jsx
│   └── index.js
├── fields
│   ├── broadcast.jsx
│   ├── category.jsx
│   ├── desk.jsx
│   ├── embargo.jsx
│   ├── expiry.jsx
│   ├── fetchedDesk.jsx
│   ├── flags.jsx
│   ├── headline.jsx
│   ├── highlights.js
│   ├── index.js
│   ├── priority.jsx
│   ├── profile.jsx
│   ├── provider.jsx
│   ├── signal.jsx
│   ├── slugline.js
│   ├── state.jsx
│   ├── takekey.jsx
│   ├── takepackage.jsx
│   ├── update.jsx
│   ├── updated.jsx
│   ├── urgency.jsx
│   ├── versioncreated.jsx
│   ├── versioncreator.jsx
│   └── wordcount.js
└── index.js

2 directories, 50 files
```


petrjasek:
just wondering - is this `svc` thing some common pattern? don't like that part too much, but not sure there is a better way (maybe using `$injector`?)

gbbr:
@petrjasek I have no idea. Never seen this before, but it was the simplest way to accomplish this task. If we were to use `$injector`, we would still have to pass it down via the props, and then there is no difference really, except with `$injector` we would have to write extra code to obtain all these dependencies in each component.

My reasoning was that the goal of this was to split up the file, and then find potential improvements later (there are many). But I'm open to other ideas if you have any.

gbbr:
I have given it some thought and I think you are right. I will try to implement a version using only the $injector. This way we can reuse any of these components by passing only that, and not having to put in all the special dependencies for each component. The component itself can obtain them internally.

PR #1184, opened 2016-12-14
feat(subscribers): implement token manager

petrjasek:
- add generate token form
- add remove token button

SDESK-437

### DIFF
--- a/scripts/apps/publish/controllers/SubscriberTokenController.js
+++ b/scripts/apps/publish/controllers/SubscriberTokenController.js
@@ -1,0 +1,55 @@
+let moment = require('moment');
+
+/**
+ * @ngdoc controller
+ * @module superdesk.apps.publish
+ * @name SubscriberTokenController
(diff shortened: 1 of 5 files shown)


scripts/apps/publish/controllers/SubscriberTokenController.js
```diff
@@ -0,0 +1,23 @@
+
+SubscriberTokenController.$inject = ['$scope', 'api'];
```
gbbr: Can we please document this?

naspeh:
now there is error in `grunt docs`, maybe I should disable this `fire:check-docs`, nobody seems are using it...

gbbr:
> now there is error in grunt docs, maybe I should disable this fire:check-docs, seems nobody is using it...

@naspeh no, I think we should all be using it, otherwise it was a waste of time to implement it. I vote YES for documentation 🚓

PR #1185, opened 2016-12-14
feature(Editor3): Add support for spellcheck on Editor3

ioanpocol:


scripts/core/editor3/components/DraftSpellchecker.jsx
```diff
@@ -0,0 +1,240 @@
...
+ * @description check the text and shos the inline spellchecker context menu
+ */
+
+import React from 'react';
+import {EditorState, Modifier} from 'draft-js';
+import {InlineSpellcheckMenu} from '../components/InlineSpellcheckMenu';
+
+export class DraftSpellchecker {
+    constructor(language) {
+        this.language = language;
```
gbbr: We should document these attributes too, even if they seem self-explanatory.

scripts/core/editor3/components/DraftSpellchecker.jsx
```diff
@@ -0,0 +1,240 @@
...
+        this.getContextMenu = this.getContextMenu.bind(this);
+
+        this.ignoreWord = this.ignoreWord.bind(this);
+        this.addWord = this.addWord.bind(this);
+        this.replaceWord = this.replaceWord.bind(this);
+    }
+
+    /** Register the context menu related actions */
+    registerContextMenuCallbacks(onOpenContextMenu, onCloseContextMenu, onReplaceWord) {
+        this.onOpenContextMenu = onOpenContextMenu;
```
gbbr: I would initialise all of these listeners in the constructor as `null`, to be able to document them and to offer a better overview of the attributes of this class. It makes it easier to understand the behaviour if all the attributes are listed in one place.

scripts/core/editor3/components/DraftSpellchecker.jsx
```diff
@@ -0,0 +1,240 @@
...
+        this.onCloseContextMenu = onCloseContextMenu;
+        this.onReplaceWord = onReplaceWord;
+    }
+
+    /** Register the dictionary related actions */
+    registerDictionaryCallbacks(ignoreWord, addWord, getSuggestions, checkText) {
+        this.onIgnoreWord = ignoreWord;
```
gbbr: Same for these ones.

scripts/core/editor3/components/DraftSpellchecker.jsx
```diff
@@ -0,0 +1,240 @@
...
+        if (this.getSuggestions) {
+            this.getSuggestions(this.wordSelection.word).then((suggestions) => {
+                this.suggestions = suggestions;
+                this.show = true;
+                if (this.onOpenContextMenu) {
+                    this.onOpenContextMenu();
+                }
+            });
+        } else {
+            this.suggestions = [];
```
gbbr: Lines 187 to 191 are the same as 180 to 184 so I would create a function here:
```js
const show = (s) => {
    this.suggestions = s;
    this.show = true;

    if (this.onOpenContextMenu) {
        this.onOpenContextMenu();
    }
};
```
and then use that:
```js
if (this.getSuggestions) {
    this.getSuggestions(this.wordSelection.word).then(show);
} else {
    show([]);
}
```
to avoid repetition...

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -39,39 +41,150 @@ export class Editor3 extends React.Component {
...
+        this.onChange(newEditorState);
+    }
+
+    /** Return if the spellchecker is initialized/active */
+    isActiveSpellchecker() {
+        return !!this.spellchecker;
+    }
+
+    /** Spelcheck the text for the current block */
+    handleSpelchecker() {
```
gbbr: There is a slight spelling mistake here int he name of the method:
`s/handleSpelchecker/handleSpellcheker`

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -39,39 +41,150 @@ export class Editor3 extends React.Component {
...
+        if (!this.isActiveSpellchecker()) {
+            return;
+        }
+        this.spellchecker.currentBlockSpellcheck(this.state.editorState).then(
+            (editorState) => this.setState({editorState})
+        );
+    }
+
+    /** Spelcheck the text for all current text from editor */
+    handleFullSpelchecker() {
```
gbbr: Here there is the same spelling mistake, missing an _L_

scripts/core/editor3/components/index.js
```diff
@@ -1 +1,2 @@
 export {Editor3} from './Editor3';
+export {DraftSpellchecker} from './DraftSpellchecker';
```
gbbr: Since you've started this, we might as well put all of them in here. And then just import them more easily.

scripts/core/editor3/controllers/Editor3Controller.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import {Editor3} from '../components/Editor3';
+import {DraftSpellchecker} from '../components/DraftSpellchecker';
```
gbbr: Then here, we could just:
```js
import {Editor3, DraftSpellchecker} from '../components';
```
or
```js
import {Editor3, DraftSpellchecker} from 'core/editor3/components';
```

petrjasek:
just wondering what is the idea for that spellchecking?
when testing I was using decorators to handle it, but can't find any decorator here, so I'm curious..

ioanpocol:
@petrjasek A 'Error" style is applied to text with errors:

+        return this.checkText(text).then((errors) =>
 +            errors.reduce(
 +            (crtContentState, error) => this.applyStyleForError(error, crtContentState, selectionState)
 +            , newContentState)
 +        );

petrjasek:
I see, might be easier to do this using decorator, imo it caches and re-runs only on blocks if there are changes, but you will see.

petrjasek:
just to illustrate what I mean - https://github.com/superdesk/editor3-prototype/commit/2b559b8010e1380da4f6cfce2365d78fc1fc02cc

scripts/core/editor3/components/index.js
ioanpocol: Done

scripts/core/editor3/controllers/Editor3Controller.jsx
```diff
@@ -11,14 +11,24 @@ import {Editor3} from '../components/Editor3';
...
-    var showToolbar = true;
+Editor3Controller.$inject = ['$element', 'spellcheck'];
+export function Editor3Controller($element, spellcheck) {
+    let showToolbar = true;
 
     if (!this.editorFormat || this.readOnly) {
         showToolbar = false;
     }
 
+    let spellchecker = new DraftSpellchecker('en');
```
gbbr: In my opinion, it doesn't make sense to initialise the spellchecker here, outside the editor. It is functionality that belongs to the editor, so it should be initialised inside. If we ever decide that we want to disable it, we can always do so via editor options.

Instead, I would simply pass the `spellcheck` service as a prop to `Editor3`, so that it can take advantage of the whole API.

It would make more sense and would be cleaner...

scripts/core/editor3/components/DraftSpellchecker.jsx
```diff
@@ -0,0 +1,244 @@
...
+      * @ngdoc method
+      * @name DraftSpellchecker#registerDictionaryCallbacks
+      * @param {Function} ignoreWord ignore word callback
+      * @param {Function} addWord add word callback
+      * @param {Function} getSuggestions return suggestions for current word
+      * @param {Function} getWords get the list of words for current language
+      * @param {Function} getIgnoredWords get the list of ignored words for current user and language
+      * @description Register the dictionary related actions
+      */
+    registerDictionaryCallbacks(ignoreWord, addWord, getSuggestions, getWords, getIgnoredWords) {
```
gbbr: This is not needed anymore, if you pass the `spellcheck` service as a prop.

scripts/core/editor3/components/DraftSpellchecker.jsx
```diff
@@ -0,0 +1,244 @@
...
+    /**
+      * @ngdoc method
+      * @name DraftSpellchecker#registerEditorCallbacks
+      * @param {Function} setContextMenu context menu initialization callback
+      * @param {Function} closeContextMenu close context menu callback
+      * @param {Function} initDecorator set the decorator callback in order to force a full refresh
+      * @param {Function} onReplaceWord replace word callback
+      * @description Register editor related actions
+      */
+    registerEditorCallbacks(setContextMenu, closeContextMenu, initDecorators, onReplaceWord) {
```
gbbr: I would remove this pattern. In my opinion it indicates that separation of concerns could be better. All of these functions that are _registered_ here could (and should) live in the context menu.

The context menu can be given access to `editorState` and it can use it to obtain anything it needs.

Additionally, the context menu (as it already is a React component) can be added into the `render()` method of editor, just like `Toolbar` has been.

scripts/core/editor3/components/DraftSpellchecker.jsx
```diff
@@ -0,0 +1,244 @@
...
+
+    /**
+      * @ngdoc method
+      * @name DraftSpellchecker#showContextMenu
+      * @param {Object} e on context menu event
+      * @param {Object} the component that generates the oncontext menu
+      * @description Show the contextual menu, init the suggestions list
+      */
+    showContextMenu(e, element) {
+        const left = e.clientX - 20;
```
gbbr: Optionally, you could consider using the (undocumented) method `getVisibleSelectionRect` from `draft-js`.

Take a look at the [`LinkPopover.jsx`](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/components/toolbar/LinkPopover.jsx#L118-L127) file, it might help.

scripts/core/editor3/components/DraftSpellchecker.jsx
```diff
@@ -0,0 +1,244 @@
...
+/** Custom overrides for "error" style. */
+const styleMap = {
+    ERROR: {
+        borderBottomWidth: '1px',
+        borderBottomStyle: 'dotted',
+        borderBottomColor: 'red'
+    }
+};
+
+export class DraftSpellchecker {
```
gbbr: Is it possible that this file and the `InlineSpellcheckerMenu` should be one and the same?

Conceptually, the spellcheck functionality is nothing but the menu, and the decorator, no? You've used this class only to register callbacks, which we can remove anyway, it will be initialised once you add `<Spellchecker />` into the Editor's JSX

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -79,20 +131,25 @@ export class Editor3 extends React.Component {
...
-        const maxDepth = 4;
-
-        this.onChange(RichUtils.onTab(e, this.state.editorState, maxDepth));
+    /**
+     * @ngdoc method
+     * @name Editor3#componentWillUpdate
+     * @description Called before update, init the current editor rectangle
+     */
+    componentWillUpdate() {
+        this.editorRect = ReactDOM.findDOMNode(this.refs.editor).getBoundingClientRect();
```
gbbr: Does this have to be here, before the update? I was thinking it could go into the render method. That way, you get the rectangle after the editor is rendered, which is the most recent size.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -18,30 +20,51 @@ import Toolbar from './toolbar';
...
 
+        this.focus = this.focus.bind(this);
+        this.onChange = this.onChange.bind(this);
+
+        this.handleKeyCommand = this.handleKeyCommand.bind(this);
+        this.onTab = this.onTab.bind(this);
+
+        this.getDecorators = this.getDecorators.bind(this);
+        this.getEditorState = this.getEditorState.bind(this);
+        this.setEditorState = this.setEditorState.bind(this);
```
gbbr: In React, state is considered internal and components are only allowed to communicate via props. For this reason, state should not be shared, but only values via properties. See this [link](https://github.com/uberVU/react-guide/blob/master/props-vs-state.md#state) for an explanation.

Alternatively, I would suggest using the pattern that I've used with toolbar, and simply share `editorState` and `onChange`.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -109,46 +166,54 @@ export class Editor3 extends React.Component {
...
                 onChange={this.onChange}
                 onTab={this.onTab}
-                readOnly={this.readOnly}
+                readOnly={this.props.readOnly}
                 ref="editor"
             />;
 
-        if (this.showToolbar) {
+        const spellchecker = () =>
+            <Spellchecker
```
gbbr: Then this becomes:
```js
<Spellchecker
    spellchecker={this.props.spellchecker}
    editorState={this.state.editorState}
    editorRect={editorRect}
    onChange={this.onChange} />
```
Similar to the toolbar. I'm not sure if sharing the `editorState` is the best pattern, but this is what I've seen as being used in the `draft-js-plugins` project and seems like a decent approach.

scripts/core/editor3/components/spellchecker/Spellchecker.jsx
```diff
@@ -0,0 +1,229 @@
...
+        const left = e.clientX;
+        const top = e.clientY;
+        var element = props.children[0];
+
+        e.preventDefault();
+        props.spellchecker.suggest(element.props.text)
+        .then((suggestions) => {
+            var {editorState} = props.getEditorState();
+
+            props.setEditorState({
```
gbbr: This is the spellchecker's state, which is this component. The spellchecker should manage the spellchecker's state. Here, you are using a static method of the spellchecker, that takes a function passed as a property, which belongs to the editor, that changes the spellcheckers state. This is very complex.

I think that the creators of draft-js intended decorators to not have access to any state or props. But I might be wrong!

I'm new and I'm not sure what the best pattern is, but this still seems like it good be simplified. I would like to suggest the following method:

* Create an [`Entity`](https://facebook.github.io/draft-js/docs/api-reference-entity.html#create) of type `TYPO` wherever there is a spelling issue.
* Because `editorState` is a prop of Spellchecker, it will cause a re-render every time anything changes. You can check the current cursor position to see if it's on a `TYPO` entity and display the context menu in that situation. I have done the same in toolbar. There are entities of type `LINK` and when the cursor is over them, the popover is displayed.

I think doing the above method will isolate all of the components and they will not depend on each other anymore. Their only common denominator will be `editorState`. It will also rid you of all the callbacks.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: It is not allowed to call it in render

scripts/core/editor3/components/spellchecker/Spellchecker.jsx
petrjasek: This is not related to decorator imo, but an action triggered on spellcheck span component, those 2 could be defined outside of `Spellchecker` which is not a component, but rather a service that gets words list and suggestions.
When you pick a suggestions it should change editor state, but not when you render the dropdown.

PR #1187, opened 2016-12-14
[SDESK-387] fix(widgets): Added ability to translate headline and description of widget

pavlovicnemanja:
SDESK-387 - Dashboard widgets: name and description are not "translatable"

### DIFF
--- a/scripts/apps/dashboard/views/workspace.html
+++ b/scripts/apps/dashboard/views/workspace.html
@@ -86,8 +86,8 @@
                             <img ng-src="{{ :: dashboard.selectedWidget.thumbnail }}">
                         </div>
                         <div class="content-box">
-                            <div class="title">{{ :: dashboard.selectedWidget.label }}</div>
-                            <div class="description">{{ :: dashboard.selectedWidget.description }}</div>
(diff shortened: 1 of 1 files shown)


scripts/apps/dashboard/views/workspace.html
```diff
@@ -86,8 +86,8 @@ <h3 class="pull-left" translate>Add new widget</h3>
                             <img ng-src="{{ :: dashboard.selectedWidget.thumbnail }}">
                         </div>
                         <div class="content-box">
-                            <div class="title">{{ :: dashboard.selectedWidget.label }}</div>
-                            <div class="description">{{ :: dashboard.selectedWidget.description }}</div>
+                            <div class="title" translate>{{ :: dashboard.selectedWidget.label }}</div>
```
petrjasek: not sure if it works like that, imo it should be:
```
<div class="title">{{ :: dashboard.selectedWidget.label | translate }}</div>
```

pavlovicnemanja: I think it works, but changed it because of extra `span` that shows in html.

PR #1198, opened 2016-12-15
feature(core/editor3): editor3 feature flag

gbbr:
When adding `editor3: true` to the `features` object of
superdesk.config.js, content profiles will now display the option to
enable Editor3 for the body_html field.


petrjasek:
just wondering what is the idea - use both at the same time? or this is just temporary?

PR #1202, opened 2016-12-16
Added content lists

Aca-jov:
SWP-439
SWP-440
SWP-442
SWP-454

### DIFF
--- a/scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
+++ b/scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
@@ -38,6 +38,16 @@ export function WebPublisherManagerController($scope, publisher, modal) {
 
         /**
          * @ngdoc method
+         * @name WebPublisherManagerController#changeListFilter
+         * @param {String} type - type of content lists
+         * @description Sets type for content lists
+         */
+        changeListFilter(type) {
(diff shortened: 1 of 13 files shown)


scripts/apps/web-publisher/controllers/WebPublisherManagerController.js
```diff
@@ -260,14 +275,105 @@ export function WebPublisherManagerController($scope, publisher, modal) {
 
         /**
          * @ngdoc method
+         * @name WebPublisherManagerController#toogleCreateListCard
+         * @param {String} listType - type of content list
+         * @description Creates a new unsaved content list card
```
gbbr: The description of this method gives no indication as to what is being toggled. Can we maybe clarify?
Is the description correct? Should we change it? If not, should we change the method name?

By looking at the implementation, I don't see anything being toggled. So maybe remove the `toggle` prefix?

Aca-jov: Y, right, I changed logic, so now its not toogle, but forgot to change name, so I'll remove prefix.

Aca-jov:
Yes, controller is big now, I planned to split it, so just for this PR I'll leave it as it is, but split it in next PR.

PR #1204, opened 2016-12-19
refactor(ingest): add dynamic feed parsers lookup

thnkloud9:


scripts/apps/ingest/services/IngestProviderService.js
```diff
@@ -16,6 +16,36 @@ export function IngestProviderService(api, $q, preferencesService, $filter, sear
...
+
+                if (result._links.next) {
+                    pg++;
+                    return _getAllFeedParsersAllowed(criteria, pg, merged);
+                }
+                return $filter('sortByLabel')(merged);
+            });
+    };
+
+    // TODO: not implemented yet.  should be used when dynamic services with custom
```
gbbr: It's a bit unclear what is not implemented... The function seems to have code, so it is implemented. Or do you mean the API endpoint? If so, maybe the comment belongs above the line `api.query` and might be worth specifying that the backend is unimplemented. And if that's the case, then maybe it's not even worth having this comment here?

gbbr:
This PR contains very little changes and it spans across 3 different commits. Please squash when merging.

scripts/apps/ingest/services/IngestProviderService.js
thnkloud9: yeah, probably you are right, better to remove the comment.

scripts/apps/ingest/tests/sdIngestSourcesContent.spec.js
```diff
@@ -28,7 +29,7 @@ describe('sdIngestSourcesContent directive', () => {
         scope.$digest();
     }));
 
-    it('initializes the list of available field names in scope', () => {
+    fit('initializes the list of available field names in scope', () => {
```
akintolga: `fit` is left here

scripts/core/filters/index.js
```diff
@@ -198,6 +198,11 @@ export default angular.module('superdesk.core.filters', [])
             _entry[propertyName].toLowerCase()
             : _entry.name.toLowerCase());
     })
+    .filter('sortByLabel', () => function(_collection, propertyName = 'label') {
```
mdhaman: There is already `filter('sortByName')` and examples are:
 https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/vocabularies/services/VocabularyService.js#L54 or https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/templates/directives/TemplatesDirective.js#L20

thnkloud9: good call, removing duplicate.

thnkloud9:
refactor(ingest): add dynamic feed parsers lookup

PR #1206, opened 2016-12-20
[SDESK-370][SDESK-515] Select/Check items using keyboard shortcut 'x' and scroll

nrvikas:
1 - use 'x' to select/check an item
2 - scroll selected element to view if it is not visible

### DIFF
--- a/scripts/apps/search/components/ItemList.jsx
+++ b/scripts/apps/search/components/ItemList.jsx
@@ -10,6 +10,8 @@ export class ItemList extends React.Component {
     constructor(props) {
         super(props);
 
+        const {keyboardManager} = this.props.svc;
+
         this.state = {itemsList: [], itemsById: {}, selected: null, view: 'mgrid', narrow: false};
 
         this.multiSelect = this.multiSelect.bind(this);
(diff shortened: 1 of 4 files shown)


scripts/apps/search/components/ItemList.jsx
```diff
@@ -29,6 +31,10 @@ export class ItemList extends React.Component {
         this.setSelectedComponent = this.setSelectedComponent.bind(this);
         this.modifiedUserName = this.modifiedUserName.bind(this);
         this.setNarrowView = this.setNarrowView.bind(this);
+
+        keyboardManager.bind('x', () => {
```
gbbr: I'm just wondering...
js```
keyboardManager.bind('x', this.multiSelectCurrentItem);
```
Would this version not work?

Comment (nrvikas) [inline scripts/apps/search/components/ItemList.jsx]: That's a good suggestion to make the code simple. I tried it, but did not work as the 'this' object is used inside multiSelectCurrentItem() function and it goes 'undefined' at the time of invocation. There could be a name for this concept (which could probably make my explanation better), but, I do not know what it is :).

Comment (gbbr) [inline scripts/apps/search/components/ItemList.jsx]: I understand, thanks for the explanation. Maybe we could also bind it like the other methods above?
```js
this.multiSelectCurrentItem = this.multiSelectCurrentItem.bind(this);
```
I wonder if that would fix it.... Just to make sure other callers don't run across this weird behaviour.

nrvikas: Thanks Gabriel. That did work. I really should have tried that out before resubmitting this PR though - my bad. Cheers.

PR #1211, opened 2016-12-21
[SDESK-420] Adding user prompt to add associated media for rewrites.

mdhaman:
@petrjasek @ioanpocol @gbbr this is part of SDESK-420 where the user is prompted to added associated media on publishing.

Depends on https://github.com/superdesk/superdesk-core/pull/704

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.js
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.js
@@ -1,6 +1,42 @@
 import * as helpers from 'apps/authoring/authoring/helpers';
 import _ from 'lodash';
 
+/**
+ * @ngdoc directive
+ * @module superdesk.apps.authoring
+ * @name sdAuthoring
+ *
+ * @requires superdesk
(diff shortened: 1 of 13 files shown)


scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -294,57 +331,96 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
...
+                    .then((result) => {
+                        if (result) {
+                            return authoring.publish(orig, item, action);
+                        }
+                        return $q.reject(false);
+                    })
+                    .then((response) => {
+                        if (!response) {
+                            notify.error(gettext('Unknown Error: Item not published.'));
+                            return false;
```
petrjasek: I think it would be better to `reject` here so that chain will stop (in case there is another `then` added later)

mdhaman: done.

PR #1212, opened 2016-12-21
fix(multiselect): Fix issue when selecting recursively

pavlovicnemanja:
SDESK-390: Improvements for multiselect

### DIFF
--- a/scripts/apps/search/components/ItemList.jsx
+++ b/scripts/apps/search/components/ItemList.jsx
@@ -102,7 +102,10 @@ export class ItemList extends React.Component {
             positionStart = _.indexOf(this.state.itemsList, this.state.selected);
         }
 
-        for (var i = positionStart; i <= positionEnd; i++) {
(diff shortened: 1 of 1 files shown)


scripts/apps/search/components/ItemList.jsx
```diff
@@ -102,7 +102,10 @@ export class ItemList extends React.Component {
             positionStart = _.indexOf(this.state.itemsList, this.state.selected);
         }
 
-        for (var i = positionStart; i <= positionEnd; i++) {
+        var start = positionStart < positionEnd ? positionStart : positionEnd,
```
petrjasek: this could be also like:
```
var start = Math.min(positionStart, positionEnd),
     end = Math.max(positionStart, positionEnd);
```
not sure if it's more readable or not?

pavlovicnemanja: you are right! It is more readable! :) tnx

PR #1214, opened 2016-12-21
[SUBSCRIBERS] [EMAIL] added hidden (BCC) recipients field

jerome-poisson:
needed for SDESK-428

### DIFF
--- a/scripts/apps/publish/views/email-config.html
+++ b/scripts/apps/publish/views/email-config.html
@@ -1,6 +1,12 @@
 <div class="field">
     <label for="recipients" translate>Recipients</label>
-    <input type="text" id="recipients" placeholder="{{:: 'Recipients'|translate }}"
-      ng-model="destination.config.recipients" required
-      ng-change="$parent.setConfig(provider)">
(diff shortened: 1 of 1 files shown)


scripts/apps/publish/views/email-config.html
```diff
@@ -1,6 +1,10 @@
 <div class="field">
     <label for="recipients" translate>Recipients</label>
-    <input type="text" id="recipients" placeholder="{{:: 'Recipients'|translate }}"
-      ng-model="destination.config.recipients" required
+    <input type="text" id="recipients" placeholder="{{:: 'Recipients (visible to everybody)'|translate }}"
```
mdhaman: looks like `required` is missing.

jerome-poisson: hi @mdhaman . required was actually removed on purpose: we need to be able to set bcc without recipient en vice versa

jerome-poisson: @mdhaman added a check to see if at least one of the field is filled, thank for the feedback

PR #1219, opened 2016-12-22
fix(content profile): Fix editing content types with custom fields

ioanpocol:


scripts/apps/workspace/content/services/ContentService.js
```diff
@@ -180,6 +180,16 @@ export function ContentService(api, superdesk, templates, desks, packages, archi
     };
 
     /**
+     * Get content type with metadata by id
+     *
+     * @param {string} id
+     * @return {Promise}
+     */
+    this.getTypeMetadata = function(id) {
```
gbbr: If you have time, it would be nice to make this doc show up in `grunt docs`. Which means that you would have to document the service too...

ioanpocol: Done

PR #1221, opened 2016-12-23
SDESK-160, SDESK-163, SDESK-165

mugurrus:
Activity reports generation and management

Depends on https://github.com/superdesk/superdesk-core/pull/708

### DIFF
--- a/scripts/apps/analytics/directives/ActivityReportPanel.js
+++ b/scripts/apps/analytics/directives/ActivityReportPanel.js
@@ -1,0 +1,125 @@
+ActivityReportPanel.$inject = [
+    '$location', 'desks', 'asset', 'config', 'metadata', 'api', 'session', 'notify', '$rootScope'
+];
+
+/**
+ * A directive that generates the sidebar containing activity report parameters
+ */
+export function ActivityReportPanel($location, desks, asset, config, metadata, api, session, notify, $rootScope) {
+    return {
+        templateUrl: asset.templateUrl('apps/analytics/views/activity-report-panel.html'),
+        scope: {},
+        link: function(scope, element, attrs, controller) {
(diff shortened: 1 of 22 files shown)


scripts/apps/analytics/directives/ActivityReportPanel.js
```diff
@@ -0,0 +1,121 @@
...
+                });
+
+                api('activity_reports', session.identity).save({}, activityReportEdit)
+                    .then(onSuccess, onFail);
+            };
+
+            /**
+             * Format given date for generate
+             */
+            function formatDate(date) {
```
mdhaman: Few places we are using `formatdate`. I might be a `filter` for formatting dates.

Moreover, all dates are stored as UTC and there is 10/11 hours difference for Sydney time. We would need timezone in `operation_date` otherwise we might not get the correct results.

mugurrus:
@gbbr Updated the documentation, please review again

scripts/apps/analytics/directives/SavedActivityReports.js
```diff
@@ -0,0 +1,119 @@
+SavedActivityReports.$inject = [
+    '$rootScope', 'api', 'session', 'modal', 'notify', 'gettext', 'asset',
+    '$location', 'desks', 'privileges', 'config', 'savedActivityReports'
+];
+
+/**
+ * @ngdoc directive
+ * @module superdesk.apps.analytics
+ * @name sdSavedActivityReports
+ * @requires $rootScope, api, session, modal, notify, gettext, asset, $location, desks,
```
gbbr: I don't think this works. It'll have to be a separate line for each one, like in the other modules.
```js
@requires $rootScope
@requires api
// .. etc
```

PR #1228, opened 2016-12-30
[SDESK-504] Alphabetically sort content filter 'field' drop-down

nrvikas:


scripts/apps/content-filters/views/filter-search.html
```diff
@@ -5,7 +5,7 @@
             <div class="field">
                 <label translate>Field</label>
                 <select class="input" ng-change="resetValues()" required ng-model="filterCondition.field">
-                    <option value="{{parameter.field}}" ng-selected="parameter.field === filterCondition.field" ng-repeat="parameter in filterConditionParameters track by $index">{{:: parameter.field | translate}}</option>
+                    <option value="{{parameter.field}}" ng-selected="parameter.field === filterCondition.field" ng-repeat="parameter in filterConditionParameters | orderBy: 'field | translate' track by $index">{{:: parameter.field | translate}}</option>
```
akintolga: `translation` may not work as expected in filter condition parameters as they are field names where translation would not make much sense. So I think `translate` can be removed from here.

PR #1233, opened 2017-01-03
Editor2: allow to drop html in the editor

vied12:
If html element is a <a>, we create a block from the url by using iframely.

Otherwise, we create the block by using the element as embed's body


naspeh:
@vied12 do you still need this `editor-drop-html` branch? If no, please remove it :)

PR #1237, opened 2017-01-04
fix(locales): use user setting to load locales

petrjasek:
so when user switches languge there are matching locales loaded

SDESK-385


Gruntfile.js
```diff
@@ -51,8 +51,7 @@ module.exports = function(grunt) {
     // The gen-importer tasks generates files that import dynamic paths read from
     // superdesk.config.js
     grunt.registerTask('ngtemplates:gen-importer', [
```
gbbr: Since this isn't grouping anything anymore, can we just remove this extra task and instead use `ngtemplates:gen-apps` everywhere instead of this `ngtemplates:gen-importer`?

PR #1240, opened 2017-01-05
Eslint: apps handle their own linter

vied12:


webpack.config.js
```diff
@@ -90,7 +90,8 @@ module.exports = function makeConfig(grunt) {
                 {
                     test: /\.jsx?$/,
                     loader: 'eslint-loader',
-                    exclude: shouldExclude
+                    // superdesk apps handle their own linter
+                    exclude: (p) => shouldExclude(p) || sdConfig.apps.some(app => p.indexOf(app) > -1),
```
gbbr: Shouldn't this be:
```js
exclude: (p) => p.indexOf('node_modules') === -1,
```
instead?

vied12: I find my solution more meaningful and less specific to npm, plus it works fine :)

gbbr: The point is that we want to exclude `node_modules` from being linted, not that we want to exclude node_modules, but not apps, or apps, which is redundant. It's like saying: _"exclude the path if it's in node_modules, but it's not an app OR if it's an app"_ versus simply saying _"exclude node_modules"_

Your version just confuses things. "It works" is not what we are after. We are after "it's maintainable and readable".

gbbr:
Seems like my inline comment doesn't show on this PR, so I will paste it here:

The point is that we want to exclude node_modules from being linted, not that we want to exclude node_modules, but not apps, or apps, which is redundant. It's like saying: "exclude the path if it's in node_modules, but it's not an app OR if it's an app" versus simply saying "exclude node_modules".

Please make the requested change.

webpack.config.js
vied12: I find my solution more maintainable as it reuses an exclusion function we already defined and it doesn't use an hard coded value to search for.
And sorry but yes, I am looking for a working solution, at least until you suggest a working one. Yours doesn't do what I'm expecting for (probably because of symbolic links created by `npm link`).

gbbr: I apologise. I made a mistake in my comment, I mean to say:
```js
exclude: (p) => p.indexOf('node_modules') !== -1,
```
Really sorry about that :-)

gbbr: That should do the job. If not, let me know.

PR #1241, opened 2017-01-05
feature(Editor3): Add support for spellcheck on Editor3

ioanpocol:


scripts/core/editor3/actions/editor3.jsx
```diff
@@ -0,0 +1,21 @@
+export function onChangeEditorState(editorState) {
```
gbbr: There is some stuttering here. We are in the editor component and this is already known. On top of that, `onChange` is common draft-js implementation approach so it should be enough to call this `onChange` instead of `onChangeEditorState`.

scripts/core/editor3/actions/editor3.jsx
```diff
@@ -0,0 +1,21 @@
...
+    return {
+        type: 'CHANGE_EDITOR_STATE',
```
gbbr: Let's set a standard here for consistency. How about we call all actions prefixed with the component name, so in this case: `EDITOR_CHANGE_STATE`

scripts/core/editor3/actions/editor3.jsx
```diff
@@ -0,0 +1,21 @@
...
+        payload: editorState
+    };
+}
+
+export function onEditorTab(e) {
```
gbbr: Same here, `onTab`

scripts/core/editor3/actions/editor3.jsx
```diff
@@ -0,0 +1,21 @@
...
+    return {
+        type: 'EDITOR_TAB',
+        payload: e
+    };
+}
+
+
+export function handleEditorKeyCommand(command) {
```
gbbr: `handleKeyCommand`

scripts/core/editor3/actions/spellchecker.jsx
```diff
@@ -0,0 +1,13 @@
+export function replaceWord(word) {
+    return {
+        type: 'REPLACE_WORD',
```
gbbr: Based on the editor approach, should we do the same here? `SPELLCHECKER_REPLACE_WORD` ?

scripts/core/editor3/actions/spellchecker.jsx
```diff
@@ -0,0 +1,13 @@
...
+        payload: word
+    };
+}
+
+export function showContextMenu(contextMenuData) {
+    return {
+        type: 'SHOW_CONTEXT_MENU',
```
gbbr: Here it would make sense too, because we might have other context menus in the future: `SPELLCHECKER_CONTEXT_MENU` ?

scripts/core/editor3/actions/spellchecker.jsx
```diff
@@ -0,0 +1,13 @@
+export function replaceWord(word) {
+    return {
+        type: 'REPLACE_WORD',
+        payload: word
+    };
+}
+
+export function showContextMenu(contextMenuData) {
```
gbbr: `data` should be enough and self-explanatory.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -20,146 +21,137 @@ import Toolbar from './toolbar';
...
-        super(props);
 
-        const initialContentState = stateFromHTML(props.value);
+export const getInitialState = (spellcheckerService, value, onChangeValue, readOnly, showToolbar, editorFormat) => {
+    const initialContentState = stateFromHTML(value);
 
-        this.state = {
-            editorState: EditorState.createWithContent(
+    return {
+        editorState: EditorState.createWithContent(
```
gbbr: This is a state object, so calling properties with the `State` suffix is stuttering imo. `editor` should be enough, and below, `spellchecker` should also be enough.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -20,146 +21,137 @@ import Toolbar from './toolbar';
...
                 initialContentState,
-                this.getDecorators()
-            )
-        };
+                getDecorators(spellcheckerService)
+        ),
+        spellcheckerState: {spellcheckerService: spellcheckerService},
```
gbbr: I think we should remove this key (spellcheckerState) and instead just pass the Angular `$injector` which will be used by other components too. It will allow obtaining any service, including the spellchecker service.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -20,146 +21,137 @@ import Toolbar from './toolbar';
  * @description Editor3 is a draft.js based editor that support customizable
  *  formatting (shortly more features will come ...)
  */
-export class Editor3 extends React.Component {
-    constructor(props) {
-        super(props);
 
-        const initialContentState = stateFromHTML(props.value);
+export const getInitialState = (spellcheckerService, value, onChangeValue, readOnly, showToolbar, editorFormat) => {
```
gbbr: Why do you need this function? You can just place the same object returned here as the second argument to `createStore` in `Editor3Controller`:
```js
let store = createStore(editorReducers, {
     // ...
});
```
Alternatively, I would move this whole logic (`let store = ...`) to a `store` folder where you can do initialisations. Take a look at [this example](https://github.com/reactjs/redux/tree/master/examples/real-world/src) in the Redux repo.

scripts/core/editor3/actions/editor3.jsx
ioanpocol: Actions can be called from entire application so it is more clear to have onChangeEditorState. OnChange is too general and also other components can have it.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -20,146 +21,137 @@ import Toolbar from './toolbar';
...
+        const {
+            readOnly,
+            showToolbar,
+            editorState,
+            onChange,
+            onTab,
+            handleKeyCommand
+        } = this.props;
 
         const editor = () =>
```
gbbr: I think we can refactor this whole `render()` function to use only JSX:
```js
render() {
    const {
        readOnly,
        showToolbar,
        editorState,
        onChange,
        onTab,
        handleKeyCommand
    } = this.props;

    var cx = classNames({
        'editor-root': true,
        'editor-single-line': !showToolbar
    });

    return (
        <div className={cx} onClick={this.focus}>
            {showToolbar ? <Toolbar editorOffset={this.editorRect} /> : null}
            <SpellcheckerContextMenu editorRect={this.editorRect} />
            <Editor
                editorState={editorState}
                handleKeyCommand={handleKeyCommand}
                onChange={onChange}
                onTab={onTab}
                readOnly={readOnly}
                ref="editor"
            />
        </div>
    );
}
```
`showToolbar` should not be the indicator of whether the editor is single-line or not. Instead we should have a prop that says `singleLine={true}` in that case.

Also, the classes need to be corrected to reflect this layout.

scripts/core/editor3/actions/editor3.jsx
gbbr: Wouldn't this make it clear:
```js
import * as editorActions from `core/editor3/actions/editor3`;
```
And then use it as `editorActions.onChange` ?

gbbr: If you prefer your version, then maybe call it `changeEditorState` because it better indicates an action. `onChange` sounds like an event.

scripts/core/editor3/components/spellchecker/SpellcheckerDecorators.jsx
```diff
@@ -0,0 +1,45 @@
...
+  * @param {Function} getEditorState
+  * @param {Function} setEditorState
+  * @param {Object} spellchecker
+  * @return {Object} returns the decorator
+  * @description Decorator for spellchecker, will highlight the words with errors
+  */
+export const getSpellcheckerDecorators = (spellcheckerService) =>
+    [{
+        strategy: spellcheckStrategy(spellcheckerService.isCorrectWord),
+        component: handleSpellcheckError,
```
gbbr: You can just inline the component here:
```js
component: (props) => <SpellcheckerError {...props} />
```

scripts/core/editor3/components/Editor3.jsx
gbbr: The point is that this object is called `state`. When you reference it you write `state.spellcheckerState` which is stuttering and redundant. That's why I suggest renaming it to `spellchecker`. That way, when you say `state.spellchecker` it makes more sense.

An analogy would be a folder called "People". Inside you would have files names "Adrian, John, etc", not files named "PeopleAdrian, PeopleJohn, etc" or "AdrianPerson, John Person, etc".

scripts/core/editor3/components/Editor3.jsx
gbbr: Sure.

scripts/core/editor3/actions/spellchecker.jsx
gbbr: The function is called `showContextMenu` and the parameter is `data`. It is obvious the data is for the context menu. This is again stuttering and only complicates things and makes code longer to read and write. I think it's about being simple, not specific.

scripts/core/editor3/actions/editor3.jsx
gbbr: Sure, whatever you prefer.

scripts/core/editor3/components/Editor3.jsx
gbbr: This is confusing. In Redux stores map to props and state is generally not in play. I think having the function here indicates that there is a state to the editor, which is false, this is the state of the store, so I would move it from here to somewhere else.

I would follow the examples provided in the Redux framework.

scripts/core/editor3/actions/editor3.jsx
ioanpocol: Done

scripts/core/editor3/actions/spellchecker.jsx
ioanpocol: Done

scripts/core/editor3/actions/spellchecker.jsx
ioanpocol: Done

scripts/core/editor3/components/Editor3.jsx
ioanpocol: Yes, usually the store is at the application level, here we use it just for Editor3. I think we can refactor the creation of the store when we extend the using of react&redux at edit content item page.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: Done

scripts/core/editor3/components/spellchecker/SpellcheckerDecorators.jsx
ioanpocol: Done

ioanpocol:
@vied12 Thank you for the trick. I will take a look at the library and I will update the code on the next PR that will contain tests too.

scripts/core/editor3/actions/editor3.jsx
ioanpocol: Done

scripts/core/editor3/actions/editor3.jsx
ioanpocol: Done

scripts/core/editor3/actions/spellchecker.jsx
ioanpocol: All over the application I used the name contextMenuData and I think it is better to keep that name in order to don't create confusion to the reader. I agree that local variable should be short but for parameters I think that a more explanatory name will help. This variable is used only once so I think it don't creates any complication or reading difficulty.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: For editor state the name in DraftJs is editorState, so I used it and I used the same pattern for spellchecker to be consistent in naming things. But yes, if is needed  it can be changed later when we add other modules and use combineReducers to combine reducers. In this case every reducer uses a key to access state. If the key is editor3, the store tree will be state.editor3.editorState and state.editor3.spellcheckerState.

PR #1243, opened 2017-01-06
[SDESK-548] Mark mandatory fields with asterix to the form (form to add a subscriber)

nrvikas:
Also, have used a slightly darker shade for 'field-info' class as it was very faint.

### DIFF
--- a/scripts/apps/publish/views/subscribers.html
+++ b/scripts/apps/publish/views/subscribers.html
@@ -32,6 +32,7 @@
                 <tabset>
                     <tab heading="{{ 'General' | translate }}" class="TabContent">
                         <div class="field">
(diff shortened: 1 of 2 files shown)


scripts/apps/publish/views/subscribers.html
```diff
@@ -32,6 +32,7 @@ <h3 translate ng-hide="subscriber._id">Add New Subscriber</h3>
                 <tabset>
                     <tab heading="{{ 'General' | translate }}" class="TabContent">
                         <div class="field">
+                            <i class="field-info pull-right" >* mandatory fields</i>
```
petrjasek: pls add `translate` here in the tag

nrvikas: Done, thanks.

PR #1244, opened 2017-01-06
[SDESK-550] Clear input text from typehead's input field after a selection is made

nrvikas:


nrvikas:
@akintolga , absolutely. I did check those 'type heads' and a few ones in the settings and configuration. They seem to work as per regular behavior. I'll continue to keep a keen eye on this. Thanks.

PR #1251, opened 2017-01-13
feat(config): allow config changes without rebuilt

petrjasek:
it will load `config.js` where user can set `superdeskConfig` dict
which is used to update config before `angular.bootstrap`.

there is empty file provided in core by default, but will use
`config.js` file from `superdesk/client` if any.

SDESK-577


takeit:
It would be nice to have that fix also in [publisher](https://github.com/superdesk/superdesk-client-core/tree/publisher) branch.

PR #1253, opened 2017-01-13
Simplify spellchecker and remove injected services

gbbr:
* Adds new service that allows sharing the app injector with anyone (`core/services/ng`).
* Removes parameter that passes around functions and dependencies.
* Isolates store logic inside it's own file.

*TODO*: documentation and more cleaning up


naspeh:
BTW: why was this PR merged before Travis and Fire checks are finished?

PR #1254, opened 2017-01-15
[SDESK-255] - Handling of new operators for filter conditions

akintolga:


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/733

PR #1255, opened 2017-01-16
Improve overall Redux architecture

gbbr:
- Move async operations to actions (via thunk middleware)
- Move all state processing logic to reducers
- Improve context menu closure logic
- Removed JSX files from dgeni. Reasoning is that dgeni has been built for Angular applications, trying to force it into parsing React code is a bit of a hustle and I would think it should be better for us to use something like [react-docgen](https://github.com/reactjs/react-docgen) instead, even if this means having 2 documentation websites. Currently it doesn't support spread operator which is extremely helpful in Redux architectures.


scripts/core/editor3/reducers/spellchecker.jsx
```diff
@@ -1,25 +1,58 @@
...
+ * @name ignoreWord
+ * @param {Object} state
+ * @param {String} word
+ * @return {Object} returns new state
+ * @description Replace the current word with the new selected one
+ */
+const ignoreWord = (state, text) => {
+    const spellcheck = ng.get('spellcheck');
+
+    spellcheck.addWord(text, true);
```
petrjasek: imo this shouldn't be here, reducers should only modify the state. such code should be in action

petrjasek: I see that's in the plan, so pls ignore my commet

gbbr: Yes, this is still WIP. And yes, you're right, I would've missed this without your comment, so it's welcomed :)

gbbr: Done

scripts/core/editor3/actions/spellchecker.jsx
```diff
@@ -21,14 +16,61 @@ export function replaceWord(word) {
...
+ * @param {String} word
+ * @return {String} action
+ * @description Creates the ignore word action
+ */
+export function ignoreWord(word) {
+    const spellcheck = ng.get('spellcheck');
+
+    spellcheck.addWord(word, true);
+
+    return replaceWord(word);
```
petrjasek: imo when you ignore word it shouldn't replace anything

scripts/core/editor3/actions/spellchecker.jsx
```diff
@@ -21,14 +16,61 @@ export function replaceWord(word) {
 
 /**
  * @ngdoc method
- * @name Spellchecker Actions#showContextMenu
- * @param {Object} contextMenuData
+ * @name replaceWord
```
petrjasek: names here don't match

scripts/core/editor3/reducers/spellchecker.jsx
```diff
@@ -28,28 +24,47 @@ const spellchecker = (state = {}, action) => {
...
 
 /**
  * @ngdoc method
- * @name Spellchecker Reducers#showContextMenu
+ * @name closeContextMenu
+ * @param {Object} state
+ * @return {Object} returns new state
+ * @description Save on store all relevant data for spellchecker context menu
+ */
+const closeContextMenu = (state) => {
```
petrjasek: imo this context menu shouldn't affect the state of the editor, it's more like button:hover, only when you click a button there it should do something and thus modify the state.

gbbr: I think you mean for the context menu to manage its own state here, which is a bit tricky given that we use Redux. The context menu's visibility has to be part of the store because it is triggered by external components (like the decorator). In this case, the state of the editor remains unaffected, only the state of the spellchecker's visibility changes.

I don't see a problem with this implementation, and I also can't think of a different solution, but if you have a suggestion I am open to it!

scripts/core/editor3/actions/spellchecker.jsx
gbbr: Copy/paste incident. It's fixed.

scripts/core/editor3/actions/spellchecker.jsx
gbbr: Here also I don't know how to implement an alternative solution. It would have to trigger the editor's change event for it to re-run the decorators. We could look at it as "_replace the word with the new word that is now in the dictionary_" :-)

Also open to suggestions here...

I have a similar issue somewhere else and if I figure out how to do this I will change this too.

petrjasek: I see. here it could use list of ignored words in state, but the issue is still the same in `addWord` where you add a new word to spellchecker dictionary. probably the redux way would be to keep dictionaries for spellchecking in state

scripts/core/editor3/reducers/spellchecker.jsx
petrjasek: yep I think it's component state rather than application state, think of eg datetime picker, that's also not affecting application state imo if you open the calendar, or when using dropdown menu.
there is a span added for errors and on hover this should be displayed, but not modify application state.
that component shouldn't be part of dom if not visible and then there is no need for state to control it.

gbbr: I think I'm lost here. This component (the context menu) is part of the editor as a whole and the editor should be aware of it.

Your suggestion raises a couple of difficulties which could potentially complicate things even more:
- If each span manages it's own menu and shows it on right click, how would other span's know to close their menu because another one has opened somewhere else?
- Whenever right-clicking a span there has to be a service call to the spellchecker to get the suggestions, which would have to be an action if we are using Redux. When 10-100+ spans are listening to this action, how do they know which one the incoming suggestions are for? Do we use IDs? This seems super complicated. We could potentially leave this out of Redux and let each decorator do the call in the React component itself, but then we aren't using Redux anymore and might confuse things even more.
- What about other actions that might affect the visibility of the context menu? Let's say the "context menu" of the link editing functionality. It seems to me that the editor should be able to manage it's menus.

I'm not sure if we should isolate this functionality...

scripts/core/editor3/actions/spellchecker.jsx
gbbr: Maybe, but keeping a huge dictionary in the state could potentially raise a performance issue.

scripts/core/editor3/reducers/spellchecker.jsx
gbbr: Also, finally, the decorator does not have access to editorState, which is required for the word to be replaced when using the context menu. This is another point in the above list.

gbbr:
@petrjasek thanks for the tips! I've greatly simplified the spellchecker following your advice. The spellchecker operates contained now, outside of Redux (except for the replace word functionality), but that's totally fine and it still stays fully testable.

Please take another looks and see if you like this more now...

PR #1259, opened 2017-01-17
[SDESK-630](fix):All renditions and metadata changes should be saved in one action for not published items.

mdhaman:
- Picture renditions and metadata are saved to database in one action for not-published items and for published items change are in memory and need to correct the item to save changes. This applies to text where media is associated.

This PR depends on [#751](https://github.com/superdesk/superdesk-core/pull/751)


gbbr:
Thank you for taking the time to add all this documentation.

PR #1261, opened 2017-01-17
fix(profiles): Fix handling of list fields defaults.

petrjasek:
Avoid saving default values twice for list type fields:

```
{
    "schema": {
        "genre": {
            "genre": "default value",
            "default": "default value"
        }
    }
}
```

SDESK-579


petrjasek:
it causes issues when publishing https://github.com/superdesk/superdesk-core/pull/739

PR #1271, opened 2017-01-19
fix(editor3): Fix editor focus after add word and ignore word actions

ioanpocol:


scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.jsx
```diff
@@ -106,9 +105,7 @@ SpellcheckerContextMenuComponent.propTypes = {
  * @description Maps the values from state to the component callback type props.
  */
 const mapDispatchToProps = (dispatch) => ({
-    addWord: (word) => dispatch(actions.ignoreWord(word)),
-    ignoreWord: (word) => dispatch(actions.addWord(word)),
-    replaceWord: (word) => setTimeout(() => dispatch(actions.replaceWord(word)), 0),
+    replaceWord: (word) => setTimeout(() => dispatch(actions.replaceWord(word)), 0)
```
petrjasek: just wondering why is there `setTimeout` ?

ioanpocol: There are 2 very closed events and the replace result is not visible.

gbbr: Instead, we could change `spellcheck.addWord` to return [this line](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/spellcheck/spellcheck.js#L372), which is already a promise if we look at the definition [here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/dictionaries/services/DictionaryService.js#L175) and only trigger the event after the promise is completed.

This way you can get rid of the timeout.

gbbr: Actually, you would have to change [this function](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/spellcheck/spellcheck.js#L414) too, so that it returns a promise in both outcomes...

Ping me on Slack and I'm happy to help with this...

gbbr: Just to follow up, I've tried the above and it doesn't work. There seems to be a problem with synchronising editor states, so the `setTimeout` part seems to be needed.

scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.jsx
```diff
@@ -62,7 +62,7 @@ class SpellcheckerContextMenuComponent extends Component {
         const {word} = this.props;
```
gbbr: Might be nicer if you put `replaceWord` here.

scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.jsx
ioanpocol: When the context menu is closed the editor automatically updates his status. Also on  word replace action the editor is updated again. If those actions are too closed, the replace will have no results. So there is this setTimeout call with 0s that solves the issue. Here are more info about setTimeout trick:
https://www.quora.com/What-does-setTimeout-with-a-0ms-delay-do

petrjasek: maybe we should avoid those 2 events, maybe there is some `event.stopPropagation` missing?

petrjasek: or keep the workaround, but pls add some info to make it obvious why it's there

PR #1273, opened 2017-01-22
[SDESK-346] - Related item widget improvements

akintolga:
- If the item has existing links, the widget will display them
- If the item has no links then the widget will display `linkable` stories


scripts/apps/archive/services/FamilyService.js
```diff
@@ -13,20 +31,150 @@ export function FamilyService(api, desks) {
...
+     * @public
+     * @description Returns any story potentially linkable
+     * @param {string} keyword - slugline to be matched
+     * @param {string} sluglineMatch - type of matching rule for slugline
+     * @param {string} eventId - event id to be searched
+     * @param {string} modificationDateAfter - filter for versioncreated
+     * @returns {Object}
+     */
+    this.fetchRelatableItems = (keyword, sluglineMatch, eventId, modificationDateAfter) => {
+        var filter = [
```
mdhaman: Do we need to add filter for the `last_published_version` for the items from the published collection?

akintolga: That's true. I was hiding the non-last versions on the view but it is better if I added here as well.

akintolga: Done.

scripts/apps/archive/related-item-widget/relatedItem.js
```diff
@@ -70,6 +73,25 @@ function RelatedItemController(
         sort: [{versioncreated: 'desc'}]
     };
 
+    $scope.loading = true;
+    familyService.fetchRelatedItems($scope.item.event_id)
```
petrjasek: imo would be better to have `item` as param to `fetchRelatedItems` and hide implementation that it's using `event_id` field

akintolga: Done.

PR #1287, opened 2017-01-26
fix(workspace): make analytics icon visibility configurable

thnkloud9:


webpack.config.js
```diff
@@ -219,6 +219,14 @@ function getDefaults(grunt) {
             useTansaProofing: false
         },
 
+        // workspace defaults
+        workspace: {
+            ingest: false,
+            content: false,
+            tasks: false,
+            analytics: true
```
petrjasek: imo this should b `false` now, otherwise we might need to modify all instances

thnkloud9: ok by me, but then should we set it to true in the superdesk repo?

petrjasek: yep we could enable it on master

thnkloud9: cool

PR #1296, opened 2017-01-31
Analytics additional fields

ancafarcas:


scripts/apps/analytics/views/activity-report-parameters.html
```diff
@@ -73,19 +81,38 @@
                 </select>
             </div>
 
+        <div class="multiple">
             <div class="field">
-                <label for="parameters-urgency" translate="">Urgency</label>
-                <select id="parameters-urgency" ng-model="activityReport.urgency"
-                    ng-options="i for i in [1,2,3,4,5]">
+                <label for="parameters-priority_start" translate="">From priority</label>
+                <select id="parameters-priority_start" ng-model="activityReport.priority_start"
```
petrjasek: imo this could work better with `<input type="range" min="1" max="6" step="1">`

ancafarcas: Yes, it looks classy

PR #1300, opened 2017-02-01
(fix): Modifying config to show-hide media metadata.

mdhaman:
- Changing config to show-hide media metadata
- Fixing the issue when a media item from monitoring is associated to an text item then metadata is not copied across to the associated item.

This would require change to the `superdesk.config.js` in the repo https://github.com/superdesk/superdesk.

### DIFF
--- a/superdesk.config.js
+++ b/superdesk.config.js
@@ -8,7 +8,29 @@
 module.exports = function(grunt) {
     return {
         defaultRoute: '/workspace',
-        requiredMediaMetadata: ['headline', 'description_text', 'alt_text'],
+        validatorMediaMetadata: {
+            headline: {
+                required: true
+            },
+            alt_text: {
+                required: true
+            },
+            description_text: {
+                required: true
+            },
+            copyrightholder: {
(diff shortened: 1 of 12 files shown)


superdesk.config.js
```diff
@@ -8,7 +8,29 @@
 module.exports = function(grunt) {
     return {
         defaultRoute: '/workspace',
-        requiredMediaMetadata: ['headline', 'description_text', 'alt_text'],
+        requiredMediaMetadata: {
+            headline: {
```
petrjasek: would probably do it like:
```
headline: true,
alt_text: true,
byline: false
```
given that the name is `requiredMediaMetadata`

mdhaman: will change the name to something appropriate. what if we have `maxlength`?

petrjasek: then it would make sense

mdhaman:
@ioanpocol can please review this changes.

ioanpocol:
Sorry for late review

PR #1304, opened 2017-02-02
fix(content types): Fix the setting of custom or default vocabulary b…

ioanpocol:
…ased fields


scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.js
```diff
@@ -108,7 +111,9 @@ export function ContentProfileSchemaEditor(gettext, metadata, content, config) {
              * @param {String} id
              */
             scope.setdefault = function(id) {
-                scope.model.schema[id].default = readonly[id][id];
+                if (id) {
```
petrjasek: this part is never triggered now, so maybe you can remove this and use `setDirty` instead

PR #1308, opened 2017-02-03
fix(featured media): Cannot drag and drop Scanpix media to "featured …

ioanpocol:
…media" field (SDNTB-379)


ioanpocol:
@mdhaman Please review this PR because it is related to the following PR/commit:
https://github.com/superdesk/superdesk-client-core/pull/1259/commits/846fdfe99e7633767fbab2413730688107636295

PR #1311, opened 2017-02-03
[SDESK-655] - Display marked stories in incoming stage

akintolga:
Originally it was "output" stage...


scripts/apps/monitoring/services/CardsService.js
```diff
@@ -67,7 +67,13 @@ export function CardsService(api, search, session, desks, config) {
 
         default:
             if (!_.isNil(card.singleViewType) && card.singleViewType === 'desk') {
-                query.filter({term: {'task.desk': card.deskId}});
+                query.filter({or: [
+                        {terms: {'marked_desks.desk_id': [card.deskId]}},
```
mdhaman: a minor one. change this filter to `term` instead of `terms` and outcome will be the same.

PR #1313, opened 2017-02-04
fix(multi-action) - Keep selected items in single desk view

akintolga:
This PR addresses few minor issues with multi selections:
- Keeps the selected items if user moves to single desk or single stage view and back
- Added a hide button for multi select header panel so user can do a search or go back to monitoring
- Changed the initial number of items in single group from 10 to 50

### DIFF
--- a/scripts/apps/monitoring/views/monitoring-view.html
+++ b/scripts/apps/monitoring/views/monitoring-view.html
@@ -155,7 +155,7 @@
                     ></div>
                 </div>
                 <div ng-if="monitoring.singleGroup && !monitoring.isDeskChanged()" class="list">
-                    <div sd-monitoring-group class="single-group" data-group="monitoring.singleGroup" data-num-items="10" data-view-type="'single_monitoring'">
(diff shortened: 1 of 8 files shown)


akintolga:
@sjunaid please let me know if increasing the size of single group from 10 to 50 would create an issue. Thanks.

scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -155,7 +155,7 @@ <h3 class="subnav__page-title">
                     ></div>
                 </div>
                 <div ng-if="monitoring.singleGroup && !monitoring.isDeskChanged()" class="list">
-                    <div sd-monitoring-group class="single-group" data-group="monitoring.singleGroup" data-num-items="10" data-view-type="'single_monitoring'">
+                    <div sd-monitoring-group class="single-group" data-group="monitoring.singleGroup" data-num-items="50" data-view-type="'single_monitoring'">
```
sjunaid: @tolga, seems like num-items not utilized in directive code. imo you can remove this for this PR. and that part 
- Changed the initial number of items in single group from 10 to 50.

I will handle it in separate PR.
Rest of changes LGTM.

PR #1329, opened 2017-02-08
fix(monitoring widget): Missing items in monitoring widget

pavlovicnemanja:
SDESK-540 - Ingest Dashboard: UI issue in "Add ingest sources" menu
SDESK-692 - Items are missing in the list in the monitoring widget

### DIFF
--- a/scripts/apps/desks/directives/StageItemListDirective.js
+++ b/scripts/apps/desks/directives/StageItemListDirective.js
@@ -31,8 +31,7 @@ export function StageItemListDirective(search, api, superdesk, desks, cards, $ti
 
             scope.page = 1;
             scope.fetching = false;
-            scope.cacheNextItems = [];
-            scope.cachePreviousItems = [];
+            scope.cache = [];
 
             /**
               * Generates Identifier to be used by track by expression.
@@ -90,8 +89,7 @@ export function StageItemListDirective(search, api, superdesk, desks, cards, $ti
                     .then((items) => {
                         scope.items = items._items;
                         scope.total = items._meta.total;
-
-                        scope.cachePreviousItems = items._items;
+                        scope.cache = items._items;
                         setNextItems(criteria);
                     })
                     .finally(() => {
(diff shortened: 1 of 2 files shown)


scripts/apps/desks/directives/StageItemListDirective.js
```diff
@@ -206,12 +206,18 @@ export function StageItemListDirective(search, api, superdesk, desks, cards, $ti
...
+                if (!scope.fetching && scope.page > 2 && criteria.source.from > 0) {
                     scope.fetching = true;
                     scope.page -= 1;
-                    criteria.source.from = scope.page > 3 ? (scope.page - 3) * criteria.source.size : 0;
                     scope.loading = true;
 
+                    if (scope.page > 3) {
+                        criteria.source.from = (scope.page - 2) * criteria.source.size;
+                    } else {
+                        criteria.source.size *= 2;
```
mdhaman: Why do we need to double size of records fetched over here?

pavlovicnemanja: While fetching previous, in last row we need to fetch double size of source.size to load last two pages together. We skip first page in `fetchPrevious`, to avoid trying to load previous items on first page.

petrjasek: issue is this will only grow, it's never reset back. and because it is used to compute `from` it would skip some items later on. better keep this constant

pavlovicnemanja: Yes, but in this case this is changed only if `scope.page` is lower than 3. Without this, with old settings it will skip 20 items at the end.

petrjasek: lets say size is 25 by default, once you get to page 2 it will be 50, so when you get to page 3 it will make from = 50, so far so good, you go to page 2 again and size = 100, you go to page 3 and from is now 100, next time it will be 200, etc

petrjasek: and it will be fetching 50/100/200/400/etc items as well

pavlovicnemanja: This is only for `fetchPrevious`, for last two pages... it will go only once, when page is 3, it will fetch last 2 pages together... This is not set for `fetchNext`... For `fetchNext` it will go page by page.

pavlovicnemanja: Case you described in above comment is for `fetchNext` functionality

petrjasek: just tested locally, after some scrolling down and up, from and size are like:
```
prev 80 40
prev 0 80
next 240 80
next 320 80
next 400 80
next 480 80
prev 320 80
prev 240 80
prev 160 80
prev 0 160
next 480 160
next 640 160
next 800 160
```
plus some errors about duplicate items

petrjasek:
hey @pavlovicnemanja can you check that failing e2e test?

pavlovicnemanja:
Working on them, had some issues with local testing instace

petrjasek:
:+1:

PR #1335, opened 2017-02-10
[SDESK-494] Export and download feature for articles

nrvikas:
Depends on https://github.com/superdesk/superdesk-core/pull/798

Refactored Preview-Format and Export functionality. Hence, needs another round of review.

### DIFF
--- a/scripts/apps/archive/directives/Export.js
+++ b/scripts/apps/archive/directives/Export.js
@@ -1,0 +1,113 @@
+class LinkFunction {
+    constructor(api, config, multi, notify, storage, session, urls, scope) {
+        this.api = api;
+        this.config = config;
+        this.multi = multi;
+        this.notify = notify;
+        this.storage = storage;
+        this.session = session;
+        this.urls = urls;
+        this.scope = scope;
+
+        this.init();
+    }
+
+    /**
+     * @ngdoc method
+     * @name sdExport#init
+     * @private
+     * @description Initializes the directive with default values for the scope
+     */
+    init() {
+        this.scope.loading = false;
+        this.scope.selectedFormatter = this.storage.getItem('selectedFormatter') ||
+                                        JSON.stringify(this.scope.exportFormatters[0]);
+        this.scope.validate = false;
+        this.scope.exportFile = this.exportFile.bind(this);
+        this.scope.failures = 0;
+        this.scope.error = false;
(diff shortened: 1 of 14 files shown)


scripts/apps/archive/directives/Export.js
```diff
@@ -0,0 +1,109 @@
...
+        this.scope.loading = true;
+        this.storage.setItem('selectedFormatter', formatterString);
+        let formatter = JSON.parse(formatterString);
+        let itemIdList = this.scope.multi ? _.map(this.multi.getItems(), '_id') : [this.scope.item._id];
+
+        this.api.save('export', {}, {item_ids: itemIdList, format_type: formatter.name, validate: validate})
+        .then((item) => {
+            if (item._id.err_msg) {
+                this.onError(item._id.err_msg);
+            } else {
```
gbbr: You can lose the `else` branch and just `return this.onError`.

scripts/apps/archive/directives/Export.js
```diff
@@ -0,0 +1,109 @@
...
+     * @param {Boolean} validate - item to be validated for publish action
+     * @description Calls 'export' endpoint with the request and downloads file if export was successful
+     */
+    exportFile(formatterString, validate) {
+        this.scope.loading = true;
+        this.storage.setItem('selectedFormatter', formatterString);
+        let formatter = JSON.parse(formatterString);
+        let itemIdList = this.scope.multi ? _.map(this.multi.getItems(), '_id') : [this.scope.item._id];
+
+        this.api.save('export', {}, {item_ids: itemIdList, format_type: formatter.name, validate: validate})
```
gbbr: I think it would give for a nicer API if you would return this promise, so that callers can tell that the export is completed. Ideally you would have to return the `item` on success in the `.then` call too.
```
@returns {Promise}
```

nrvikas: Hi Gabriel, I have made some changes. Just noticed might need some updates. Sorry about that. Will update it tomorrow.

scripts/apps/archive/directives/Export.js
nrvikas: Same here :D

hlmnrmr:
@nrvikas We have posted a comment to the corresponding user story in JIRA. We would like to understand better the use case for this before approving it.

PR #1336, opened 2017-02-10
added wufoo service provider

jerome-poisson:
Wufoo is a form building external service, and integration was
requested by NTB.
This commit add it client-side, and hide feed parsed and article-type
when it is selected. This is currently hardcoded, but future improvment
is planed to do it generically in the backend (SDESK-716).

needed for SDNTB-377

### DIFF
--- a/scripts/apps/ingest/directives/IngestSourcesContent.js
+++ b/scripts/apps/ingest/directives/IngestSourcesContent.js
@@ -312,6 +312,11 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                 } else {
                     $scope.provider.config = {};
                 }
+                if ($scope.provider.feeding_service === 'wufoo') {
+                    // TODO(jerome-poisson): temporary hardcoded in client, will move to backend
+                    //                       when SDESK-716 will be fixed
+                    $scope.provider.feed_parser = 'wufoo';
+                }
             };
 
             /**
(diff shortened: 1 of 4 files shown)


scripts/apps/ingest/views/settings/wufoo.html
```diff
@@ -0,0 +1,12 @@
+<div class="field">
+    <label for="wufoo_username" translate>Login</label>
+    <input type="text" id="wufoo_username" placeholder="{{:: 'Wufoo login'|translate }}"
+      ng-model="provider.config.wufoo_username" required
```
gbbr: The indentations here seem off...

scripts/apps/ingest/directives/IngestSourcesContent.js
```diff
@@ -312,6 +312,11 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                 } else {
                     $scope.provider.config = {};
                 }
+                if ($scope.provider.feeding_service === 'wufoo') {
+                    // FIXME: temporary hardcoded in client, will move to backend
```
gbbr: Isn't `TODO(jerome-poisson): ....` the standard?

scripts/apps/ingest/views/settings/wufoo.html
jerome-poisson: Hi @gbbr thanks for the review.

I'm not sure about the indentation rules used in the client, but I've used the same as in other views (e.g. https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/ingest/views/settings/emailConfig.html). How would you change that?

scripts/apps/ingest/directives/IngestSourcesContent.js
jerome-poisson: Didn't you about that, will change.

scripts/apps/ingest/views/settings/wufoo.html
gbbr: I would do like the file above this one for example.

PR #1337, opened 2017-02-10
fix monitoring widget

petrjasek:
rebased https://github.com/superdesk/superdesk-client-core/pull/1329


petrjasek:
it didn't do the trick

PR #1342, opened 2017-02-13
fix(search): handle list of search providers by ids

petrjasek:


scripts/apps/search/views/item-repo.html
```diff
@@ -11,7 +11,7 @@
     </div>
     <div ng-repeat="provider in providers track by provider._id">
         <label class="sd-radio">
-            <input type="radio" ng-value="provider.search_provider" ng-model="repo.search" ng-checked="isDefault(provider)" ng-change="toggleRepo(provider.search_provider)">
+            <input type="radio" ng-value="provider._id" ng-model="repo.search" ng-change="toggleRepo(provider._id, true)">
```
mdhaman: @petrjasek if `provider._id` is used then `repo.search` scope member points to `ObjectID` of the search provider and there are checks in the markup using `search_provider` name and those checks fail. 
https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/search/views/search-parameters.html#L154
https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/search/views/search-parameters.html#L180

petrjasek: that's ok for now, it is not replacing those old implementations

petrjasek: my bad, that would break it obviously

petrjasek:
not ready yet

PR #1344, opened 2017-02-13
fix(search): Fix filter for custom fields (SDNTB-342)

ioanpocol:


scripts/apps/search/services/SearchService.js
```diff
@@ -33,6 +33,8 @@ export function SearchService($location, gettext, config, session, multi) {
         {field: 'genre.name', label: gettext('Genre')}
     ];
 
+    var self = this;
```
gbbr: Move this to line 94 please - it's only used there.

ioanpocol: It is not working because it is already imbricated in other function.
Also we already used this pattern over the application and we always declared the self at the beginning.

gbbr: Ok, sure

PR #1349, opened 2017-02-14
[SDESK-751] - Make the word counts consistent between ui and server

akintolga:


akintolga:
Could this be smallest possible change for a PR?   :)

PR #1358, opened 2017-02-17
Fix typos and grammar in source strings

danielhjames:
Just some errors that I found on Transifex while reviewing strings.


danielhjames:
You're welcome :-)

PR #1367, opened 2017-02-21
feat(profile): Change of content profile should overwrite item values…

ioanpocol:
… with defaults from the CP

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.js
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.js
@@ -114,6 +114,10 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
             $scope.proofread = false;
             $scope.referrerUrl = referrer.getReferrerUrl();
 
+            content.getTypes().then(() => {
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -754,6 +758,27 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
...
             };
 
+            /**
+            * On change content profile the default values from new content profile
+            * will overwrite the current values from item
+            *
+            * @function changeProfile
+            * @param {Object} item - ucurrent edited content items
+            */
+            $scope.changeProfile = function(item) {
```
petrjasek: imo it should at least check if the value is same as default for previous profile, otherwise we might override user work

ioanpocol: I agree with this, but this was the requirement to override anyway, even if the user already has set a different value.

petrjasek: I see, ignore my comment then

PR #1380, opened 2017-02-23
[SDESK-746] - Wiring UI for new history items

akintolga:
- Backward compatibility is done in `HistoryController`:
  - If there's no new history items for a story it will drive history from versions (as it was before)
  - If the new history starts from version > 1 then new history will be merged to old
  - If new history starts from version 1 then new history will be displayed
- History items for system activated actions will have 'System' [SDESK-768]
- Removed duplicates and history tabs for 'Archived' items


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/810

PR #1383, opened 2017-02-27
[SDESK-796] Bug fix export feature

nrvikas:


mdhaman:
@petrjasek this should be merged with 1.7 not 1.6 milestone.

PR #1384, opened 2017-02-27
fix(ui): fix sdModal closing via keyboard

petrjasek:
model reseting doesn't work everywhere, so better do this explicitly

SDESK-694


petrjasek:
follow up https://github.com/superdesk/superdesk-client-core/pull/1363

PR #1387, opened 2017-03-02
[SDESK-794] - Display 'marked for desk' items as a separate filtered …

sjunaid:
…view in Monitoring
- Reverted https://github.com/superdesk/superdesk-client-core/pull/1311 changes, rather implementation,
- Requires setting up (say:) 'Attention Sports Desk' [Global] saved search via global search (for marked items), i.e. saving search for selected desk in 'Marked Desk' filter with only Production and Published repos selected. Then turning ON that Attention Desk saved search via 'desk monitoring settings' will display Attention Desk list (marked items) in monitoring view of the desk.
- updated e2e to simulate the process

### DIFF
--- a/scripts/apps/monitoring/services/CardsService.js
+++ b/scripts/apps/monitoring/services/CardsService.js
@@ -72,13 +72,7 @@ export function CardsService(api, search, session, desks, config) {
 
         default:
             if (!_.isNil(card.singleViewType) && card.singleViewType === 'desk') {
-                query.filter({or: [
-                        {terms: {'marked_desks.desk_id': [card.deskId]}},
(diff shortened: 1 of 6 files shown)


mdhaman:
@hlmnrmr  as discussed made changes so that mark desk will use saved search. It can be done using pre-confiured saved search like `desk output` in desk settings. But currently relying on the admin or user with privileges to configure mark desk saved search for the required desk.

Please review and give us you feedback.

mdhaman:
@hlmnrmr  why do we want to limit only user with permissions to search marked items? if you wish to remove the field from the search that can be done via config.

hlmnrmr:
@mdhaman please proceed as discussed this morning over the call

PR #1391, opened 2017-03-03
[SDESK-834] Show take-key, related item count, and missing-link in authoring header pane if applicable to the story.

nrvikas:


scripts/apps/authoring/views/authoring-header.html
```diff
@@ -23,10 +23,10 @@
             <label translate>SOURCE</label>
             <div id="item-source" class="data">{{ item.source }}</div>
         </div>
-        <div ng-if="item.anpa_take_key && !schema.relatedItems">
+        <div ng-if="item.anpa_take_key">
```
mdhaman: remove this `anpa_take_key`. there is already input field for `anpa_take_key`.

nrvikas: Done @mdhaman

PR #1395, opened 2017-03-05
[SDESK-775] - Allow inclusion of spiked content in global search

akintolga:
- Replaced `In Spiked` switch with a dropdown in search parameters.

### DIFF
--- a/scripts/apps/search/views/search-parameters.html
+++ b/scripts/apps/search/views/search-parameters.html
@@ -110,24 +110,26 @@
             </div>
         </div>
 
-        <div class="multiple">
-            <div class="field" ng-if="repo.archive && search_config.creator">
-                <label for="search-creator">
-                    {{:: 'Creator' | translate}}
-                </label>
-                <select id="search-creator"
-                        ng-model="fields.original_creator"
-                        ng-options="user._id  as user.display_name for user in userList track by user._id">
-                <option value="" label=""></option>
-                </select>
-            </div>
+        <div class="field" ng-if="repo.archive && search_config.creator">
+            <label for="search-creator">
+                {{:: 'Creator' | translate}}
+            </label>
+            <select id="search-creator"
+                    ng-model="fields.original_creator"
+                    ng-options="user._id  as user.display_name for user in userList track by user._id">
+            <option value="" label=""></option>
+            </select>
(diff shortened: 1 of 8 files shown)


scripts/apps/search/views/search-parameters.html
```diff
@@ -110,24 +110,25 @@
...
-                {{:: 'In Spiked' | translate}}
-                </label>
-                <span sd-switch ng-model="fields.spike"></span>
-            </div>
+        <div class="field" ng-if="search_config.spike">
+            <label for="spike-options">
+                {{:: 'Spiked Content' | translate}}
+            </label>
+            <select id="spike-options" name="spike-options" ng-model="fields.spike">
+                <option value="include" label="Include spiked content"></option>
```
petrjasek: pls translate those labels as well

akintolga:
Added translations to the labels

scripts/apps/search/views/search-parameters.html
```diff
@@ -110,24 +110,25 @@
...
-                {{:: 'In Spiked' | translate}}
-                </label>
-                <span sd-switch ng-model="fields.spike"></span>
-            </div>
+        <div class="field" ng-if="search_config.spike">
+            <label for="spike-options">
+                {{:: 'Spiked Content' | translate}}
+            </label>
+            <select id="spike-options" name="spike-options" ng-model="fields.spike">
+                <option value="include" label="{{:: 'Include spiked content' | translate }}"></option>
```
mdhaman: @akintolga  Once you select `Include spiked content` or `Spiked only content`. You cannot exclude spiked content i.e. the default search without spiked content. You have to clear filters. Can you include the option i.e. exclude spiked content?

akintolga: You can remove it by removing the tag without clearing filters. 

I originally put that in but since the default is already "excluded" it looked redundant and removed. Let me know if you think it would be better with it still.

Michael-Gourlay: @akintolga IMO better with it in and selected as default - as per the Jira ticket.

akintolga: No problem, I'll put it in.

PR #1403, opened 2017-03-08
feat(widget): Rewrite monitoring widget with ReactJS

ioanpocol:


scripts/apps/monitoring/directives/WidgetGroup.js
```diff
@@ -0,0 +1,364 @@
...
+                criteria = cards.criteria(scope.stage, queryString);
+                if (!scope.fetching) {
+                    scope.setLoading(true);
+                }
+
+                oldSize = criteria.source.size;
+                criteria.source.from = 0;
+
+                if (scope.page > 0 && criteria.source) {
+                    criteria.source.size = (scope.page - 1) * oldSize;
```
petrjasek: imo we should change from and not size, otherwise it could be too much for server after some scrolling

ioanpocol: Looks that the same solution is used on monitoring: 
https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/monitoring/directives/MonitoringGroup.js#L364
If you don't load all items it will appear issues if an item is deleted from list (ex: moved to other stage or publish).
I think for server is not much difference on returning 100 or 1000 rows, and usually a user don't scroll more than 40 pages.
Due to unique keys on items React is fast on rendering so I think that all old vs new list processing on search.updateItems is not needed anymore for items from monitoring groups.

akintolga: I think there would be a difference in server returning 100 or 1000 rows. Payload is increasing with that many rows is one thing that makes elastic server and network busy. Also when the stories are fetched we call `enhance_items` functions to attach additional information, this increases CPU utilization. If few users are scrolling then it would affect overall performance.

I agree with Petr the we should try to get 25-50 items everytime but on client we can keep a larger number of items.

PR #1426, opened 2017-03-16
[SDESK-896] Added time--short CSS class to have an option to display time in limited width ( no date only time) used in planning

nrvikas:


ioanpocol:
Please rebase in order to pass the tests

PR #1432, opened 2017-03-20
[SDESK-900] - Add 'Get From External Source To' action for external s…

sjunaid:
…ource items.

Depends on https://github.com/superdesk/superdesk-core/pull/847

### DIFF
--- a/scripts/apps/authoring/views/send-item.html
+++ b/scripts/apps/authoring/views/send-item.html
@@ -18,6 +18,10 @@
             class="panel-header {{currentUserAction === vm.userActions.duplicate_to ? 'panel-header-active' : ''}}"
             ng-show="!canPublishItem() && currentUserAction === vm.userActions.duplicate_to"
             translate>Duplicate To</h3>
+        <h3 id="panel-externalsource"
+            class="panel-header {{currentUserAction === vm.userActions.externalsource_to ? 'panel-header-active' : ''}}"
+            ng-show="!canSendItem() && !canPublishItem() && currentUserAction === vm.userActions.externalsource_to"
+            translate>Get From External Source To</h3>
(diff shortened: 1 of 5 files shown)


scripts/apps/authoring/views/send-item.html
```diff
@@ -100,6 +104,27 @@
         </ul>
     </div>
 
+    <div class="slide-pane__footer" sd-loading="loading" ng-show="currentUserAction === vm.userActions.externalsource_to">
+        <ul class="btn-list">
+            <li class="full-width" ng-if="showSendButtonAndDestination() && !canSendItem()">
+                <button id="externalsource-open-btn" class="btn btn--light-gray"
+                        ng-click="send(true)"
+                        ng-disabled="disableFetchAndOpenButton()"
```
akintolga: Please double check if `disableFetchAndOpenButton` is relevant if the story is retrieved from external source.

scripts/apps/authoring/views/send-item.html
```diff
@@ -100,6 +104,27 @@
...
+                        translate>
+                    get from external source and open
+                </button>
+            </li>
+            <li class="full-width" ng-if="showSendButtonAndDestination()">
+                <button id="externalsource-btn" class="btn btn--primary"
+                        ng-click="send()"
+                        ng-disabled="disableSendButton()"
```
akintolga: Like above, please double check if `disableSendButton` is relevant if the story is retrieved from external source.

scripts/apps/authoring/views/send-item.html
sjunaid: @akintolga , yes double checked, so if this action `Get From External source To` action is available then that would similar to `Fetch To'. so kept same checks. Thanks

PR #1442, opened 2017-03-22
[SDESK-847] Add 'Used' tab for media items if the item is used in any story

nrvikas:
Depends on: https://github.com/superdesk/superdesk-core/pull/850


mdhaman:
@nrvikas please rebase

PR #1452, opened 2017-03-23
[SEARCH PROVIDERS] fixed use of source instead of label where search …

jerome-poisson:
…providers are displayed

SDNTB-385

### DIFF
--- a/scripts/apps/search/directives/ItemRepo.js
+++ b/scripts/apps/search/directives/ItemRepo.js
@@ -24,7 +24,7 @@ export function ItemRepo(
                 scope.search_config = metadata.search_config;
 
                 searchProviderService.getAllowedProviderTypes().then((providerTypes) => {
-                    scope.searchProviderTypes = providerTypes;
+                    scope.providerLabels = searchProviderService.getProviderLabels(providerTypes);
                 });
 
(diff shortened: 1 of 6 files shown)


scripts/apps/search/directives/ItemRepo.js
```diff
@@ -25,6 +25,10 @@ export function ItemRepo(
 
                 searchProviderService.getAllowedProviderTypes().then((providerTypes) => {
                     scope.searchProviderTypes = providerTypes;
+                    scope.providerLabels = {};
+                    scope.searchProviderTypes.forEach((type) => {
```
petrjasek: imo this would be better done in `searchProviderService.getAllowedProviderTypes()` so you don't have to duplicate this code. it could populate `type.label` field and just use it.

jerome-poisson:
@petrjasek getAllowedProviderTypes() return an array, and we need to do reverse map by search provider.source and getting label. label is not saved in search_provider service in the api, so we can't get it directly using provider.label in the lines modified by this patch.

I've put the Label reverse mapping code in the service, so it's not duplicated anymore. An other way would be to add label in search_provider Service, but that would break existing instances (as they don't have label), and I'm not sure if it worth it.

PR #1463, opened 2017-03-27
chore(angular): update angular to 1.6

petrjasek:
main issue was https://github.com/angular/angular.js/commit/4f6f2bce4
so it needs an extra handling in auth interceptor for protractor to
detect when page is done with requests.

SDESK-623


naspeh:
:+1:

PR #1470, opened 2017-03-29
[SDESK-1004] - Remember last duplicate to destination

akintolga:
Also fixes:
- [SDESK-1005] - Send panel is visible along with duplicate to
- Missing clear button for embargo


akintolga:
@mdhaman I see, I'll make sure that `Publish` won't remember it then. Thanks for the notification.

akintolga:
Publish panel panel will revert to current desk (no memory) but send, duplicate and fetch panels will share the persisted desk and stage.

PR #1471, opened 2017-03-30
[SDESK-969] - On changing the original in the modify crop screen it c…

MarkLark86:
…reates duplicate crops.


petrjasek:
depends on https://github.com/superdesk/superdesk-core/pull/869

PR #1485, opened 2017-04-03
fix(datetime): use moment-timezone as moment service

jerome-poisson:
to make sure .tz is defined on it


jerome-poisson:
this is a cherry-pick of a patch from @petrjasek on master, to fix installation issue on 1.6 branch. Successfully tested on wpp-1.6 superdesk branch.

PR #1495, opened 2017-04-06
feat(superdesk-ui-framework): UI framework integration

pavlovicnemanja:
Main idea is to move all UI components outside of main superdesk-client-core repo, and put them in separate repo so all future superdesk plugins can rely on it:

Documentation: https://superdesk.github.io/superdesk-ui-framework/examples/
GIT Repo:  https://github.com/superdesk/superdesk-ui-framework

Framework is not complete, we will need to add more code there, also to do some "cleaning" of code that is not using anymore in superdesk.


gbbr:
@pavlovicnemanja @fritzSF can you guys please put together an overview or an outline of the idea and changes intended with this PR?

It's hard to get a general feeling, given that you have a PR that adds 1780 lines and removes 7k...

gbbr:
I can see it's mostly standardizing CSS class names and improvements, but also some HTML has been added. So a description to the PR would be super appreciated.

I've taken the liberty to add the `wip` label while we work it out, I hope you don't mind.

pavlovicnemanja:
@gbbr it is not just renaming things, lot of stuff is removed from Superdesk and refactored inside Superdesk UI framework and like that is included in Superdesk. There is a lot of leftovers that we will need to remove from Superdesk.

petrjasek:
what is the idea regarding translations? is this happening in the framework and we need to import it somehow?

pavlovicnemanja:
Translations are done on superdesk side. We are sending translated string to some UI framework directive.

petrjasek:
would be good imo to use a prefix for directives, so it won't conflict with other tools.

pavlovicnemanja:
we are using, it is `sd-`. You mean different then one defined in Superdesk?

package.json
```diff
@@ -106,7 +105,8 @@
     "sass-loader": "^4.0.2",
     "style-loader": "^0.13.1",
     "webpack": "^1.13.1",
-    "webpack-dev-server": "1.14.1"
+    "webpack-dev-server": "1.14.1",
+    "superdesk-ui-framework": "git://github.com/superdesk/superdesk-ui-framework#master"
```
gbbr: imo it's nicer to have:
```js
"superdesk-ui-framework": "superdesk/superdesk-ui-framework"
```
It has the same outcome, I think...

gbbr:
@darconny you can update the same commit, without pushing new ones, by using `amend`:
```js
git commit -a --amen
git push -f <branch_name>
```

PR #1499, opened 2017-04-07
html/css for new publishing pane in publisher monitoring.

tomaszrondio:
This PR is only for Aca to pick html changes. 
Don't merge it.


gbbr:
He can access this branch by doing:
```bash
git remote add tomasz https://github.com/tomaszrondio/superdesk-client-core
git fetch tomasz
git checkout tomasz/publisher
```
No need to open a PR. The above should do the job, unless I misunderstood what you're doing here.

gbbr:
P.S. Or you can share [this link](https://github.com/superdesk/superdesk-client-core/compare/publisher...tomaszrondio:publisher) for the exact same outcome, without engaging CI

PR #1500, opened 2017-04-09
[SDESK-1056] - Correct and Kill buttons missing when workqueue item opened

akintolga:
During a correction or kill if the story is minimized (or another story is opened) it loses the context and the correct publish button disappears. To solve the issue now we're:

- Saving the `action` when locking the story
- Using the `action` when story is opened from workqueue

### DIFF
--- a/scripts/apps/authoring/authoring/services/AuthoringService.js
+++ b/scripts/apps/authoring/authoring/services/AuthoringService.js
@@ -50,8 +50,9 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
      * @param {string} _id Item _id.
      * @param {boolean} readOnly
      * @param {string} repo - repository where an item whose identifier is _id can be found.
+     * @param {string} action - action performed to open the story: edit, correct or kill
      */
-    this.open = function openAuthoring(_id, readOnly, repo) {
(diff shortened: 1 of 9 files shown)


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/877

scripts/apps/authoring/authoring/services/AuthoringService.js
```diff
@@ -50,8 +50,9 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
      * @param {string} _id Item _id.
      * @param {boolean} readOnly
      * @param {string} repo - repository where an item whose identifier is _id can be found.
+     * @param {string} action - action performed to open the story: edit, correct or kill
      */
-    this.open = function openAuthoring(_id, readOnly, repo) {
+    this.open = function openAuthoring(_id, readOnly, repo, action) {
```
gbbr: Should there be a default value for this new parameter? Or is `undefined` ok?

akintolga: I think it should be ok. ArchiveLock endpoint handles the null value for action and saves as 'edit' if it is null.

PR #1505, opened 2017-04-10
fix(templates) - Handling orphan desks

akintolga:
- If the desk is missing the templates settings page becomes broken


akintolga:
@mdhaman This should address the issue in UAT in template settings page

PR #1508, opened 2017-04-11
update angular-embed to version for angular 1.6

petrjasek:
SDESK-1063

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "angular-bootstrap-npm": "jusopi/angular-bootstrap-npm#0.13.1",
     "angular-contenteditable": "0.0.2",
     "angular-dynamic-locale": "0.1.32",
-    "angular-embed": "superdesk/angular-embed#2afd7397a7179f25b66c4b5f13ea2a0daf799821",
+    "angular-embed": "superdesk/angular-embed#b2b4a69",
     "angular-embedly": "Urigo/angular-embedly#0.0.8",
     "angular-gettext": "2.3.9",
     "angular-history": "decipherinc/angular-history#v0.8.0",
@@ -127,6 +127,7 @@
     "karma-webpack": "~1.8.1",
     "protractor": "~4.0.14",
     "react-addons-test-utils": "~15.5.1",
+    "react-test-renderer": "~15.5.1",
     "sinon": "~2.1.0"
   },
   "scripts": {


scripts/core/editor2/add-embed.ctrl.js
```diff
@@ -1,6 +1,16 @@
 var embedCodeHandlers = require('./embedCodeHandlers.js');
 
-angular.module('superdesk.apps.editor2.embed', []).controller('SdAddEmbedController', SdAddEmbedController);
+angular.module('superdesk.apps.editor2.embed', []).controller('SdAddEmbedController', SdAddEmbedController)
+    .config(['$sceDelegateProvider', ($sceDelegateProvider) => {
+        $sceDelegateProvider.resourceUrlWhitelist([
```
naspeh: this list repeats 2 times in https://github.com/superdesk/angular-embed/pull/5/files and here ones

maybe it should be in variable, what do you think?

scripts/core/editor2/editor.js
```diff
@@ -698,7 +698,8 @@ angular.module('superdesk.apps.editor2', [
     'superdesk.apps.editor2.utils',
     'superdesk.apps.spellcheck',
     'superdesk.apps.authoring',
-    'angular-embed'
+    'angular-embed',
+    'angular-embed.handlers'
```
gbbr: This seems unusual. Should `angular-embed` not automatically include `angular-embed.handlers`?

petrjasek: it seems to be the other way around https://github.com/superdesk/angular-embed/blob/master/dist/angular-embed.js#L28

scripts/core/editor2/add-embed.ctrl.js
petrjasek: I've removed it from there, should be enough to have it in lib

package.json
```diff
@@ -33,7 +33,7 @@
     "angular-bootstrap-npm": "jusopi/angular-bootstrap-npm#0.13.1",
     "angular-contenteditable": "0.0.2",
     "angular-dynamic-locale": "^0.1.32",
-    "angular-embed": "superdesk/angular-embed#2afd7397a7179f25b66c4b5f13ea2a0daf799821",
+    "angular-embed": "petrjasek/angular-embed#79c1038",
```
naspeh: I'll change this after merging PR in `superdesk/angular-embed`, right?

petrjasek: updated it

naspeh: great! BTW: I checked on http://sdcpr-1508.test.superdesk.org/ instance and it's working now :)

scripts/core/editor2/editor.js
gbbr: Maybe we should have only one then... ?

petrjasek: I've removed the later, it wasn't there before and it is not needed for the fix

package.json
```diff
@@ -127,6 +127,7 @@
     "karma-webpack": "~1.8.1",
     "protractor": "~4.0.14",
     "react-addons-test-utils": "~15.5.1",
+    "react-test-renderer": "~15.5.1",
```
naspeh: this the fix for `npm test`, right?

petrjasek: yes

PR #1511, opened 2017-04-12
fix(dependencies): use exact versions for dependencies

petrjasek:
so there are no surprises in production. allow minor version
updates for development dependencies.


naspeh:
with such version pinning, there is a process needed to update these dependencies during releases... because if keep them outdated it could be difficult to update in the future releases.

but in general it's good idea.

I was thinking about another approach, to upload `dist` (after `grunt build`) directory during releases and simplify deployment for releases (just serve this `dist` directory with `config.js`).

Releases could have assets: https://developer.github.com/v3/repos/releases/#list-assets-for-a-release

petrjasek:
yep `~` should be default, but last time our build broke was when moment-timezone released 0.5.12, while 0.5.11 works fine..

petrjasek:
@naspeh yep but the point here is that what you have on eg qa instance will be the same what you get in production one week later. which won't use any uploaded files on github etc. but avoiding building client on install would make sense, this would speed up the whole process a lot :+1:

naspeh:
yep, good addition about "qa instance", agreed.

akintolga:
I agree that more control on dependencies is a good idea. 👍 

Stating the obvious but I think that version upgrades should also be testable in itself so they need to be done within a ticket so that -say- when moment is upgraded qa can test the timezone related areas.

PR #1517, opened 2017-04-18
[SDESK-908] - Simplify workflow for "Correct" and "Kill"

MarkLark86:


mdhaman:
👍 Looks good. But not sure if `Cancel` action should go back to the view item or close the item.

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -664,6 +664,15 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
                 });
             };
 
+            /**
+             * Cancel an action, i.e. go back to view mode.
+             * This is used in the 'EDIT AND CORRECT' and 'EDIT AND KILL' actions.
+             * When the user clicks on the 'CANCEL' button it takes them back to the view action on the item
+             */
+            $scope.cancel = function() {
```
gbbr: Is `cancel` the best name for this? In this context of the scope of the _authoring directive_ which we are in, we should expect (judging by name) that this method should cancel authoring, but it doesn't, instead it opens the original item - so should we instead call it `viewOriginalItem` or something more specific? To avoid confusion.

Just a suggestion....

mdhaman:
@hlmnrmr @fritzSF  can you please review these changes....

PR #1518, opened 2017-04-18
[SDESK-1137] - Show the profile drop down for text items only.

MarkLark86:


scripts/apps/authoring/metadata/metadata.js
```diff
@@ -13,7 +13,8 @@ function MetadataCtrl(
         $scope.userLookup = desks.userLookup;
     });
 
-    $scope.change_profile = config.item_profile && config.item_profile.change_profile === 1;
+    $scope.change_profile = config.item_profile && config.item_profile.change_profile === 1 &&
+                            $scope.origItem.type === 'text';
```
gbbr: I think most places we've used 4 space indentation. I've seen some occurrences of this too, but afaik this is more appropriate for Python.

Odd that ESLint didn't complain.

PR #1519, opened 2017-04-18
feat(editor3): floating toolbar [SDESK-879]

gbbr:
When content is too long and scrolling down, float toolbar at top of the page container (just like editor2 does now):

apr-18-2017 12-29-21

### DIFF
--- a/scripts/core/editor3/components/Editor3.jsx
+++ b/scripts/core/editor3/components/Editor3.jsx
@@ -36,8 +36,11 @@ export class Editor3Component extends React.Component {
 
         this.editorRect = {top: 0, left: 0};
         this.readOnly = props.readOnly;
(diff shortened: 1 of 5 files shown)


scripts/core/editor3/components/Editor3.jsx
```diff
@@ -36,8 +36,11 @@ export class Editor3Component extends React.Component {
 
         this.editorRect = {top: 0, left: 0};
         this.readOnly = props.readOnly;
+        this.pageContainer = $('.page-content-container');
```
vied12: is this selector hardcoded on purpose ? maybe this could be set in the component parameters if we ever plan to reuse this editor somewhere else ?

gbbr:
@vied12 thanks for the review, you're right. I've added an attribute on the editor directive to take a string selector, which will be the scroll container.

PR #1523, opened 2017-04-19
[SDESK-1106] - When creating a new desk then a stage, macros are not …

MarkLark86:
…available unless the user saves and revisits


akintolga:
Nice work 👍 
One minor thing: you can remove `browser.pause` lines from the e2e test

PR #1527, opened 2017-04-19
fix(Missing Link): Deactivate "Missing Link" logic for related items…

ioanpocol:
… if noTakes is activated


mdhaman:
@ioanpocol @petrjasek we require this `missing_link` as this is not dependent on the `no_takes` flag. It is visual indicator for user that they have not linked the story via (Update or Takes). 
Can we control it via profile?

ioanpocol:
Fixed, here is the PR: https://github.com/superdesk/superdesk-client-core/pull/1532

PR #1555, opened 2017-05-02
[SDESK-1206] - Allow desks and templates to deselect a profile

akintolga:


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/909

PR #1565, opened 2017-05-09
[SDESK-1259](fix): Fixing markup for preview and export.

mdhaman:


scripts/apps/authoring/styles/authoring.scss
```diff
@@ -1244,4 +1244,14 @@ feature-image {
             opacity: .2;
         }
   }
-}
\ No newline at end of file
+}
+
+// Styling for formatted preview popup
+// -------------------------------------------------------------
+.sd-line-input {
```
akintolga: Just wondering, Do we need to make changes to UI framework from now on?

mdhaman: Maybe not. It got clarified in the standup now.

fritzSF: Everything that can be reused elsewhere will be added to the framework at some point. The framework will be maintained and extended only by a couple of people though, in order to keep it tight and consistent. New CSS that has to be added to Superdesk during development should follow the BEM logic from the framework. Main pointers to follow: 1) no nesting of CSS, so that the code can be reused elsewhere. 2) naming the classes in a way that it's clear what it does 3) following the logic of appending modifier classes to the base class.

PR #1570, opened 2017-05-10
Big icon font implementation

darconny:
- from SD UI framework, refactoring


scripts/apps/authoring/comments/comments.scss
```diff
@@ -59,8 +59,8 @@
               width: 56px;
               height: 48px;
               opacity: 0.2;
-              background-size: 100%;
-              margin: 8px auto 0 auto;
+              margin: 14px auto 0 auto;
```
vied12: what about using `rem` here  ?

scripts/apps/authoring/comments/comments.scss
```diff
@@ -59,8 +59,8 @@
...
+              font-size: 52px;
```
vied12: and here too ?

darconny: We can use rem's, yes. Not sure if we refactor everything at this point.

PR #1572, opened 2017-05-11
[SDESK-774]-(fix) - Send button should not be active if the current d…

sjunaid:
…esk/stage of the item is selected

### DIFF
--- a/scripts/apps/authoring/authoring/directives/SendItem.js
+++ b/scripts/apps/authoring/authoring/directives/SendItem.js
@@ -200,11 +200,10 @@ export function SendItem($q, api, desks, notify, authoringWorkspace,
              * @returns {Boolean}
              */
             scope.disableSendButton = function() {
-                if (scope.item && scope.item.task) {
-                    return !scope.selectedDesk ||
-                           scope.mode !== 'ingest' && scope.selectedStage && scope.mode !== 'spike' &&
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/directives/SendItem.js
```diff
@@ -200,11 +200,10 @@ export function SendItem($q, api, desks, notify, authoringWorkspace,
...
              */
             scope.disableSendButton = function() {
-                if (scope.item && scope.item.task) {
-                    return !scope.selectedDesk ||
-                           scope.mode !== 'ingest' && scope.selectedStage && scope.mode !== 'spike' &&
-                           scope.selectedStage._id === scope.item.task.stage;
-                }
+                return !scope.selectedDesk ||
+                        scope.mode !== 'ingest' && scope.selectedStage && scope.mode !== 'spike' &&
+                        (_.get(scope.item, 'scope.item.task.stage') === scope.selectedStage._id ||
```
mdhaman: should this be `_.get(scope.item, 'task.stage')`

petrjasek: good point, maybe even `_.get(scope, 'item.task.stage')` given there was this check `if (scope.item && ...)`

sjunaid:
Thanks @petrjasek @mdhaman ,
fix updated.

PR #1575, opened 2017-05-11
fix(stages): show error when saving item on readonly stage

petrjasek:
SDESK-1243


petrjasek:
requires https://github.com/superdesk/superdesk-core/pull/926

akintolga:
👍 

Just curious, how do we get to save a story on `readonly` stage in the first place? Is it by editing (which edit actions should have been disabled) or by creating a new one on that stage?

petrjasek:
@akintolga creating a new item when working stage is readonly (not very practical :wink: )

PR #1577, opened 2017-05-12
feat(cvs): allow item removal

petrjasek:
cv must have at least 1 more item or schema defined. and using tables for rendering

SDESK-1241


petrjasek:
schema definition https://github.com/superdesk/superdesk-core/pull/927

PR #1586, opened 2017-05-16
[SDESK-1234]- Fix the markup in the Filter search screen.

sjunaid:


mdhaman:
@petrjasek is affecting version 1.7

petrjasek:
sure np

PR #1589, opened 2017-05-16
fix(editor3): optimize html conversion

gbbr:
Improves HTML conversion performance from 2s to ~10ms for 800+ words.


scripts/core/editor3/html/to-html/HTMLGenerator.js
```diff
@@ -192,9 +174,12 @@ export class HTMLGenerator {
      * this instance.
      */
     html() {
-        return this.contentState
-            .getBlockMap()
-            .map(this.convertBlock)
-            .join('');
+        let html = '';
+
+        this.blocks.forEach((block, id) => {
```
petrjasek: imo looks like reducing:
```
return this.blocks.reduce((html, block, id) => html + this.convertBlock(block, id), '');
```

PR #1591, opened 2017-05-17
feat(keywords): Move keywords field on authoring view, header section

ioanpocol:


scripts/apps/authoring/metadata/metadata.js
```diff
@@ -556,6 +556,33 @@ function MetaWordsListDirective() {
...
+        templateUrl: 'scripts/apps/authoring/metadata/views/metadata-keywords-list.html',
+        link: function(scope, element) {
+            scope.words = scope.list ? scope.list.join(' ') : '';
+
+            scope.changeList = function() {
+                var wordList = scope.words.split(' ');
+                var o = {};
+
+                o[scope.field] = wordList;
+                _.extend(scope.item, o);
```
petrjasek: not sure why `o` is used here, why not just `scope.item[scope.field] = scope.words.split(' ' );` ?

petrjasek: probably should be used with `.filter((x) => x);` to remove empty strings (for input with multiple spaces)

scripts/apps/authoring/metadata/views/metadata-widget.html
```diff
@@ -37,7 +37,7 @@
             </li>
             <li class="terms-box" ng-if="!$root.config.features.alchemy">
                 <label translate>Keywords</label>
-                <div sd-meta-words-list data-item="item" data-field="keywords" data-change="autosave(item)" data-header="_editable" data-list="metadata.keywords" data-disabled="!_editable"></div>
```
mdhaman: @ioanpocol we have list of predefined keywords 
https://github.com/superdesk/superdesk-aap/blob/master/server/data/vocabularies.json#L233

The `sd-meta-words-list` allows you to define arbitrary keywords as well as predefined one. Does the new one does the same?

PR #1616, opened 2017-05-23
feat(config): use client_config resource to get allowed auth methods

petrjasek:
SDESK-1327

uses https://github.com/superdesk/superdesk-core/pull/942


petrjasek:
I might put it here, but I'm struggling with naming - would like to add a service for this, but we already use config..

petrjasek:
@naspeh added it here

petrjasek:
ups, we have even `serverConfig` service already. will call it `serverClientConfig` or something :wink:

naspeh:
Maybe we can combine two of them to one? We can set that it requires authorization for modifying the `superdesk.server_config`, but all setting can be fetched without auth... What do you think?

petrjasek:
renamed it to `deployConfig` for now

PR #1643, opened 2017-05-31
feat(compare versions): Highlight the differences between versions

ioanpocol:


scripts/apps/authoring/compare-versions/CompareVersionsService.js
```diff
@@ -11,7 +11,7 @@
  * Compare versions screen has its boards and displays at least 2 of them.
  * Every board can be popuplated with one version of opened item.
  */
-const MIN_BOARDS = 2; // at least two boards to display for versions comparision.
+const MIN_BOARDS = 1; // at least two boards to display for versions comparision.
```
gbbr: The comment doesn't match the number...

scripts/apps/authoring/compare-versions/directives/CompareVersionsDropdownInnerDirective.js
```diff
@@ -22,7 +22,15 @@ class LinkFunction {
         // watches the selected versions of an item and filter out unselected versions
         // to provide in inner dropdown as a rest of available versions.
         this.scope.$watch(() => this.compareVersions.items, (items) => {
-            this.compareVersionsItems = _.map(this.compareVersions.items, (board) => board.article);
+            if (this.scope.compareWith) {
```
gbbr: Please fix this logic.

gbbr: For example:
```js
if (this.scope.compareWith) {
   this.compareVersionsItems = this.compareWithVersion ? [this.compareWithVersion] : [];
} else {
   this.compareVersionsItems = _.map(this.compareVersions.items, (board) => board.article);
}
```

ioanpocol: fixed

scripts/apps/authoring/compare-versions/CompareVersionsService.js
```diff
@@ -11,7 +11,7 @@
  * Compare versions screen has its boards and displays at least 2 of them.
```
gbbr: The comment here might be misleading too :-)

karelpetrak:
Some feedback from user testing on http://sdcpr-1643.test.superdesk.org/:
- we should add some caption to the three dot menu, something like "Show differences..."
- it seems to be quite counter-intuitive to select a version to compare and not to see the two versions side by side... I would expect that if I am looking at version 1 and I choose version 5 for comparison, I will see these two side by side and nothing else. So I would limit the diff  view only to two versions on the screen... 
- it should not be possible to select the same version for comparison
- highlighting of the differences (new text) fails to have the green underground colour if there is more than one paragraph of new text, the new text is only underlined
- Errors: sometimes I get something like this in the comparison view: `caption" data-reactid=".7.0.2.0.$4.$1"headline" class="media__caption--title" data-reactid=".7.0.2.0.$4.$0">Te Araroa Trail in New Zealand` ... in this case for a deleted paragraph
- the "Remove item" button overlaps with the hamburger menu and if it is clicked it only hides the version but the pane remains there

ioanpocol:
>>we should add some caption to the three dot menu, something like "Show differences..."
<<The caption was there: "Select previous version"

>>it seems to be quite counter-intuitive to select a version to compare and not to see the two versions side by side... I would expect that if I am looking at version 1 and I choose version 5 for comparison, I will see these two side by side and nothing else. So I would limit the diff view only to two versions on the screen...
<<This was the initial decision on design and it will be refactored after the final design decision will be take.

>>it should not be possible to select the same version for comparison
<<It will be fixed if it will be still the case after refactoring

>>highlighting of the differences (new text) fails to have the green underground colour if there is more than one paragraph of new text, the new text is only underlined
<<Fixed

>>Errors: sometimes I get something like this in the comparison view: caption" data-reactid=".7.0.2.0.$4.$1"headline" class="media__caption--title" data-reactid=".7.0.2.0.$4.$0">Te Araroa Trail in New Zealand ... in this case for a deleted paragraph
<<I think this is because of copy&paste from headline. We have to add a task to investigate this.

>>the "Remove item" button overlaps with the hamburger menu and if it is clicked it only hides the version but the pane remains there UPDATE/EDIT: and I found an option to remove it only after writing this comment ;)
<<this is from initial version of differences, not affected by this PR; it will be fixed after refactoring if it is still the case

petrjasek:
@hlmnrmr @karelpetrak any update?

ioanpocol:
All required changes are done

PR #1651, opened 2017-06-05
(fix) Fixing related item query so that it looks for both published and not published items.

mdhaman:
- `Missing_Link` was displayed even though there was no content with same slugline.

### DIFF
--- a/scripts/apps/archive/services/ArchiveService.js
+++ b/scripts/apps/archive/services/ArchiveService.js
@@ -91,7 +91,16 @@ export function ArchiveService(desks, session, api, $q, search, $location, confi
         query.size(200);
 
         if (itemId) {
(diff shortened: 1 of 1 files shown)


scripts/apps/archive/services/ArchiveService.js
```diff
@@ -91,7 +91,16 @@ export function ArchiveService(desks, session, api, $q, search, $location, confi
         query.size(200);
 
         if (itemId) {
-            query.filter({not: {term: {_id: itemId}}});
+            let filter = {
+                bool: {
+                    must_not: [
```
petrjasek: wouldn't this work with:

```
must_not: [
    {term: {_id: itemId}},
    {term: {item_id: itemId}}
]
```
?

PR #1659, opened 2017-06-07
[SETTINGS] [INGEST] [FTP] added move path error

jerome-poisson:
ftp_move_path_error allow to enter path for file which haven't been
ingested correctly. This field can be empty, in which case the files will
not be moved on error

SDESK-1452

### DIFF
--- a/scripts/apps/ingest/views/settings/ftp-config.html
+++ b/scripts/apps/ingest/views/settings/ftp-config.html
@@ -25,10 +25,14 @@
     <input type="checkbox" ng-model="provider.config.passive" ng-change="$parent.setConfig(provider)">
 </div>
 <div class="field">
-    <label translate>Move articles after ingestion</label>
+    <label translate>Move items after ingestion</label>
     <input type="checkbox" ng-model="provider.config.move" ng-change="$parent.setConfig(provider)">
 </div>
 <div class="field" ng-show="provider.config.move">
-    <label for="ftp_move_path" translate>Move ingested articles to</label>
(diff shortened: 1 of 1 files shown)


scripts/apps/ingest/views/settings/ftp-config.html
```diff
@@ -25,10 +25,14 @@
...
     <input type="checkbox" ng-model="provider.config.move" ng-change="$parent.setConfig(provider)">
 </div>
 <div class="field" ng-show="provider.config.move">
-    <label for="ftp_move_path" translate>Move ingested articles to</label>
+    <label for="ftp_move_path" translate>Move ingested items to</label>
     <input type="text" id="ftp_move_path" placeholder="{{ :: 'FTP Server Path'|translate }}" ng-model="provider.config.move_path" ng-required="provider.config.move" ng-change="$parent.setConfig(provider)">
 </div>
+<div class="field" ng-show="provider.config.move">
+    <label for="ftp_move_path_error" translate>Move *NOT* ingested items (i.e. on error) to</label>
+    <input type="text" id="ftp_move_path_error" placeholder="{{ :: 'FTP Server Path, can be empty if files must not be moved on error'|translate }}" ng-model="provider.config.move_path_error" ng-required="provider.config.move" ng-change="$parent.setConfig(provider)">
```
petrjasek: with this `ng-required` it seems like it can't be empty

jerome-poisson: right, I've added the empty feature afterwards and didn't remove it, good catch thanks!

PR #1671, opened 2017-06-12
feat(editor3): added JIRA-like tooltips to Editor3

gbbr:
Uses `superdesk-ui-framework` tooltips.

### DIFF
--- a/scripts/core/editor3/components/toolbar/StyleButton.jsx
+++ b/scripts/core/editor3/components/toolbar/StyleButton.jsx
@@ -17,6 +17,22 @@ const StyleIcons = {
     orderedlist: 'icon-ordered-list'
 };
 
+const StyleTooltips = {
+    bold: gettext('Bold (Ctrl+B)'),
+    italic: gettext('Italic (Ctrl+I)'),
+    underline: gettext('Underline (Ctrl+U)'),
+    h1: gettext('H1'),
+    h2: gettext('H2'),
+    h3: gettext('H3'),
+    h4: gettext('H4'),
+    h5: gettext('H5'),
+    h6: gettext('H6'),
+    quote: gettext('Quote'),
+    unorderedlist: gettext('Unordered list'),
+    orderedlist: gettext('Ordered list'),
+    TH: gettext('Toggle Header')
+};
+
 /**
  * @ngdoc React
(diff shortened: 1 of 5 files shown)


darconny:
@gbbr Can we use tooltips from UI framework?

fritzSF:
If they work correctly for this we should definitely try to use the CSS only tooltips from the UI framework. @gbbr please try if they will work in this case. here is the page with examples: https://superdesk.github.io/superdesk-ui-framework/examples/

gbbr:
Oh! Right. I didn't know about those. Unfortunately they don't work. React doesn't allow adding custom attributes, unless they are of the form `data-*`. So it would be great if the UI framework would allow both, in this case `sd-tooltip` _and_ `data-sd-tooltip`.

darconny:
We added support for 'data-sd-tooltip' in UI framework, its merged.

gbbr:
@darconny thanks! That's great! Do you think we can also add the `flow` attribute? In this case the tooltips need to show at the bottom, while here the default is at the top.

P.S. Ideally all ui-framework attributes should support the `data-` prefix to allow us to use them with React components, but for now, if you can just add `data-flow` that would be great.

darconny:
@gbbr Now you can use 'data-flow' along with 'data-sd-tooltip'. Let me know if this works.

scripts/core/editor3/components/toolbar/StyleButton.jsx
```diff
@@ -17,6 +17,22 @@ const StyleIcons = {
     orderedlist: 'icon-ordered-list'
 };
 
+const StyleTooltips = {
```
petrjasek: would be good to use `gettext` for values

gbbr: Damn! I keep forgetting this, sorry!

gbbr:
@darconny Done with all changes. Thanks!

P.S. Wouldn't it be better to prefix all UI framework attributes the same? (`sd-*`). Why isn't flow prefixed I wonder? I'm sure it makes sense, I'm just curious...

gbbr:
@darconny @fritzSF since you've asked for changes to this PR do you mind submitting a review and approving it, considering you are happy with the current state?

PR #1672, opened 2017-06-12
ui(editor3/embed): allow adding embed codes

gbbr:
This is a highly insecure implementation which allows pasting random
HTML into the embed input, which alternatively gets injected into the
editor's content.

It's a temporary solution after discussing with @karelpetrak. Assuming
this software is used only internally and that potential employee X with
malicious intentions can find worse ways to harm the company, it should
be ok for now.

The 'correct' implementation would be to sanitize the input and see if
it's a valid embed code for any of the supported providers, which would
prove a lot more work and would be harder to maintain due to the fact
that embed codes might change all the time. Alternatively, we could also
simply santitize the URLs presented in the HTML, making sure they
represent trusted sources.

Change-Id: I63fe77e32b69038a86b0e7c927cd8b9728295df5

### DIFF
--- a/scripts/core/editor3/components/embeds/EmbedInput.jsx
+++ b/scripts/core/editor3/components/embeds/EmbedInput.jsx
@@ -76,6 +76,12 @@ export class EmbedInputComponent extends Component {
      */
     onSubmit() {
         const {value} = this.refs.txt;
+
+        if (!value.startsWith('http://') && !value.startsWith('https://')) {
+            this.props.embedCode(value);
+            return this.onCancel();
+        }
+
         const config = ng.get('config');
         const apiKey = config.iframely.key || fallbackAPIKey;
 
@@ -101,7 +107,7 @@ export class EmbedInputComponent extends Component {
 
         return (
             <form onSubmit={this.onSubmit} className="embed-dialog" onKeyUp={this.onKeyUp}>
-                <input type="url" ref="txt" placeholder="Enter a URL to embed" />
+                <input type="url" ref="txt" placeholder={gettext('Enter URL or code to embed')} />
                 <div className="input-controls">
                     <i className="svg-icon-ok" onClick={this.onSubmit} />
                     <i className="icon-close-small" onClick={onCancel} />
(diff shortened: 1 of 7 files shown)


scripts/core/editor3/components/embeds/EmbedInput.jsx
```diff
@@ -76,6 +76,12 @@ export class EmbedInputComponent extends Component {
      */
     onSubmit() {
         const {value} = this.refs.txt;
+
+        if (value.substring(0, 4) !== 'http') {
```
petrjasek: what about `value.startsWith('http')` ?

scripts/core/editor3/components/embeds/EmbedInput.jsx
```diff
@@ -101,7 +107,7 @@ export class EmbedInputComponent extends Component {
 
         return (
             <form onSubmit={this.onSubmit} className="embed-dialog" onKeyUp={this.onKeyUp}>
-                <input type="url" ref="txt" placeholder="Enter a URL to embed" />
+                <input type="url" ref="txt" placeholder="Enter URL or code to embed" />
```
petrjasek: this placeholder should use `gettext`

gbbr:
@petrjasek I've added the changes you've requested (I actually didn't know about [startsWith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith)!).

Not sure I understand your question:
> not sure if we should run such code in the authoring, it has potential to break the ui

Do you mean in the preview? I guess there is always that chance, unless we make sure there is appropriate CSS. All embed blocks are wrapped in a `div.embed-block`. Does that help?

Sorry if I misunderstood your question, in which case maybe you can rephrase it or be more specific.

petrjasek:
yep you got it right. if there is some script code or styles it can in general break the page. maybe we can try using shadow dom for that if available.

gbbr:
Not sure I understand how that would help. It would still be a container just like we currently have the `div`. I did set a width limitation on iframe's, and by default it should be 100%. Since we are allowing technically _any_ content the possibility of breaking the layout is quite high. But then again if people stick to the classical embed codes, I think we should be good.

Worst case scenario we can handle the issues as we go. Proper QA should also help identify edge-cases.

petrjasek:
yep let qa figure this out

PR #1677, opened 2017-06-14
fix(featuremedia): Fix drag and drop to feature media an archived con…

ioanpocol:
…tent item (search -> archived)

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ItemAssociationDirective.js
+++ b/scripts/apps/authoring/authoring/directives/ItemAssociationDirective.js
@@ -48,8 +48,7 @@ export function ItemAssociationDirective(superdesk, renditions, config, authorin
                 let item = angular.fromJson(event.originalEvent.dataTransfer.getData(dataType));
 
                 if (item._type !== 'externalsource') {
-                    return api.find('archive', item._id)
-                        .then((result) => result);
+                    return api.find(item._type, item._id);
                 }
                 return $q.when(item);
             }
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/directives/ItemAssociationDirective.js
```diff
@@ -47,7 +47,10 @@ export function ItemAssociationDirective(superdesk, renditions, config, authorin
             function getItem(event, dataType) {
                 let item = angular.fromJson(event.originalEvent.dataTransfer.getData(dataType));
 
-                if (item._type !== 'externalsource') {
+                if (item._type === 'archived') {
+                    return api.find('archived', item._id)
```
petrjasek: seems like this `.then()` is useless there.

scripts/apps/authoring/authoring/directives/ItemAssociationDirective.js
```diff
@@ -47,7 +47,10 @@ export function ItemAssociationDirective(superdesk, renditions, config, authorin
...
+                        .then((result) => result);
+                } else if (item._type !== 'externalsource') {
                     return api.find('archive', item._id)
```
petrjasek: maybe it should just do `if (item._type !== 'externalsource') { return api.find(item._type, item._id); }`

jerome-poisson:
@ioanpocol is this patch fixing the issue for context panel too? cf. https://dev.sourcefabric.org/secure/attachment/32138/archived_picture_bug.webm at the end

ioanpocol:
Now yes

PR #1683, opened 2017-06-19
Removal of takes from core components

akintolga:


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/957

PR #1687, opened 2017-06-21
[SDESK-1501](fix): Fixing item history for scheduled items.

mdhaman:
depends on https://github.com/superdesk/superdesk-core/pull/964


scripts/apps/authoring/versioning/history/views/history.html
```diff
@@ -14,7 +14,9 @@
             </div>
 
             <div ng-if="historyItem.operation === 'update'">
-                <strong>{{ :: 'Updated by' | translate }}</strong> {{ :: historyItem.displayName }}
+                <strong ng-if="historyItem.update.operation === 'deschedule'">{{ :: 'Deschedule by' | translate }}</strong>
```
akintolga: `Descheduled by` would be better here.

PR #1692, opened 2017-06-23
feat(compare versions): Highlight the differences between versions

ioanpocol:


scripts/apps/authoring/compare-versions/directives/CompareVersionsArticleDirective.js
```diff
@@ -59,6 +72,109 @@ class LinkFunction {
...
+
+    /**
+     * @ngdoc method
+     * @name sdCompareVersionsArticle#setVersionsDifference
+     * @param {Object} item - current item version
+     * @param {Object} oldItem - old item to compare with
+     * @description Changes 'headline', 'abstract', 'body_footer', 'body_html', 'byline' from item
+     * in order to highlight the differences from oldItem
+     */
+    setVersionsDifference(item, oldItem) {
```
petrjasek: imo would be better to return a diff and not modify the item, otherwise you have to keep a copy somewhere

scripts/apps/authoring/compare-versions/directives/CompareVersionsArticleDirective.js
```diff
@@ -59,6 +72,109 @@ class LinkFunction {
...
+
+        let text = this.diffMatchPatch
+            .diff_prettyHtml(diffs)
+            .replace(/&lt;/g, '<')
+            .replace(/&gt;/g, '>')
+            .replace(/<span><p><\/span>/g, '<p>')
+            .replace(/<span><\/p><\/span>/g, '</p>')
+            .replace(/&nbsp;/g, ' ');
+
+        if (this.config.features.editor3) {
```
petrjasek: why is the diff editor specific?

scripts/core/editor3/directive.js
```diff
@@ -111,6 +115,14 @@ class Editor3Directive {
 
         const store = createStore(this);
 
+        $scope.$watch('vm.value', (newValue, oldValue) => {
```
petrjasek: why is this needed?

scripts/apps/authoring/compare-versions/directives/CompareVersionsArticleDirective.js
ioanpocol: Is it already a copy

scripts/apps/authoring/compare-versions/directives/CompareVersionsArticleDirective.js
ioanpocol: Editor3 don't support 'ins' tag and we used 'code' instead.

scripts/core/editor3/directive.js
ioanpocol: To detect if the value sent to editor3 wrapper directive has been changed and then to update the editor3 state.

scripts/apps/authoring/compare-versions/directives/CompareVersionsArticleDirective.js
ioanpocol: We will move this code on Editor3. @gbbr

scripts/apps/authoring/compare-versions/directives/CompareVersionsArticleDirective.js
petrjasek: that's what I mean

scripts/core/editor3/directive.js
petrjasek: so it's using editor to render the diff?

ioanpocol: yes

petrjasek: why?

ioanpocol: It was more simpler

petrjasek: imo that's confusing, it shouldn't be editable right?
wouldn't it work with just `ng-bind-html` ? why is it simpler with editor?

ioanpocol: When I started to work on it it was already using the editor, I just added the highlights. The editor is readonly.

petrjasek: I see

PR #1695, opened 2017-06-26
[SDESK-1512] - Refactoring archive_link endpoint to use archive_rewrite

akintolga:


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/969

PR #1699, opened 2017-06-27
feat(cvs): add option to create new vocabulary

petrjasek:
it has predefined schema with name, qcode and parent for now.
stores data into subject field.

SDESK-1483

requires https://github.com/superdesk/superdesk-core/pull/970

### DIFF
--- a/scripts/apps/authoring/metadata/metadata.js
+++ b/scripts/apps/authoring/metadata/metadata.js
@@ -654,9 +654,9 @@ function MetaTermsDirective(metadata, $filter, $timeout) {
 
                 scope.terms = filterSelected(scope.list);
                 if (scope.cv) { // filter out items from current cv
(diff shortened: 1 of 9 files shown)


scripts/apps/archive/directives/MarkedItemTitle.js
```diff
@@ -57,7 +57,7 @@ class LinkFunction {
         this.scope.$watch('item', (item) => {
             if (item) {
                 this.scope.marks = [];
-                if (item[this.scope.markField].length) {
+                if (!_.isEmpty(item[this.scope.markField])) {
```
akintolga: @pavlovicnemanja has also made a fix for this line: 
https://github.com/superdesk/superdesk-client-core/pull/1700/files#diff-7ffbaf915ec369a4c5eb37a2ad9cca95R60

petrjasek: reverted it

PR #1704, opened 2017-06-28
feat(cvs): add option to create new vocabulary

petrjasek:
it has predefined schema with name and qcode for now.
will store data into subject field.

SDESK-1483

recreating https://github.com/superdesk/superdesk-client-core/pull/1699 which got stuck on our ci


petrjasek:
this was reviewed already https://github.com/superdesk/superdesk-client-core/pull/1699

PR #1715, opened 2017-07-04
SDFID-26 Add custom field "Disclaimer" to content profiles

mugurrus:
move usageterms from item header to body

### DIFF
--- a/scripts/apps/authoring/views/article-edit.html
+++ b/scripts/apps/authoring/views/article-edit.html
@@ -371,3 +371,19 @@
          data-display-user="{{ getSignOffMapping() }}"
     ></div>
 </div>
+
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/views/article-edit.html
```diff
@@ -371,3 +371,19 @@
          data-display-user="{{ getSignOffMapping() }}"
     ></div>
 </div>
+
+<div class="field" ng-if="schema.usageterms" sd-width="{{editor.usageterms.sdWidth || 'full'}}" order="{{editor.usageterms.order}}" sd-validation-error="error.usageterms" data-required="schema.usageterms.required">
+  <label translate>Usage Terms</label>
+  <input type="text"
```
petrjasek: I'm not sure if this should be input, imo in general there will be more text, so textarea would be better.

PR #1718, opened 2017-07-04
fix(tabindex): SDESK-1553 Fixed issue with tabindex on authoring

pavlovicnemanja:


pavlovicnemanja:
Waiting for https://dev.sourcefabric.org/browse/SDESK-1599 to be fixed, so tests can pass

PR #1719, opened 2017-07-04
fix(templates): Missing items in content templates in desk settings

pavlovicnemanja:
If there is some other way to return all data from database, please let me know. By default it is returning only 25 items.


mdhaman:
The default is 25 records and max per request is 200 records. `PAGINATION_LIMIT` of 200 is set on the API. If you want more than 200 records then the recursive call is required.

pavlovicnemanja:
Ok, I guess it will ok for 200 records for now. I guess once we will have more than 200 templates we will need to change UI for this dropdown 😄

PR #1724, opened 2017-07-07
fix(editor3): make image captions text only [SDESK-1458]

gbbr:
Adds a dependency on [react-textarea-autosize](https://github.com/andreypopp/react-textarea-autosize) which is similar to what we have as [`sd-autoheight`](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/ui/autoheight-directive.js).

Additionally fixes some errors in unit tests about missing props on the EmbedInput component.

### DIFF
--- a/scripts/core/editor3/components/images/ImageBlock.jsx
+++ b/scripts/core/editor3/components/images/ImageBlock.jsx
@@ -1,6 +1,7 @@
 import React, {Component} from 'react';
 import {connect} from 'react-redux';
 import * as actions from '../../actions';
+import Textarea from 'react-textarea-autosize';
 
 /**
  * @ngdoc React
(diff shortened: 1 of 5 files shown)


scripts/core/editor3/components/images/ImageBlock.jsx
```diff
@@ -70,13 +71,13 @@ export class ImageBlockComponent extends Component {
             <div className="image-block" onClick={(e) => e.stopPropagation()}>
                 <div className="image-block__wrapper">
                     <img src={rendition.href} alt={alt} onClick={this.onClick} />
-                    <div contentEditable={true} placeholder="Description"
-                        ref={(el) => {
-                            this.captionInput = el;
-                        }}
+                    <Textarea
+                        placeholder="Description"
```
petrjasek: it should use gettext here

ioanpocol:
On show version difference page we have to add some html tags to caption text in order to highlight the differences. Is this working with the new changes?

scripts/core/editor3/components/images/ImageBlock.jsx
gbbr: Done.

gbbr:
@ioanpocol I'm afraid not like you would expect. I think we will have to find another way to do that for the images. Let's discuss it.

ioanpocol:
Probably you can show a div when the editor is readonly

gbbr:
@petrjasek PTAL

ioanpocol:
@gbbr This PR break the version comparing.

PR #1725, opened 2017-07-07
feat(editor3): Disable image drag and drop if the editor is read only…

ioanpocol:
…, single line or if the image is not supported

### DIFF
--- a/scripts/core/editor3/components/Editor3.jsx
+++ b/scripts/core/editor3/components/Editor3.jsx
@@ -94,18 +94,21 @@ export class Editor3Component extends React.Component {
     /**
      * @ngdoc method
      * @name Editor3#onDragOver
-     * @returns {Boolean} Returns true if the item is permitted.
+     * @returns {Boolean} Returns true if the item is not permitted.
      * @description Checks if the dragged over item is allowed.
      */
     onDragOver(e) {
+        const {editorFormat, readOnly, singleLine} = this.props;
         const mediaType = e.originalEvent.dataTransfer.types[0] || '';
-
-        return [
+        const isValidMedia = [
             'application/superdesk.item.picture',
             'application/superdesk.item.graphic',
             'application/superdesk.item.video',
             'text/html',
-        ].indexOf(mediaType) === -1;
+        ].indexOf(mediaType) !== -1;
+        const supportsImages = !readOnly && !singleLine && editorFormat.indexOf('picture') !== -1;
+
+        return !(supportsImages && isValidMedia);
     }
 
     keyBindingFn(e) {
@@ -248,14 +251,22 @@ Editor3Component.propTypes = {
     onTab: React.PropTypes.func,
     dragDrop: React.PropTypes.func,
     scrollContainer: React.PropTypes.string,
-    singleLine: React.PropTypes.bool
+    singleLine: React.PropTypes.bool,
+    editorFormat: React.PropTypes.array
+};
+
+Editor3Component.defaultProps = {
+    readOnly: false,
+    singleLine: false,
+    editorFormat: []
 };
 
 const mapStateToProps = (state) => ({
     readOnly: state.readOnly,
     showToolbar: state.showToolbar,
     editorState: state.editorState,
-    locked: state.locked
+    locked: state.locked,
+    editorFormat: state.editorFormat
 });
 
 const mapDispatchToProps = (dispatch) => ({
--- a/scripts/core/editor3/components/tests/editor3.spec.jsx
+++ b/scripts/core/editor3/components/tests/editor3.spec.jsx
@@ -12,7 +12,7 @@ describe('editor3.component', () => {
     });
 
     it('should not accept dragging over invalid items', () => {
-        const wrapper = shallow(<Editor3 />);
+        const wrapper = shallow(<Editor3 editorFormat={['picture']}/>);
         const {onDragOver} = wrapper.instance();
         const makeEvent = (t) => ({originalEvent: {dataTransfer: {types: [t]}}});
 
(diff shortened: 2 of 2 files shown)


scripts/core/editor3/components/Editor3.jsx
```diff
@@ -94,12 +94,17 @@ export class Editor3Component extends React.Component {
...
      * @name Editor3#onDragOver
-     * @returns {Boolean} Returns true if the item is permitted.
+     * @returns {Boolean} Returns true if the item is not permitted.
      * @description Checks if the dragged over item is allowed.
      */
     onDragOver(e) {
+        const {editorFormat, readOnly, singleLine} = this.props;
         const mediaType = e.originalEvent.dataTransfer.types[0] || '';
 
+        if (readOnly || singleLine || !(editorFormat && editorFormat.indexOf('picture') !== -1)) {
```
gbbr: This should suffice:
```js
    onDragOver(e) {
        const {editorFormat, readOnly, singleLine} = this.props;
        const mediaType = e.originalEvent.dataTransfer.types[0] || '';

        return !readOnly && editorFormat.indexOf('picture') > -1 && [
            'application/superdesk.item.picture',
            'application/superdesk.item.graphic',
            'application/superdesk.item.video',
            'text/html',
        ].indexOf(mediaType) === -1;
    }
```
Reason being that `singleLine` editors don't have formatting options anyway, so it's not needed to check that. It's also not needed to validate that `editorFormat` is non-null if you add this at the bottom:
```js
Editor3Component.defaultProps = {
    editorFormat: []
};
```
I would also write a test for this, where you call `onDragOver` directly and assert whether it returns `true` or `false` based on different props.

scripts/core/editor3/components/tests/editor3.spec.jsx
```diff
@@ -33,6 +33,51 @@ describe('editor3.component', () => {
...
+        const {onDragOver} = wrapper.instance();
+        const makeEvent = (t) => ({originalEvent: {dataTransfer: {types: [t]}}});
+
+        [
+            'application/superdesk.item.picture',
+            'application/superdesk.item.graphic',
+            'application/superdesk.item.video',
+            'text/html'
+        ].forEach((validType) => {
+            expect(onDragOver(makeEvent(validType))).toBe(true);
```
gbbr: This is not a valid test because you don't know which return branch resulted in the value `true` and whether the drag was processed or not (both your branches return true in `onDragOver`).

scripts/core/editor3/components/tests/editor3.spec.jsx
```diff
@@ -33,6 +33,51 @@ describe('editor3.component', () => {
...
```
gbbr: Same goes for this test.

scripts/core/editor3/components/tests/editor3.spec.jsx
```diff
@@ -33,6 +33,51 @@ describe('editor3.component', () => {
...
+        });
+    });
+
+    it('should not accept dragging when editor does not support images', () => {
+        const wrapper = shallow(<Editor3 readOnly={false} editorFormat={[]} singleLine={false}/>);
```
gbbr: You can remove all props here because you will get the same values if you don't set them. If not, add them to defaultProps.

scripts/core/editor3/components/tests/editor3.spec.jsx
```diff
@@ -33,6 +33,51 @@ describe('editor3.component', () => {
...
+        const {onDragOver} = wrapper.instance();
+        const makeEvent = (t) => ({originalEvent: {dataTransfer: {types: [t]}}});
+
+        [
+            'application/superdesk.item.picture',
+            'application/superdesk.item.graphic',
+            'application/superdesk.item.video',
+            'text/html'
+        ].forEach((validType) => {
+            expect(onDragOver(makeEvent(validType))).toBe(true);
```
gbbr: Same for this test. You can also use `.toBeTruthy()` and `.toBeFalsy()` instead of `.toBe(true)` and `.toBe(false)` if you want.

scripts/core/editor3/components/tests/editor3.spec.jsx
```diff
@@ -33,6 +33,51 @@ describe('editor3.component', () => {
...
+        });
+    });
+
+    it('should not accept dragging when editor is single line', () => {
+        const wrapper = shallow(<Editor3 readOnly={false} editorFormat={['picture']} singleLine={true}/>);
```
gbbr: Just do:
```js
const wrapper = shallow(<Editor3 editorFormat={['picture']} singleLine />);
```
Note that you don't need to write `={true}` for boolean props.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: Editor3 component don't have any default value yet, but yes it is a good idea to add them now. 
Sometimes very long conditions are difficult to read and understood, for example your proposed solution is wrong, it should be || instead &&  (return !readOnly && editorFormat.indexOf('picture') > -1 || ....). Probably a third solution will be better.
Related to singleLine I think it is more stable the fact that a single line don't contain an image than the fact that a singleLine don't support any formatting. Also the singleLine condition is more clear.
I think the current tests covers all cases for onDragOver.

scripts/core/editor3/components/tests/editor3.spec.jsx
ioanpocol: Yes, I will fix it.

scripts/core/editor3/components/tests/editor3.spec.jsx
ioanpocol: There are not defaultProps at all, but yes it is a good idea to add them.

scripts/core/editor3/components/tests/editor3.spec.jsx
ioanpocol: OK, good to know.

scripts/core/editor3/components/tests/editor3.spec.jsx
ioanpocol: There are not defaultProps at all, but yes it is a good idea to add them.

scripts/core/editor3/components/tests/editor3.spec.jsx
ioanpocol: Probably we can add this as a rule for style checker. In current file is used only .toBe(true) form and over the Editor3 component is used 30 times .toBe(true) form  and 5 times .toBeThruly() form.

scripts/core/editor3/components/tests/editor3.spec.jsx
ioanpocol: Actually the test is good, it can be read as: even if the files are good and even if the picture option is present in menu, if the editor is read only it will not accept drag and drop.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -94,18 +94,21 @@ export class Editor3Component extends React.Component {
...
-     * @returns {Boolean} Returns true if the item is permitted.
+     * @returns {Boolean} Returns true if the item is not permitted.
      * @description Checks if the dragged over item is allowed.
      */
     onDragOver(e) {
+        const {editorFormat, readOnly, singleLine} = this.props;
         const mediaType = e.originalEvent.dataTransfer.types[0] || '';
-
-        return [
+        const goodFile = [
```
gbbr: I would name this `isValidMedia` because it checks the validity of the media type, not a file.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -94,18 +94,21 @@ export class Editor3Component extends React.Component {
...
             'application/superdesk.item.picture',
             'application/superdesk.item.graphic',
             'application/superdesk.item.video',
             'text/html',
-        ].indexOf(mediaType) === -1;
+        ].indexOf(mediaType) !== -1;
+        const supportDragAndDrop = !readOnly && !singleLine && editorFormat.indexOf('picture') !== -1 && goodFile;
```
gbbr: Here I would do:
```js
const supportsImages = !readOnly && !singleLine && editorFormat.indexOf('picture') !== -1;

return supportsImages && isValidMedia;
```
In my opinion in reads nicer.

PR #1728, opened 2017-07-11
Add app "extension-points"

markusherzog:
This adds the ability for adding more elements the core UI from within external superdesk apps.
Usage:

1.  Place this tag in a view where you'd like to add an extension:
`<span sd-extension-point="MY_TYPE"></span>`

2. Your extension then needs to register one or more components that get placed at the tag from the previous step.
Inject 'extensionPointsProvider' into your module and then (in the config phase):
 `extensionPointsProvider.register('MY_TYPE', MyComponent, an_array);`
 where MyComponent is a React component and an_array is an array with names of variables that your component will receive as props from the parent scope, for example: ['item']


scripts/apps/authoring/views/authoring.html
```diff
@@ -10,6 +10,7 @@
                         <label translate>Modified</label>
                         <time sd-datetime data-date="item.versioncreated"></time>
                     </div>
+                    <span sd-extension-point sd-extension-point-type="authoring-sticky-detailed"></span>
```
petrjasek: maybe this could be simplified `<span sd-extension-point="authoring-sticky-detailed"></span>`

scripts/apps/extension-points/index.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ *    where MyComponent is a React component and an_array is an array with names
+ *    of variables that your component will receive as props from the parent
+ *    scope, for example: ['item']
+ *
+ */
+
+angular.module('superdesk.apps.extension-points', [])
+    .directive('sdExtensionPoint', directive.ExtensionPointDirective)
+    .service('extensionPoints', svc.ExtensionPointService);
```
petrjasek: I would prefer provider here, so that you can register these in `.config` phase

scripts/apps/extension-points/index.js
```diff
@@ -0,0 +1,25 @@
+import * as directive from './directives';
+import * as svc from './services';
+
+/*
```
gbbr: We have a documentation system (not sure how well it's working at the moment), but the syntax for it is as follows:
```js
/**
 * @ngdoc module
 * @module superdesk.apps.extension-points
 * @name superdesk.apps.extension-points
 * @packageName superdesk.apps
 * @description DESCRIPTION GOES HERE
 */
```
Please try and use the same pattern throughout, when documenting both directives and services. You can get inspired from other modules in the codebase.

PR #1729, opened 2017-07-11
fix(editor3): Fix drag&drop allow check for Firefox

ioanpocol:


scripts/core/editor3/components/Editor3.jsx
```diff
@@ -108,7 +110,28 @@ export class Editor3Component extends React.Component {
...
 
-        return !(supportsImages && isValidMedia);
+        return supportsImages && isValidMedia;
+    }
+
+    /**
+     * @ngdoc method
+     * @name Editor3#onDragOver
+     * @returns {Boolean} Returns true if the item is not permitted.
+     * @description Checks if the dragged over item is allowed.
```
gbbr: The `@description` documentation for this method is the same as for `allowItem` even though one returns the opposite value of the other. Can you please try and amend the documentation so that it correctly explains what is going on? Perhaps if it's worth mentioning that this is a behaviour that varies cross-browser?

PR #1733, opened 2017-07-13
Expands Notifications for desktop notifications SDESK-55

vied12:
It sends a desktop notification when a new activity is received

It broadcast a 'notification:click' when an notif has been clicked in
the user's activities list.

depends on

https://github.com/superdesk/superdesk-core/pull/988
https://github.com/superdesk/superdesk-core/pull/994


vied12:
I could ask for permission at the application start-up, but I don't know where would be the best place for that.

mdhaman:
We have a busy system and it may generate a lot of notifications. @vied12 it would be good to have some system config which administrators can control to turn off notification globally or based it on user preference so that we can do a trial for some users.

petrjasek:
imo it should ask for permissions when user enables this on his profile

vied12:
I added a toggle control in the user profile in order to enable/disable the desktop notification globally

PR #1734, opened 2017-07-15
SDESK-1190 As a user I want to see the scheduled date/time for scheduled content

mugurrus:


scripts/apps/search/components/fields/state.jsx
```diff
@@ -1,12 +1,19 @@
...
 
 export function state(props) {
-    const {$filter, gettextCatalog} = props.svc;
+    const {$filter, gettextCatalog, datetime} = props.svc;
 
     if (props.item.state !== undefined && props.item.state !== null) {
+        var title = $filter('removeLodash')(props.item.state);
+
+        if (props.item.state === 'scheduled') {
+            title = $filter('removeLodash')(gettextCatalog.getString('Scheduled on ') +
```
petrjasek: here the filter is not needed, it's there for states like `in_progress`

PR #1739, opened 2017-07-18
feat(actions): SDESK-1503 Removed labels and grouped duplicate items

pavlovicnemanja:


scripts/apps/archive/index.js
```diff
@@ -135,8 +135,8 @@ angular.module('superdesk.apps.archive', [
                     return authoring.itemActions(item).unspike;
                 }]
             })
-            .activity('duplicate', {
-                label: gettext('Duplicate'),
+            .activity('duplicateHere', {
+                label: gettext('Duplicate Here'),
```
mdhaman: `Duplicate Here` makes sense if the user is in the monitoring view of the desk. It would be confusing when you are in the custom workspace or global search. The duplicate item could land on your default desk as well.

fritzSF: Would _Duplicate in place_ make sense?

mdhaman: let merge it and get some feedback.

PR #1742, opened 2017-07-20
fix(build): create index.html in dist folder

petrjasek:
should fix build after https://github.com/superdesk/superdesk-client-core/pull/1731/commits/6233c8ea8833c9d3db7d04703a312b6cdfc8e954
@gbbr pls check

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -64,7 +64,6 @@ module.exports = function(grunt) {
     grunt.registerTask('server', [
         'clean',
         'ngtemplates:index',
-        'copy:index',
         'copy:config',
         'copy:locales',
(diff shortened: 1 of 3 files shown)


gbbr:
I think `copy:index` was needed when `grunt server` was running in main `superdesk` repo. But I might be wrong.

Please check that `grunt build` and `grunt server` continues to work correctly in both repos. If so, then all is well. When checking, make sure you don't have any left over files that were generated from previous builds but are not in the actual Git repo, such as .gitignore'd files (ie. index.html)

petrjasek:
it seems like it didn't work in main superdesk repo before, at least not grunt build

petrjasek:
breaks grunt server

PR #1752, opened 2017-07-27
attachments ui

petrjasek:
using https://github.com/superdesk/superdesk-client-core/pull/1745 and connecting it to backend


petrjasek:
requires https://github.com/superdesk/superdesk-core/pull/998

PR #1753, opened 2017-07-27
fix(upload): use inputs instead of contenteditable

petrjasek:
to avoid html being pasted there

SDESK-1458

### DIFF
--- a/scripts/apps/archive/views/upload.html
+++ b/scripts/apps/archive/views/upload.html
@@ -39,28 +39,28 @@
 
                     <div ng-if="validator.headline" class="row">
                         <label ng-class="{'label-asterisk':validator.headline.required}" translate>Title</label>
-                        <div ng-model="item.meta.headline" contenteditable="true" class="contenteditable-input contenteditable-input--side-padding"></div>
+                        <input type="text" ng-model="item.meta.headline" />
                     </div>
                     <div class="other-info">
                         <div class="row" ng-if="validator.description_text">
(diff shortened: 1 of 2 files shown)


scripts/apps/archive/views/upload.html
```diff
@@ -39,28 +39,28 @@ <h3 class="modal__heading">{{ activity.label|translate }}</h3>
...
                     <div ng-if="validator.headline" class="row">
                         <label ng-class="{'label-asterisk':validator.headline.required}" translate>Title</label>
-                        <div ng-model="item.meta.headline" contenteditable="true" class="contenteditable-input contenteditable-input--side-padding"></div>
+                        <input type="text" ng-model="item.meta.headline" />
                     </div>
                     <div class="other-info">
                         <div class="row" ng-if="validator.description_text">
                             <label ng-class="{'label-asterisk':validator.description_text.required}" translate>Caption</label>
-                            <div ng-model="item.meta.description_text" contenteditable="true" class="contenteditable-input contenteditable-input--side-padding"></div>
+                            <input type="text" ng-model="item.meta.description_text" />
```
mdhaman: Can we have `textarea` with auto-height instead of `input` for all fields except byline?

petrjasek: even for title/headline?

mdhaman: oops deleted your comment. headline and title can be input text.

PR #1757, opened 2017-07-31
feat(assignments): Add assignments list menu

ioanpocol:


petrjasek:
I think we should add an API for this eventually

PR #1760, opened 2017-07-31
chore: update webpack

petrjasek:
and other dependencies to make it work


gbbr:
@petrjasek what remains to be done here? Can I help? I was wondering if we could jump straight to Webpack 2 and if it's worth the trouble

petrjasek:
I have to check how it works with planning, so far couldn't do it

PR #1768, opened 2017-08-02
[SDESK-1720](Fix):Hide 'Edit in new Window' and 'Open in new Window' from item action menu

mdhaman:


petrjasek:
would rather go with some sort of whitelist config instead so we don't have to add new config for each action

mdhaman:
@petrjasek Good point. With whitelist you have to opt-in and with blacklist you have to opt-out.
I have modified the code using the activity and so that we can opt-out. Please let me know your thoughts on it.

scripts/core/activity/activity.js
```diff
@@ -172,6 +173,18 @@ function SuperdeskProvider($routeProvider, _) {
                 return privileges.userHasPrivileges(activity.privileges);
             }
 
+            function checkActivityEnabled(activity) {
+                if (!_.get(config, 'activity')) {
+                    return true;
+                }
+
+                if (_.isUndefined(config.activity[activity._id])) {
```
petrjasek: I think I would rather return `false` here because that will happen with new fields..
so either you have them all or you hand pick and then imo there should be no surprises..

petrjasek: maybe it's better the way you have it, it would be too much config once you put there anything

PR #1773, opened 2017-08-03
fix(metadata): add 'Add new Text field' button

petrjasek:
it didn't survive ui refactoring

SDESK-1636


petrjasek:
@fritzSF is it ok to have 2 buttons there? or should it be some dropdown?

fritzSF:
@petrjasek , 2 buttons are OK. I would go with a dropdown for 3 and more

PR #1782, opened 2017-08-09
feat(attachments): add download button

petrjasek:
SDESK-1723


darconny:
We should add 'three dots' button with dropdown menu, instead of having three actions in a row. Not sure if this would be a separate task, though.

PR #1785, opened 2017-08-11
fix(crops) - Allow initial crops available to save when it's new imag…

sjunaid:
…e plus increased spacing b/w metadata fields to prevent labels sticking with char-counts above

- [SDESK-1782]

### DIFF
--- a/scripts/core/upload/image-crop-directive.js
+++ b/scripts/core/upload/image-crop-directive.js
@@ -276,11 +276,15 @@ export default angular.module('superdesk.core.upload.imagecrop', [
 
                             // Store the Jcrop API in the jcropApi variable
                             jcropApi = self;
-
                             // define onSelect once Jcrop initialized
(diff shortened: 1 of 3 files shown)


scripts/apps/authoring/authoring/controllers/ChangeImageController.js
```diff
@@ -32,6 +33,19 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
         }
     });
 
+    $scope.hasCrops = $scope.data.renditions.some(
+        (crop) => $scope.data.cropData && !_.isEmpty($scope.data.cropData[crop.name])
+    );
+
+    _.defer(() => {
```
petrjasek: wondering why would you need both `defer` and `$timeout` ?

petrjasek:
cherry-picked to master

PR #1791, opened 2017-08-16
Extension points

markusherzog:
Add props parameter so we can use a redux store


scripts/apps/extension-points/services/ExtensionPoints.js
```diff
@@ -0,0 +1,32 @@
...
+    var extensions = {};
+
+    this.register = function(type, componentClass, props, data) {
+        if (typeof extensions[type] === 'undefined') {
+            extensions[type] = [];
+        }
+        extensions[type].push({type, componentClass, props, data});
+    };
+
+    this.$get = function() {
```
gbbr:
```js
this.$get = () => extensions;
```
should do.

scripts/apps/extension-points/directives/ExtensionPointDirective.js
```diff
@@ -0,0 +1,37 @@
...
+ * Place this tag in a view where you'd like to add an extension:
+ *   <span sd-extension-point="MY_TYPE"></span>
+ * See also ExtensionPointsProvider.
+ */
+ExtensionPointDirective.$inject = ['extensionPoints'];
+export function ExtensionPointDirective(extensionPoints) {
+    function _buildCompoment(extension, scope) {
+        // for easy access put values from parent scope into the redux store of the extension
+        _.forEach(extension.data, (value) => {
+            extension.props.store.getState()[value] = scope.$parent.$eval(value);
```
gbbr: This looks like a neat API.

gbbr:
Also, is there any way to test this? More specifically that the store props get evaluated on the parent $scope correctly.

markusherzog:
@gbbr Thanks for reviewing. extension-points has been merged, but now there's a couple of small changes to it (the last 2 commits). I was confused by the way github showed *all* the commits from my branch in this PR again, there probably went sth wrong somewhere :-(
As for testing this, I'm going to provide a little example superdesk-addon that shows how to use it

markusherzog:
@gbbr I think this looks better now, had to merge upstream/master into out fork and branch again and it looks like there are no conflicts anymore

scripts/apps/extension-points/services/ExtensionPoints.js
```diff
@@ -8,25 +8,21 @@
  * the core UI.
  * Inject 'extensionPointsProvider' into your module and then:
  *
- * extensionPointsProvider.register('MY_TYPE', MyComponent, an_array); // do this in config phase
+ * extensionPointsProvider.register('MY_TYPE', MyConnectedComponent, props, an_array); // do this in config phase
```
gbbr: Here it says provider in the docs, yet you've now changed it into a service.

scripts/apps/extension-points/directives/ExtensionPointDirective.js
```diff
@@ -15,26 +15,18 @@ import React from 'react';
...
 export function ExtensionPointDirective(extensionPoints) {
-    function propsFromData(data, scope) {
-        var json = {};
-
-        _.forEach(data, (value) => {
-            json[value] = scope.$parent.$eval(value);
+    function _buildCompoment(extension, scope) {
+        // for easy access put values from parent scope into the redux store of the extension
+        _.forEach(extension.data, (value) => {
+            extension.props.store.getState()[value] = scope.$parent.$eval(value);
```
gbbr: This implementation makes it necessary for all React components to implement the Redux architecture. This is a bit limiting in my (h) opinion. I suggest the following:
- If a `store` prop is there, use the logic that is here
- If a `store` prop is not there, use the previous logic where the values get passed directly as props (as before)

scripts/apps/extension-points/directives/ExtensionPointDirective.js
```diff
@@ -15,26 +15,18 @@ import React from 'react';
...
-            propsFromData(extension.data, scope)
-        );
+        return React.createElement(extension.componentClass, extension.props);
     }
 
     return {
         link: function(scope, elem, attr) {
-            var registeredExtenstions = extensionPoints[attr.sdExtensionPoint];
-            var components = _.map(registeredExtenstions, (extension) => buildCompoment(extension, scope));
+            var registeredExtenstions = extensionPoints.extensions[attr.sdExtensionPoint];
```
gbbr: - I think you mean `registeredExtension` here (the perils of auto-complete 😄 ). Additionally, it seems a bit unorthodox to me to access this property directly on the service. Since we already have a `register` method, why not also a getter (ie. `get`) ?

- Let's not assume this value exists in the object and check for it (and perhaps `throw` if not). This could be checked in the method suggested in my previous point.

scripts/apps/extension-points/services/ExtensionPoints.js
```diff
@@ -8,25 +8,21 @@
...
  */
-export function ExtensionPointsProvider() {
-    var extensions = {};
+export function ExtensionPointsService() {
+    this.extensions = {};
 
-    this.register = function(type, componentClass, data) {
-        if (typeof extensions[type] === 'undefined') {
-            extensions[type] = [];
+    this.register = function(type, componentClass, props, data) {
```
gbbr: Please rename `data` to `an_array`, like in the documentation.

scripts/apps/extension-points/directives/ExtensionPointDirective.js
```diff
@@ -15,26 +15,18 @@ import React from 'react';
...
 ExtensionPointDirective.$inject = ['extensionPoints'];
 export function ExtensionPointDirective(extensionPoints) {
-    function propsFromData(data, scope) {
-        var json = {};
-
-        _.forEach(data, (value) => {
-            json[value] = scope.$parent.$eval(value);
+    function _buildCompoment(extension, scope) {
+        // for easy access put values from parent scope into the redux store of the extension
+        _.forEach(extension.data, (value) => {
```
gbbr: You can use plain JS Array.forEach here, no need for underscore.

scripts/apps/extension-points/directives/ExtensionPointDirective.js
```diff
@@ -15,26 +15,18 @@ import React from 'react';
...
-        );
+        return React.createElement(extension.componentClass, extension.props);
     }
 
     return {
         link: function(scope, elem, attr) {
-            var registeredExtenstions = extensionPoints[attr.sdExtensionPoint];
-            var components = _.map(registeredExtenstions, (extension) => buildCompoment(extension, scope));
+            var registeredExtenstions = extensionPoints.extensions[attr.sdExtensionPoint];
+            var components = _.map(registeredExtenstions, (extension) => _buildCompoment(extension, scope));
```
gbbr: Please use Array.map, no need for underscore here either. It will work just fine.

scripts/apps/extension-points/directives/ExtensionPointDirective.js
markusherzog: I'm adding a check now, but it will return an empty array if there aren't any registered extension for that specific extension point. It is not an error if not all of the possible extension points are actually used.

scripts/apps/extension-points/services/ExtensionPoints.js
markusherzog: I'd like to use data as the parameter's name, so I changed the documention accordingly

scripts/apps/extension-points/services/ExtensionPoints.js
```diff
@@ -6,27 +6,30 @@
...
         }
-        extensions[type].push({type: type, componentClass: componentClass, data: data});
+        this.extensions[type].push({type, componentClass, props, data});
     };
 
-    this.$get = function() {
-        return extensions;
+    this.get = (type) => {
+        if (typeof this.extensions[type] === 'undefined') {
+            return [];
```
gbbr: It would be nice to let the user know if this happens. Either by `console.error`, or `throw`. Otherwise it can cause hard to debug problems.

markusherzog: I think it's the best thing to do to ignore this case silently. Let me try to explain why:
The purpose of this plugin is to make the superdesk UI extendable. For this to work you'd need to add "extension points" (i.e. an angular directive) in all of the places that potentially someone else might want to hook in to. This piece of code is only used by that angular directive, not by the developer that is actually writing an extension. So if we sprinkle extension points all over the UI, probably only some of them are actually used by an extension - maybe even none, depending on the setup. So when there is an extension point that is not used, the above happens. But that's not an error, it is rather the default

gbbr: I see. My reasoning was that you'd want to know in case you potentially mistype the name of the component.

gbbr:
@markusherzog You have to rebase before this can be merged.

PR #1793, opened 2017-08-16
[users] added author user type

jerome-poisson:
New author user type has been added. It is selected by default for new
users.

When author is selected, 2 new fields "biography" and "job title" appear.

A user type filter as been added on the left side so it's easy to select
only the desired users.

SDESK-1754


jerome-poisson:
closed as the feature has been modified, the P.R. https://github.com/superdesk/superdesk-client-core/pull/1798 replaces it.

PR #1798, opened 2017-08-18
[users] added author flag

jerome-poisson:
New author flag has been added. It is selected by default for new
users.

When author is selected, 2 new fields "biography" and "job title" are
enabled.

Filters have been added on the left side so it's easy to select
only the desired users.

SDESK-1754

### DIFF
--- a/scripts/apps/users/views/edit-form.html
+++ b/scripts/apps/users/views/edit-form.html
@@ -120,6 +120,18 @@
                   <input type="checkbox" ng-true-value="'administrator'" ng-false-value="'user'" ng-model="user.user_type" name="user_type" id="user_type">
                 </div>
 
+                <div sd-info-item class="user_author" ng-if="privileges.users">
+                  <label for="user_author" translate>author</label>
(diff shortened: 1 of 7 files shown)


scripts/apps/users/views/edit-form.html
```diff
@@ -120,6 +120,18 @@
                   <input type="checkbox" ng-true-value="'administrator'" ng-false-value="'user'" ng-model="user.user_type" name="user_type" id="user_type">
                 </div>
 
+                <div sd-info-item class="user_author" ng-if="privileges.users">
+                  <label for="user_author" translate>author</label>
+                  <input type="checkbox" ng-model="user.is_author" name="user_author" id="user_author">
+                  <div sd-info-item>
+                    <label for="job_title">{{ 'Job Title'|translate }}</label>
+                    <input type="text" name="jid" id="job_title" ng-model="user.job_title" ng-disabled="!user.is_author">
```
petrjasek: name seems wrong here, but that probably doesn't affect anything

jerome-poisson: oh right, thx

PR #1812, opened 2017-08-31
fix(editor3): Fix file upload from external folders (SDESK-1681)

ioanpocol:


scripts/core/menu/menu.js
```diff
@@ -31,7 +31,14 @@ angular.module('superdesk.core.menu', [
             this.flags = superdeskFlags.flags;
 
             this.openUpload = function openUpload(files) {
```
gbbr: This function can take extra parameters (you may look into the [official docs](https://github.com/danialfarid/ng-file-upload#file-select-and-drop)). You can capture the drop event here, but you will have to make a change in `scripts/core/menu/views/superdesk-view.html`:
```html
<div id="main-container" ng-class="superdesk.flags" ng-if="!popup && superdesk.flags.workspace === true" ngf-drop="superdesk.openUpload($files, $event)">
```
Notice the second `$event` param I've added at the very end of the line. Now, you can change this line to:
```js
this.openUpload = function openUpload(files, event) {
```
and use [`event.target.contains`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains) function to find out if the drop target is `div.public-DraftEditor-content`, in which case do not upload. It would also be nice to leave a comment like `// Editor3 handles its own file drop events`.

Then you can use the editor3 actions to handle the upload, without using these global variables. Just like it works now.

gbbr:
Here is the code to prevent dropping on Editor3:
```diff
diff --git a/scripts/core/menu/menu.js b/scripts/core/menu/menu.js
index c132864a..6c06d004 100644
--- a/scripts/core/menu/menu.js
+++ b/scripts/core/menu/menu.js
@@ -30,8 +30,13 @@ angular.module('superdesk.core.menu', [
         function SuperdeskViewController(superdeskFlags, superdesk) {
             this.flags = superdeskFlags.flags;
 
-            this.openUpload = function openUpload(files) {
-                superdesk.intent('upload', 'media', files);
+            this.openUpload = function openUpload(files, e) {
+                const draftNodes = document.getElementsByClassName('Editor3-root');
+                const onEditor3 = Array.prototype.some.call(draftNodes, (node) => node.contains(e.target));
+
+                if (!onEditor3) {
+                    superdesk.intent('upload', 'media', files);
+                }
             };
         }
 
diff --git a/scripts/core/menu/views/superdesk-view.html b/scripts/core/menu/views/superdesk-view.html
index b28b4cff..85c656e5 100644
--- a/scripts/core/menu/views/superdesk-view.html
+++ b/scripts/core/menu/views/superdesk-view.html
@@ -3,7 +3,7 @@
 
 <div class="loading" ng-if="loading"></div>
 
-<div id="main-container" ng-class="superdesk.flags" ng-if="!popup && superdesk.flags.workspace === true" ngf-drop="superdesk.openUpload($files)">
+<div id="main-container" ng-class="superdesk.flags" ng-if="!popup && superdesk.flags.workspace === true" ngf-drop="superdesk.openUpload($files, $event)">
     <div id="workspace-container" ng-view sd-splitter-widget></div>
     <div id="authoring-container" ng-if="superdesk.flags.workspace" sd-authoring-container></div>
     <div id="superdesk-workqueue" ng-if="superdesk.flags.workspace" sd-workqueue></div>
```

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -133,8 +144,19 @@ export class Editor3Component extends React.Component {
...
             // Firefox ignores the result of onDragOver and accept the item in all cases
             // Here will be tested again if the item is allowed
-            this.props.dragDrop(e);
+
+            if (this.isFileUpload(e)) {
+                const files = e.originalEvent.dataTransfer.files;
+
+                this.props.insertImages(files);
+            } else {
+                this.props.dragDrop(e);
```
gbbr: I think the Redux friendly way here is to leave the original implementation, and instead do this logic inside the action, like this:
```diff
diff --git a/scripts/core/editor3/actions/editor3.jsx b/scripts/core/editor3/actions/editor3.jsx
index a8b540e8..d9e20267 100644
--- a/scripts/core/editor3/actions/editor3.jsx
+++ b/scripts/core/editor3/actions/editor3.jsx
@@ -46,6 +46,12 @@ export function handleEditorTab(e) {
  * @description Creates the editor drop action.
  */
 export function dragDrop(e) {
+    const transfer = e.originalEvent.dataTransfer;
+
+    if (transfer.types[0] === 'Files') {
+        return insertImages(transfer.files);
+    }
+
     return {
         type: 'EDITOR_DRAG_DROP',
         payload: e
```

scripts/core/editor3/actions/toolbar.jsx
```diff
@@ -84,11 +84,11 @@ export function removeLink() {
  * @description Displays the external upload dialog and returns the insert images
  * action.
```
gbbr: Would be cool to document this new parameter up here...

ioanpocol: done

PR #1813, opened 2017-09-01
[SDESK-1754] - Author management

pavlovicnemanja:


petrjasek:
what is this PR actually doing? it seems to do a lot not related to authors management

hlmnrmr:
It should provide new organization to the user form. Separating user info from author info. Two tabs are now available and the fields are separated. The author tab is only available if the author checkbox is checked. That's from the functional point of view, not sure what other changes have been added, @fritzSF can clarify...

fritzSF:
@petrjasek , there was a lot of general cleaning up of markup and CSS, which has gone into this PR.

PR #1817, opened 2017-09-05
Fix e2e test for "Custom text fields"

naspeh:
It was caused by race condition in merging two pull requests
- #1800 which added a test at Aug 22, 2017, but  merged at Sep 1, 2017
- #1805 which added `text helper` at Aug 24, 2017 and merged at Aug 25, 2017


naspeh:
Thank you @petrjasek, just added `;` %)

naspeh:
Update for `superdesk-core` is also needed :)

PR #1822, opened 2017-09-07
fix(Editor3): Fix the inline editing of caption

ioanpocol:


scripts/core/editor3/components/images/ImageBlock.jsx
```diff
@@ -60,6 +60,7 @@ export class ImageBlockComponent extends Component {
         const entityKey = block.getEntityAt(0);
 
         changeCaption(entityKey, target.value, target.placeholder);
+        this.setState({refreshTrigger: target.value});
```
gbbr: Please use `this.forceUpdate()` instead.

PR #1828, opened 2017-09-12
fix(tansa): Load tansa module in async mode

ioanpocol:


scripts/index.js
```diff
@@ -13,7 +13,7 @@ const appConfig = __SUPERDESK_CONFIG__;
 import externalApps from 'external-apps';
 
 if (appConfig.features.useTansaProofing) {
-    require('apps/tansa');
+    setTimeout(require('apps/tansa'), 0);
```
gbbr: What is the effect of this? Perhaps an explanation why you did this would be helpful to others. Maybe a link where we can find more about this approach. I think `require.ensure` is for async, in WebPack.

ioanpocol:
http://www.hiddenwebgenius.com/blog/guides/understanding-javascripts-asynchronous-code/

petrjasek:
if the issue with with loading tansa script, imo there should be `<script async>` somewhere rather than generating the `<script>` tag using timeout which would then stop anyhow

petrjasek:
sorry closed by mistake

gbbr:
The logic you're looking for is in [`app/tansa/index.js`](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/tansa/index.js). As far as I see it's already loaded async. It's lacking the `async` attribute though. So you may add it there.

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -534,22 +534,22 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
             var deregisterTansa = $rootScope.$on('tansa:end', afterTansa);
 
             $scope.runTansa = function() {
-                const editor = editorResolver.get();
+                if (window.tansa && window.RunTansaProofing && window.tansaIsLoaded) {
```
gbbr: Will `if (window.tansaIsLoaded) {` not be enough?

ioanpocol: I think it is safer, for example RunTansaProofing is an object created by tansa script.

PR #1835, opened 2017-09-14
CI: fixes for npmtest and e2e

naspeh:
- try to increase `browserNoActivityTimeout` for karma
- use `browser.wait` in `dashboard_spec.js`

### DIFF
--- a/spec/dashboard_spec.js
+++ b/spec/dashboard_spec.js
@@ -44,7 +44,12 @@ describe('dashboard', () => {
         monitoring.nextSearches();
         monitoring.nextReorder();
         monitoring.saveSettings();
-        browser.sleep(300);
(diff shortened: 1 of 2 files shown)


spec/dashboard_spec.js
```diff
@@ -44,7 +44,7 @@ describe('dashboard', () => {
         monitoring.nextSearches();
         monitoring.nextReorder();
         monitoring.saveSettings();
-        browser.sleep(300);
+        browser.sleep(3000);
```
petrjasek: it's better to use `wait` so it can run faster if possible, check `spec/helpers/utils.js`

naspeh: yep it's better for sure, done

PR #1845, opened 2017-09-19
fix(tansa): Load tansa async

ioanpocol:


scripts/index.js
```diff
@@ -13,7 +13,7 @@ const appConfig = __SUPERDESK_CONFIG__;
 import externalApps from 'external-apps';
 
 if (appConfig.features.useTansaProofing) {
-    require('apps/tansa');
+    setTimeout(require('apps/tansa'), 0);
```
gbbr: Could you live a short note or better yet, the link to that article where this sorcery is explained? 🎩

PR #1849, opened 2017-09-20
feat(desk): Add helper text to the desk type in Desk management when …

ioanpocol:
…publish in authoring is not allowed

### DIFF
--- a/scripts/apps/desks/directives/DeskeditBasic.js
+++ b/scripts/apps/desks/directives/DeskeditBasic.js
@@ -1,8 +1,10 @@
 import {limits} from 'apps/desks/constants';
 import _ from 'lodash';
 
-DeskeditBasic.$inject = ['gettext', 'desks', 'WizardHandler', 'metadata', '$filter', '$interpolate', '$rootScope'];
-export function DeskeditBasic(gettext, desks, WizardHandler, metadata, $filter, $interpolate, $rootScope) {
+DeskeditBasic.$inject = ['gettext', 'desks', 'WizardHandler', 'metadata', 'config',
+    '$filter', '$interpolate', '$rootScope'];
+export function DeskeditBasic(gettext, desks, WizardHandler, metadata, config,
+    $filter, $interpolate, $rootScope) {
     return {
         link: function(scope, elem, attrs) {
             scope.limits = limits;
@@ -91,6 +93,10 @@ export function DeskeditBasic(gettext, desks, WizardHandler, metadata, $filter, 
                 if (!_.isNil(scope.desk.edit.name)) {
                     scope._errorLimits = scope.desk.edit.name.length > scope.limits.desk ? true : null;
                 }
+            };
+
+            scope.showNoPublishOnAuthoringDesk = function(deskType) {
+                return deskType === 'authoring' && config.features.noPublishOnAuthoringDesk;
             };
 
             if (metadata.values.desk_types) {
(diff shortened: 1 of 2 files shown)


scripts/apps/desks/directives/DeskeditBasic.js
```diff
@@ -93,6 +95,12 @@ export function DeskeditBasic(gettext, desks, WizardHandler, metadata, $filter,
                 }
             };
 
+            scope.getNoPublishOnAuthoringDesk = function(deskType) {
```
gbbr: It would be nice to leave a comment. I'm not familiar with this functionality (but would like to be) and I have no idea what "NoPublishOnAuthoringDesk" means.

Additionally, some points:
- `config.features` is always defined in the [default app configuration](https://github.com/superdesk/superdesk-client-core/blob/master/webpack.config.js#L225) so you don't need to check
- `config.features.noPublishOnAuthoringDesk` should be enough, no need to use the `in` operator.
- Because this function is getting the value of a feature and not an item, it might be a bit confusing it to prefix it with _get_

My recommendation would be to make this:
```js
// Checks if the given deskType can publish on the authoring desk.
scope.canPublishOnAuthoring = (deskType) =>
    deskType === 'authoring' && config.features.noPublishOnAuthoringDesk;
```

gbbr: There was a mistake in my code snippet, I have edited the above comment.

scripts/apps/desks/views/desk-config-modal.html
```diff
@@ -52,6 +52,9 @@ <h3 class="modal__heading" ng-show="desk.edit._id" translate>Edit "{{desk.edit.n
                                     <option id="{{type.qcode}}" ng-repeat="type in deskTypes" value="{{type.qcode}}" ng-selected="desk.edit.desk_type === type.qcode"
                                         >{{ type.name | translate}}</option>
                                 </select>
+                                <div ng-show="getNoPublishOnAuthoringDesk(desk.edit.desk_type)" class="column">
```
gbbr: This would become:
```html
<div ng-show="!canPublishOnAuthoring(desk)" class="column">
```

scripts/apps/desks/directives/DeskeditBasic.js
ioanpocol: I think config.feature can be empty, it is just your assumption that it is already prsent.

gbbr: I have already linked to the code where `config.features` is defined, illustrating that it is never empty. I don't understand why you say it's an assumption.

gbbr: Here it is again, in case you've missed it:
https://github.com/superdesk/superdesk-client-core/blob/master/webpack.config.js#L225

It gets assigned here, on app init:
https://github.com/superdesk/superdesk-client-core/blob/0d233de155b60ccc029dd7de480e3e28e7c49a23/scripts/index.js#L10

gbbr:
Added [more comments](https://github.com/superdesk/superdesk-client-core/pull/1849#discussion_r139945179) to the hidden thread.

scripts/apps/desks/directives/DeskeditBasic.js
ioanpocol: Yes, I missed that link on the first time

scripts/apps/desks/directives/DeskeditBasic.js
```diff
@@ -93,6 +95,10 @@ export function DeskeditBasic(gettext, desks, WizardHandler, metadata, $filter,
                 }
             };
 
+            scope.showNoPublishOnAuthoringDesk = function(deskType) {
```
gbbr: In your other PR #1850 you've called it `canPublishOnDesk`. Which is a much better name. Please use the same here for consistency.

ioanpocol: Here the meaning is to show the information about "NoPublishOnAuthoringDesk"

PR #1850, opened 2017-09-20
feat(authoring): Disable publishing in "authoring" desk type when con…

ioanpocol:
…fig.features.noPublishOnAuthoringDesk is true


scripts/apps/authoring/authoring/directives/SendItem.js
```diff
@@ -367,10 +368,21 @@ export function SendItem($q, api, desks, notify, authoringWorkspace,
             };
 
             /**
+             * Check if it is allowed to publish on current desk
+             * @returns {Boolean}
+             */
+            scope.canPublishOnDesk = function() {
```
gbbr: This doesn't seem to be used in the template, does it have to be on the `scope`?

PR #1852, opened 2017-09-20
Fix false failing in authoring_spec.js

naspeh:
Even it wasn't failing because of `protractor-flake`, it was usually
passing during second run, but it increase the time significantly,
because `authoring_spec.js` is huge...

From logs:
- https://test.superdesk.org/logs/3/all/20170920-113653-46-sdcpr-1801/check-e2e--part1.log.htm
- https://test.superdesk.org/logs/3/all/20170920-104852-52-sdcpr-1849/check-e2e--part1.log.htm

### DIFF
--- a/spec/authoring_spec.js
+++ b/spec/authoring_spec.js
@@ -665,7 +665,6 @@ describe('authoring', () => {
         ctrlKey('x');
         authoring.sendAndpublish('Sports Desk');
         authoring.confirmSendTo(); // confirm unsaved changes
-        browser.sleep(1000);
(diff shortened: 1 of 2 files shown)


naspeh:
it uses `browser.wait`, so it works without delay if element exists or wait for 1-2sec

naspeh:
BTW: usually `e2e--part1` is running in 25 minutes, when false failing occurs in `authoring_spec.js` it takes 35 minutes...

gbbr:
Ah, I see. My bad. Looks good! :)

PR #1859, opened 2017-09-28
Do prepolulation request only once

naspeh:
Concurrent requests to `prepopulate` endpoint just mess thnigs up.

### DIFF
--- a/spec/helpers/backend.js
+++ b/spec/helpers/backend.js
@@ -13,8 +13,16 @@ function getBackendUrl(uri) {
     return constructUrl(browser.params.baseBackendUrl, uri);
 }
 
-function backendRequest(params, callback) {
+/**
+ * Do request to backend with retry on error
+ *
+ * @param {Object} params
+ * @param {function} callback
+ * @param {Integer} retry - how many times it will try to request before throwing error
+ */
+function backendRequest(params, callback, retry = 3) {
     let cb = callback || function() { /* no-op */ };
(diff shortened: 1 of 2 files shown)


spec/helpers/backend.js
```diff
@@ -26,7 +26,7 @@ function backendRequest(params, callback) {
     }
 
     // how many times it will try to request before throwing error
-    var ttl = 3;
+    let ttl = retry && 3 || 0;
```
gbbr: This is too fancy, please use the ternary:
```js
let ttl = retry ? 3 : 0;
```
Or just delete this variable and declare the function as:
```js
function backendRequest(params, callback, ttl = 0) {
```
Then when calling it, instead of true or false just use the ttl value

naspeh: thanks, I'll change to your second version with `ttl = 0`

naspeh:
```
/arch/naspeh/works/sourcefabric/superdesk-dev/client-core/spec/helpers/backend.js
  45:13  error  Assignment to function parameter 'ttl'  no-param-reassign

✖ 1 problem (1 error, 0 warnings)
```
because below we have
```
ttl -= 1;
```
So I'll use different parameter name then

naspeh: Done

gbbr: You're right, I didn't see that initially, LGTM! Thanks

PR #1863, opened 2017-10-09
SDESK-1965 As a user I want to use custom CVs when I create a "filter condition"

mugurrus:


scripts/apps/content-filters/controllers/FilterConditionsController.js
```diff
@@ -20,6 +20,7 @@ export function FilterConditionsController($scope, contentFilters, notify, modal
     $scope.valueLookup = {};
     $scope.valueFieldLookup = {};
     $scope.loadedFilters = false;
+    $scope.labels = {};
```
gbbr: Given that this never changes, it doesn't have to live on the $scope. Every time we put something into the $scope we should consider that we are adding extra computations into each Angular cycle and can affect performance.

_Only_ put items that need to be two-way bound to the view on the scope. Most stuff doesn't have to be, so it can be a plain controller property accessible through the template.

If I'm wrong and the labels can change, ignore this.

PR #1868, opened 2017-10-10
feat(ui-framework): Update to the latest version of ui-framework

pavlovicnemanja:


package.json
```diff
@@ -111,7 +111,7 @@
     "sass-loader": "6.0.6",
     "shortid": "^2.2.8",
     "style-loader": "0.18.2",
-    "superdesk-ui-framework": "superdesk/superdesk-ui-framework#195e5ca",
+    "superdesk-ui-framework": "superdesk/superdesk-ui-framework#957eac552",
```
petrjasek: maybe it's time to release some 1.0 version, what do you think?

pavlovicnemanja: we have it, v1.0, it's final version before refactoring to webpack and adding react. After I fix all issues, I will create new release and add it :)

petrjasek: do you push this to npm? it should be set then like `"superdesk-ui-framework": "1.0.0",`

pavlovicnemanja: still not, but I will make that also! :)

PR #1870, opened 2017-10-11
feat(push): Add support for secret token for push subscriber

ioanpocol:


scripts/apps/publish/views/http-push-config.html
```diff
@@ -13,3 +13,7 @@
     <label for="http_host" translate>Assets URL</label>
     <input type="text" id="http_host" placeholder="{{ :: 'Assets URL'|translate }}" ng-model="destination.config.assets_url" ng-change="$parent.setConfig(destination)">
 </div>
+<div class="field">
+    <label for="http_host" translate>Secret Token</label>
+    <input type="text" id="http_host" placeholder="{{ :: 'Secret Token'|translate }}" ng-model="destination.config.secret_token" ng-change="$parent.setConfig(destination)">
```
petrjasek: the ids should be unique otherwise labels won't work. also an issue with assets url

PR #1874, opened 2017-10-18
ansa changes should go to master

petrjasek:
same thing like https://github.com/superdesk/superdesk-core/pull/1055

### DIFF
--- a/scripts/apps/archive/controllers/BaseListController.js
+++ b/scripts/apps/archive/controllers/BaseListController.js
@@ -8,10 +8,10 @@ export class BaseListController {
         $scope.selected = {};
 
         $scope.fetchNext = (from) => {
-            var criteria = this.getQuery();
+            const source = this.getQuery(null, $scope.repo.archive || false);
 
-            criteria.from = from;
-            this.fetchItems({source: criteria}, true);
+            source.from = from;
+            this.fetchItems({source}, true);
         };
 
         $scope.$on('$routeUpdate', (e, data) => {
--- a/scripts/apps/authoring/authoring/directives/ItemAssociationDirective.js
+++ b/scripts/apps/authoring/authoring/directives/ItemAssociationDirective.js
@@ -17,8 +17,8 @@
  */
 
 ItemAssociationDirective.$inject = ['superdesk', 'renditions', 'config', 'authoring', '$q',
-    'api', 'notify', 'gettext'];
-export function ItemAssociationDirective(superdesk, renditions, config, authoring, $q, api, notify, gettext) {
+    'api', 'notify', 'gettext', 'send'];
+export function ItemAssociationDirective(superdesk, renditions, config, authoring, $q, api, notify, gettext, send) {
     return {
         scope: {
             rel: '=',
@@ -63,8 +63,13 @@ export function ItemAssociationDirective(superdesk, renditions, config, authorin
                 let item = angular.fromJson(event.originalEvent.dataTransfer.getData(dataType));
 
                 if (item._type !== 'externalsource') {
+                    if (item._type === 'ingest') {
+                        return send.one(item);
+                    }
+
                     return api.find(item._type, item._id);
                 }
+
                 return $q.when(item);
             }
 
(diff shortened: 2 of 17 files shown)


gbbr:
Do you want this reviewed, or not yet? It would be nice to rewrite the commit history a bit and perhaps summarise in the commit message.

petrjasek:
@gbbr yep should be good to go, summary:

- allow multiple values in place field
- allow only single widget with specific id
- fetch item from external source before editing on dbl click
- add publish action to multi action toolbar
- allow extra modules to be within superdesk repo

plus some tweaks to lists/criteria/css

scripts/apps/archive/controllers/BaseListController.js
```diff
@@ -8,9 +8,10 @@ export class BaseListController {
         $scope.selected = {};
 
         $scope.fetchNext = (from) => {
-            var criteria = this.getQuery();
+            var criteria = this.getQuery(null, $scope.repo.archive || false);
 
             criteria.from = from;
+            criteria.size = 50;
```
gbbr: You are already setting the size to 50 in the other file, so you can remove this line.

scripts/apps/archive/controllers/BaseListController.js
```diff
@@ -8,9 +8,10 @@ export class BaseListController {
         $scope.selected = {};
 
         $scope.fetchNext = (from) => {
-            var criteria = this.getQuery();
+            var criteria = this.getQuery(null, $scope.repo.archive || false);
```
gbbr: I would call this `source` instead of `criteria`. It is confusing when you look at it from the perspective of the `fetchItems` function where `criteria` means the parent object.

scripts/apps/authoring/authoring/directives/ItemAssociationDirective.js
```diff
@@ -62,9 +62,15 @@ export function ItemAssociationDirective(superdesk, renditions, config, authorin
             function getItem(event, dataType) {
                 let item = angular.fromJson(event.originalEvent.dataTransfer.getData(dataType));
 
+
```
gbbr: Accidental blank line.

scripts/apps/archive/controllers/BaseListController.js
```diff
@@ -8,9 +8,10 @@ export class BaseListController {
...
 
             criteria.from = from;
+            criteria.size = 50;
             this.fetchItems({source: criteria}, true);
```
gbbr: Then you can have `this.fetchItems({source}, true);` here.

scripts/apps/ingest/controllers/IngestListController.js
```diff
@@ -17,6 +17,7 @@ export class IngestListController extends BaseListController {
             $scope.loading = true;
             criteria.aggregations = 1;
             criteria.es_highlight = search.getElasticHighlight();
+            criteria.source.size = 50;
```
gbbr: I wonder if this could be done differently, so that we don't have it in two places...

scripts/apps/search/controllers/MultiActionBarController.js
```diff
@@ -130,4 +130,21 @@ export function MultiActionBarController(
...
         return multi.getItems().every((item) => authoring.itemActions(item).mark_item_for_highlight);
     };
+
+    /**
+     * Publish all items
+     */
+    this.publish = () => {
+        $q.all(multi.getItems().map((item) => authoring.publish(item, item)))
+            .then((responses) => {
+                let errors = responses.filter((response) => response.status >= 400);
```
gbbr: You can use `some` instead of `filter`. It'll return a boolean and it will stop as soon as it finds the first one.

tasks/options/ngtemplates.js
```diff
@@ -78,21 +78,10 @@ module.exports = {
                     return 'export default [];\r\n';
                 }
 
-                // abs returns the correct absolute path to be used with a require call. On Windows,
```
gbbr: What about the Windows problem?

petrjasek: these should be gone because it doesn't put there abs path but module name and let webpack to resolve it

PR #1880, opened 2017-10-24
feat(custom fields): Add option to support image, video or image&vide…

ioanpocol:
…o files

### DIFF
--- a/scripts/apps/archive/controllers/UploadController.js
+++ b/scripts/apps/archive/controllers/UploadController.js
@@ -9,6 +9,8 @@ export function UploadController($scope, $q, upload, api, archiveService, sessio
     $scope.enableSave = false;
     $scope.currentUser = session.identity;
     $scope.uniqueUpload = $scope.locals && $scope.locals.data && $scope.locals.data.uniqueUpload === true;
+    $scope.allowPicture = !($scope.locals && $scope.locals.data && $scope.locals.data.allowPicture === false);
+    $scope.allowVideo = !($scope.locals && $scope.locals.data && $scope.locals.data.allowVideo === false);
     $scope.validator = config.validatorMediaMetadata;
 
     var uploadFile = function(item) {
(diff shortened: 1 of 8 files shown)


scripts/apps/vocabularies/views/vocabulary-config-modal.html
```diff
@@ -66,6 +66,14 @@ <h3 class="modal__heading" ng-show="!vocabulary._id"><span translate>Create</spa
                             </label>
                         </div>
                     </li>
+                    <li>
+                        <div>
+                            <label>
+                                <input type="radio" name="field_type" value="picture_video" ng-model="vocabulary.field_type">
```
petrjasek: I'm not sure it's a good idea to have such a field type, imo would be better to have field_type=media which has some custom logic like it's a drag & drop with upload on click and use some other field (in options) for supported mimetypes. otherwise it will complicate the logic all over the place

petrjasek: it would also imply there could be `video_audio`, `picture_audio`, `picture_video_audio` etc

scripts/apps/authoring/views/article-edit.html
```diff
@@ -526,7 +527,8 @@
      data-required="schema[field._id].required">
     <label translate>{{ field.display_name }}</label>
     <div sd-item-association
-         data-allow-video="field.field_type === 'video'"
+         data-allow-video="field.field_type === 'video' || field.field_type === 'media'"
```
petrjasek: I thought we would use filed type media also for picture and video, so those would not be needed. and there would be something like `field.field_options.allowed_types` with eg `{picture: 1, video: 1, audio: 0}` or just `{picture: 1}`

scripts/apps/archive/controllers/UploadController.js
```diff
@@ -81,8 +83,12 @@ export function UploadController($scope, $q, upload, api, archiveService, sessio
         if (!files.length) {
             return false;
         }
+
         _.each(files, (file) => {
             if (/^image/.test(file.type)) {
+                if (!$scope.allowPicture && $scope.allowVideo) {
+                    $scope.errorMessage = 'Only video files are allowed';
```
petrjasek: missing gettext here and below

PR #1882, opened 2017-10-29
[SDESK-2037] Display custom error message on Archive Item lock fail

nrvikas:


scripts/apps/authoring/authoring/services/LockService.js
```diff
@@ -13,7 +15,9 @@ export function LockService($q, api, session, privileges, notify) {
                 item.lock_action = action || 'edit';
                 return item;
             }, (err) => {
-                notify.error(gettext('Failed to get a lock on the item!'));
+                let msg = _.get(err, 'data._message') || 'Failed to get a lock on the item!';
```
petrjasek: would be better to have gettext around `'Failed to get...' otherwise the tool can't extract it

PR #1890, opened 2017-11-02
fix(authoring): Fix required errors for custom text and media fields

ioanpocol:


scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -417,7 +416,11 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
                                     .split(',');
 
                                 for (var i = 0; i < modifiedErrors.length; i++) {
-                                    notify.error(_.trim(modifiedErrors[i]));
+                                    var message = _.trim(modifiedErrors[i]);
+                                    var field = helpers.getFirstWord(message).substr(1);
```
gbbr: You can just use `message.split(" ")[0]` instead of creating this new method `getFirstWord` which does the same thing.

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -417,7 +416,11 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
...
+
+                                    $scope.error[field.toLowerCase()] = true;
+                                    notify.error(message);
```
gbbr: What is going on here? Why are we displaying only the first word without the first character? It's not self-explanatory.

scripts/apps/authoring/views/article-edit.html
```diff
@@ -501,7 +512,7 @@
     sd-width="{{ editor[field._id].sdWidth || 'full' }}"
     data-required="schema[field._id].required"
     order="{{ editor[field._id].order }}"
-    sd-validation-error="error.extra[field._id]">
+    sd-validation-error="error[field._id.toLowerCase()]">
```
mugurrus: If field._id has the same value as the name of a field from the item schema (e.g. headline) this will create confusion.

scripts/apps/authoring/views/article-edit.html
```diff
@@ -522,7 +533,7 @@
 <div class="field" ng-repeat="field in fields track by field._id"
      ng-if="field.field_type === 'media'"
      sd-width="{{editor[field._id].sdWidth || 'full'}}"
-     sd-validation-error="error.extra[field._id]"
+     sd-validation-error="error[field._id.toLowerCase()]"
```
mugurrus: Same here.

scripts/apps/authoring/views/article-edit.html
ioanpocol: The errors are added in a dictionary by id lowercase. Why is this a problem?

mugurrus: Say you have a vocabulary with id 'headline'. If you got an error on the vocabulary field it will look as an error on the 'headline' field from the item. I know it's an edge case, but why not keep custom vocabulary errors in extra?

ioanpocol: When it comes from backend I don't know if it is an custom extra (custom vocabulary or field) error or not. This is more a vocabulary and custom field validation.

PR #1892, opened 2017-11-02
fix(editor3): Show video and audio on editor3

ioanpocol:


ioanpocol:
@gbbr  Should we rename ImageBlock to MediaBlock? Other suggestions?

gbbr:
This is looking good. Indeed I also think it would be a nice idea to rename the `images` folder to `media` and the `ImageBlock` to `MediaBlock`. But I'm afraid there are more changes with regards to terminology throughout:

- Rename the content profile entry from `picture` to `media`
https://github.com/gbbr/superdesk-client-core/blob/095f47234c3485f006a1a19523fc7fd02f608580/scripts/core/editor3/components/toolbar/index.jsx#L154
https://github.com/gbbr/superdesk-client-core/blob/4c0c9f364c7378e3fe62b122427342be98b083f5/scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.js#L68
- Rename the block type from `IMAGE` to `MEDIA`
https://github.com/gbbr/superdesk-client-core/blob/4c0c9f364c7378e3fe62b122427342be98b083f5/scripts/core/editor3/components/blockRenderer.jsx#L20
- Rename actions and their names (where appropriate), such as:
https://github.com/gbbr/superdesk-client-core/blob/4c0c9f364c7378e3fe62b122427342be98b083f5/scripts/core/editor3/actions/toolbar.jsx#L126-L137

I might have missed some. Perhaps a good idea is to case-insensitively search the entire folder `scripts/core/editor3` for the term `images` and check.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -85,7 +85,7 @@ export class Editor3Component extends React.Component {
             'text/html',
             'Files'
         ].indexOf(mediaType) !== -1;
-        const supportsImages = !readOnly && !singleLine && editorFormat.indexOf('picture') !== -1;
+        const supportsImages = !readOnly && !singleLine && editorFormat.indexOf('media') !== -1;
```
gbbr: You've missed one here `const supportsMedia`, make sure you replace where it is used too.

scripts/core/editor3/html/to-html/AtomicBlockParser.js
```diff
@@ -66,19 +66,28 @@ export class AtomicBlockParser {
...
-        const alt = img.alt_text || '';
+        const alt = media.alt_text || '';
+        const mediaType = media.type;
 
-        let html = `<div class="image-block"><img src="${href}" alt="${alt}" />`;
+        let html = '<div class="media-block">';
 
-        html += img.description_text
-            ? `<span class="image-block__description">${img.description_text}</span>`
+        if (mediaType === 'video') {
```
gbbr: Please consider using a switch case here, it will look friendlier.

PR #1899, opened 2017-11-09
[SDESK-2088]: Fixing the context menu for highlights and mark desks actions.

mdhaman:
This pr handles the multi-user scenario where one user is acting on an item using three-dot menu and another user is acting on highlights/markdesks of some other item.

### DIFF
--- a/scripts/apps/search/components/HighlightsInfo.jsx
+++ b/scripts/apps/search/components/HighlightsInfo.jsx
@@ -7,8 +7,8 @@ import {HighlightsList} from 'apps/search/components';
 
 import {
     closeActionsMenu,
-    isCheckAllowed,
-    renderToBody
+    openActionsMenu,
+    isCheckAllowed
 } from 'apps/search/helpers';
 
 export class HighlightsInfo extends React.Component {
@@ -26,7 +26,7 @@ export class HighlightsInfo extends React.Component {
             event.stopPropagation();
         }
 
-        closeActionsMenu();
+        closeActionsMenu(this.props.item._id);
         this.renderDropdown();
     }
 
@@ -120,7 +120,7 @@ export class HighlightsInfo extends React.Component {
         var icon = ReactDOM.findDOMNode(this).getElementsByClassName('icon-star')[0] ||
         ReactDOM.findDOMNode(this).getElementsByClassName('icon-multi-star')[0];
 
-        renderToBody(elem, icon);
+        openActionsMenu(elem, icon, this.props.item._id);
     }
 }
 
--- a/scripts/apps/search/components/HighlightsList.jsx
+++ b/scripts/apps/search/components/HighlightsList.jsx
@@ -11,6 +11,7 @@ export class HighlightsList extends React.Component {
         this.removeHighlight = this.removeHighlight.bind(this);
         this.stopTimeout = this.stopTimeout.bind(this);
         this.close = this.close.bind(this);
+        this.closeMenu = this.closeMenu.bind(this);
     }
 
     removeHighlight(highlight) {
@@ -20,6 +21,7 @@ export class HighlightsList extends React.Component {
         return function(event) {
             event.stopPropagation();
             highlightsService.markItem(highlight._id, this.props.item);
+            this.closeMenu();
 
             if (scope.viewType === 'highlights' && this.props.item.highlights.length === 1) {
                 $rootScope.$broadcast('multi:remove', this.props.item._id);
@@ -41,10 +43,14 @@ export class HighlightsList extends React.Component {
         this.timeout = $timeout.cancel(this.timeout);
     }
 
+    closeMenu() {
+        closeActionsMenu(this.props.item._id);
+    }
+
     close() {
         const {$timeout} = this.props.svc;
 
-        this.timeout = $timeout(closeActionsMenu, 2000, false);
+        this.timeout = $timeout(this.closeMenu, 2000, false);
     }
 
     render() {
@@ -80,7 +86,7 @@ export class HighlightsList extends React.Component {
                 ),
                 React.createElement(
                     'button',
-                    {className: 'dropdown__menu-close', onClick: closeActionsMenu},
+                    {className: 'dropdown__menu-close', onClick: this.closeMenu},
                     React.createElement(
                         'i',
                         {className: 'icon-close-small icon--white'}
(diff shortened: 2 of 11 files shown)


scripts/apps/search/components/HighlightsInfo.jsx
```diff
@@ -28,6 +31,24 @@ export class HighlightsInfo extends React.Component {
 
         closeActionsMenu();
         this.renderDropdown();
+        this.setActionMenuState(true);
+    }
+
+    componentWillUnmount() {
+        if (this.state.isActionMenuOpen) {
```
gbbr: Instead of adding in this state, why not just change `scripts/apps/search/helpers.js`? It could be:
```js
export function closeActionsMenu() {
    let el = menuHolderElem();

    if (el) {
        ReactDOM.unmountComponentAtNode(el);
    }
}
```
Then you can remove all this checking and state setting.

scripts/apps/search/components/MarkedDesksInfo.jsx
```diff
@@ -23,9 +23,12 @@ export class MarkedDesksInfo extends React.Component {
     constructor(props) {
         super(props);
 
+        this.state = {isActionMenuOpen: false};
+
         this.toggle = this.toggle.bind(this);
         this.getMarkedDesks = this.getMarkedDesks.bind(this);
         this.renderDropdown = this.renderDropdown.bind(this);
+        this.setActionMenuState = this.setActionMenuState.bind(this);
```
gbbr: Same here, please remove this state.

scripts/apps/search/components/HighlightsList.jsx
```diff
@@ -41,10 +43,17 @@ export class HighlightsList extends React.Component {
         this.timeout = $timeout.cancel(this.timeout);
     }
 
+    closeMenu() {
+        this.props.onClose && this.props.onClose(false);
```
gbbr: Instead of doing this, just set a default prop for it and do:
```js
this.props.onClose(false);
```
and further down:
```js
HighllightsList.defaultProps = {
    onClose: () => {}
}
```

scripts/apps/search/components/MarkedDesksList.jsx
```diff
@@ -39,6 +41,11 @@ export class MarkedDesksList extends React.Component {
         this.stopTimeout();
     }
 
+    closeMenu() {
+        this.props.onClose && this.props.onClose(false);
```
gbbr: Same here, use defaultProps

scripts/apps/search/components/HighlightsInfo.jsx
mdhaman: Just checking if the menu element exists won't work. It will close the action menu of any other item. Hence the state is set for that particular item.

mdhaman:
Thanks @gbbr.

scripts/apps/search/components/HighlightsInfo.jsx
petrjasek: I think it's true that there is only one menu at a time, so if we set somewhere what item has the menu opened it should be enough instead of having the state all over the place

scripts/apps/search/components/HighlightsList.jsx
```diff
@@ -107,4 +116,10 @@ HighlightsList.propTypes = {
     item: PropTypes.any,
     highlights: PropTypes.any,
     highlightsById: PropTypes.any,
+    onClose: PropTypes.func
+};
+
+
+HighlightsList.defaultProps = {
+    onClose: () => {} // eslint-disable-line no-empty-function
```
gbbr: You can get rid of the `eslint-disable` comment by rewriting this as:
```js
    onClose: () => { /* no-op */ }
```

scripts/apps/search/components/MarkedDesksList.jsx
```diff
@@ -48,7 +55,9 @@ export class MarkedDesksList extends React.Component {
     close() {
         const {$timeout} = this.props.svc;
 
-        this.timeout = $timeout(closeActionsMenu, 2000, false);
+        this.timeout = $timeout(() => {
```
gbbr: `$timeout` takes a function, so this should also work:
```js
this.timeout = $timeout(this.closeMenu, 2000, false)
```

scripts/apps/search/components/MarkedDesksList.jsx
```diff
@@ -89,4 +98,9 @@ MarkedDesksList.propTypes = {
     svc: PropTypes.object.isRequired,
     item: PropTypes.any,
     markedDesksById: PropTypes.any,
+    onClose: PropTypes.func
+};
+
+MarkedDesksList.defaultProps = {
+    onClose: () => {} // eslint-disable-line no-empty-function
```
gbbr: Do the same here, please.

scripts/apps/search/components/MarkedDesksInfo.jsx
```diff
@@ -35,6 +38,24 @@ export class MarkedDesksInfo extends React.Component {
...
+
+    componentWillUnmount() {
+        if (this.state.isActionMenuOpen) {
+            closeActionsMenu();
+        }
+    }
+
+    componentWillReceiveProps(nextProps) {
+        if (this.state.isActionMenuOpen && nextProps.item !== this.props.item) {
+            this.setActionMenuState(false);
```
gbbr: Is this hack here because it's not appropriate to set state in this method? If so, I think we should find a better way. This is very hacky as is. Plus, we don't need this flag to be in the state. Every time you set the state this component will be re-rendered.

Instead, can you not just use a property on the class? Something like `this.isActionMenuOpen` ?

scripts/apps/search/components/HighlightsInfo.jsx
gbbr: A quick solution might be to add an ID to the open menu element, and optionally pass it to `closeActionMenu`. If it's provided, then only close the menu with that ID (if it's open).

mdhaman: I will make change accordingly. @petrjasek, @gbbr

scripts/apps/search/components/MarkedDesksInfo.jsx
mdhaman: Not a hack. You can set state in `componentwillreceiveprops` method (see https://reactjs.org/docs/react-component.html#componentwillreceiveprops).

gbbr: Still, it doesn't have to be on the state. But you seem to have accepted Petr's and my suggestion already so this will have to be removed anyway.

PR #1901, opened 2017-11-09
fix(editor3): Activate save button after changes on image description…

ioanpocol:
… and on changes on table cell

### DIFF
--- a/scripts/core/editor3/reducers/editor3.jsx
+++ b/scripts/core/editor3/reducers/editor3.jsx
@@ -55,20 +55,37 @@ export const forceUpdate = (state) => {
 /**
  * @ngdoc method
  * @name onChange
+ * @param {Object} state
  * @param {Object} editorState
+ * @param {Bool} force When true, forces an editor update regardless of whether the content has changed.
+ * This is used because currently it is impossible to detect changes happening solely on entity data.
+ * In Draft v0.11.0 we will be able to request all entities from the content and could compare them to get a
+ * more accurate result.
+ * See https://draftjs.org/docs/api-reference-content-state.html#getentitymap"
  * @return {Object} returns new state
  * @description Handle the editor state has been changed event
  */
-export const onChange = (state, newState) => {
+export const onChange = (state, newState, force = false) => {
+    // TODO(x): Remove `force` once Draft v0.11.0 is in
     let editorState = newState, activeHighlight = null;
     let contentChanged = state.editorState.getCurrentContent() !== newState.getCurrentContent();
 
     if (state.allowsHighlights) {
         ({editorState, activeHighlight} = updateHighlights(state.editorState, newState));
     }
-    if (contentChanged) {
+
+    if (contentChanged || force) {
         state.onChangeValue(editorState.getCurrentContent());
     }
+
+    if (force) {
+        return forceUpdate({
+            ...state,
+            editorState,
+            activeHighlight
+        });
+    }
+
     return {
         ...state,
         editorState,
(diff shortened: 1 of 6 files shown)


scripts/core/editor3/reducers/toolbar.jsx
```diff
@@ -153,8 +153,8 @@ export const addMedia = (editorState, media) => {
  * @param {Object} data Contains the entityKey and the new image data.
  * @description Updates the given entityKey with the new image data.
  */
-const updateImage = (state, {entityKey, img}) => {
-    Entity.replaceData(entityKey, {img});
+const updateImage = (state, {entityKey, media}) => {
+    Entity.replaceData(entityKey, {media});
```
gbbr: Please do not use the `Entity` API. It is being deprecated in 0.11 and it will be hard to update DraftJS afterwards. Use `contentState.replaceEntityData` instead. Check the usage in `processCells` or other functions in this file.

scripts/core/editor3/reducers/editor3.jsx
```diff
@@ -55,20 +55,32 @@ export const forceUpdate = (state) => {
 /**
  * @ngdoc method
  * @name onChange
+ * @param {Object} state
  * @param {Object} editorState
+ * @param {Bool} update if true force change value and editor refresh
```
gbbr: Let's call this parameter `force`. It makes it obvious that it _forces_ the onChange operation. Maybe we can document it as such:

"_When true, forces an editor update regardless of whether the content has changed. This is used because currently it is impossible to detect changes happening solely on entity data. In Draft v0.11.0 we will be able to request all entities from the content and could compare them to get a more accurate result.
See https://draftjs.org/docs/api-reference-content-state.html#getentitymap_"

Then we can remove the comments further down in the code because this explains everything.

scripts/core/editor3/reducers/editor3.jsx
```diff
@@ -55,20 +55,32 @@ export const forceUpdate = (state) => {
...
  * @return {Object} returns new state
  * @description Handle the editor state has been changed event
  */
-export const onChange = (state, newState) => {
+export const onChange = (state, newState, update = false) => {
```
gbbr: Please add a line here:
```js
// TODO(x): Remove `force` once Draft v0.11.0 is in
```
I have other TODOs like this in editor3, making it easier for us to update later on.

scripts/core/editor3/reducers/table.jsx
```diff
@@ -162,10 +162,9 @@ const processCells = (state, fn) => {
         data: fn(cells, numCols, numRows, i, j, withHeader)
     });
```
gbbr: Let's do the same here:
```js
const entityDataHasChanged = true;
```
And use it as the last param to `onChange`.

scripts/core/editor3/reducers/editor3.jsx
```diff
@@ -165,15 +177,21 @@ const changeImageCaption = (state, {entityKey, newCaption, field}) => {
     const {editorState} = state;
     const contentState = editorState.getCurrentContent();
     const entity = contentState.getEntity(entityKey);
-    let {img} = entity.getData();
+    let {media} = entity.getData();
+
+    const update = field === 'Title' ?
```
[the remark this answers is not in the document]
ioanpocol: Can be the situation when you just use the arrow keys for navigation and you will get the save button activated.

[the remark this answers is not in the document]
ioanpocol: Actually no.

scripts/core/editor3/reducers/table.jsx
ioanpocol: fixed

scripts/core/editor3/reducers/editor3.jsx
```diff
@@ -165,15 +182,19 @@ const changeImageCaption = (state, {entityKey, newCaption, field}) => {
     const {editorState} = state;
     const contentState = editorState.getCurrentContent();
     const entity = contentState.getEntity(entityKey);
-    let {img} = entity.getData();
+    let {media} = entity.getData();
+    let entityDataHasChanged = true;
```
gbbr: Please can you put this after line 194, exactly after the entity data changes. It makes more sense that way.

scripts/core/editor3/reducers/table.jsx
```diff
@@ -152,6 +152,7 @@ const processCells = (state, fn) => {
         return state;
     }
 
+    const entityDataHasChanged = true;
```
gbbr: Same with this one, please put it after line 164. If it stays here, it's not quite accurate because the entity data has not _yet_ changed.

scripts/core/editor3/reducers/editor3.jsx
ioanpocol: done

PR #1902, opened 2017-11-10
fix(media fields): Fix the show of audio player on custom media fields

ioanpocol:


scripts/apps/authoring/views/item-association.html
```diff
@@ -11,6 +11,11 @@
                 ng-repeat="(key, rendition) in related.renditions"
                 ng-if="isVideo(rendition)" html5vfix>
     </video>
+    <audio ng-if="related && related.type === 'audio'" controls="controls">
+        <source vsrc="{{ rendition.href }}"
```
petrjasek: not sure what `vsrc` is..

scripts/apps/authoring/views/item-association.html
```diff
@@ -7,10 +7,15 @@
         </button>
     </div>
     <video ng-if="related && related.type === 'video'" controls="controls">
-        <source vsrc="{{ rendition.href }}"
+        <source src="{{ rendition.href }}"
                 ng-repeat="(key, rendition) in related.renditions"
                 ng-if="isVideo(rendition)" html5vfix>
     </video>
+    <audio ng-if="related && related.type === 'audio'" controls="controls">
+        <source src="{{ rendition.href }}"
```
petrjasek: I think it would be better to use [ngSrc](https://docs.angularjs.org/api/ng/directive/ngSrc)

petrjasek:
it seems it actually should be `vsrc`, it's used by this html5vfix directive, sorry

PR #1905, opened 2017-11-10
feat(editor3): Read annotation types from annotation_types vocabulary

ioanpocol:


scripts/core/editor3/components/annotations/AnnotationInput.jsx
```diff
@@ -62,21 +63,33 @@ export class AnnotationInput extends Component {
     }
 
     componentDidMount() {
+        const annotationTypes = ng.get('metadata').values.annotation_types;
```
gbbr: Is there any reason why you are doing this here? I was thinking it would make more sense to do it in the constructor. It feels odd to set the state _again_ after the component mounts. Let's just start with the state already prepared before this happens.

scripts/core/editor3/components/annotations/AnnotationInput.jsx
```diff
@@ -15,7 +16,7 @@ export class AnnotationInput extends Component {
     constructor(props) {
         super(props);
 
-        this.state = {body: '', type: 'regular'};
+        this.state = {body: '', annotationTypes: []};
```
gbbr:
```js
this.state = {
    body: '',
    annotationTypes: ng.get('metadata').values.annotation_types || [],
    type: ''
};
```

scripts/core/editor3/components/annotations/AnnotationInput.jsx
```diff
@@ -62,21 +63,33 @@ export class AnnotationInput extends Component {
...
 
     render() {
         const {onCancel} = this.props;
-        const {type} = this.state;
+        const {type, annotationTypes} = this.state;
 
         return (
             <div className="annotation-input">
                 <Dropdown open={true}>
                     <div className="sd-line-input sd-line-input--is-select">
```
gbbr: Is it possible to have no annotation types set in the CV? If yes, then I'm thinking we should hide this `div` when that happens.

PR #1906, opened 2017-11-13
duplication of sd-toggle directive removed

tomaszrondio:


scripts/core/ui/ui.js
```diff
@@ -1201,7 +1176,7 @@ export default angular.module('superdesk.core.ui', [
     }])
 
     .directive('sdShadow', ShadowDirective)
-    .directive('sdToggleBox', ToggleBoxDirective)
+    // .directive('sdToggleBox', ToggleBoxDirective)
```
gbbr: Why is this commented out? Just remove it please. We shouldn't commit dead code.

tomaszrondio: true. We are having pressure here before presentation :)

gbbr:
It would be really cool if you could squash (after you rebase on master) and rephrase your commit a bit better, for example:
```
core/ui: remove duplicate sd-toggle directive
```
Or at the very least remove the commit called "merge"

PR #1908, opened 2017-11-14
SDESK-2057 Enable removing custom fields from Metadata settings

mugurrus:


scripts/apps/vocabularies/controllers/VocabularyConfigController.js
```diff
@@ -78,6 +80,32 @@ export function VocabularyConfigController($scope, $route, $routeParams, vocabul
         $rootScope.$broadcast('vocabularies:updated');
     };
 
+    $scope.remove = (vocabulary) => {
```
gbbr: This function is extremely complicated, JavaScript is a difficult enough language as it is, so it'd be nice if you could break it into smaller functions and explain what it does. 
```js
// remove is the UI callback for deleting a vocabulary entry
$scope.remove = (vocabulary) => {
	modal.confirm(gettext('Please confirm you want to delete the vocabulary.'))
		.then(() => api.remove(vocabulary, {}, 'vocabularies'))
		.then(vocabularyRemoved(vocabulary), errorRemoving);
};

// vocabularyRemoved updates the UI after the given vocabulary has been removed via the API.
const vocabularyRemoved = (vocabulary) =>
	() => {
		_.remove($scope.vocabularies, (v) => v._id === vocabulary._id);
		$scope.reloadList();
		notify.success(gettext('Vocabulary was deleted.'));
	};

// errorRemoving alerts the user of an error that occurred while attempting to remove a vocabulary.
const errorRemoving = (response) => {
	const issues = response.data._issues;
	if (!issues) {
		notify.error(gettext('Error removing the vocabulary.'));
	} else if (issues._message) {
		notify.error(issues._message);
	} else if (issues.content_types) {
		const contentTypes = _.reduce(issues.content_types, (k, v) => result ? `${k}, ${v.label}` : v.label, null);

		notify.error(gettext('The vocabulary is used in the following content types:') + ' ' + contentTypes);
	}
};
```
Also, why does `remove` have to be on the `$scope`?

PR #1909, opened 2017-11-15
fix(editor3): Fix the delete of unordered list that have a table abov…

ioanpocol:
…e it


scripts/core/editor3/components/Editor3.jsx
```diff
@@ -141,7 +141,21 @@ export class Editor3Component extends React.Component {
 
         let newState;
 
-        if (command === 'soft-newline') {
+        if (command === 'backspace') {
```
gbbr: Would you mind turning this into a `swith` case?

PR #1915, opened 2017-11-17
SDESK-2041 As a user I want to have the main left navigation toolbar in the Planning component

mugurrus:


petrjasek:
seems like this is not doing what it says in the title

PR #1916, opened 2017-11-20
fix(ngtemplates): Fixed issue with ngtemplates on windows

pavlovicnemanja:
There is issue with adding external apps on windows platform because of slash (not recognizing `\`), so I modified gen-apps task a little.

Now it test if it is windows, then it convert all `\` to `/`.

### DIFF
--- a/tasks/options/ngtemplates.js
+++ b/tasks/options/ngtemplates.js
@@ -78,7 +78,17 @@ module.exports = {
                     return 'export default [];\r\n';
                 }
 
-                let abs = (p) => path.join(process.cwd(), 'node_modules', p);
+                // abs returns the correct absolute path to be used with a require call. On Windows,
+                // it correctly converts backslash (\) characters to forward-slash (/).
+                let abs = (p) => {
+                    let abspath = path.join(process.cwd(), 'node_modules', p);
+
+                    if (/^win/.test(process.platform)) {
+                        return abspath.replace(/\\/g, '/');
+                    }
+                    return abspath;
+                };
+
                 let data = 'export default [\r\n\trequire("' + abs(paths[0]) + '").default.name';
 
                 for (var i = 1; i < paths.length; i++) {


tasks/options/ngtemplates.js
```diff
@@ -78,7 +78,14 @@ module.exports = {
                     return 'export default [];\r\n';
                 }
 
-                let abs = (p) => path.join(process.cwd(), 'node_modules', p);
+                let abs = (p) => {
+                    let abspath = path.join(process.cwd(), 'node_modules', p);
+                    if (/^win/.test(process.platform)) {
```
gbbr: This should never happen. `path.join` will always use the correct OS separator based on the NodeJS [documentation](https://nodejs.org/api/path.html#path_path_join_paths). In case it doesn't, there is a [path.normalize](https://nodejs.org/api/path.html#path_path_normalize_path) command which does the same thing as your code, so perhaps this suggestion might work:
```js
let abspath = path.join(process.cwd(), 'node_modules', path.normalize(p));
```

petrjasek: I think it might come from `process.cwd()` and not `path.join()`

petrjasek: I think the issue is actually different, the path is alright for windows, but it doesn't work with later `require` call. there it expects unix paths but it gets windows ones

gbbr: I would've assumed it was probably bad user input (`p`), but even if it comes from `process.cwd`, we can still use `path.normalize`.

It's best that @pavlovicnemanja checks.

gbbr: > I think the issue is actually different, the path is alright for windows, but it doesn't work with later require call. there it expects unix paths but it gets windows ones

Ah, I see. You might be right. That makes sense. In that case we should leave an inline comment explaining why this logic is there.

pavlovicnemanja: Hey, I tried with `path.normalize` or `path.resolve`, and both are returning same thing. Yes, I agree with @petrjasek issue is on `require` side. I will add inline comment.

gbbr: Thanks @pavlovicnemanja. Sorry, I didn't realise that. Perhaps something like this should cut it:
```js
// abs returns the correct absolute path to be used with a require call. On Windows, 
// it correctly converts backslash (\) characters to forward-slash (/).
```

pavlovicnemanja:
hey @petrjasek can we merge this? 🙂

PR #1917, opened 2017-11-21
fix(editor3): Change the debounce from 3000 to 500

ioanpocol:


scripts/apps/authoring/views/article-edit.html
```diff
@@ -300,7 +300,7 @@
          data-value="item.body_html"
          data-read-only="!_editable"
          data-on-change="autosave(item, 0)"
-         data-debounce="3000"
+         data-debounce="500"
```
gbbr: I'm not sure why this value was so big. I do know that it was added because on huge articles the editor was being slow. This should still be ok but it might be worth testing it.

PR #1919, opened 2017-11-21
[AUTHORING] [METADATA]: raise limit of authors to 200

jerome-poisson:
authors fetching was using the default max_results (i.e. 25), resulting
in trouble when a instance has more authors.
This patch is a quick fix to raise this limit to the maximum (200)
before checking a more clean solution which need discussion (should we
fetch all the pages, or use pagination?).

SDFID-195


jerome-poisson:
@karelpetrak this is a temporary fix, I need to check with team if it's a good idea to fetch all results for big number of authors (200+), or if we should use pagination instead (or if 200 is good enough for now).

PR #1930, opened 2017-11-27
fix(editor3): Fix the save of last change in editor3

ioanpocol:


scripts/core/editor3/directive.js
```diff
@@ -59,17 +59,21 @@ class Editor3Directive {
 
             /**
              * @type {Object}
-             * @description A JSON object representing the Content State of the Draft
-             * editor. When available, it is used to show content, using `convertFromRaw`.
-             * Either this, or value have to be set. Use this for most accurate behavior.
+             * @description Item which is being edited
              */
-            editorState: '=?',
```
petrjasek: won't this break annotations?

scripts/apps/authoring/views/article-edit.html
```diff
@@ -520,7 +521,8 @@
         data-item="item.extra[field._id]" data-html="true" data-limit="schema[field._id].maxlength"></span>
     <div sd-editor3
         data-single-line="field.field_options.single"
-        data-value="item.extra[field._id]"
```
petrjasek: this didn't work for assignment?

scripts/core/editor3/store/index.js
```diff
@@ -79,14 +86,20 @@ function onChange(content) {
  * use available HTML. If none are available, an empty ContentState is created.
  */
 function getInitialContent(props) {
-    // we have an editor state stored in the DB
-    if (props.editorState && typeof props.editorState === 'object') {
-        return applyInlineStyles(convertFromRaw(props.editorState)).contentState;
+    // we have an editor state stored in the DB for body field
+    if (props.field === 'body_html' && props.item.editor_state &&
```
petrjasek: why do it here and make editor do some field specific behavior?

petrjasek:
pls add some description what is the issue you've found

scripts/apps/authoring/views/article-edit.html
ioanpocol: The problem was that on angular on value was not updated with the last html from editor3. So I sent the item and the field name and the html is initialized/saved from/to item.

scripts/core/editor3/store/index.js
ioanpocol: The editor_state is used only for body_html. We can add a new parameter to specify if the editor_state is used or not.

ioanpocol:
As PR title specified the last change in editor3 was never saved. This problem appear because of some synchronize problem of directive bidirectional parameters.

scripts/core/editor3/store/index.js
petrjasek: ok so we could use the old editorState param and only set it for body_html right?

ioanpocol: I will use a boolean parameter for this.

PR #1938, opened 2017-11-30
fix(editor3): let editor handle annotations

petrjasek:
- keep body_html in sync with annotations

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,8 @@
   },
   "scripts": {
     "test": "grunt ci:travis",
-    "lint": "eslint --ext .js --ext .jsx scripts spec tasks *.js"
+    "lint": "eslint --ext .js --ext .jsx scripts spec tasks *.js",
+    "protractor": "protractor protractor.conf.js"
   },
   "engines": {
     "node": ">=6.6.0 <7.0.0",
--- a/scripts/core/editor3/html/index.js
+++ b/scripts/core/editor3/html/index.js
@@ -4,11 +4,12 @@ import {HTMLParser} from './from-html';
 /**
  * @name toHTML
  * @param {Object} contentState
+ * @param {Object} options
  * @description Converts DraftJS ContentState to HTML.
  * @returns {string} HTML
  */
-export function toHTML(contentState) {
-    return new HTMLGenerator(contentState).html();
+export function toHTML(contentState, options) {
+    return new HTMLGenerator(contentState).html(options);
 }
 
 /**
--- a/scripts/core/editor3/html/to-html/HTMLGenerator.js
+++ b/scripts/core/editor3/html/to-html/HTMLGenerator.js
@@ -1,6 +1,7 @@
 import {BlockEntityWrapper} from '.';
 import {BlockInlineStyleWrapper} from '.';
 import {AtomicBlockParser} from '.';
+import {Modifier, SelectionState, convertFromRaw} from 'draft-js';
 
 /**
  * @type {Object}
@@ -15,9 +16,9 @@ const BlockStyleTags = {
     'header-four': 'h4',
     'header-five': 'h5',
     'header-six': 'h6',
-    blockquote: 'quote',
+    blockquote: 'blockquote',
     'unordered-list-item': 'li',
-    'ordered-list-item': 'li'
+    'ordered-list-item': 'li',
 };
 
 /**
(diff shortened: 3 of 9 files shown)


package.json
```diff
@@ -55,6 +55,7 @@
     "d3": "3.5.17",
     "diff-match-patch": "^1.0.0",
     "draft-js": "0.10.2",
+    "draft-js-export-html": "^1.2.0",
```
gbbr: Can we not use the `HTMLGenerator` that we've created?

petrjasek:
yep we can use that. the real problem here though is the update of annotations and probably this should also handle updates in associations. do you see any downsides @gbbr ?

gbbr:
Please allow me some time and I will take a look later today. If I understand correctly, the problem you are solving here is rendering the annotations in the HTML output, right?

scripts/core/editor3/html/to-html/HTMLGenerator.js
```diff
@@ -169,11 +171,53 @@ export class HTMLGenerator {
...
+                const selectionState = SelectionState.createEmpty();
+                const contentStateWithEntity = this.contentState.createEntity(
+                    'ANNOTATION',
+                    'INMUTABLE',
+                    {id}
+                );
+
+                const entityKey = contentStateWithEntity.getLastCreatedEntityKey();
+
+                this.contentState = Modifier.applyEntity(
```
gbbr: This won't work because it may overwrite other entities (in our case links).

scripts/core/editor3/html/index.js
```diff
@@ -4,11 +4,12 @@ import {HTMLParser} from './from-html';
...
  * @name toHTML
  * @param {Object} contentState
+ * @param {bool} withAnnotations
  * @description Converts DraftJS ContentState to HTML.
  * @returns {string} HTML
  */
-export function toHTML(contentState) {
-    return new HTMLGenerator(contentState).html();
+export function toHTML(contentState, withAnnotations) {
+    return new HTMLGenerator(contentState).html(withAnnotations);
```
gbbr: I'm thinking that from an API point of view it might be more friendly and would make more sense to add a second param for options to the `HTMLGenerator` so that it looks something like:
```js
return new HTMLGenerator(contentState, opts).html();
```
Just in case more stuff gets added later, and it might be needed earlier than the `html()` call.

scripts/core/editor3/html/to-html/HTMLGenerator.js
petrjasek: yep, but not sure how it would work in the output anyhow. you can't have those 2 combined really, clicks would only work for 1

gbbr: Sure, but I think the suggestion in the review comment would work better, unless you see a problem with that. Also, if I remember correctly, the annotation gets triggered by clicking a small icon (I saw some design at some point) - so the icon would be separate from the link, making both available for clicking.

petrjasek: at first I was handling it in the content block, but you can have annotations going over multiple blocks which makes it hard to handle given the rendering per block. so that's why it seemed much cleaner using the entity

gbbr: Yes, but the annotation HTML only needs to show up _at the end_ of the actual selection, so you only care about the last character in the annotation. It doesn't matter where it starts and whether it's the same block or another. Take a look [here](https://dev.sourcefabric.org/browse/SDFID-79), there only needs to be an icon at the end of the annotation selection, you don't need to make the whole text clickable.

gbbr: I am pointing to the screen shots in [SDFID-79](https://dev.sourcefabric.org/browse/SDFID-79). Please excuse me if anything has changed while I was away or if I'm missing something.

petrjasek: I don't think this is in the requirements, the icon itself will be added by the other system, probably via css, so we need the `<span>`s around the annotated text

gbbr: Ok, that makes sense. But I still feel it's a bit unnatural to change the content state just for the sake of generating the HTML. We could still use the same pattern and open a span where an annotation starts and close it where it ends (or if it goes across blocks then when the block ends too).

I do understand that the implementation may be more complicated but it's more correct.

This is of course just my opinion so please feel free to chose whichever solution you like more...

gbbr: P.S. There is of course the link problem too. And who knows what other entities might show up in the future, so I'd advise against it.

petrjasek: imo it's problematic anyhow if we allow the links and annotations to be mixed
because then we could have html like:
```
<span annotation-id="1">some <a href="">link</span> and link continues</a>
```
or vice versa

gbbr: That's true. That's not very semantic HTML. It would work though, nevertheless. I will let you decide here. To me it seems a bit hacky to create those entities.

petrjasek: seems like this is not required (more info on superdesk-editorial slack channel)

ioanpocol:
@petrjasek Please rebase the PR.

petrjasek:
not now, it's feature freeze

petrjasek:
this is now obsolete as the assocations storage format will change

PR #1941, opened 2017-12-05
fix(editor3): Set empty string when the all is deleted from editor3

ioanpocol:


scripts/core/editor3/store/index.js
```diff
@@ -69,6 +69,9 @@ function onChange(content) {
     this.$scope.$apply(() => {
         this.editorState = convertToRaw(cleanedContent);
         this.value = toHTML(cleanedContent);
+        if (this.value === '<p></p>') {
```
petrjasek: imo this fix should be part of that `toHTML` function that's where it's added. maybe it can be avoided there

PR #1943, opened 2017-12-05
SDESK-1767 Replace the 'update' statement to 'duplicated' for duplicated items

mugurrus:


scripts/apps/search/components/fields/takekey.jsx
```diff
@@ -4,6 +4,10 @@ import PropTypes from 'prop-types';
 export function takekey(props) {
     const {gettextCatalog} = props.svc;
 
+    if (props.item.rewrite_of) {
```
gbbr: Would you mind converting these to JSX? Perhaps something like:

```js
export const takekey = ({item, svc}) => {
    let getString = svc.gettextCatalog.getString;

    if (item.rewrite_of) {
        return <span className="takekey" key="takekey">{getString('duplicated')}</span>;
    }

    if (item.anpa_take_key) {
        return <span className="takekey" key="takekey">{getString(item.anpa_take_key)}</span>;
    }

    return null;
}
```

scripts/apps/search/components/fields/takekey.jsx
```diff
@@ -4,6 +4,10 @@ import PropTypes from 'prop-types';
...
+        return React.createElement('span', {className: 'takekey', key: 'takekey'},
+            gettextCatalog.getString('duplicated'));
+    }
     if (props.item.anpa_take_key) {
         return React.createElement('span', {className: 'takekey', key: 'takekey'},
             gettextCatalog.getString(props.item.anpa_take_key));
```
gbbr: What does this function return if none of the conditions are met? Is it ok that it doesn't render anything? At least we should add a `return null`.

petrjasek:
seems like this does not replace the updated label, but adds another one.

gbbr:
I'm not sure how this component is thought out, but it seems a bit redundant to add a new component for every field where the only difference is a conditional and a label.

PR #1946, opened 2017-12-06
fix(editor3): Prevent the merge of editor state in backend by wrappin…

ioanpocol:
…g it in a list


gbbr:
This looks to me like something that needs to be addressed in the backend.

ioanpocol:
Yes, it can be but on backend there are multiple places where the content item is write or read. Here on frontend there is a single place where we read/save the editor content.

PR #1947, opened 2017-12-07
feat(vocabularies): Add support for vocabulary translation on article…

ioanpocol:
… metadata

### DIFF
--- a/scripts/apps/authoring/metadata/metadata.js
+++ b/scripts/apps/authoring/metadata/metadata.js
@@ -17,7 +17,10 @@ function MetadataCtrl(
                             _.get($scope, 'origItem.type') === 'text';
 
     metadata.initialize().then(() => {
-        $scope.metadata = metadata.values;
+        $scope.$watch('item.language', (language) => {
+            $scope.metadata = metadata.getVocabulariesLocale(metadata.values, language);
+        });
+
         return preferencesService.get();
     })
         .then(setAvailableCategories)
@@ -356,7 +359,7 @@ function MetaDropdownDirective($filter) {
                 }
             };
 
-            scope.$watch(':: list', () => {
+            scope.$watch('list', () => {
                 scope.values = _.keyBy(scope.list, 'qcode');
             });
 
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/metadata/metadata.js
```diff
@@ -17,7 +17,12 @@ function MetadataCtrl(
                             _.get($scope, 'origItem.type') === 'text';
 
     metadata.initialize().then(() => {
-        $scope.metadata = metadata.values;
+        $scope.metadata = metadata.getVocabulariesLocale(metadata.values, $scope.item.language);
```
petrjasek: imo this is not needed, it will be triggered via watch anyhow

scripts/apps/authoring/metadata/metadata.js
```diff
@@ -356,8 +361,9 @@ function MetaDropdownDirective($filter) {
                 }
             };
 
-            scope.$watch(':: list', () => {
+            scope.$watch('list', () => {
                 scope.values = _.keyBy(scope.list, 'qcode');
+                scope.$applyAsync();
```
petrjasek: why?

scripts/apps/authoring/metadata/metadata.js
```diff
@@ -650,6 +656,8 @@ function MetaTermsDirective(metadata, $filter, $timeout) {
                 scope.tree = tree;
                 scope.activeTree = tree.null;
                 scope.combinedList = _.union(scope.list, scope.item[scope.field] ? scope.item[scope.field] : []);
+
+                scope.$applyAsync();
```
petrjasek: why?

scripts/apps/authoring/metadata/metadata.js
ioanpocol: Is not updated the set value

scripts/apps/authoring/metadata/metadata.js
ioanpocol: Is not updated the set value

petrjasek: that's strange, if it's in watch callback, there should be digest in progress

PR #1954, opened 2017-12-11
[API] added getAll method

jerome-poisson:
added a generic getAll method to retrieve all pages of a query.
Use this method to retrieve all authors in authoring.

fix SDESK-2168
fix SDFID-195

### DIFF
--- a/scripts/apps/content-filters/services/ContentFiltersService.js
+++ b/scripts/apps/content-filters/services/ContentFiltersService.js
@@ -26,9 +26,7 @@ export function ContentFiltersService(api, $filter) {
         return api.remove(item);
     };
 
-    this.getAllFilterConditions = function(page, items) {
-        return _getAll('filter_conditions', page, items);
-    };
+    this.getAllFilterConditions = () => api.getAll('filter_conditions');
 
     this.getFilterSearchResults = function(inputParams) {
         // call api to get search results
@@ -36,9 +34,7 @@ export function ContentFiltersService(api, $filter) {
             .then(angular.bind(this, (resultSet) => resultSet._items));
     };
 
-    this.getAllContentFilters = function(page, items) {
-        return _getAll('content_filters', page, items);
-    };
+    this.getAllContentFilters = () => api.getAll('content_filters');
 
     this.saveContentFilter = function(orig, diff) {
         return api.save('content_filters', orig, diff);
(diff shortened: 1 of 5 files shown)


scripts/core/api/api-service.js
```diff
@@ -246,6 +246,31 @@ function APIProvider() {
             });
         };
 
+        Resource.prototype.getAll = function(params) {
+            var p = $q.when();
+            var deferred = $q.defer();
```
petrjasek: it's better to work with the promise you get from query, it would also handle errors in fetching and would reject the promise

scripts/core/api/api-service.js
```diff
@@ -246,6 +246,31 @@ function APIProvider() {
...
+
+            function _getAll(page = 1, items = []) {
+                this.query(Object.assign({max_results: 200, page: page}, params))
```
petrjasek: so here should be `return this.query(...`

scripts/core/api/api-service.js
```diff
@@ -246,6 +246,31 @@ function APIProvider() {
...
+                    .then((result) => {
+                        let pg = page;
+                        let merged = items.concat(result._items);
+
+                        if (result._links.next) {
+                            pg++;
+                            p = p.then(_getAll.call(this, pg, merged));
```
petrjasek: here `return _getAll.call(...`

scripts/core/api/api-service.js
```diff
@@ -246,6 +246,31 @@ function APIProvider() {
...
+                        } else {
+                            deferred.resolve(merged);
```
petrjasek: and here `return merged`

scripts/core/api/api-service.js
```diff
@@ -246,6 +246,31 @@ function APIProvider() {
...
+                        }
+                    });
+
+                return deferred.promise;
+            }
+
+            p = _getAll.call(this);
+            return p.then((res) => res);
```
petrjasek: this would be then `return _getCall.call(this);`

scripts/apps/content-filters/services/ContentFiltersService.js
```diff
@@ -26,20 +26,12 @@ export function ContentFiltersService(api, $filter) {
         return api.remove(item);
     };
 
-    this.getAllFilterConditions = function(page, items) {
-        return _getAll('filter_conditions', page, items);
```
petrjasek: not sure why not just put `api.getAll` here instead of modifying service api

jerome-poisson: I initially did by modifying `ContentFiltersService` api actually, but as a second though it seemed better to me to use directly getAll as it is as simple. I can change it again no pb.

jerome-poisson: changing the API is breaking current tests anyway, so I'll go back to first implementation.

PR #1961, opened 2017-12-15
SDANSA-95 Create CV for publishing destinations

mugurrus:


petrjasek:
this could be a custom field, any reason to put this to item schema?

mugurrus:
@petrjasek I tried putting it in extra but sd-meta-terms doesn't work this way

petrjasek:
I think it should be custom cv so it would be stored in subject field

PR #1962, opened 2017-12-17
Imp(Contacts): Some improvements in Contacts grid view

sjunaid:
- Includes Job title & Mobile fields in card view
- made facebook and website as link to open in new tab
- increased card dimension to include more information
- multiple values of email, mobile, phone is accomodated
- Increased no. of characters to display in Notes field.


scripts/apps/contacts/components/ContactHeader.jsx
```diff
@@ -1,17 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import {JobTitle} from 'apps/contacts/components/fields';
```
petrjasek: I think it's better to use relative imports if it's part of same package, like from `./fields` here, but not a big deal

PR #1971, opened 2017-12-19
fix(spellchecker): Fix the initialization of spellchecker on Editor3

ioanpocol:


scripts/core/spellcheck/spellcheck.js
```diff
@@ -431,56 +435,60 @@ function SpellcheckService($q, api, dictionaries, $rootScope, $location, _) {
...
+    this.isAuto = false;
+    this.runSpellchecker = runSpellchecker;
+    this.pushSettings = pushSettings;
+    var self = this;
 
     /**
      * Force spell ckecking
      */
-    function spellcheck() {
+    function runSpellchecker() {
```
petrjasek: why is the name changed here?

ioanpocol: I use also spellcheck service and I wanted to avoid any confusion or name clash.

PR #1974, opened 2017-12-21
fix(correction): Update dateline on correction

ioanpocol:
Reverts superdesk/superdesk-client-core#1973


ioanpocol:
AAP confirmed that dateline can be changed on corrections.

PR #1976, opened 2017-12-21
fix(editor3): fix 3 dots menu not working after item publishing

petrjasek:
menu was rendered but only for a few miliseconds, click handler
on HighlightsPopup removed it right away, because it was never removed
once registered.

SDESK-2241


scripts/core/editor3/components/HighlightsPopup.jsx
```diff
@@ -157,6 +159,12 @@ export class HighlightsPopup extends Component {
         setTimeout(() => this.props.highlight ? this.renderCustom() : this.unmountCustom(), 0);
     }
 
+    componentWillUnmount() {
+        if (this.rendered) {
```
gbbr: Does this throw an error if we remove the check?

PR #1984, opened 2017-12-27
fix(vocabulary): Refresh the selected items and the option lists when…

ioanpocol:
… the current language is changed


ioanpocol:
Moved from master to 1.9 branch

PR #1985, opened 2017-12-28
fix(vocabularies): Refactor the vocabulary translations and just find…

ioanpocol:
… on html the locale name of terms


petrjasek:
imo this means we will have to translate it for each output, would be better to store it using the current language. otherwise if translation changes meanwhile there could be something different in the output

PR #1987, opened 2017-12-28
SDANSA-109 Support for multiple items on custom media fields

mugurrus:


petrjasek:
pls add some description how it works

jerome-poisson:
Hi, I've tried the patch, it's working but I found a bug: if you drag and drop several files (tried with 3 files) at once, you see 3 panels to enter metadata as expected, but when saving only one file is visible.
If you want to add several files, you have to add them one by one.
Also it would be nice to activate multi-selection when uploading files, for now we can only select one at a time.

scripts/apps/authoring/authoring/directives/ItemAssociationDirective.js
```diff
@@ -278,15 +293,31 @@ export function ItemAssociationDirective(superdesk, renditions, config, authorin
...
+
+                                function editNextFile() {
+                                    if (imagesWithIds.length > 0) {
+                                        var imageWithId = imagesWithIds.shift();
+
+                                        scope.edit(imageWithId.image, imageWithId.id, editNextFile);
+                                    }
+                                }
+
+                                _.forEach(images, (image) => {
```
petrjasek: do we actually need user to edit all images right after uploading them?

ioanpocol: This was the flow when you uploaded one image

mugurrus: This was how the functionality was implemented. I guess we should ask if we want to change that, but I'd do that in another PR.

petrjasek: I see

PR #1988, opened 2017-12-28
fix(vocabularies): Refactor the vocabulary translations and just find…

ioanpocol:
… on html the locale name of terms


ioanpocol:
Copied from master

PR #1994, opened 2018-01-09
fix(authoring): Change the word counter on frontend to match the code…

ioanpocol:
… counter on backend


petrjasek:
pls add some test so we can see what's expected behaviour

ioanpocol:
Done

PR #1996, opened 2018-01-10
[AUTHORING] display dynamic word count in authoring header, instead o…

jerome-poisson:
…f value from backend

word count in header was displaying item.word_count, i.e. the value
stored in the backend, resulting in confustion for user. This patch
change it by displaying the same counter as the one on top of body.
sdWordCount directive has a new argument "countOnly" to only display the
counter (and not " words" after it).

fixes SDFID-194


ioanpocol:
@jerome-poisson  Please rebase in order to get the fix for npmtest

petrjasek:
I've restarted the test, that should be enough

PR #1999, opened 2018-01-10
feat(authoring): Add support for date custom field

ioanpocol:


ioanpocol:
Depends on https://github.com/superdesk/superdesk-core/pull/1141

PR #2003, opened 2018-01-11
Embed user object into /activity API responses in order to display avatar images

tomaskikutis:


ioanpocol:
Please delete from PR package-lock.json file.

tomaskikutis:
Embed user object into /activity API responses in order to display avatar images

PR #2006, opened 2018-01-11
[AUTHORING] Use backend settings to override editor note

jerome-poisson:
According to backend settings, editor note will be overriden or not.
If a template is specified in backend, it will be used, else the legacy
template is used.

SDFID-261

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringEmbeddedDirective.js
+++ b/scripts/apps/authoring/authoring/directives/AuthoringEmbeddedDirective.js
@@ -1,7 +1,7 @@
 import * as helpers from 'apps/authoring/authoring/helpers';
 
-AuthoringEmbeddedDirective.$inject = ['api', 'notify', 'gettext', '$filter', 'config'];
-export function AuthoringEmbeddedDirective(api, notify, gettext, $filter, config) {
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/directives/AuthoringEmbeddedDirective.js
```diff
@@ -1,14 +1,35 @@
...
+                 * @param {string} template - template to use (set in backend)
+                 */
+                let date = $filter('formatLocalDateTimeString')(scope.item.versioncreated, config.view.dateformat +
+                    ' ' + config.view.timeformat);
+
+                if (template == null) {
+                    // no template specified in backend, we use default one
+                    let slugline = scope.item.slugline ? '"' + scope.item.slugline + '" ' : '';
+
+                    scope.item.ednote = gettext('In the story ') + slugline + gettext('sent at: ') + date +
```
petrjasek: I know it was like that before, but imo would be better to use $interpolate so there is single string for translation:
```
scope.item.ednote = $interpolate(gettext('In the story {{ slugline }} sent at: {{ date }}\r\n\r\nThis is corrected repeat.'))({slugline, date});
```

jerome-poisson: yes it was like that before, but you are right, I will change it.

PR #2007, opened 2018-01-12
SDFID-260 Display translations for metadata values in preview

pablopunk:
I created a function just like [this one](https://github.com/superdesk/superdesk-client-core/blob/0238618032a55590d901377d0a4984aad836ab4c/scripts/apps/authoring/metadata/metadata.js#L842) and then exposed it for the preview html.

Also, I had to apply a `filter` to `item.subject` because the previous one was only extracting the property `name`, and I needed the actual object

### DIFF
--- a/scripts/apps/authoring/authoring/directives/FullPreviewDirective.js
+++ b/scripts/apps/authoring/authoring/directives/FullPreviewDirective.js
@@ -44,6 +44,19 @@ export function FullPreviewDirective(api, $timeout, config, content) {
                 }, 200, false);
                 return false;
             };
+
+            scope.getLocaleName = function(term) {
+                if (!term) {
+                    return 'None';
+                }
+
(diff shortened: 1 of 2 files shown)


pablopunk:
@petrjasek you mean like translate them already in the controller and not doing anything in the view?

petrjasek:
imo it should use translated name when you select a term in authoring, using some of those sd-meta- directives from https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/metadata/metadata.js

petrjasek:
sorry closed by mistake

petrjasek:
it was implemented like that initially, not sure why it was changed @ioanpocol ?

ioanpocol:
It was too complicated and didn't worked well. What we can do is to update it on backend when the item is saved.

PR #2010, opened 2018-01-12
fix(custom date): Fix the format for custom date fields

ioanpocol:


ioanpocol:
Depends on https://github.com/superdesk/superdesk-core/pull/1144

PR #2011, opened 2018-01-12
SDFID-105 Show correct timestamp on package item preview

pablopunk:
Date was always `few seconds ago` because the variable didn't exist and `data-from-now` was taking that as "0 secs" I guess.

I'm still not sure if the date I'm displaying now it's the one that we want, I think it's the time the package was created, not added to the current item

### DIFF
--- a/scripts/apps/packaging/views/sd-package-item-preview.html
+++ b/scripts/apps/packaging/views/sd-package-item-preview.html
@@ -21,8 +21,8 @@
     </div>
 
     <div class="package-item__item-creator">
-      Created <time sd-datetime data-date="item.firstcreated" data-from-now="true"></time>
-	 <span ng-if="userLookup[data.original_creator]" translate>by <b>{{userLookup[data.original_creator].display_name}}</b></span>
+      Created <time sd-datetime data-date="data.versioncreated" data-from-now="true"></time>
+   <span ng-if="userLookup[data.original_creator]" translate>by <b>{{userLookup[data.original_creator].display_name}}</b></span>
     </div>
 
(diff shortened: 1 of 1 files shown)


scripts/apps/packaging/views/sd-package-item-preview.html
```diff
@@ -21,8 +21,8 @@
     </div>
 
     <div class="package-item__item-creator">
-      Created <time sd-datetime data-date="item.firstcreated" data-from-now="true"></time>
-	 <span ng-if="userLookup[data.original_creator]" translate>by <b>{{userLookup[data.original_creator].display_name}}</b></span>
+      Created <time sd-datetime data-date="data.versioncreated" data-from-now="true"></time>
```
petrjasek: it should be `.firstcreated` otherwise it would be Last Updated time not Created

pablopunk: @petrjasek then I think I need to bring that property from [here](https://github.com/superdesk/superdesk-client-core/blob/eb288a0026678618d0b69bad02a1f6934cabd0a8/scripts/apps/packaging/services/PackagesService.js#L114) with `versioncreated: item.versioncreated`, but that will only work for new packages, not the ones already added.

pablopunk: Actually, there's no property `firstcreated`, there's one called `_created`

petrjasek: we should have it, imo the issue there was `item.` instead of `data.`

pablopunk: @petrjasek you're right, I spent a lot of time looking for how to bring that property but we had it already, I'll push it now

PR #2017, opened 2018-01-16
[AUTHORING] [WORD COUNT] synchronized word count algorithm with backe…

jerome-poisson:
…nd one

Word count should now be the same in backend and client, making numbers
coherent in differents places where it's used. In some cases the numbers
may still differ because backend and client ways to clean HTML are not
the same, but in most cases the current behaviour should be alright.

SDFID-194


petrjasek:
imo the best would be to add some tests on both client/server with same inputs so we're sure it's in sync

ioanpocol:
@jerome-poisson One unittest is failing

PR #2027, opened 2018-01-19
Fix text decoration(bold, italic, underline) in editor3

tomaskikutis:
SDESK-2322

### DIFF
--- a/scripts/core/editor3/reducers/editor3.jsx
+++ b/scripts/core/editor3/reducers/editor3.jsx
@@ -54,6 +54,26 @@ export const forceUpdate = (state) => {
     };
 };
 
+
+function noCharactersAreSelectedAndCursorIsInsideCommentOrAnnotation(editorState) {
+    var selection = editorState.getSelection();
+    var character = editorState
+        .getCurrentContent()
+        .getBlockForKey(selection.getAnchorKey())
+        .getCharacterList()
+        .get(selection.getAnchorOffset());
+
+    if (character === undefined) {
+        return false;
+    }
+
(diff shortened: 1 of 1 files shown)


scripts/core/editor3/reducers/editor3.jsx
```diff
@@ -54,6 +54,26 @@ export const forceUpdate = (state) => {
...
+        .get(selection.getAnchorOffset());
+
+    if (character === undefined) {
+        return false;
+    }
+
+    var characterStyle = character.getStyle();
+
+    return selection.getAnchorKey() === selection.getFocusKey()
+        && selection.getAnchorOffset() === selection.getFocusOffset()
```
petrjasek: this seems like [selection.isCollapsed()](https://draftjs.org/docs/api-reference-selection-state.html#iscollapsed)

tomaskikutis: Yeah, it seems so. Anyway, in the meantime, I've found a better solution, so I'll close this one.

scripts/core/editor3/reducers/editor3.jsx
```diff
@@ -54,6 +54,26 @@ export const forceUpdate = (state) => {
...
+        && (characterStyle.contains('COMMENT') || characterStyle.contains('ANNOTATION'));
```
petrjasek: here it could be [character.hasStyle](https://draftjs.org/docs/api-reference-character-metadata.html#hasstyle) so we stick with draftjs api

tomaskikutis:
Closed in favor of #2030

PR #2029, opened 2018-01-19
fix(vocabularies): Fixed issue with save button always active

pavlovicnemanja:
SDESK-2260 - Cancel button saves changes in metadata management

### DIFF
--- a/scripts/apps/vocabularies/views/vocabulary-config-modal.html
+++ b/scripts/apps/vocabularies/views/vocabulary-config-modal.html
@@ -172,7 +172,7 @@
 
         <div class="modal__footer">
             <button id="save-edit-btn" type="submit" class="btn btn--primary pull-right" ng-click="save()"
-                    ng-disabled="editForm.$invalid" translate>Save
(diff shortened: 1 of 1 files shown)


scripts/apps/vocabularies/views/vocabulary-config-modal.html
```diff
@@ -172,7 +172,7 @@ <h3 class="modal__heading" ng-show="!vocabulary._id"><span translate>Create</spa
 
         <div class="modal__footer">
             <button id="save-edit-btn" type="submit" class="btn btn--primary pull-right" ng-click="save()"
-                    ng-disabled="editForm.$invalid" translate>Save
+                    ng-disabled="!isDirty() || editForm.$invalid" translate>Save
```
petrjasek: there is also `editForm.$dirty` that should be set on first change

petrjasek: and also [editFrom.$pristine](https://docs.angularjs.org/api/ng/type/form.FormController#$pristine) which should be like !dirty

PR #2035, opened 2018-01-23
fix(editor3): fix annotations type select box styling

petrjasek:
`.sd-line-input__label` got overriden by `.main-article .field label` styles

SDESK-2329


petrjasek:
we should refactor `.main-articles` styles to use bem to avoid this, it might affect other components as well

PR #2040, opened 2018-01-29
SDFID-284 Labels on stories in a package

mugurrus:


scripts/apps/packaging/views/sd-package-item-preview.html
```diff
@@ -19,10 +19,12 @@
 	  <div ng-show=":: data.anpa_take_key">{{ :: data.anpa_take_key}}</div>
       <div ng-hide=":: data.headline || data.slugline || data.description_text" translate>Blank headline received</div>
     </div>
-
-    <div class="package-item__item-creator">
-      Created <time sd-datetime data-date="data.firstcreated" data-from-now="true"></time>
-   <span ng-if="userLookup[data.original_creator]" translate>by <b>{{userLookup[data.original_creator].display_name}}</b></span>
+    <div>
+      <span ng-if="label" class="package-item__item-creator">label: <span id="package-item-label"><b>{{label.name}}</b></span></span>
```
petrjasek: there should be `<span class="package-item-label">`

scripts/apps/packaging/views/sd-package-item-preview.html
```diff
@@ -19,10 +19,12 @@
 	  <div ng-show=":: data.anpa_take_key">{{ :: data.anpa_take_key}}</div>
       <div ng-hide=":: data.headline || data.slugline || data.description_text" translate>Blank headline received</div>
     </div>
-
-    <div class="package-item__item-creator">
-      Created <time sd-datetime data-date="data.firstcreated" data-from-now="true"></time>
-   <span ng-if="userLookup[data.original_creator]" translate>by <b>{{userLookup[data.original_creator].display_name}}</b></span>
+    <div>
+      <span ng-if="label" class="state-label state-in_progress">label: <span id="package-item-label"><b>{{label.name}}</b></span></span>
```
petrjasek: there is missing `translate` and there should be no id in `<span id="package-item-label"` because there can be multiple items in package with a label

scripts/apps/packaging/services/PackagesService.js
```diff
@@ -123,6 +123,34 @@ export function PackagesService(api, $q, archiveService, lock, autosave, authori
...
     };
 
+    this.setItemLabel = function(item, label) {
+        var pkg = authoringWorkspace.getItem();
+
+        _.forEach(pkg.groups, (group) => {
+            var ref = _.find(group.refs, {guid: item._id});
+
+            if (ref) {
+                ref.label = label ? label.qcode : null;
```
petrjasek: could we store label as `{qcode, name}` like we do for other cvs ?

scripts/apps/archive/index.js
```diff
@@ -198,6 +199,26 @@ angular.module('superdesk.apps.archive', [
                 groupLabel: gettext('Duplicate'),
                 groupIcon: 'copy'
             })
+            .activity('label', {
+                label: gettext('Set label in current package'),
+                priority: 30,
+                icon: 'label',
+                monitor: true,
+                disableInList: true,
```
petrjasek: this should be probabaly `list: false` to be consistent with other options like monitor and dropdown

scripts/apps/packaging/services/PackagesService.js
mugurrus: What if the label name changes? Should I scan the archive to replace the name everywhere?

scripts/apps/packaging/services/PackagesService.js
```diff
@@ -123,6 +123,34 @@ export function PackagesService(api, $q, archiveService, lock, autosave, authori
...
+                $rootScope.$broadcast('package:updateGroupRef', {ref: ref, group: group});
+            }
+        });
+    };
+
+    this.isSetItemLabel = function(item, label) {
```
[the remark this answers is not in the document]
mugurrus: To disable the already set label in the dropdown list.

PR #2042, opened 2018-01-30
feat(editor3) Remove media blocks with a button

pablopunk:
SDFID-152

Added a "delete button" to MediaBlock

Now there's a removeBlock action and reducer. The reducer will just take the `blockKey`, get all blocks as an array, remove the one with that key, and push the new content.

I was thinking maybe I could extend `MediaComponent` inside `blockRenderer.jsx` so the button would be just in one place, and we could use it for Tables/Embeds too.


scripts/core/editor3/reducers/toolbar.jsx
```diff
@@ -168,6 +170,25 @@ const updateImage = (state, {entityKey, media}) => {
...
+ * @name removeBlock
+ * @param {Object} data Contains the key from of block to remove
+ * @description Removes block from editor
+ */
+const removeBlock = (state, {blockKey}) => {
+    const {editorState} = state;
+    const contentState = editorState.getCurrentContent();
+
+    const blocksArray = contentState.getBlocksAsArray();
+    const newBlocksArray = blocksArray.filter((block) => block.getKey() !== blockKey);
```
tomaskikutis: Use [removeRange](http://facebook.github.io/draft-js/docs/api-reference-modifier.html#removerange) to remove a block. I've found a discussion mentioning issues with 

> redo/undo, next selectionState, removing entities, etc.

 when working with arrays. See https://github.com/facebook/draft-js/issues/168#issuecomment-274941419

pablopunk: @tomaskikutis actually that's [exactly](https://github.com/facebook/draft-js/issues/168#issuecomment-274937682) the discussion I based my PR on. I did try to use `removeRange` but it would throw many errors on render. I think I just don't know how to get the SelectionState correctly for the block I'm trying to remove.

tomaskikutis: Yeah, I had a hard time understanding how to get the selection state as well. In this case, you can create it manually using a constructor. Take a look at https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-drag-n-drop-plugin/src/modifiers/removeBlock.js
You will need to modify the function, remove the `else` and make sure the selection you create takes into account if the block is first, last or somewhere in the middle.

pablopunk: @tomaskikutis I saw also draft-js-plugins and I didn't try to look at how they do it 😭. Thanks Tomas! I will try this today.

scripts/core/editor3/reducers/toolbar.jsx
```diff
@@ -169,6 +171,35 @@ const updateImage = (state, {entityKey, media}) => {
...
+    const contentState = editorState.getCurrentContent();
+
+    const afterKey = contentState.getKeyAfter(blockKey);
+    const targetRange = new SelectionState({
+        anchorKey: blockKey,
+        anchorOffset: 0,
+        focusKey: afterKey,
+        focusOffset: 0
+    });
+    let newContentState = Modifier.setBlockType(
```
petrjasek: what does it do?

pablopunk: @petrjasek it just gets the selection state for that block and then sets that range to unstyled. After that it removes the block with the selection range ([source](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-drag-n-drop-plugin/src/modifiers/removeBlock.js#L30))

petrjasek: why it sets the range to unstyled?

PR #2043, opened 2018-01-30
Insert empty line before/after the embed only if the embed is the first/last item in the editor

tomaskikutis:
SDESK-2165

### DIFF
--- a/scripts/core/editor3/helpers/insertAtomicBlockWithoutEmptyLinesAroundIt.js
+++ b/scripts/core/editor3/helpers/insertAtomicBlockWithoutEmptyLinesAroundIt.js
@@ -13,7 +13,6 @@ var List = Immutable.List,
     Repeat = Immutable.Repeat;
 
 function getInsertionTarget(contentState, selectionState) {
-
     var afterRemoval = Modifier.removeRange(contentState, selectionState, 'backward');
     var targetSelection = afterRemoval.getSelectionAfter();
     var afterSplit = Modifier.splitBlock(afterRemoval, targetSelection);
--- a/scripts/core/editor3/helpers/insertAtomicBlockWithoutEmptyLinesAroundIt.js
+++ b/scripts/core/editor3/helpers/insertAtomicBlockWithoutEmptyLinesAroundIt.js
@@ -1,0 +1,96 @@
+import {
+    BlockMapBuilder,
+    CharacterMetadata,
+    ContentBlock,
+    Modifier,
+    EditorState,
+    genKey
+} from 'draft-js';
+
+const Immutable = require('immutable');
+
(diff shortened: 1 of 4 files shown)


scripts/core/editor3/helpers/insertAtomicBlockWithoutEmptyLinesAroundIt.js
```diff
@@ -0,0 +1,95 @@
...
+
+/*
+    insertAtomicBlockWithoutEmptyLinesAroundIt: function(
+        editorState: EditorState,
+        entityKey: string,
+        character: string
+    ): EditorState
+*/
+
+function insertAtomicBlockWithoutEmptyLinesAroundIt(editorState, entityKey, character) {
```
petrjasek: as much as I like descriptive names this seems a bit too much :wink:

scripts/core/editor3/helpers/insertAtomicBlockWithoutEmptyLinesAroundIt.js
```diff
@@ -0,0 +1,95 @@
...
+    var selectionState = editorState.getSelection();
+    var target = getInsertionTarget(editorState.getCurrentContent(), selectionState);
+    var asAtomicBlock = Modifier.setBlockType(target.contentState, target.selectionState, 'atomic');
+    var charData = CharacterMetadata.create({entity: entityKey});
+    var fragmentArray = [];
+
+    if (asAtomicBlock.getFirstBlock().getKey() === target.selectionState.getAnchorKey()) {
+        fragmentArray.push(new ContentBlock({
```
petrjasek: we keep annotations/comments data on first block atm, would this still work if there is new block added?

scripts/core/editor3/helpers/insertAtomicBlockWithoutEmptyLinesAroundIt.js
tomaskikutis: I wanted to make it clear it's not the same as the original `insertAtomicBlock` from `draft-js` helpers I based this one on. Do you suggest to keep the name the same?

petrjasek: I think `insertAtomicBlockWithoutEmptyLines` would be enough

scripts/core/editor3/helpers/insertAtomicBlockWithoutEmptyLinesAroundIt.js
tomaskikutis: I've tried to break it by inserting a table in various positions when having active highlights. Seems to work fine.
highlights

PR #2050, opened 2018-01-31
Separate application and development dependiences

tomaskikutis:


tomaskikutis:
When you do `npm install` both `dependencies` and `devDependencies` get downloaded. Why do you think it won't work?

petrjasek:
because when you do npm install in superdesk repo client folder it will only get core dependencies

tomaskikutis:
Do you mean when you run `npm install` in this folder https://github.com/superdesk/superdesk/tree/master/client ? Then, yes, it won't get `devDependencies` because `npm` expects that your dependencies are already built. I guess it's easier to revert this one than to fix the build process.

PR #2054, opened 2018-01-31
feat(suggestions): Add support to create add and delete suggestions

ioanpocol:


scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -0,0 +1,348 @@
...
+    let content;
+    let selection;
+    let newState;
+
+    newState = changeEditorSelection(editorState, -1, 0, false);
+    content = newState.getCurrentContent();
+    selection = newState.getSelection();
+
+    content = Modifier.removeRange(content, selection, 'forward');
+    newState = EditorState.push(newState, content, 'delete-suggestions');
```
petrjasek: the change types are given by draft - https://github.com/facebook/draft-js/blob/master/src/model/immutable/EditorChangeType.js

ioanpocol:
Paste, re-edit a suggestion in edit&suggestion modes and reducers tests will be added in next PR.

scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -0,0 +1,360 @@
...
+const createAddSuggestion = (state, {text, data}) => {
+    let {editorState} = state;
+    const selection = editorState.getSelection();
+    const length = selection.getEndOffset() - selection.getStartOffset();
+    let i;
+
+    if (length !== 0) {
+        // if text is selected, apply delete suggestion for every selected character
+        editorState = changeEditorSelection(editorState, length, 0, false);
+        for (i = 0; i < length; i++) {
```
tomaskikutis: Declare a local iterator instead of reusing the one in the upper scope. This is not a big deal in this case, but it's a good practice not to use variables from upper scopes when you don't have to.

scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -0,0 +1,360 @@
...
+    const selection = editorState.getSelection();
+    const length = selection.getEndOffset() - selection.getStartOffset();
+    let i;
+
+    if (length === 0) {
+        editorState = setDeleteSuggestionForCharacter(editorState, data);
+    } else {
+        // if text is selected, apply delete suggestion for every selected character
+        editorState = changeEditorSelection(editorState, length, 0, false);
+        for (i = 0; i < length; i++) {
```
tomaskikutis: Declare a local iterator instead of reusing the one in the upper scope. This is not a big deal in this case, but it's a good practice not to use variables from upper scopes when you don't have to.

scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -0,0 +1,360 @@
...
+const getKeyAndEntityAtOffset = (editorState, offset) => {
+    const content = editorState.getCurrentContent();
+    const selection = editorState.getSelection();
+    const types = ['DELETE_SUGGESTION', 'ADD_SUGGESTION'];
+    let tmpKey = null;
+    let key = null;
+    let entity = null;
+
+    tmpKey = getEntityKeyByOffset(content, selection, offset);
+    if (tmpKey) {
```
tomaskikutis: Use triple equals when the value inside `if` is not a boolean

scripts/core/editor3/reducers/utils/composite-entity.jsx
```diff
@@ -0,0 +1,135 @@
...
+    const entity = entityKey ? content.getEntity(entityKey) : null;
+    let data;
+
+    if (!entity || entity.get('type') === type) {
+        return Modifier.applyEntity(content, selection, null);
+    }
+
+    data = entity.getData();
+    if (entity.get('type') === 'COMPOSITE' && data[type]) {
+        if (data.length() > 2) {
```
tomaskikutis: What is `data`, why should it be `> 2`?

scripts/core/editor3/reducers/utils/composite-entity.jsx
```diff
@@ -0,0 +1,135 @@
...
+            return content.replaceEntityData(entityKey,
+                data.filter((crtType) => crtType !== type)
+            );
+        }
+
+        data.forEach((crtType, entityKey) => {
+            if (crtType !== type) {
+                // TODO: clean unused composite entities
```
tomaskikutis: Do a TODO? ;D

PR #2057, opened 2018-02-01
[INGEST] added Ritzau config panel

jerome-poisson:
SDESK-2328

### DIFF
--- a/scripts/apps/ingest/index.js
+++ b/scripts/apps/ingest/index.js
@@ -201,4 +201,8 @@ angular.module('superdesk.apps.ingest', [
             label: 'Wufoo Forms',
             templateUrl: asset.templateUrl('apps/ingest/views/settings/wufoo.html')
         });
+        ingestSources.registerFeedingService('ritzau', {
+            label: 'Ritzau Webservice',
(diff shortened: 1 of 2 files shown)


scripts/apps/ingest/index.js
```diff
@@ -201,4 +201,12 @@ angular.module('superdesk.apps.ingest', [
             label: 'Wufoo Forms',
             templateUrl: asset.templateUrl('apps/ingest/views/settings/wufoo.html')
         });
+        ingestSources.registerFeedingService('ritzau', {
+            label: 'Ritzau Webservice',
+            templateUrl: asset.templateUrl('apps/ingest/views/settings/ritzauConfig.html')
+        });
+        ingestSources.registerFeedingService('ntb_ritzau', {
```
petrjasek: is there anything ntb specific on the service?

jerome-poisson: just small stuff, the same that we do for other feeds (default category, utenriks or sport), but still it is a specific behaviour.

petrjasek: isn't it done with the parser?

jerome-poisson: Oh right, got confused between service and parser :-/ . Will fix that thanks!

PR #2058, opened 2018-02-06
feat (editor3): Remove embed block (SDFID-152)

pablopunk:
* Uses `react-redux` in `EmbedBlock`
* Fixes tests to fit previous change
* Adds remove button to embeds (+ css)

This should have been done in https://github.com/superdesk/superdesk-client-core/pull/2042, so it's the same feature.


ioanpocol:
please rebase

PR #2061, opened 2018-02-07
Preserve data while drag-and-droping an image to the body

tomaskikutis:
SDESK-1508

### DIFF
--- a/scripts/apps/search/constants.js
+++ b/scripts/apps/search/constants.js
@@ -104,7 +104,13 @@ export const CORE_PROJECTED_FIELDS = {
         'queue_state',
         'alt_text',
         'description_text',
-        'assignment_id'
+        'assignment_id',
+
+        // Added in order not to lose data while drag-and-droping an image to the body | SDESK-1508
+        'byline',
+        'copyrightholder',
+        'copyrightnotice',
+        'usageterms'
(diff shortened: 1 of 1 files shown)


petrjasek:
I think it might make sense to fetch the item from server on drop to get the data, instead of fetching it all the time

tomaskikutis:
Ok, I will try to do that. Anyway, why are we not fetching everything every time? It would prevent such bugs and it doesn't seem it would put significantly more load on the database if it's a single object in mongo anyway?

petrjasek:
it's there to limit the response size

tomaskikutis:
Does it matter if it's 20kb or 200kb?

tomaskikutis:
I agree, that it's nice to have smaller payloads, and use less data, but if we can't safely do that without causing bugs, that 200kb doesn't sound so bad anymore.

petrjasek:
thing is it's not fetched once, but all the time by all users

tomaskikutis:
Yeah, I know, but it still takes <300ms to fetch on a fast network. I guess the problems arise when internet connections aren't that fast.

petrjasek:
yep those fields won't be probably populated usually so atm it shouldn't be a big deal

PR #2066, opened 2018-02-08
feat(editor3): Add support for paste on suggesting mode (SDFID-279)

ioanpocol:


scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -77,24 +71,77 @@ const createDeleteSuggestion = (state, {data}) => {
...
+    let {editorState} = state;
+    const blockMap = content.getBlockMap();
+
+    editorState = deleteCurrentSelection(editorState, data);
+
+    blockMap.forEach((block) => {
+        if (block.getType() !== 'atomic') {
+            const text = block.getText();
+
+            for (let i = 0; i < text.length; i++) {
```
petrjasek: I think you could use here `block.characterList.forEach` and then you can check the styles using [CharacterMetadata.hasStyle](https://draftjs.org/docs/api-reference-character-metadata.html#hasstyle)

scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -77,24 +71,77 @@ const createDeleteSuggestion = (state, {data}) => {
...
+ * @description Add a new suggestion of type ADD based on content.
+ */
+const pasteAddSuggestion = (state, {content, data}) => {
+    const acceptedStyles = ['BOLD', 'ITALIC', 'UNDERLINE'];
+    let {editorState} = state;
+    const blockMap = content.getBlockMap();
+
+    editorState = deleteCurrentSelection(editorState, data);
+
+    blockMap.forEach((block) => {
```
petrjasek: I don't understand how this works - it seems like it removes the selected characters - ok - but then it set add suggestion for all non atomic blocks and every character there?

ioanpocol: Yes, the suggestion entity key is set for every character.

petrjasek: but how is this limited to only content which was pasted?

ioanpocol: Html is converted in an editor content structure that is received here(content parameter).

petrjasek: so content is a contentState with pasted content only?

ioanpocol: yes

PR #2067, opened 2018-02-08
SDFID-268 Add style button to show paragraph marks

pablopunk:
* Adds a new decorator for spaces that can be toggled within the new reducer
* Adds a class to custom `UnstyledBlock` to append the paragraph mark

> BTW new button to show marks doesn't exist yet in ui-framework icons, I'm waiting on it to be added.

This might not be the best way to do it, let me know what you guys think.

***

paragraphs


scripts/core/editor3/reducers/toolbar.jsx
```diff
@@ -223,6 +229,38 @@ const applyEmbed = (state, code) => {
...
 
+
+/**
+ * @ngdoc method
+ * @name toggleParagraphMark
+ * @param {Object} state
+ * @return {Object} returns new state
+ * @description Enable/Disable the paragraph marks
+ */
+const toggleParagraphMark = (state) => {
```
petrjasek: I think it would toggle it off when you toggle spellchecker after activating this

pablopunk: @petrjasek hmmm it does... why is that? 🤔

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.js
```diff
@@ -66,7 +66,7 @@ export function ContentProfileSchemaEditor(content, config) {
 
             scope.formattingOptionsEditor3 = [
                 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'ordered list', 'unordered list', 'quote', 'media', 'link',
-                'embed', 'underline', 'italic', 'bold', 'table'
+                'embed', 'underline', 'italic', 'bold', 'table', 'paragraph mark'
```
tomaskikutis: Since you also added support for displaying spaces, it might be good to rename 'paragraph mark' to 'invisibles' or something similar.

scripts/core/editor3/components/spaces/SpaceComponent.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const SpaceComponent = (props) => {
+    const spanProps = {...{}, ...props};
```
tomaskikutis: I checked babel repl and both `{...{}, ...props}` and `{ ...props}` generate the same output, so it seems you don't need `...{}`

scripts/core/editor3/reducers/toolbar.jsx
```diff
@@ -223,6 +229,38 @@ const applyEmbed = (state, code) => {
...
+    const {paragraphMark, editorState} = state;
+    const spellcheck = ng.get('spellcheck');
+    const decorators = [LinkDecorator];
+    const newParagraphMark = !paragraphMark;
+
+    if (state.spellcheckerEnabled || spellcheck.isAutoSpellchecker) {
```
tomaskikutis: Why do you need to know about the spellchecker here? Can't you retrieve all existing decorators, add yours and create a composite without explicitly listing the other ones?

scripts/core/editor3/reducers/toolbar.jsx
petrjasek: because [there it's also changing decorators](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/reducers/spellchecker.jsx#L61) but it has no idea about this new one, it should be probably somehow combined together

pablopunk:
On my last commit I deleted code in the reducer. Now it doesn't change decorators at all. `SpaceDecorator` is always rendered. It just displays the character when the paragraph marks are on, so performance in long articles is waaay better 💪🏻

Also when the spellchecker is toggled it still shows the invisibles

PR #2073, opened 2018-02-13
feat(suggestions): Disable the edition of suggestions on edit mode an…

ioanpocol:
…d allow the user to change only his suggestions in suggesting mode


scripts/core/editor3/components/Editor3.jsx
```diff
@@ -173,9 +174,25 @@ export class Editor3Component extends React.Component {
         case 'soft-newline':
             newState = RichUtils.insertSoftNewline(editorState);
             break;
+        case 'delete':
+            if (!this.allowEditSuggestion('delete')) {
```
tomaskikutis: I wish it were possible to use an abstraction for deletion like `handleBeforeInput`. Since we aren't using one, we are limited to supporting only two ways of removing text. For example when using `ctrl+x` the text is removed, but not marked as a deletion.

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -173,9 +174,25 @@ export class Editor3Component extends React.Component {
...
+
+            if (suggestingMode) {
+                onCreateDeleteSuggestion('delete');
+                return 'handled';
+            }
+
+            newState = RichUtils.handleKeyCommand(editorState, command);
+            break;
         case 'backspace': {
+            if (!this.allowEditSuggestion('backspace')) {
```
tomaskikutis: Are we allowing to edit suggestions when not in suggesting mode?

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -217,10 +234,48 @@ export class Editor3Component extends React.Component {
...
+                anchorOffset: selection.getStartOffset(),
+                focusOffset: selection.getEndOffset() + chars.length,
+                isBackward: false
+            });
+            let newContentState = editorState.getCurrentContent();
+            let newEditorState;
+
+            newContentState = Modifier.insertText(newContentState, selection, chars);
+            inlineStyle.forEach((style) => {
+                if (styles.indexOf(style) !== -1) {
```
tomaskikutis: You should exclude the styles you don't want instead of including the ones you want. If `strikethrough` or any other style gets added, we'd have a bug here.

scripts/core/editor3/components/handlePastedText.js
```diff
@@ -101,3 +108,54 @@ const atomicBlock = (data, entity) => new ContentBlock({
...
+ * @returns {Boolean} True if the current text don't contains a noneditable suggestion.
+ * @description Check if the current text don't contain a noneditable suggestion.
+ */
+export function allowEditSuggestion(action) {
+    const {suggestingMode, editorState} = this.props;
+    const content = editorState.getCurrentContent();
+    const selection = editorState.getSelection();
+    const types = ['DELETE_SUGGESTION', 'ADD_SUGGESTION'];
+
+    if (selection.getStartOffset() !== selection.getEndOffset()) {
```
tomaskikutis: You probably want to use `SelectionState.isCollapsed()` which also takes into account that a selection doesn't span multiple blocks. Do we support suggestions spanning multiple blocks by the way?

scripts/core/editor3/components/handlePastedText.js
```diff
@@ -101,3 +108,54 @@ const atomicBlock = (data, entity) => new ContentBlock({
...
+        for (let i = selection.getStartOffset(); i < selection.getEndOffset(); i++) {
+            const key = getEntityKey(content, getEntityKeyByOffset(content, selection, i), types);
```
tomaskikutis: Wouldn't a built-in method `ContentBlock.getEntityAt()` be simpler?

scripts/core/editor3/components/handlePastedText.js
```diff
@@ -101,3 +108,54 @@ const atomicBlock = (data, entity) => new ContentBlock({
...
+            }
+        }
+
+        return true;
+    }
+
+    const keyBefore = getEntityKey(content, getEntityKeyByOffset(content, selection, -1), types);
+    const keyAfter = getEntityKey(content, getEntityKeyByOffset(content, selection, 0), types);
+
+    return allowEditForKey(content, keyBefore, suggestingMode) && (action === 'backspace' || action === 'insert') ||
```
tomaskikutis: If every line can be wrapped in parentheses, do so. It will be more explicit what it is you are checking.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: Prevent to edit of suggestions in editing mode.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: On user interface we support only the list of styles from 'styles' list.

scripts/core/editor3/components/handlePastedText.js
ioanpocol: This is related to composite entity, I will refactor after the custom styles implementation is done.

scripts/core/editor3/components/Editor3.jsx
tomaskikutis: That's fair, but then try to use the same list so it doesn't go out of sync. For example there is a list at `/superdesk-client-core/scripts/core/editor3/components/toolbar/InlineStyleButtons.jsx` which has no connection to your `styles` variable.

scripts/core/editor3/components/handlePastedText.js
ioanpocol: Looks like there isn't support for multiple blocks suggestions. It will be fixed in other PR,  I created a task for this (SDNTB-472).

tomaskikutis: You can wait until I finish with multiple highlights, and if you make suggestions use that, you'd get multi-block suggestions by default.

scripts/core/editor3/components/Editor3.jsx
ioanpocol: Fixed ctrl-x issue.

scripts/core/editor3/components/handlePastedText.js
ioanpocol: OK

PR #2084, opened 2018-02-16
[SDESK-2199] - Contact Form - As a User, I would like to add a new Contact fr…

sjunaid:
…om the Event form.


scripts/apps/contacts/components/Form/ContactNumberInput.jsx
```diff
@@ -0,0 +1,112 @@
...
+        }
+        this.props.onChange(field, value);
+    }
+
+    render() {
+        const {value, field, remove, onChange, readOnly, usages} = this.props;
+
+        return (
+            <Row flex={true}>
+                <LineInput readOnly={readOnly}
```
petrjasek: imo we should introduce components for LineInputText, LineInputSelect etc. so it would handle the whole sd-line markup and generate the label which we can connect to the field

scripts/apps/contacts/components/Form/ContactNumberInput.jsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Row, LineInput, Input, SelectInput, Toggle, Label} from './';
```
petrjasek: would consider moving this to `core/components/form` so we can reuse this later on

scripts/apps/contacts/components/Form/ProfileDetail.jsx
```diff
@@ -0,0 +1,440 @@
...
+        const {svc, contact} = this.props;
+        const {metadata, $filter} = svc;
+
+        let stateNames = [];
+
+        if (metadata.values.contact_states) {
+            stateNames = $filter('sortByName')(metadata.values.contact_states);
+        }
+
+        this.setState({
```
petrjasek: why is this in `componentDidMount` ? seems like it's not related to dom

scripts/apps/contacts/constants.js
```diff
@@ -27,3 +27,34 @@ export const DEFAULT_LIST_CONFIG = {
     ]
 };
 
+export const KEYCODES = {
```
petrjasek: this could go to core

petrjasek:
ideally this would be 2 PRs - first adding the form components into core and the second with the contacts app. and imo we should move those form components to core

petrjasek:
regarding core components - idea is to move [planning components](https://github.com/superdesk/superdesk-planning/tree/master/client/components/UI) into core and use that when available, so the current components are temporary and should not be put into core/components.

PR #2085, opened 2018-02-16
SDESK-1983 | Hide keywords in metadata if they are not enabled in content profile

pablopunk:


petrjasek:
maybe we should remove that field from the sidebar, when it's in schema it's also in the autoring isn't it?

pablopunk:
@petrjasek what is the authoring 🤔

petrjasek:
keywords
it's there on the left - the field with focus - and in the sidebar

pablopunk:
@petrjasek oh gotcha. Yes you're right, that's what I was talking about in the meeting, but maybe someone uses the sidebar all the time even if we have it in the authoring already.

PR #2089, opened 2018-02-20
SDESK-1412 Trim html tags in open article headline

pablopunk:
Removing tags on save was not a good idea because we might need those tags on the headline for styling.

I just created an angular filter to trim html tags and used it on the headline at the bottom bar.

### DIFF
--- a/scripts/apps/authoring/authoring/helpers.js
+++ b/scripts/apps/authoring/authoring/helpers.js
@@ -1,3 +1,5 @@
+import {stripHtmlTags} from '../../../core/utils';
+
 export const CONTENT_FIELDS_DEFAULTS = Object.freeze({
     headline: '',
     slugline: '',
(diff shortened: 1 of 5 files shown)


scripts/core/filters/index.js
```diff
@@ -214,4 +214,8 @@ export default angular.module('superdesk.core.filters', [])
         var cleanedValue = value || '';
 
         return cleanedValue.replace('_', ' ');
-    });
+    })
+    .filter('trimHtmlTags', () => function(value) {
```
petrjasek: we have something defined already, maybe it should go into some `core/utils` and be used in both
https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/authoring/helpers.js#L99

pablopunk: @petrjasek okay, I'll do that. You mean creating a new `core/utils/index.js` and exporting a `stripHtmlRaw` function that I can use in the filter as well as the tests right?

petrjasek: yep, maybe just make it `core/utils.js` not sure if we need more now

scripts/apps/authoring/authoring/helpers.js
```diff
@@ -1,3 +1,5 @@
+import {stripHtmlTags} from '../../../core/utils';
```
petrjasek: you can import from `core/utils`

PR #2092, opened 2018-02-20
SDESK-1881 Load the JID field in the user profile when XMPP auth enabled only

mugurrus:


scripts/apps/users/directives/UserEditDirective.js
```diff
@@ -26,6 +26,10 @@ export function UserEditDirective(api, gettext, notify, usersService, userList,
             scope.dirty = false;
             scope.errorMessage = null;
 
+            api.query('client_config').then((result) => {
```
petrjasek: you should use [deployConfig.get](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/config.js#L43) so it can fetch the config just once

PR #2103, opened 2018-02-23
feat(highlights): Adapt highlights store to support suggestions

ioanpocol:


scripts/core/editor3/components/MultipleHighlights.jsx
```diff
@@ -119,70 +119,132 @@ export class MultipleHighlights extends React.Component {
...
+
+        if (applyChanges) {
+            this.props.onHighlightChange(newEditorState, this.exportState());
+        }
+
+        return newEditorState;
+    }
+
+    initHighlights(editorState) {
+        const applyChanges = editorState === undefined;
```
tomaskikutis: `editorState` shouldn't be able to be undefined at this point. If it is, you should have thrown an `Error` in the calling function. ~~I don't think the bool `applyChanges` has to exist in the first place. If this is to be stateless why wouldn't we want to always apply changes?~~ Ok, it seems `applyChanges` means a different thing than I thought. I though it's a flag for deciding if we want to push a new highlights state into the `editorState`. It seems like it's a flag if we want to call `this.props.onChange` which in turn would trigger a reducer. We want to have this configurable, but it shouldn't be checked by comparing `editorState` to `undefined`. Instead it should be passed as a parameter.

scripts/core/editor3/components/MultipleHighlights.jsx
```diff
@@ -119,70 +119,132 @@ export class MultipleHighlights extends React.Component {
...
+        };
+
+        const newEditorState = RichUtils.toggleInlineStyle(editorState, styleName);
+
+        return {newEditorState, newHighlightsState};
+    }
+
+
+    addHighlight(highlightType, highlightData, editorState) {
+        const {newEditorState, newHighlightsState, applyChanges} = this.initHighlights(editorState);
```
tomaskikutis: I think it'd be better if you used `this.getHighlightsStates` here since it's all `initHighlights` does.

scripts/core/editor3/components/MultipleHighlights.jsx
```diff
@@ -119,70 +119,132 @@ export class MultipleHighlights extends React.Component {
         return state;
     }
 
-    addHighlight(highlightType, highlightData) {
-        // NICE-TO-HAVE: remove associated data when highlights are removed by deleting characters in the editor
+    setHighlightsState(editorState, highlightsState) {
+        // TODO: save to first block
+        this.setState(highlightsState);
```
tomaskikutis: If it's stateless we can remove all instances of `this.state` or `this.setState`. So this function would simply return a new `editorState`

scripts/core/editor3/components/MultipleHighlights.jsx
```diff
@@ -119,70 +119,132 @@ export class MultipleHighlights extends React.Component {
...
-    removeHighlight(styleName) {
+            highlightsData: {
+                ...highlightsState.highlightsData,
+                [styleName]: highlightData
+            }
+        };
+
+        const newEditorState = RichUtils.toggleInlineStyle(editorState, styleName);
+
+        return {newEditorState, newHighlightsState};
```
tomaskikutis: `newHighlightsState` should be put inside the first block of the `editorState`, so we shouldn't return it here.

scripts/core/editor3/components/MultipleHighlights.jsx
```diff
@@ -119,70 +119,132 @@ export class MultipleHighlights extends React.Component {
...
+    getHighlightsStates(editorState) {
+        // TODO: get from first block
+        return this.state;
+    }
+
+    saveHighlights({newEditorState, newHighlightsState}, applyChanges) {
+        this.setHighlightsState(newEditorState, newHighlightsState);
+
+        if (applyChanges) {
+            this.props.onHighlightChange(newEditorState, this.exportState());
```
tomaskikutis: We don't need `exportState` anymore since we are storing it all in the first block of the `editorState`

scripts/core/editor3/components/MultipleHighlights.jsx
ioanpocol: Yes, this is the intention, because of this I added the TODO. For now I just added some code in order to keep the comments working. We can implement the set and get here or we can receive the functions in props.

tomaskikutis: I think the component is not that big and it would be better to perform the refactoring in one go, than to merge this half-baked PR. I'm not sure if that's what you intend to do.

scripts/core/editor3/components/MultipleHighlights.jsx
ioanpocol: The MultipleHighlights functions has editorState as last parameter. for now this parameter is not sent from comments re: code and it this case it will be taken from props. If we change this and add the editorState parameter we can also add the check.

scripts/core/editor3/components/MultipleHighlights.jsx
ioanpocol: The same code appear 4 times or more, so I created a function for it.

scripts/core/editor3/components/MultipleHighlights.jsx
ioanpocol: The newHighlightsState is returned to  addHighlight where the save function is called.

scripts/core/editor3/components/MultipleHighlights.jsx
ioanpocol: If we send all the time the editorState, yes, we can drop initHighlights.

scripts/core/editor3/components/MultipleHighlights.jsx
ioanpocol: Yes, the intention was to keep the code working, and have something to review and agree on.
And yes, we can do this in a single PR.

scripts/core/editor3/components/MultipleHighlights.jsx
```diff
@@ -114,105 +91,146 @@ export class MultipleHighlights extends React.Component {
...
+        if (applyChanges) {
+            this.props.onHighlightChange(newEditorState, newHighlightsState);
+        }
+
+        return newEditorState;
     }
 
-    addHighlight(highlightType, highlightData) {
-        // NICE-TO-HAVE: remove associated data when highlights are removed by deleting characters in the editor
+    initHighlights(editorState) {
```
tomaskikutis: `Init` is usually used as a constructor - called a single time for the lifecycle of the object. This is not how it's used here, so a better name might help IMO. How about `getHighlights`?

PR #2104, opened 2018-02-23
fix(monitoring): introduce preferences for swimlane config

petrjasek:
use user preferences for swimlane switch, can be overridden per session.

SDFID-296


petrjasek:
requires https://github.com/superdesk/superdesk-core/pull/1189

PR #2108, opened 2018-02-23
add coding conventions

tomaskikutis:


petrjasek:
imo this should be in the root folder, called [contributing](https://github.com/blog/1184-contributing-guidelines)

pablopunk:
I agree with @petrjasek, `/contributing.md` is more like a standard

tomaskikutis:
I thought I was putting it in the root :) Well, I moved it to `CONTRIBUTING.md`

pablopunk:
I guess this could be merge already, if someone has more suggestions they'll just submit another PR for discussion

PR #2115, opened 2018-03-01
fix(editor3): scroll authoring while dragging media item up/down

petrjasek:
SDESK-2420


ioanpocol:
Please delete yarn.lock file

petrjasek:
it's there on purpose

petrjasek:
@ioanpocol removed it after all, now it was obsolete

PR #2120, opened 2018-03-05
fix(suggestions): Add the changes associated with suggestions actions…

ioanpocol:
… as a single change on undo


scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -40,6 +40,27 @@ const toggleSuggestingMode = (state) => {
     };
 };
 
+/**
+ * @ngdoc method
+ * @name saveEditorStatus
+ * @param {Object} state
+ * @param {Object} editorState
```
pablopunk: this should be tmpEditorState

PR #2121, opened 2018-03-05
feat(carousel): Added carousel functionality

pavlovicnemanja:


package.json
```diff
@@ -86,6 +86,7 @@
     "moment-timezone": "0.5.14",
     "ng-file-upload": "12.2.13",
     "node-sass": "4.7.2",
+    "owl.carousel": "^2.2.0",
```
petrjasek: pls remove that `^`

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
```diff
@@ -0,0 +1,55 @@
...
+    return {
+        scope: {
+            items: '='
+        },
+        transclude: true,
+        templateUrl: 'scripts/apps/authoring/views/item-carousel.html',
+        link: function(scope, elem) {
+            let carousel;
+
+            $timeout(() => {
```
petrjasek: why is the timeout used there?

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
```diff
@@ -0,0 +1,55 @@
...
+                initCarousel();
+            }, 500);
```
petrjasek: there should be probably `}, 500, false);` to avoid digest

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
petrjasek: actually not sure if this part is needed given that there is that `scope.$watch('items')` later

package.json
pavlovicnemanja: 👍

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
```diff
@@ -0,0 +1,68 @@
...
+
+            function initCarousel() {
+                carousel = elem.find('.sd-media-carousel__content').owlCarousel({
+                    items: 1,
+                    autoHeight: true,
+                    loop: true
+                });
+            }
+
+            elem.find('.sd-media-carousel__thumb-strip').on('dragover', (event) => {
```
petrjasek: you should remove listeners on `$destroy`

PR #2124, opened 2018-03-06
Data for suggestions history

tomaskikutis:


scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -133,13 +135,37 @@ const pasteAddSuggestion = (state, {content, data}) => {
...
-const processSuggestion = (state, {selection}, accepted) => {
+const processSuggestion = (state, {suggestion}, accepted) => {
+    const {selection} = suggestion;
     const start = selection.getStartOffset();
     const end = selection.getEndOffset();
     let {editorState} = state;
     let style;
     let data;
 
+    const resolvedSuggestions = getCustomDataFromEditor(
```
ioanpocol: You can extract the code that deals with suggestion history in a separate function or even in a separate file.

PR #2126, opened 2018-03-07
Fix the global search for AAP Multimedia Provider

marwoodandrew:


marwoodandrew:
@petrjasek the search parameter pane selects based on the search_provider string.

petrjasek:
will try again later, I need it to support also search provider instances defined using new api https://github.com/superdesk/superdesk-core/blob/master/superdesk/search_provider.py

petrjasek:
btw @marwoodandrew do you have an instance configured in settings > search providers? I would prefer to keep the id in $location but also expose the type and use it in templates for those custom fields

PR #2129, opened 2018-03-07
Remove/generate highlightsStyleMap when saving/loading to/from server

tomaskikutis:
So it's possible to change styles for existing highlights

### DIFF
--- a/scripts/core/editor3/store/index.js
+++ b/scripts/core/editor3/store/index.js
@@ -63,7 +63,7 @@ export default function createEditorStore(props, isReact = false) {
 
 /**
  * @name onChange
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/store/index.js
```diff
@@ -67,15 +68,18 @@ export default function createEditorStore(props, isReact = false) {
  * current content states and updates the values of the host controller. This function
  * is bound to the controller, so 'this' points to controller attributes.
  */
-function onChange(content) {
+function onChange(contentState) {
```
petrjasek: docs `@param` name is not updated

PR #2130, opened 2018-03-08
fix(search): expose providerType to search scope

petrjasek:
and use it to toggle provider specific functionality


petrjasek:
hi @marwoodandrew can you pls check if that works for you any better?

marwoodandrew:
@petrjasek I've tested against our search provider and it works well! Thanks for that! (Even using the name  in the side nav, nice!)

PR #2133, opened 2018-03-08
New widget: Suggestions history

pablopunk:
Also refactored some things about inline-comments history.

This is how it's looking. We still need the [big icon for edit](https://wiki.sourcefabric.org/display/NR/Suggesting+mode?preview=/85590454/85590461/Suggesting-mode-07.png)

captura de pantalla 2018-03-08 a las 17 21 30


scripts/apps/authoring/track-changes/suggestions.js
```diff
@@ -0,0 +1,100 @@
...
+
+    users.forEach((user) => {
+        usersObj[user._id] = user;
+    });
+
+    return usersObj;
+}
+
+function getUsersFromAPI(api, query) {
+    return api.query('users', {where: JSON.stringify(query)});
```
petrjasek: instead of using api you could use [userList](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/users/services/UserListService.js#L15) service which has all users cached

scripts/apps/authoring/track-changes/suggestions.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function getTypeText(type) {
+    switch (type) {
+    case 'ADD_SUGGESTION':
+        return 'Added';
```
petrjasek: here it would need gettext for strings, you can get it injected in controller

petrjasek: or you can use the global one, I see you use translate in the template so it's only needed for string extraction

scripts/apps/authoring/track-changes/suggestions.js
pablopunk: In that service, how can I get the users of several `_id` at the same time?

scripts/apps/authoring/track-changes/suggestions.js
pablopunk: I use `{{ getTypeText(type) | translate }}` in the template. Is `gettext` in the controller a better approach?

scripts/apps/authoring/track-changes/suggestions.js
petrjasek: you can use `getAll` and do the filtering here

scripts/apps/authoring/track-changes/suggestions.js
petrjasek: for the translation it doesn't matter much, but we also need to extract strings for tools to translate (like transifex) and when it's scanning the source code it has no idea what `getTypeText(type)` will be, but it can get string from `gettext('Added')`

PR #2135, opened 2018-03-09
Rename "invisibles" to "formatting marks"

tomaskikutis:
SDFID-321

### DIFF
--- a/scripts/core/editor3/components/toolbar/StyleButton.jsx
+++ b/scripts/core/editor3/components/toolbar/StyleButton.jsx
@@ -35,7 +35,7 @@ const StyleTooltips = {
     'ordered list': gettext('Ordered list'),
     TH: gettext('Toggle Header'),
     suggestions: gettext('Toggle Suggestions Mode'),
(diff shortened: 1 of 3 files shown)


scripts/core/editor3/components/toolbar/StyleButton.jsx
```diff
@@ -35,7 +35,7 @@ const StyleTooltips = {
     'ordered list': gettext('Ordered list'),
     TH: gettext('Toggle Header'),
     suggestions: gettext('Toggle Suggestions Mode'),
-    invisibles: gettext('Toggle Non-Printing characters')
+    invisibles: gettext('Toggle formatting marks')
```
petrjasek: here it's using wrong key after the change above

tomaskikutis: It's still called `invisibles` in the code, the change is only for the UI. I've checked that hovering the icon displays the new title.

petrjasek: I see, my bad

PR #2144, opened 2018-03-13
Prevent publishing items with open suggestions

tomaskikutis:


scripts/apps/authoring/authoring/directives/SendItem.js
```diff
@@ -1,13 +1,14 @@
 import _ from 'lodash';
+import {hasUnresolvedSuggestions} from '../../../../core/editor3/helpers/highlights';
```
petrjasek: you should be able to import from `core/editor3/...`

scripts/apps/authoring/authoring/directives/SendItem.js
```diff
@@ -50,6 +51,30 @@ export function SendItem($q, api, desks, notify, authoringWorkspace,
...
+                            && fieldValue.length === 1
+                            && typeof fieldValue[0].blocks === 'object';
+
+                        if (isDraftjsField === false) {
+                            return false;
+                        }
+
+                        const rawState = fieldValue[0];
+
+                        return hasUnresolvedSuggestions(rawState);
```
petrjasek: maybe it would better api if we have `hasUnresolvedSuggestions(scope.item)` and handle editor state inside.

tomaskikutis: Yeah, I was thinking about this one myself. I want to keep `text-highlights` implementation modular and agnostic to our business logic e.g. the fields of the news item. Does it work if I put the logic to `authoring/helpers` as per 6346e6d ?

PR #2147, opened 2018-03-13
SDFID-297 Display inline comments selected text

pablopunk:
captura de pantalla 2018-03-13 a las 15 32 51


scripts/apps/authoring/track-changes/views/inline-comments-widget.html
```diff
@@ -19,6 +19,12 @@
                     </div>
                 </li>
             </ul>
+            <div ng-show="!!item.data.commentedText" class="commented-text">
+                <div>Selected text:</div>
```
petrjasek: missing translate

pablopunk: Every time... 😭

PR #2149, opened 2018-03-14
 Prevent extending a text-highlight when inserting text right after it

tomaskikutis:
SDESK-2491

### DIFF
--- a/scripts/core/editor3/components/Editor3Component.jsx
+++ b/scripts/core/editor3/components/Editor3Component.jsx
@@ -85,8 +85,6 @@ export class Editor3Component extends React.Component {
         this.onDragDrop = this.onDragDrop.bind(this);
         this.handleKeyCommand = this.handleKeyCommand.bind(this);
         this.handleBeforeInput = this.handleBeforeInput.bind(this);
-        this.handleBeforeInputCombined = this.handleBeforeInputCombined.bind(this);
-        this.handleBeforeInputHighlightsBound = handleBeforeInputHighlights.bind(null, this.props.onChange);
         this.keyBindingFn = this.keyBindingFn.bind(this);
         this.allowEditSuggestion = allowEditSuggestion.bind(this);
     }
@@ -184,18 +182,6 @@ export class Editor3Component extends React.Component {
         return getDefaultKeyBinding(e);
     }
 
-    handleBeforeInputCombined() {
-        const handlers = [this.handleBeforeInput, this.handleBeforeInputHighlightsBound];
-
-        for (let i = 0; i < handlers.length; i++) {
-            if (handlers[i].apply(this, arguments) === 'handled') {
-                return 'handled';
-            }
-        }
-
-        return 'not-handled';
-    }
-
     /**
      * @ngdoc method
      * @name Editor3#handleKeyCommand
@@ -273,8 +259,8 @@ export class Editor3Component extends React.Component {
      * at the end of content, any character added will continue to be part of the link.
      * This logic stops that behavior.
      */
-    handleBeforeInput(chars) {
-        const {editorState, onChange, suggestingMode, onCreateAddSuggestion} = this.props;
+    handleBeforeInput(chars, editorState) {
+        const {onChange, suggestingMode, onCreateAddSuggestion} = this.props;
 
         if (!this.allowEditSuggestion('insert')) {
             return 'handled';
@@ -332,6 +318,10 @@ export class Editor3Component extends React.Component {
 
             onChange(newEditorState);
 
+            return 'handled';
+        }
+
+        if (handleBeforeInputHighlights(this.props.onChange, chars, editorState) === 'handled') {
             return 'handled';
         }
 
@@ -413,7 +403,7 @@ export class Editor3Component extends React.Component {
                     <Editor editorState={editorState}
                         handleKeyCommand={this.handleKeyCommand}
                         keyBindingFn={this.keyBindingFn}
-                        handleBeforeInput={this.handleBeforeInputCombined}
+                        handleBeforeInput={this.handleBeforeInput}
                         blockRenderMap={blockRenderMap}
                         blockRendererFn={getBlockRenderer({svc: this.props.svc})}
                         customStyleMap={{...customStyleMap, ...this.props.highlightsManager.styleMap}}
--- a/scripts/core/editor3/components/Editor3Component.jsx
+++ b/scripts/core/editor3/components/Editor3Component.jsx
@@ -27,6 +27,7 @@ import UnstyledBlock from './UnstyledBlock';
 import UnstyledWrapper from './UnstyledWrapper';
 import {isEditorBlockEvent} from './BaseUnstyledComponent';
 import {acceptedInlineStyles} from '../helpers/inlineStyles';
+import {handleBeforeInputHighlights} from '../helpers/highlights';
 
 const VALID_MEDIA_TYPES = [
     'application/superdesk.item.picture',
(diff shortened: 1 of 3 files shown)


scripts/core/editor3/helpers/highlights.js
```diff
@@ -486,6 +486,33 @@ export function getRangeAndTextForStyle(editorState, style) {
     };
 }
 
+export function handleBeforeInputHighlights(onChange, chars, editorState) {
```
pablopunk: This function is missing `@ngdoc`, I think it's nice to put it when we export a function

scripts/core/editor3/components/Editor3Component.jsx
```diff
@@ -181,6 +184,18 @@ export class Editor3Component extends React.Component {
         return getDefaultKeyBinding(e);
     }
 
+    handleBeforeInputCombined() {
+        const handlers = [this.handleBeforeInput, this.handleBeforeInputHighlightsBound];
+
+        for (let i = 0; i < handlers.length; i++) {
+            if (handlers[i].apply(this, arguments) === 'handled') {
+                return 'handled';
```
pablopunk: This is returning `handled` just for one handler, and will not execute the next handlers, is that correct?

scripts/core/editor3/helpers/highlights.js
tomaskikutis: I only put these for more complex functions. Do you use any tooling which depends on `@ngdoc`?

scripts/core/editor3/components/Editor3Component.jsx
tomaskikutis: Right, if any handler claims to have handled the event, we don't execute others.

scripts/core/editor3/helpers/highlights.js
pablopunk: not really, but in case someone does

scripts/core/editor3/components/Editor3Component.jsx
ioanpocol: When we add a new handleBefore function we need to take a look at all other handleBefore function to know where to add the new function in the list. Also can be situations when this kind of ordering is just not working.
It will be better if we extract some code from  handleBeforeInput as functions and keep here the logic when we call different handlers.

tomaskikutis: > It will be better if we extract some code from handleBeforeInput as functions and keep here the logic when we call different handlers.

Sounds the same as I've implemented, except that I hadn't refactored any code from the original `handleBeforeInput`. You can extract it, add it to the list of handlers and it'd be the same. You'd still need to keep track of ordering of handlers even if it was a single function.

It would be an improvement if we could forget about ordering. We could do that, by changing how a handler works. So instead of
```
handleBeforeInput?: (chars: string, editorState: EditorState) => DraftHandleValue
```

we could have
```
handleBeforeInput?: (chars: string, editorState: EditorState) =>
    CombineHandlers(Array<Handler>) => DraftHandleValue
```
```
Handler: (chars: string, editorState: EditorState) =>
    {
        chars: string,
        editorState: EditorState
        handled: DraftHandleValue
    }
```

So we would pipe the output of one handler to the input of another, and in the end would return 'handled' if at least on handler handled the event.

ioanpocol: "You'd still need to keep track of ordering of handlers even if it was a single function."
Yes, but you don't have to know what is doing every handler when you add a new one, you just manage the conditions when a handler should be called. Usually only one should be called.

"So we would call all handlers pipe the output of one handler to the input of another, and in the end would return 'handled' if at least on handler handled the event."
Even now there are cases when only one handler should be called even if multiple handlers return 'handled'. It is too error prone to prevent the handlers to clash between them.

ioanpocol: Yes, the current situation with those 2 handlers is not working, they already clashed.
And  handleBeforeInput case is not the same as render. The render of components don't clash between them.

tomaskikutis: I'm afraid I'm not understanding you. When I asked
>Do you see cases where it wouldn't work this way?

I was asking about the proof of concept I described when all handlers being called, and all can transform the `editorState` and pipe it to next handlers.

You mentioned that 2 handlers that are in place in this PR have already clashed. Could you point me to the exact place where it's not working as it is supposed to?

ioanpocol: Just reverse the order of the handler in the list and some features will don't work well. Your proposal don't solve this. This is the problem I pointed to.

Not related to your proposal, a part of the original handler is doing the same thing like the new one. We have to eliminate this kind of duplicate code.

tomaskikutis: Well, in the current implementation of this PR, handlers must be ordered, so it works according to the design.

Regarding my proposal, could you give an example where it wouldn't work?

ioanpocol: "Well, in the current implementation of this PR, handlers must be ordered, so it works according to the design."
That is the problem, handlers must be ordered. Here is my previous comment that explain why:

When we add a new handleBefore function we need to take a look at all other handleBefore function to know where to add the new function in the list. Also can be situations when this kind of ordering is just not working.

"Regarding my proposal, could you give an example where it wouldn't work?"
I already responded to this in my previous comment:
"Just reverse the order of the handler in the list and some features will don't work well. Your proposal don't solve this. This is the problem I pointed to."

tomaskikutis: > When we add a new handleBefore function we need to take a look at all other handleBefore function to know where to add the new function in the list. Also can be situations when this kind of ordering is just not working.

I think we are stuck in a circular reasoning. You are saying, that handlers need to be ordered, because you >need to take a look at all other handleBefore function(...)

What I'm asking is why do you need to look at other functions? What I'm claiming with my proposal is that you don't and that handlers would work in any order. I might be wrong here since it's not an easy problem, but I'm interested if we know an exact use case where it wouldn't work?

ioanpocol: Sorry I don't see any circular reasoning here.

"What I'm asking is why do you need to look at other functions? "
Because currently the order of handler matters when you add a new handler you have to take a look at all others to don't brake the code.

'but I'm interested if we know an exact use case where it wouldn't work?'
For current case with those 2 handlers the solution is not working. If I want to add a new handler I have to look at the other 2 to be sure that I don't brake the code.

tomaskikutis: > Because currently the order of handler matters(...)
> For current case(...)

I agree, that in the current implementations, the order does matter. The reason it matters in my 2-handler function, is because I return after the first 'handled' response, and don't even call other handlers. The reason it matters in the single function implementation, is that function immediately returns after calling a handler as well.

Why would it matter in a situation, where every handler is called and has a chance to transform `editorState`?

petrjasek: > It would be an improvement **if we could forget about ordering**.

seems like we can't forget about the ordering. is that right?

tomaskikutis: I gave it another thought, and I think in theory we can't since there might be one handler which replaces letter `a` with `z` and another handler which replaces `a` to `y`, so they can't work at the same time.

I guess let's go with Ioan's solution then as it might be more clear to read since reading an array of handlers doesn't imply order.

I'll refactor it tomorrow :)

scripts/core/editor3/helpers/highlights.js
```diff
@@ -518,6 +518,38 @@ function addCommentsForServer(editorState) {
     return setCustomDataForEditor(editorState, editor3DataKeys.__PUBLIC_API__comments, comments);
 }
 
+/**
+ * @ngdoc method
+ * @name handleBeforeInputHighlights
+ * @description prevents inheriting of highlight styles
+ */
+export function handleBeforeInputHighlights(onChange, chars, editorState) {
```
ioanpocol: On handleBeforeInput at line 272 (https://github.com/tomaskikutis/superdesk-client-core/blob/e61169293cb6cf5addc8519d1f9960a9f78aef15/scripts/core/editor3/components/Editor3Component.jsx#L272) starts a code that seams to do the same thing like the code on this function. Can you refactor the code from handleBeforeInput  and add a call to this function.

PR #2151, opened 2018-03-14
Support tracking changes of multiple editors

pablopunk:
If in the future we have comments/suggestions in other fields like Title, etc, we can show the history for this fields also.

### DIFF
--- a/scripts/core/editor3/helpers/editor3CustomData.js
+++ b/scripts/core/editor3/helpers/editor3CustomData.js
@@ -42,10 +42,18 @@ export function setCustomDataForEditor(editorState, key, value) {
 }
 
 function getDataFromArrayState(editorStateArr, key) {
-    const editorState = editorStateArr[0];
-    const firstBlock = editorState.blocks[0];
+    let data = [];
 
-    return firstBlock.data[key];
+    // for each editor state we collect all data as one single array
+    editorStateArr.forEach((editorState) => {
+        const firstBlock = editorState.blocks[0];
+
+        if (firstBlock.data[key]) {
+            data = [...data, ...firstBlock.data[key]];
+        }
+    });
+
+    return data;
 }
 
 export function getCustomDataFromEditor(editorState, key) {


scripts/core/editor3/helpers/editor3CustomData.js
```diff
@@ -42,10 +42,18 @@ export function setCustomDataForEditor(editorState, key, value) {
 }
 
 function getDataFromArrayState(editorStateArr, key) {
-    const editorState = editorStateArr[0];
-    const firstBlock = editorState.blocks[0];
+    let data = [];
 
-    return firstBlock.data[key];
+    // for each editor state we collect all data as one single array
+    editorStateArr.forEach((editorState) => {
```
ioanpocol: Do we already save the editorState for every field that uses editor3?

pablopunk: We don't, the array is always one element right now (the body editor). I assumed when we have suggestions/comments in multiple fields we'll have several items in this array.

scripts/core/editor3/helpers/editor3CustomData.js
```diff
@@ -42,10 +42,18 @@ export function setCustomDataForEditor(editorState, key, value) {
...
+        const firstBlock = editorState.blocks[0];
+
+        if (firstBlock.data[key]) {
+            data = [...data, ...firstBlock.data[key]];
```
tomaskikutis: Is there a way to know which field an array item belongs to? I think it'd be better to make this change after data storage on the back-end is implemented.

I think it'd be best to have an optional `meta` field for every primary field like so:
```
{
    "title": "<span>Hello world!</span>",
    "title_meta": {
        "editorState": {}
    },
    "body": "<span>Hello world!</span>",
    "body_meta": {
        "editorState": {}
    }
}
```
We could then store `editorState` or any other metadata we need in the future.

ioanpocol: I agree that it is better to add this functionality after the save/restore of editorState for all editor3 fields is implemented.

petrjasek: we have to stick with lists because of the way how eve works,
but if we extend that editor state field to be like:
```
[
    {id: 'headline', editorState: ...},
    {id: 'body_html', editorState: ...}.
]
```
it should work

pablopunk: But if we have extra fields in the item as Tomas proposed, it should still work right?

petrjasek: yep, we could have `{id: 'foo', extra: true}` then or something
or `{id: 'extra.foo'}`

pablopunk: But I mean, that list you were talking about before, should it be inside the `item` object? Like `editor_states` or something?

tomaskikutis: Petr, I think we are talking about different objects. What I suggest is to add fields to the `article` object, so we shouldn't need any lists for Eve. Here's a sample article object I'm referring to.
```
{
    "force_unlock": true,
    "dateline": {},
    "body_html": "<p>What is Lorem Ipsum?</p>",
    "event_id": "tag:sd-master.test.superdesk.org:2018:f7cb6f96-ae71-4c75-913e-15bff8117dc9",
    "_current_version": 5,
    "sign_off": "fl",
    "_id": "urn:newsml:sd-master.test.superdesk.org:2018-03-13T22:00:01.156786:09de6b86-c781-454e-9ee4-44bcb4fa4324",
    "lock_time": "2018-03-16T10:27:48+0000",
    "_created": "2018-03-14T11:19:55+0000",
    "_etag": "58c5ca354db581c6373b6ca99f81d7a2b1caebcb",
    "schedule_settings": {
        "utc_embargo": null,
        "time_zone": null,
        "utc_publish_schedule": null
    },
    "versioncreated": "2018-03-14T19:55:35+0000",
    "family_id": "urn:newsml:sd-master.test.superdesk.org:2018-03-13T22:00:01.156786:09de6b86-c781-454e-9ee4-44bcb4fa4324",
    "format": "HTML",
    "version": 1,
    "slugline": "galoga",
    "language": "en",
    "sms_message": null,
    "urgency": 3
}
```

pablopunk: Also, regarding this PR, I'm dealing with a property of that article object:
```js
{
  ...,
  editor_state: [
    { blocks: [{ data: ... }] },
    // I guess here we'll have more editor states in the future (one per field)
  ]
}
```

And then I'm collecting all comments/suggestions from each element of `editor_state` (which is only the body state right now).

What Tomas suggested was having each state for each field in a separated property for the article (`body_meta: { editorState }`, etc)

petrjasek: yep I'm talking about the same thing, it's better if it's a single field otherwise it would be hard to define elastic mapping for it, and if it's a single field it's better if it's a list because dict doesn't work well with eve. so I'm suggesting to replace editor_state field with eg field_metadata where it would have an dict for each field with editor state etc.

tomaskikutis: How come `schedule_settings` seems to work?

Anyway, if it doesn't, how about this?
```
{
    "item": {
        "body": "<span>Hello world!</span>",
        "title": "<span>Hello world!</span>",
        "field_metadata": [
            {
                "body": {
                    "editorState": {},
                    "anythingElse": {}
                },
                "title": {
                    "editorState": {},
                    "anythingElse": {}
                }
            }
        ]
    }
}
```

pablopunk: So that list would always have only one object?

tomaskikutis: Yes, to please Eve. The apple wasn't enough.

petrjasek: that would work too, but I don't see then why not do it like:
```
{field: 'body', editorState: {}, etc: {}},
```
so it's a list of items. schedule_settings work because their structure is predefined, but in field metadata we would add new keys for new fields and that's the issue.

petrjasek: it's also a pattern we use for other fields like anpa_category, subject, genre, signal, authors, place

pablopunk: Ok I think we can go with that solution. Also, I'm not sure where `editor_state` is being injected into the `item`

tomaskikutis: I was trouble understanding where exactly would you place it in the item object, but I've looked up `anpa_category` and I think I understand it now. Is this what you're trying to describe?
```
{
    "item": {
        "heading": [
            {
                "value": "<span>Hello world!</span>",
                "editorState": {},
                "anythingElse": {}
            }
        ],
        "body": [
            {
                "value": "<span>Hello world!</span>",
                "editorState": {},
                "anythingElse": {}
            }
        ]
    }
}
```

If so, I'm fine with it. Suggested a separate field thinking that we'd get a better backwards compatibility by doing that.

petrjasek: no, I mean:
```
{
    "item": {
          "headline": "foo",
          "slugline": "bar",
          "field_meta": [
                {"field": "headline", "editorState": {}},
                {"field": "slugline", "editorState": {}},
          ]
    }
}
```

tomaskikutis: :smile: Ok, it's clear at last. Yeah, we can go with this one as well. I suggested a single object inside the array because retrieving the value from the object is easier, since you don't have to iterate.

ioanpocol: Why we are not using the existing editor_state field?
Also we have to add in a list only the individual value of a single editorState to prevent the eve update issue.
"
{
    "item": {
          "headline": "foo",
          "slugline": "bar",
          "editor_state": {
                "headline": [<editorState dictionary>],
                "slugline": [<editorState dictionary>],
          }
    }
}
"

Eve update issue:
We have to put the editor state in a list because on updating a dictionary eve performs by default a merge on multiple levels between current and updated dictionary. In the case of editor state if we delete something, the old value will be still presented in editor_state.  A list stops that behavior and replace the current list with the new one.

tomaskikutis: It's possible to use `editor_state`, but in any way, we'll have to make changes on the backed to adjust to the new data structure as well as write a deployment script to transform current data, so we might as well rename it, so it's easy to use it in the future if we decide to have other kinds of metadata, not just `draftjs` state

ioanpocol: OK, so something like this:

{
    "item": {
          "headline": "foo",
          "slugline": "bar",
          "field_meta": {
                "headline": { editorState: [<editorState dictionary>]},
                "slugline": {editorState: [<editorState dictionary>],
          }
    }
}

tomaskikutis: Why does the structure Petr posted doesn't work?

If I understand correctly, eve merges only the first dictionary, so in your snippet "headline" and "slugline" would get merged and we wouldn't be able to remove fields.

editorState would no longer be required to be wrapped inside the array if we wrapped the root object as in Petr's snippet

ioanpocol: If you will update only one field you still have to send the current values for all fields.
The eve merge is good in this case, it is just not good for updating a single editor state structure.

tomaskikutis: I guess both structures are fine then, I can use the one you posted if you think it's better. Isn't there a way to tell eve to replace an object instead of mapping?

ioanpocol: I already explained that not both structures are fine. I will try to explain again here:
In order to support partial update for editor fields (ex: update the editor state for one editor3 field will not require the editor state of all other fields), field_metadata should be a dictionary not a list.
But for every editor3 field the editor_state should be a list that contains the dictionary data generated by draft js. In this way it will be prevented partial updates at editor state level, that will corrupt the editor state.
By default eve try to perform a merge on all possible levels so to wrap the dictionary in a list is a small workaround that solved the problem.

tomaskikutis:
Let's first implement multiple editor states on the article itself

PR #2154, opened 2018-03-16
fix(editor3): Small bug fix and add the option to move by positive an…

ioanpocol:
…d negative offset from start or end of selection

### DIFF
--- a/scripts/core/editor3/helpers/highlights.js
+++ b/scripts/core/editor3/helpers/highlights.js
@@ -406,6 +406,7 @@ export function getCharByOffset(editorState, selection, offset) {
  * from beggining of selection if startFromEnd is false or from end of selection otherwise.
  */
 const getBlockAndOffset = (content, selection, offset, startFromEnd = false) => {
+    const noValue = {block: null, newOffset: null};
     let newOffset;
     let block;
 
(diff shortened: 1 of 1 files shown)


tomaskikutis:
Don't merge it just yet, I'll check the tests

tomaskikutis:
Tests pass, you can merge. The only ones which don't are for a feature to limit extending to a single block, and returning alternative selection instead of `null`.

It's strange that for out-of-range cases you are returning an array of nulls, why a single `null` isn't enough?

ioanpocol:
I didn't tried but I suppose that when you destruct a null you will get an error.

tomaskikutis:
I've checked the destrcuting, it returns `undefined` no matter if there is something inside the array or if it's empty. You could return `return {block: null, newOffset: null};` to make it consistent.

ioanpocol:
OK, I seen now, I forgot to update those returns.

ioanpocol:
I updated the code

tomaskikutis:
Saw it. All good. If we were to be pedantic, we could define a variable `noValue` and use it in those 3 places, but it's just a code style thing. I don't want to cause you work, so feel free to merge it as is.

ioanpocol:
OK, let's be pedantic :)

tomaskikutis:
Ok, let's merge it, I have no more to suggest :smile:

PR #2157, opened 2018-03-20
SDESK-2389 Editor3 - Change the generation of html to include annotations

mugurrus:


scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -193,11 +193,37 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
...
+                            }
+                        });
+                    });
+                }
+            }
+
             /**
              * Create a new version
              */
             $scope.save = function() {
```
tomaskikutis: It would be much simpler and cleaner if you removed this code, and instead when you need the annotation when generating HTML, you could use `contentBlock.getInlineStyleAt()`

scripts/core/editor3/helpers/highlights.js
```diff
@@ -262,12 +262,16 @@ export function getHighlightDataAtOffset(editorState, types, selection, offset)
...
+    if (highlightsState.lastHighlightIds && _.has(highlightsState.lastHighlightIds, type)) {
+        newIndex = highlightsState.lastHighlightIds[type] + 1;
+    }
+    const styleName = type + '-' + newIndex;
 
     const newHighlightsState = {
         lastHighlightIds: {
             ...highlightsState.lastHighlightIds,
-            [type]: highlightsState.lastHighlightIds[type] + 1
+            [type]: newIndex
```
tomaskikutis: Is this change only to check for null/undefined values? `highlightsState.lastHighlightIds[type]` should never be null/undefined by design, so check might not be required

scripts/core/editor3/html/to-html/HTMLGenerator.js
```diff
@@ -29,11 +31,12 @@ const BlockStyleTags = {
  * tables.
  */
 export class HTMLGenerator {
-    constructor(contentState, disabled = []) {
+    constructor(contentState, editorState, disabled = []) {
```
tomaskikutis: I think `contentState` you already had is enough for generating HTML. `editorState` is the same as `contentState` except it has a `selectionState` which you will no be using for HTML generation anyway.

scripts/core/editor3/html/to-html/HTMLGenerator.js
```diff
@@ -160,12 +163,26 @@ export class HTMLGenerator {
      * @description Returns an array of annotations sorted by offset.
      */
     getAnnotations(contentBlock) {
-        let ranges = contentBlock.getData().get('inlineStyleRanges');
+        let annotationIds = [];
+        let ranges = [];
+        let styleRanges = contentBlock.findStyleRanges((charMeta) => {
```
tomaskikutis: Were there any problems using/extending `BlockInlineStyleWrapper` as I've suggested before? Your approach might work, but considering the ease of maintenance, it's better not to have multiple pieces of code doing very similar things.

petrjasek:
we should avoid changing the output - so it should produce the same html like [it did on server](https://github.com/superdesk/superdesk-core/commit/8c23362b16f751ecc914dac9b9f5a4aedc1ab83a#diff-448d51b04f91591e4739b31dc6ba2ac3R677) plus it was also populating [annotations](https://github.com/superdesk/superdesk-core/commit/8c23362b16f751ecc914dac9b9f5a4aedc1ab83a#diff-448d51b04f91591e4739b31dc6ba2ac3R680) field which seems missing here

scripts/core/editor3/html/to-html/HTMLGenerator.js
```diff
@@ -29,11 +31,12 @@ const BlockStyleTags = {
...
  */
 export class HTMLGenerator {
-    constructor(contentState, disabled = []) {
+    constructor(contentState, editorState, disabled = []) {
         this.contentState = contentState;
+        this.editorState = editorState;
         this.listTags = [];
         this.lastDepth = 0;
-        this.atomicBlockParser = new AtomicBlockParser(contentState, disabled);
+        this.atomicBlockParser = new AtomicBlockParser(contentState, editorState, disabled);
```
tomaskikutis: `const editorState = EditorState.createWithContent(contentState)` [docs](https://draftjs.org/docs/api-reference-editor-state.html#createwithcontent)

scripts/apps/authoring/authoring/services/AuthoringService.js
```diff
@@ -248,6 +251,35 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         return promise;
     };
 
+    /**
+     * Generate item annotations field
+     *
+     * @param {Object} item
+     */
+    this.generateAnnotations = (item) => {
```
tomaskikutis: Avoid using the internals of how annotations are implemented. Get them using the interface as it's done in `addCommentsForServer`

PR #2179, opened 2018-03-29
fix(suggestions): Show correctly the replace suggestion on suggestion…

ioanpocol:
…s list


scripts/apps/authoring/track-changes/views/suggestions-widget.html
```diff
@@ -10,12 +10,20 @@
             </figure>
             <div>{{users[item.suggestionInfo.author].display_name || users[item.suggestionInfo.author].username}}</div>
             <div class="date" sd-reldate ng-model="item.suggestionInfo.date"></div>
-            <div class="content">
+            <div class="content" ng-show="!item.oldText">
```
tomaskikutis: It would be more clear to use if/else by `item.suggestionInfo.type === 'replace'`

PR #2180, opened 2018-03-29
Make suggestions work on any custom text field

tomaskikutis:
SDFID-345

MUST BE MERGED TOGETHER WITH SERVER CHANGES


scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.js
```diff
@@ -66,31 +67,11 @@ export function ContentProfileSchemaEditor(content, config) {
 
             scope.formattingOptionsEditor3 = [
                 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'ordered list', 'unordered list', 'quote', 'media', 'link',
-                'embed', 'underline', 'italic', 'bold', 'table', 'formatting marks'
+                'embed', 'underline', 'italic', 'bold', 'table', 'formatting marks', 'annotation', 'comments',
+                'suggestions'
             ];
 
-            if (config.features.editorHighlights) {
```
ioanpocol: It was a requirement to be able to enable/disable highlights functionality from config file. Why did you dropped this feature?

scripts/apps/authoring/views/article-edit.html
```diff
@@ -390,7 +394,7 @@
      data-required="schema.footer.required"
      order="{{editor.footer.order}}"
      sd-width="{{editor.media_description.sdWidth|| 'full'}}">
-    <label class="field__label" translate>Helplines/Contact Information</label>
+    <label class="field__label"translate>Helplines/Contact Information</label>
```
petrjasek: missing space

scripts/core/editor3/components/Editor3.jsx
```diff
@@ -28,7 +28,6 @@ Editor3Base.propTypes = Editor3Component.propTypes = {
 const mapStateToProps = (state) => ({
     readOnly: state.readOnly,
     showToolbar: state.showToolbar,
-    highlightsEnabled: state.allowsHighlights,
```
ioanpocol: There was the option to disable highlights from config file. Why did you dropped this feature.

tomaskikutis: The feature was only working forward, meaning once you enable it, disabling wouldn't work properly  - you could no longer add new highlights, but existing ones would be visible, but nothing would happen if you clicked it.

I thought since now highlights can be configured for any field, we might not need to hardcode it in the file itself. If we still need this feature I can restore it, but we'd need to fix it.

PR #2188, opened 2018-04-04
fix(vocabularies): Increase the number of vocabularies returned

ioanpocol:


scripts/apps/vocabularies/services/VocabularyService.js
```diff
@@ -15,7 +15,7 @@ import _ from 'lodash';
 VocabularyService.$inject = ['api', '$q', '$filter', '$rootScope'];
 export function VocabularyService(api, $q, $filter, $rootScope) {
     var self = this;
-    const MAX_RESULTS = 200;
+    const MAX_RESULTS = 1000;
```
petrjasek: there is [api.getAll](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/api/api-service.js#L401) method which should handle it

PR #2199, opened 2018-04-05
feat(photo-grid): Added photo grid in global search

pavlovicnemanja:


scripts/apps/search/components/Item.jsx
```diff
@@ -114,11 +114,20 @@ export class Item extends React.Component {
             'sd-grid-item sd-grid-item--with-click' :
             'media-box media-' + item.type;
 
+        let multiSelected = this.props.item.selected;
+
+        // If an item is both multi-selected and is active selection in the list
+        // remove the 'selected' styling class to distinguish it as the active item
+        if (this.props.item.selected && this.props.flags.selected) {
+            multiSelected = !multiSelected;
```
tomaskikutis: Replace `!multiSelected` with `false` since it's the only possible value at this line.

Since you already created a variable, `multiSelected` use it in the `if` condition as well.

scripts/apps/search/components/Item.jsx
```diff
@@ -114,11 +114,20 @@ export class Item extends React.Component {
             'sd-grid-item sd-grid-item--with-click' :
             'media-box media-' + item.type;
 
+        let multiSelected = this.props.item.selected;
```
petrjasek: imo would be better to avoid this variable altogether, because it's not what it says and the way it's used is kind of confusing. so for `div.selected` there would be:
`selected: this.props.item.selected || this.props.flags.selected,`
and for `li.selected`:
`selected: this.props.item.selected && !this.props.flags.selected,`

PR #2200, opened 2018-04-06
Create isolate scopes

tomaskikutis:
SDFID-364

Create isolate scopes, so a directive can be used multiple times with the same controller. It used to use the scope of the controller and if you had multiple instances, they would all display the same date from the controller scope.

### DIFF
--- a/scripts/apps/archive/views/metadata-view.html
+++ b/scripts/apps/archive/views/metadata-view.html
@@ -86,16 +86,16 @@
         </dl>
         <dl ng-if="item.firstcreated">
             <dt translate>Created</dt>
-            <dd sd-reldate-complex ng-model="item.firstcreated"></dd>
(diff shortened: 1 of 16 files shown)


petrjasek:
would be better to avoid using ng-model here, we only need to watch the value and not change it, there is some overhead with parsers/validators

tomaskikutis:
Is there a different way to tell a directive to have the variable in its own scope?

petrjasek:
the scope is fine, but there should be no `require: 'ngModel'` and the scope param name should be different, because `ng-model` is a directive which has some overhead and handles input events,
you can make it like:
```
scope: {datetime: '='},
link: (scope) => {
    scope.$watch('datetime', (datetime) => {
         // here goes code from ngModel.$render
    });
}
```

tomaskikutis:
Got it, thanks for the example.

PR #2208, opened 2018-04-10
[SDESK-2746] Remove all suggestions when pressing DELETE key in suggestion mode

pablopunk:
### The problem

When you do this:

kapture 2018-04-10 at 16 16 27

When you press the DELETE key it would only delete the text `suggestion` but not the suggestion itself. So whenever you wanted to publish the item, the suggestion would exist there forever and you couldn't publish as it was unresolved.

### The solution

As the processing of the DELETE key is handled char by char, it now removes the highlight if there's no more text for that suggestion ID.

### DIFF
--- a/scripts/core/editor3/reducers/suggestions.jsx
+++ b/scripts/core/editor3/reducers/suggestions.jsx
@@ -347,6 +347,7 @@ const deleteCurrentSelection = (editorState, data) => {
     let newState = editorState;
 
     newState = initSelectionIterator(newState, backward);
+
     while (hasNextSelection(newState, selection, backward)) {
         newState = setDeleteSuggestionForCharacter(newState, data);
     }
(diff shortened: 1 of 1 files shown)


scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -436,7 +437,16 @@ const setDeleteSuggestionForCharacter = (editorState, data) => {
     if (currentData != null && currentData.type === 'ADD_SUGGESTION' &&
         currentData.author === data.author) {
         // if current character already a suggestion of current user, delete the character
-        return deleteCharacter(editorState);
+        let newState = deleteCharacter(editorState);
+
+        // also delete the suggestion if it was the last character of that suggestion
```
ioanpocol: It will be better if you move this code on deleteCharacter, because this code should run every time deleteCharacter is called.

PR #2210, opened 2018-04-11
fix(editor3): Fix the reject suggestion that contain a list and fix t…

ioanpocol:
…he delete of list in suggestion mode

### DIFF
--- a/scripts/core/editor3/reducers/suggestions.jsx
+++ b/scripts/core/editor3/reducers/suggestions.jsx
@@ -511,13 +511,14 @@ const resetSuggestion = (editorState, style) => {
  * @description Delete the current character.
  */
 const deleteCharacter = (editorState, style = null) => {
-    let content;
-    let selection;
-    let newState;
+    let newState = Highlights.changeEditorSelection(editorState, -1, 0, false);
+    let content = newState.getCurrentContent();
+    const selection = newState.getSelection();
+    const block = content.getBlockForKey(selection.getStartKey());
 
-    newState = Highlights.changeEditorSelection(editorState, -1, 0, false);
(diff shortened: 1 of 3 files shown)


scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -535,4 +536,44 @@ const deleteCharacter = (editorState, style = null) => {
...
+            anchorKey: block.getKey(),
+            focusOffset: block.getLength(),
+            focusKey: block.getKey(),
+            isBackward: false
+        });
+
+        content = Modifier.removeRange(content, selection, 'forward');
+    }
+
+    return EditorState.push(editorState, content, 'backspace-character');
```
petrjasek: change type should be `delete-character` if it's forward

scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -535,4 +536,44 @@ const deleteCharacter = (editorState, style = null) => {
     return newState;
 };
 
+const deleteBlock = (editorState, blockKey) => {
```
tomaskikutis: There's a function `removeBlock` in `/scripts/core/editor3/reducers/toolbar.jsx`. Could it be merged into one? It seems that your function does something different than just removes a block. I see that if the block is `null` is picks the block after and next selection depends on which block was picked. If it's not the same as `removeBlock` it'd be useful to give it a more descriptive name. Suggestions that contain a list are mentioned in the PR name, but nowhere in the code is there any reference to it. Could you add a comment before calling `deleteBlock` to explain why it's called from a function which should only delete a character?

PR #2211, opened 2018-04-11
fix(editor3): fix paste performance when suggesting

petrjasek:
handle whole pasted content at once, not each character separate.

SDFID-361

### DIFF
--- a/scripts/core/editor3/reducers/suggestions.jsx
+++ b/scripts/core/editor3/reducers/suggestions.jsx
@@ -1,4 +1,4 @@
-import {EditorState, Modifier, RichUtils} from 'draft-js';
+import {EditorState, Modifier, RichUtils, SelectionState} from 'draft-js';
 import {onChange} from './editor3';
 import {acceptedInlineStyles} from '../helpers/inlineStyles';
 import {changeSuggestionsTypes, styleSuggestionsTypes} from '../highlightsConfig';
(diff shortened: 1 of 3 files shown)


scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -206,26 +247,41 @@ const createChangeBlockStyleSuggestion = (state, {blockType, data}) => {
...
+    // push new content
+    editorState = EditorState.push(editorState, mergedContent, 'insert-fragment');
 
-    return saveEditorStatus(state, editorState, 'change-block-data');
+    // store current selection for later
+    const finalSelection = editorState.getSelection();
+
+    // select pasted content
+    editorState = EditorState.acceptSelection(editorState, editorState.getSelection().merge({
+        anchorKey: initialSelection.getAnchorKey(),
```
ioanpocol: Is this working well when the initial selection is backward?
Usually I use getStartKey/Offset and getEndKey/Offset and isBackward: false.

scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -195,6 +195,47 @@ const createChangeBlockStyleSuggestion = (state, {blockType, data}) => {
...
+            }
+        );
+
+        // remove any entities
+        block.findEntityRanges(
+            () => true,
+            (start, end) => {
+                const selection = getSelection(block, start, end);
+
+                output = Modifier.applyEntity(output, selection, null);
```
ioanpocol: Not sure if we have to delete the Entity, for example when you copy  a link from the same editor.

ioanpocol:
:+1:

scripts/core/editor3/reducers/suggestions.jsx
petrjasek: I think it doesn't, will push an update

scripts/core/editor3/reducers/suggestions.jsx
petrjasek: yep we could keep that, I was testing with external links and those didn't work so I've rather disabled it for now

PR #2214, opened 2018-04-13
Add configuration option for fields in gridview

tomaskikutis:


scripts/core/utils.js
```diff
@@ -3,3 +3,9 @@ export function stripHtmlTags(value) {
 
     return value ? String(value).replace(htmlRegex, '') : '';
 }
+
+export function getAngularService(name) {
+    return angular.element(document.body)
```
pablopunk: Couldn't we use `ng.get` to get a service?

```js
import ng from './services/ng'
ng.get(name)
```

tomaskikutis:
```
console.log(ng);
console.log(ng.get('content'));
```

```
ProviderService {}
ng.js:48 Uncaught ng: $injector not registered for core/services/ng
```

Unfortunately not. You can only use `ng` when you have angular stuff already configured in that scope. It doesn't work outside angular for example in React components unless you pass some props all the way down from the place you are calling `ReactDOM.render`.

@petrjasek do you think it would be safe to replace the implementation of `scripts/core/services/ng.js` with my implementation of `getAngularService` except that I'd only return the injector and not the servivce itself?

petrjasek: yep it will only work after angular config phase, but this should be the same with your implementation.
so this shouldn't work neither
```
import {getAngularService} from 'core/utils';
getAngularService('content');
```

scripts/core/helpers/ReactRenderAsync.jsx
```diff
@@ -0,0 +1,56 @@
...
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            loading: true,
+            mappedProps: {}
+        };
+    }
+    componentWillMount() {
+        Promise.all(this.props.promises)
```
petrjasek: angular has this `$q.all` which handles also objects of promises, this could simplify the api so you woudn't have to provide the mapping function:
```js
const promises = {content: () => content.getCustomFields()};
```
would populate content prop when resolved. what do you think?

scripts/core/utils.js
tomaskikutis: The code you posted does work. I'd guess it wouldn't work if angular hasn't finished it's config phase/bootstraping/mouting whatever it's called, but once it does - it works everywhere - which is why it's especially useful in React components as it allows accessing services without having to pass props 10 levels deep

petrjasek: I think they both do the same thing, so if you do in your code:
```js
const getPromises = () => [ng.get('content').getCustomFields()];
```
it will work the same. it's only called when rendering so angular should be configured at that time

scripts/core/helpers/ReactRenderAsync.jsx
tomaskikutis: The API you are suggesting looks good, but I don't like introducing dependencies where they are not required.

I suggest we leave it as is, or write something like
```
var promises = {
    one: new Promise(),
    two: new Promise()
};

var keys = Object.keys(promises);
var promisesArray = keys.map((key)=> promises[key]);
Promise.all(promisesArray).then(()=>{
    return keys.reduce((obj, key, i)=>{
        obj[keys[i]] = arguments[i];
    }, {});
});
```
This is proof of concept, I haven't ran the code and have probably left a mistake somewhere.

petrjasek: yep I was thinking about replicating the api, not using `$q` there. your code looks good

tomaskikutis: Ok, I will implement it then

scripts/core/utils.js
tomaskikutis: I checked it and you are right, I'll use ng.

scripts/core/helpers/ReactRenderAsync.jsx
tomaskikutis: I've noticed that we don't support custom fields in list view, so I refactored and simplified the PR, not to handle them in the grid view as well.

PR #2215, opened 2018-04-13
Fix code navigation for vscode

tomaskikutis:


jsconfig.json
```diff
@@ -1,5 +1,9 @@
 {
     "compilerOptions": {
-        "baseUrl": "./scripts",
-    }
+        "target": "es6",
+        "jsx": "preserve"
+    },
+    "exclude": [
+        "node_modules"
```
petrjasek: you could also add here `bower_components` `dist` and `po`

PR #2219, opened 2018-04-16
fix(search-providers): fix search providers menu

petrjasek:
- filter out closed providers
- put default provider on top

SDESK-2644 SDANSA-136


scripts/apps/search/directives/RepoDropdown.js
```diff
@@ -2,10 +2,14 @@
...
         scope: {
-            providers: '=',
             activeProvider: '=',
             toggleProvider: '=',
         },
         template: require('../views/repo-dropdown.html'),
+        link: (scope) => {
+            scope.isActive = (provider) => scope.activeProvider ?
+                scope.activeProvider._id === provider._id :
+                !provider._id;
```
tomaskikutis: Negating non-boolean values makes it hard to understand what are you doing. It'd be simpler if you used `provider._id == null` which is what I guess you actually mean by `!provider._id`. I think the same stands for type coercion in general, so you could also improve the ternary by adding a value to compare to `scope.activeProvider != null ?`

PR #2224, opened 2018-04-17
restore focus before adding a highlight, so the selection is visible after undo

tomaskikutis:
SDFID-376

### DIFF
--- a/scripts/core/editor3/helpers/highlights.js
+++ b/scripts/core/editor3/helpers/highlights.js
@@ -363,7 +363,18 @@ export function getHighlightDataAtCurrentPosition(editorState, types) {
  * @description add a new highlight for the current selection.
  */
 export function addHighlight(editorState, type, data) {
-    const highlightsState = getHighlightsState(editorState);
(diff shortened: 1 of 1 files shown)


scripts/core/editor3/helpers/highlights.js
```diff
@@ -363,7 +363,18 @@ export function getHighlightDataAtCurrentPosition(editorState, types) {
  * @description add a new highlight for the current selection.
  */
 export function addHighlight(editorState, type, data) {
-    const highlightsState = getHighlightsState(editorState);
+    var nextEditorState = editorState;
```
pablopunk: You're assigning nextEditorState to editorState here but reassigning on the next line, should it be like this?:

```diff
- var nextEditorState = editorState;
+ const nextEditorState = EditorState.acceptSelection(
+   editorState,
+   editorState.getSelection().merge({
+     hasFocus: true
+   })
+ );
```

tomaskikutis: You are right, there's no need to reassign here :)

PR #2228, opened 2018-04-19
SDESK-2763 Empty Qcode not checked for some vocabularies

mugurrus:


scripts/apps/vocabularies/directives/VocabularyConfigModalItems.jsx
```diff
@@ -48,7 +74,9 @@ export function VocabularyConfigModalItems(gettext) {
             // re-render on items changes
             scope.$watch('vocabulary.items', (items) => {
                 if (items && component) {
-                    component.setState({items});
+                    itemsValidation = validateItems(items);
+                    scope.itemsValidation.valid = validItems(itemsValidation);
```
petrjasek: you can use [ngForm.$setValidity](https://docs.angularjs.org/api/ng/type/form.FormController#$setValidity) here so it works as other form controls

scripts/apps/vocabularies/directives/VocabularyConfigModalItems.jsx
```diff
@@ -8,16 +8,42 @@ export function VocabularyConfigModalItems(gettext) {
...
                 const updates = {[key]: value};
 
                 // sync scope
                 scope.$applyAsync(() => {
                     ngForm.$setDirty();
-                    scope.vocabulary.items = scope.vocabulary.items.map((_item) =>
-                        _item === item ? Object.assign({}, item, updates) : _item
-                    );
+                    let index = 0;
+                    scope.vocabulary.items = scope.vocabulary.items.map((_item) => {
```
petrjasek: you can get index via second param ```items.map((_item, index) => {```

PR #2234, opened 2018-04-19
fix(editor3): Prevent infinite loop on editor3 processing

ioanpocol:


petrjasek:
isn't it an issue with `hasNextSelection` ?

ioanpocol:
It was an issue on hasNextSelection too, but it has been fixed on https://github.com/superdesk/superdesk-client-core/pull/2233

PR #2235, opened 2018-04-20
[SDESK-2777] Only support users can view support users

pablopunk:
Following https://github.com/superdesk/superdesk-core/pull/1252

### DIFF
--- a/scripts/apps/users/controllers/UserListController.js
+++ b/scripts/apps/users/controllers/UserListController.js
@@ -75,7 +75,7 @@ export function UserListController($scope, $location, api, _, session, usersServ
     }
 
     function initCriteria(search, filter) {
-        let query = {};
+        const query = {};
         const canSeeSupportUsers = usersService.isSupport(session.identity);
 
         if (!canSeeSupportUsers) {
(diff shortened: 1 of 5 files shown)


scripts/apps/users/directives/UserListDirective.js
```diff
@@ -25,6 +25,15 @@ export function UserListDirective(keyboardManager, usersService, asset) {
                 bindKeys();
             };
 
+            scope.$watch('users', (users) => {
```
petrjasek: I would rather put this logic into `UserListController` where is the rest of the filtering (active/online/etc) so it's all in one place

pablopunk:
I found an error, don't merge this

scripts/apps/users/controllers/UserListController.js
```diff
@@ -117,6 +117,15 @@ export function UserListController($scope, $location, api, _) {
             break;
         }
 
+        if (!usersService.isSupport(session.identity)) {
+            // If user is not support type, they can't see support users
+            query.$or = [
```
petrjasek: I think it should be like `query.is_support: {$ne: true}`

scripts/apps/users/controllers/UserListController.js
```diff
@@ -75,14 +75,28 @@ export function UserListController($scope, $location, api, _, session, usersServ
...
 
         if (search.q) {
             query = {
                 $or: [
-                    {username: {$regex: search.q, $options: '-i'}},
-                    {display_name: {$regex: search.q, $options: '-i'}},
-                    {email: {$regex: search.q, $options: '-i'}},
+                    {
+                        username: {$regex: search.q, $options: '-i'},
+                        is_support: {$ne: !canSeeSupportUsers},
```
petrjasek: I think you don't have to add it in `$or`, the query should work in a way that on top level it's AND so it must pass is_support + some of the $or filters

PR #2236, opened 2018-04-22
SDFID-313 Change email address field not to be case sensitive

mugurrus:


scripts/apps/users/services/UsersService.js
```diff
@@ -19,9 +19,9 @@ export function UsersService(api, $q, notify) {
     usersService.save = function save(user, data) {
         return api.save('users', user, data)
             .then((updates) => {
-                angular.extend(user, data);
-                angular.extend(user, updates);
-                return user;
+                let newUser = _.assign({}, user, data, updates)
```
pablopunk: You could use `const` instead of `let` as the value won't change

PR #2242, opened 2018-04-24
[SDESK-2807] Add button to editor3 toolbar to remove all formatting

pablopunk:
Removes both inline and block styles from selection except from atomic blocks (images, tables and embeds)


petrjasek:
the tests finished ok, but it didn't report it to github

PR #2248, opened 2018-04-26
(fix): Fixing the unit test for planning.

mdhaman:
Unit test for planning was failing as `__SUPERDESK_CONFIG__` is not defined.  Using the `config` service instead.


mdhaman:
required by https://github.com/superdesk/superdesk-planning/pull/784

PR #2249, opened 2018-04-26
[SDESK-2808] Pasted text won't include disabled styles in editor3

pablopunk:


ioanpocol:
@karelpetrak Should we just hide the disabled stile or we have to also delete it from pasted text?

PR #2250, opened 2018-04-26
fix(editor3): On insert on edit mode don't split already existing sug…

ioanpocol:
…gestion but also don't extend them on the edges


scripts/core/editor3/helpers/highlights.js
```diff
@@ -246,22 +246,26 @@ export function canAddHighlight(editorState, highlightType) {
  * @param {Integer} offset
```
pablopunk: This is missing docs for 2 last parameters

ioanpocol: Done

PR #2252, opened 2018-04-26
[INGEST] disabled autocompletion on password for feeding service

jerome-poisson:
autocomplete="off" on `<input>` is not respected anymore by modern
browsers, and autocomplete="off" on `<form>` is not working either.
Furthermore, Chromium respect autocomplete="new-password" but Firefox
is not (as of 59.0.2).
A hack is used for Firefox with dummy login/password field, which
disable autocompletion, as it seems to be the only working way (beside
setting password with code).
For Chromium, autocomplete="new-password" should be used for each new
password input.

SDESK-2745

### DIFF
--- a/scripts/apps/ingest/views/settings/apConfig.html
+++ b/scripts/apps/ingest/views/settings/apConfig.html
@@ -4,7 +4,7 @@
 </div>
 <div class="field">
     <label for="ap_password" translate>Password</label>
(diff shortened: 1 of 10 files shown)


petrjasek:
what about `autocomplete="nope"` https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#Disabling_autocompletion would that work in firefox?

PR #2264, opened 2018-05-04
Wait for media to fire load events instead of setting a timeout

tomaskikutis:
SDESK-2731

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
+++ b/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
@@ -2,6 +2,10 @@
 
 import 'owl.carousel';
 import * as ctrl from '../controllers';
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
```diff
@@ -46,19 +50,25 @@ export function ItemCarouselDirective($timeout) {
...
                 if (carousel) {
-                    interval = 0;
                     carousel.trigger('destroy.owl.carousel');
                 }
 
-                if (items.length > 1) {
-                    $timeout(() => {
-                        initCarousel();
-                    }, interval, false);
+                const carouselImages = Array.from(document.querySelectorAll(`${carouselContainerSelector} img`));
```
petrjasek: just wondering if this works for multiple multi media fields at the same time? shouldn't it use `elem.find` rather?

tomaskikutis: You're right, I've pushed a fix.

PR #2265, opened 2018-05-04
SDESK-2832 It is not possible to edit and save some CVs

mugurrus:


scripts/apps/vocabularies/tests/vocabularies.spec.js
```diff
@@ -1,4 +1,4 @@
-describe('vocabularies', () => {
+fdescribe('vocabularies', () => {
```
ioanpocol: Please change back to describe

PR #2268, opened 2018-05-04
[SDANSA-40] - Notifications within the desks on the routing

fritzSF:


fritzSF:
The commented out markups in menu.html is the part that should be implemented. It is left like this intentionally to make it easier implementation.

PR #2271, opened 2018-05-04
feat(editor3): Add support for merge paragraph suggestion and improve…

ioanpocol:
… the add split paragraph suggestion

### DIFF
--- a/scripts/core/editor3/helpers/highlights.js
+++ b/scripts/core/editor3/helpers/highlights.js
@@ -6,9 +6,9 @@ import {getDraftCharacterListForSelection} from './getDraftCharacterListForSelec
 import {getDraftSelectionForEntireContent} from './getDraftSelectionForEntireContent';
 import {resizeDraftSelection} from './resizeDraftSelection';
 import {clearInlineStyles} from './clearInlineStyles';
-import {changeSuggestionsTypes, blockStylesDescription, blockChangeSuggestionTypes} from '../highlightsConfig';
+import {changeSuggestionsTypes, blockStylesDescription, paragraphSuggestionTypes} from '../highlightsConfig';
 
-export const paragraphSeparator = 'π';
+export const paragraphSeparator = '¶';
 
 export const availableHighlights = Object.keys(highlightsConfig).reduce((obj, key) => {
     obj[key] = highlightsConfig[key].draftStyleMap;
@@ -685,10 +685,9 @@ function getLeftRangeAndTextForStyle(editorState, style) {
     let found;
     let newBlock = false;
 
-    if (blockChangeSuggestionTypes.indexOf(type) !== -1) {
+    if (paragraphSuggestionTypes.indexOf(type) !== -1) {
         startText = block.getText()
-            .substring(0, startOffset)
-            .replace(/ + paragraphSeparator + /g, '') + ' \\ ';
+            .substring(0, startOffset) + ' \\ ';
 
         return {startOffset, startBlock, startText};
     }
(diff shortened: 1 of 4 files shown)


scripts/core/editor3/helpers/highlights.js
```diff
@@ -682,6 +683,14 @@ function getLeftRangeAndTextForStyle(editorState, style) {
     let found;
     let newBlock = false;
 
+    if (blockChangeSuggestionTypes.indexOf(type) !== -1) {
+        startText = block.getText()
+            .substring(0, startOffset)
+            .replace(/π/g, '') + ' \\ ';
```
tomaskikutis: Why Pi(π) here? Wouldn't it be better to use a named constant?

scripts/core/editor3/helpers/selection/selectionIterator.js
```diff
@@ -51,10 +51,10 @@ export function hasNextSelection(editorState, selection, backward = false) {
     }
 
     if (backward) {
-        return selection.getStartOffset() !== crtSelection.getStartOffset() ||
+        return selection.getStartOffset() < crtSelection.getStartOffset() ||
```
tomaskikutis: Let's not use abbreviations

scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -245,6 +246,27 @@ const createChangeBlockStyleSuggestion = (state, {blockType, data}) => {
     return saveEditorStatus(state, editorState, 'change-block-type', true);
 };
 
+
+/**
+ * @ngdoc method
+ * @name isMergeParagraph
+ * @param {Object} editorState
+ * @param {Object} data - info about the author of suggestion
```
tomaskikutis: How about naming the argument `authorInfo` instead of `data`, so there's no need to comment on what it contains

scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -670,13 +746,21 @@ const setMergeParagraphSuggestion = (editorState, data) => {
...
-    const currentData = currentStyle != null ? Highlights.getHighlightData(editorState, currentStyle) : null;
 
     if (selection.getStartOffset() === 0) {
         return setMergeParagraphSuggestion(editorState, data);
     }
 
+    const paragraphStyle = Highlights.getHighlightStyleAtOffset(editorState, paragraphSuggestionTypes, selection, -1);
+
+    if (paragraphStyle != null) {
+        // if current character is marked as paragraph suggestion, skip
```
tomaskikutis: It's a bit late to comment on this, but it seems to me that it's a bad idea to implement `split-paragraph-suggestion` by inserting a character into the text itself. Inline styles can be put on this character, you can comment and add annotations on it. I feel it would be a constant edge-case in all work related to editor3

scripts/core/editor3/reducers/suggestions.jsx
ioanpocol: It is suggestion related data, it is not only about author.

tomaskikutis: How about suggestionAuthorInfo? :smile_cat:

pablopunk: I think this is all around the code, it's just metadata for the suggestion, but I think currently it only hold the author info

scripts/core/editor3/reducers/suggestions.jsx
ioanpocol: This is the agreed solution with POs. Let's see how it works and we will see if we will change or not.
Right now the only problem I see is when add a comment or an annotation on this single character. The comment/annotation will be deleted on paragraph suggestion accept/reject. I will fix it if it will be needed.

scripts/core/editor3/reducers/suggestions.jsx
ioanpocol: For block style contains also the set style. Probably suggestionData or suggestionInfo is a better name.
Anyway all over the app is used data so it is better this change to be performed in a different PR.

PR #2277, opened 2018-05-08
wait for angular to render elements before mounting the carousel, return only if both images AND videos aren't present

tomaskikutis:
SDESK-2731

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
+++ b/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
@@ -50,26 +50,30 @@ export function ItemCarouselDirective($timeout) {
                 scope.carouselItems = _.sortBy(_.filter(items, (item) => item[item.fieldId]),
                     [(item) => item[item.fieldId].order]);
 
-                if (carousel) {
-                    carousel.trigger('destroy.owl.carousel');
-                }
+                setTimeout(() => {
+                    // waiting for angular to render items
 
-                const carouselImages = Array.from(elem.get(0).querySelectorAll(`${carouselContainerSelector} img`));
-                const carouselAudiosAndVideos = Array.from(
-                    elem.get(0).querySelectorAll(
-                        `${carouselContainerSelector} video, ${carouselContainerSelector} audio`
-                    )
-                );
+                    if (carousel) {
+                        carousel.trigger('destroy.owl.carousel');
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
```diff
@@ -50,26 +50,30 @@ export function ItemCarouselDirective($timeout) {
...
+
+                    if (items.length < 1 || (carouselImages.length + carouselAudiosAndVideos.length < 1)) {
+                        return;
+                    }
+
+                    Promise.all([
+                        waitForImagesToLoad(carouselImages),
+                        waitForAudioAndVideoToLoadMetadata(carouselAudiosAndVideos),
+                    ]).then(() => {
+                        initCarousel();
```
pablopunk: This could be `then(initCarousel)`

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
```diff
@@ -50,26 +50,28 @@ export function ItemCarouselDirective($timeout) {
...
-                if (items.length < 1 || (carouselImages.length < 1 && carouselAudiosAndVideos.length < 1)) {
-                    return;
-                }
-
-                Promise.all([
-                    waitForImagesToLoad(carouselImages),
-                    waitForAudioAndVideoToLoadMetadata(carouselAudiosAndVideos),
-                ]).then(() => {
-                    initCarousel();
+                setTimeout(() => {
```
petrjasek: it's probably better to use `$timeout(..., 0)` this will run as soon as angular digest is over

tomaskikutis: According to the [docs](https://docs.angularjs.org/api/ng/service/$timeout) `$timeout` is a wrapper for `window.setTimeout` the difference being that it wraps the function in a try-catch and returns a promise. Are you sure there's a difference to the digest whether `window.setTimeout` or `$timeout` is used?

petrjasek: afaik it will only run in next digest, but not sure atm. there is also [scope.$applyAsync](https://docs.angularjs.org/api/ng/type/$rootScope.Scope#$applyAsync) that might do it

PR #2287, opened 2018-05-10
fix(editor3): Fix the paste suggestion on the first block of the editor

ioanpocol:


scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -371,20 +372,31 @@ const pasteAddSuggestion = (state, {content, data}) => {
...
+        focusKey: finalSelection.getStartKey(),
+        focusOffset: finalSelection.getStartOffset(),
+        hasFocus: true,
+        isBackward: false
+    });
+
+    // for the first block recover the initial block data because on replaceWithFragment the block data is
+    // replaced with the data from pasted fragment
+    if (firstBlock != null && initialSelection.getStartKey() === firstBlock.getKey()) {
+        const newSelection = initialSelection.merge({
```
tomaskikutis: Use `getCustomDataFromEditor` / `setCustomDataForEditor`

ioanpocol: Why?

tomaskikutis: Because you are doing the same, except in low-level APIs which were already abstracted away in `getCustomDataFromEditor` / `setCustomDataForEditor`.

We are already hacking stuff by storing data on the first block, which breaks stuff as we can see from this PR, so let's at least have the hackery in one place :smiley_cat:

ioanpocol: I think we have already multiple keys. Probably a new set of set get for all data is needed so in this way we hide the implementation details of getCustomDataFromEditor / setCustomDataForEditor.

tomaskikutis: That's right, I didn't notice it requires a key for argument. Then add a method in `/scripts/core/editor3/helpers/editor3CustomData.js` to get/set all data at once to keep the implementation in one place.

PR #2290, opened 2018-05-11
fix(editor3): fix pasting when something is selected in e3

petrjasek:
SDFID-402

### DIFF
--- a/scripts/core/editor3/reducers/suggestions.jsx
+++ b/scripts/core/editor3/reducers/suggestions.jsx
@@ -349,9 +349,7 @@ const createSplitParagraphSuggestion = (state, {data}) => {
  */
 const pasteAddSuggestion = (state, {content, data}) => {
     let {editorState} = state;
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/reducers/suggestions.jsx
```diff
@@ -374,8 +375,8 @@ const pasteAddSuggestion = (state, {content, data}) => {
     // store current selection for later
     const finalSelection = editorState.getSelection();
     const newSelection = initialSelection.merge({
-        anchorKey: initialSelection.getEndKey(),
```
ioanpocol: I think that only the new text should be set as add suggestion. The old text from getStartOffset to getEndOffset will be set as delete suggestion.

petrjasek: yep it is, but it has to select if from start which is where initial selection started, old text is pushed further by replaceWithFragment

PR #2292, opened 2018-05-14
[SDESK-2836] Move superdesk-planning UI components to client-core

nrvikas:
Need to update documentation for components.


package.json
```diff
@@ -95,6 +95,7 @@
     "rangy": "1.3.0",
     "raven-js": "3.22.3",
     "react": "16.2.0",
+    "react-bootstrap": "^0.32.1",
```
petrjasek: pls avoid `^` so there are no surprises..

PR #2293, opened 2018-05-14
chore - update superdesk-code-style version

petrjasek:


package.json
```diff
@@ -95,12 +95,12 @@
     "rangy": "1.3.0",
     "raven-js": "3.22.3",
     "react": "16.2.0",
+    "react-debounce-input": "^3.2.0",
```
tomaskikutis: Remove ^

petrjasek: removed all `^` from dependencies

PR #2296, opened 2018-05-14
fix(planning): add norwegian planning translation

petrjasek:


scripts/core/services/translate.js
```diff
@@ -20,19 +20,27 @@ export default angular.module('superdesk.core.translate', [
         function(gettextCatalog, config, $location, $rootScope, SESSION_EVENTS, tmhDynamicLocale) {
             $rootScope.$on(SESSION_EVENTS.IDENTITY_LOADED, (event) => {
                 if ($rootScope.$root.currentUser
-                    && gettextCatalog.strings.hasOwnProperty($rootScope.$root.currentUser.language)) {
+                    && config.profileLanguages.indexOf($rootScope.$root.currentUser.language) !== -1) {
```
pablopunk: Not specific to this commit but, I see all around the code `array.indexOf(x) !== -1` and I think it's more intuitive `array.includes(x)` and also [better sometimes](https://dev.to/adroitcoder/includes-vs-indexof-in-javascript)

petrjasek: good point, will fix that

PR #2300, opened 2018-05-16
[SDESK-2880] Don't reload carousel if items don't change

pablopunk:
Everytime a field in editor3 is updated, the autosave function executes a function that processes media fields which would trigger a reload on carousel causing an annoying flickering when you type on any other field. This solves it by not reloading items whenever they are the same as before

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
+++ b/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
@@ -7,6 +7,8 @@ import {waitForImagesToLoad, waitForAudioAndVideoToLoadMetadata} from 'core/help
 
 const carouselContainerSelector = '.sd-media-carousel__content';
 
+let previousItemsString = null;
+
 /**
  * @ngdoc directive
  * @module superdesk.apps.authoring
@@ -14,7 +16,6 @@ const carouselContainerSelector = '.sd-media-carousel__content';
  *
  * @requires $timeout
  */
-
 ItemCarouselDirective.$inject = ['$timeout'];
 export function ItemCarouselDirective($timeout) {
     return {
@@ -41,9 +42,14 @@ export function ItemCarouselDirective($timeout) {
              * otherwise carousel height is messed up
              */
             scope.$watch('items', (items) => {
-                if (!items) {
+                const itemsString = JSON.stringify(items);
+
+                // Don't execute if there are no items or they are the same as before
+                if (!items || previousItemsString === itemsString) {
                     return false;
                 }
+
+                previousItemsString = itemsString;
 
                 let field = _.find(items, (item) => !item[item.fieldId]);
 


scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
```diff
@@ -41,10 +42,15 @@ export function ItemCarouselDirective($timeout) {
              * otherwise carousel height is messed up
              */
             scope.$watch('items', (items) => {
-                if (!items) {
+                const itemsString = JSON.stringify(items);
```
petrjasek: you can use [angular.toJson](https://docs.angularjs.org/api/ng/function/angular.toJson) which would remove all angular added properties (if any), not sure if that's what @tomaskikutis was talking about, if not then ignore this

pablopunk: @petrjasek oh I didn't know about that, thanks.

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
```diff
@@ -7,14 +7,15 @@ import {waitForImagesToLoad, waitForAudioAndVideoToLoadMetadata} from 'core/help
 
 const carouselContainerSelector = '.sd-media-carousel__content';
 
+let previousItemsString = null;
```
tomaskikutis: You can't put it here, because it will break in cases where there are multiple instances of the directive in a single page(and also when one article is closed and another opened this will be cached). I think you have to attach it to `scope`.

tomaskikutis: Make sure the scope isn't shared with the controller [like this](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/datetime/reldate-directive.js#L12) or it will break when you have multiple instanced of that directive on the same controller.

pablopunk: @tomaskikutis you mean that I should put `scope: { items: '=' }` because that's already done.

tomaskikutis: scope: { itemsSinceLastCarouselRerender: '=' }

tomaskikutis: or `scope: { previousItemsString: '=' }` and then `scope.previousItemsString = itemsString;`

petrjasek: I don't think it has to be on scope, link is called for each directive instance so local variable should be fine

tomaskikutis: @petrjasek If you have two instances of the directive and you write the "local" variable, it will update it in all directives, isn't it so?

tomaskikutis: I mean if you wrote to the variable local to the `link` function than it would be really local, but if you write it and the top of the file, it's scoped to the module and isn't really local

petrjasek: yep local I mean inside link function, otherwise it's shared

tomaskikutis: If `link` is called only once per directive then I agree with your original point. I thought it was like react's `render` where you can't store local variables which live longer than a single render.

petrjasek: right it's called only once, not like render

PR #2301, opened 2018-05-16
fix(authoring): fix missing annotations

petrjasek:
it was using old storage for editor state, make it use helpers
to get it from the right place.

SDESK-2910

### DIFF
--- a/scripts/apps/authoring/authoring/authoring.spec.js
+++ b/scripts/apps/authoring/authoring/authoring.spec.js
@@ -66,69 +66,90 @@ describe('authoring', () => {
             word_count: 3,
             slugline: 'item5',
             body_html: '<p>lorem ipsum dolor</p>',
-            editor_state: [{
-                entityMap: {},
-                blocks: [{
-                    depth: 0,
-                    key: '2sso6',
-                    entityRanges: [],
-                    data: {},
-                    type: 'unstyled',
-                    text: 'lorem ipsum dolor',
-                }],
-            }],
         };
         let updates = {
-            editor_state: [{
-                entityMap: {},
-                blocks: [{
-                    depth: 0,
-                    key: '2sso6',
-                    entityRanges: [],
-                    data: {
-                        MULTIPLE_HIGHLIGHTS: {
-                            lastHighlightIds: {ANNOTATION: 2},
-                            highlightsData: {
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/authoring.spec.js
```diff
@@ -66,69 +66,90 @@ describe('authoring', () => {
...
+                            entityRanges: [],
+                            data: {
+                                MULTIPLE_HIGHLIGHTS: {
+                                    lastHighlightIds: {ANNOTATION: 2},
+                                    highlightsData: {
+                                        'ANNOTATION-1': {
+                                            type: 'ANNOTATION',
+                                            data: {
+                                                date: '2018-03-30T14:57:53.172Z',
+                                                msg: angular.toJson({
```
tomaskikutis: Use `JSON.stringify` for plain objects

scripts/apps/authoring/authoring/authoring.spec.js
```diff
@@ -66,69 +66,90 @@ describe('authoring', () => {
...
+                            type: 'unstyled',
+                            text: 'lorem ipsum dolor',
+                        }],
                     }],
-                    type: 'unstyled',
-                    text: 'lorem ipsum dolor',
-                }],
-            }],
+                },
+            },
```
tomaskikutis: I can't comment on line 156 since it hadn't been changed, but I instead of checking `updates.annotations` we need to test if the annotations are in HTML, or we will end up with the same situation as this one where annotations weren't outputted, but the tests were all green.

petrjasek: that's tested in [to-html.spec.js](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/html/tests/to-html.spec.js#L89)

tomaskikutis: A test for data needs to be added then because we can still bump into the same problem, that tests pass, but no data is present in the html, only the ids

petrjasek: I think ideally it would handle both when processing annotations to generate that html, not sure why it's separate atm, but that would involve a bit of refactoring

PR #2307, opened 2018-05-18
Annotations are not updated in the content api ninjs when correcting updating an item (SDESK 2917)

tomaskikutis:
SDESK-2917

### DIFF
--- a/scripts/apps/authoring/authoring/services/AutosaveService.js
+++ b/scripts/apps/authoring/authoring/services/AutosaveService.js
@@ -1,19 +1,15 @@
 import * as helpers from 'apps/authoring/authoring/helpers';
-import {convertFromRaw} from 'draft-js';
-import {toHTML} from 'core/editor3';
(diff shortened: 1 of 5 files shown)


scripts/apps/authoring/authoring/services/AutosaveService.js
```diff
@@ -61,6 +67,36 @@ export class AutosaveService {
         return this.timeouts[id];
     }
 
+    /**
+     * Generate item annotations field
+     *
+     * @param {Object} item
+     */
+    generateAnnotations(item) {
```
petrjasek: I don't think it should be in autosave service. imo best would be if it's handled in the editor on change

tomaskikutis: Awesome suggestion, thanks :+1: It also required less code to put it in the editor, and that's where it belongs as well.

PR #2310, opened 2018-05-21
[SDFID-388] Retrieve annotation type name for qcode

pablopunk:


scripts/core/editor3/components/annotations/AnnotationPopup.jsx
```diff
@@ -10,7 +10,9 @@ import {HighlightsPopupPresentation} from '../HighlightsPopupPresentation';
 import {UserAvatar} from 'apps/users/components/UserAvatar';
 
 const Annotation = ({annotation, editorNode, showPopup, highlightId, highlightsManager}) => {
-    const {author, avatar, date, msg, annotationType: type} = annotation.data;
+    const {author, avatar, date, msg, annotationType} = annotation.data;
+    const annotationTypes = ng.get('metadata').values.annotation_types;
```
pablopunk: @petrjasek is there any way I can avoid using `ng` to retrieve metadata values?

ioanpocol: We can add it on store when we init it.

pablopunk: So I should use `ng` when we initialize the store then?

ioanpocol: Yes, I think it is already used there.

PR #2311, opened 2018-05-22
SDESK-715 Feeding service labels and fields should be set server side, not in client

mugurrus:


scripts/apps/ingest/views/settings/service-config.html
```diff
@@ -0,0 +1,83 @@
+<div ng-repeat="field in currentFeedingService.fields track by field.id">
+    <div class="field" ng-if="field.type === 'text'" ng-show="isConfigFieldVisible(field)">
+        <label for="{{getConfigFieldId(field.id)}}" translate>{{ field.label | translate }}</label>
```
petrjasek: there should be no translate in the `<label>`

PR #2315, opened 2018-05-23
fix(editor3): Fix the delete of multiple paragraphs

ioanpocol:


tomaskikutis:
What was the issue with multiple paragraphs?

PR #2316, opened 2018-05-23
SDESK-2832 It is not possible to edit and save some CVs

mugurrus:


jerome-poisson:
closing after talking with @petrjasek and @mugurrus as string conversion may break output. We have restored old behaviour with https://github.com/superdesk/superdesk-core/pull/1274 and we'll do an other PR to specify type validation in vocabularies

PR #2317, opened 2018-05-24
[SDESK-2905] Fix issue when pasting from Google Docs on Windows

pablopunk:
I [submitted a PR](https://github.com/aem/docs-soap/pull/36) to fix the library we use to paste from gdocs but we can use my package until that's merged


package.json
```diff
@@ -56,7 +57,6 @@
     "css-loader": "0.28.10",
     "d3": "3.5.17",
     "diff-match-patch": "1.0.0",
-    "docs-soap": "1.2.0",
```
petrjasek: would be better imo to use `"docs-soap": "pablopunk/docs-soap#v1.2.1"` so we don't have to change the imports

pablopunk: @petrjasek I tried that but it throws an error because `npm run build` is not executed when downloading the repo, and when I upload it to NPM the build is already done. Maybe I should write a `"postinstall"` script?

petrjasek: build of docs-soap? I think you could build it locally, push it to your repo and then we could use it

pablopunk: @petrjasek okay I will do that

PR #2318, opened 2018-05-24
Edit embed codes in the text editor (SDFID 325)

tomaskikutis:
SDFID-325


scripts/core/helpers/ReactRenderAsync.jsx
```diff
@@ -52,4 +53,16 @@ export function connectPromiseResults(getPromises) {
             );
         };
     };
+}
+
+export function connectServices(component, services) {
```
petrjasek: seems like this one is not used

PR #2320, opened 2018-05-25
LBSD-2196: fix: adding a few css classes in order to apply individual styles per field type

eos87:
In Liveblog we need to customize some fields individually. The easiest way would be to have some particular classes per field. One example of this is that we don't use the `language` field in user's profile so we need to get rid of it.

In the past, we forked the project and commented the whole html piece. In my opinion, that makes it harder to maintain since removing the whole field from the html is something we would need to do on each `superdesk-client-core` upgrade and won't be mergeable with main project. 

So, I added a few css classes to fields in order to achieve what we need without the need of having our own `superdesk-client-core` version :)

### DIFF
--- a/scripts/apps/users/views/edit-form.html
+++ b/scripts/apps/users/views/edit-form.html
@@ -130,7 +130,7 @@
                 </div>
             </div>
 
-            <div sd-info-item class="user_role field__language form__row">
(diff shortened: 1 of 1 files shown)


eos87:
Hi guys, at the moment only `language` is hidden. I've updated the PR content. Thanks for checking :)

PR #2325, opened 2018-05-25
chore(editor3): Add unit tests for multi-block delete

ioanpocol:


scripts/core/editor3/reducers/tests/suggestions_delete_blocks.spec.jsx
```diff
@@ -0,0 +1,416 @@
...
+        type: 'unstyled',
+        text: '',
+        data: {},
+    };
+
+    let newRawContent = {
+        ...{
+            blocks: [],
+            entityMap: {},
+        },
```
pablopunk: Is that different from this?

```js
let newRawContent = {
  blocks: [],
  entityMap: {},
  ...rawContent,
};
```

PR #2334, opened 2018-05-29
chore(editor3): Add unit tests for style suggestions

ioanpocol:


scripts/core/editor3/reducers/tests/suggestion_setup.jsx
```diff
@@ -0,0 +1,141 @@
...
+    return newRawContent;
+}
+
+export function getInitialEditorState(rawContent) {
+    const content = convertFromRaw(prepareRawContent(rawContent));
+
+    return EditorState.createWithContent(content);
+}
+
+export function applySelection(editorState, startBlockIndex, startOffset, endBlockIndex, endOffset) {
```
tomaskikutis: This seems simpler to me
https://github.com/superdesk/superdesk-client-core/blob/687d4259f6bbc619d62a6961d79f744308c4cd88/scripts/core/editor3/components/tests/multipleHighlights.spec.jsx#L78

scripts/core/editor3/reducers/tests/suggestion_style.spec.jsx
```diff
@@ -0,0 +1,244 @@
...
+
+        expect(Highlights.getHighlightData(editorState, 'TOGGLE_BOLD_SUGGESTION-1')).toEqual({
+            date: date,
+            author: 'author_id',
+            type: 'TOGGLE_BOLD_SUGGESTION',
+            originalStyle: '',
+        });
+    });
+
+    it('CREATE_CHANGE_STYLE_SUGGESTION - set BOLD suggestion for a bold text', () => {
```
tomaskikutis: When creating a test case, describe the expected outcome. Write the "it" case so it's gramatically correct when you read it including "it".

`it('Should set BOLD suggestion even if the text is already bold')` or
`it('Should ignore BOLD suggestion if the text is already bold')`

The same applies to other tests like "set BOLD suggestion 2 times" where it's not specified what is the expected behaviour after setting it 2 times.

scripts/core/editor3/reducers/tests/suggestion_setup.jsx
ioanpocol: I think the current solution is fine, it cover also the selection of all blocks.

scripts/core/editor3/reducers/tests/suggestion_style.spec.jsx
```diff
@@ -53,7 +53,7 @@ describe('editor3.reducers.suggestion.CREATE_CHANGE_STYLE_SUGGESTION', () => {
         });
     });
 
-    it('CREATE_CHANGE_STYLE_SUGGESTION - set BOLD suggestion for a bold text', () => {
+    it('should set BOLD suggestion for a bold text, the BOLD style not present anymore for selected text', () => {
```
tomaskikutis: Sorry for picking on the details, but I think it important to have correct and properly structured sentences

* it should create a `BOLD` suggestion
* it should create a `REMOVE-BOLD` suggestion when toggling `BOLD` on text which was already bold before enabling suggestion mode
* it should not add any suggestions after toggling 'BOLD' twice
* ...

scripts/core/editor3/reducers/tests/suggestion_style.spec.jsx
```diff
@@ -141,7 +141,7 @@ describe('editor3.reducers.suggestion.CREATE_CHANGE_STYLE_SUGGESTION', () => {
         }
     });
 
-    it('CREATE_CHANGE_STYLE_SUGGESTION - set BOLD partially overlapping suggestions', () => {
+    it('should overwrite overlapped section when set 2 BOLD overlapped suggestions', () => {
```
tomaskikutis: Clarify the description, I can't understand what does it do from the description alone

scripts/core/editor3/reducers/tests/suggestion_style.spec.jsx
ioanpocol: Fix is in the next PR

scripts/core/editor3/reducers/tests/suggestion_style.spec.jsx
ioanpocol: fix is in the next PR

PR #2340, opened 2018-05-31
Editor3 pasting links swaps the order of lines (SDESK 2965)

tomaskikutis:
SDESK-2965


tomaskikutis:
I've sent the PR to draft as well, let's see if they merge it. Anyway, I think we should definitely come up with a workflow to easily fork the libraries since it's easier to add a single line(or two) in the library itself than to spend hours or days dancing around the public API which is not always possible.

pablopunk:
I agree

PR #2341, opened 2018-05-31
Add unresolved comments to sidebar widget

pablopunk:
The widget has now 2 tabs, `UNRESOLVED | RESOLVED` so you can see all
unresolved changes too.

One thing to keep in mind is that unresolved comments can be seen in the widget but you cannot execute any action on them. I'm not sure if that's gonna be useful for the user but it's something to keep in mind.

SDFID-347


pablopunk:
If you're going to review this PR I recommend turning the white space diff off, it's way easier. Toggle this ON:

captura de pantalla 2018-06-01 a las 10 33 48

PR #2343, opened 2018-06-01
Close tag instead of self-closing div

pablopunk:
Not sure if this is the problem, I'll check fireq build when it's ready and see if it's fixed.

SDESK-2962


petrjasek:
we can also try some tool like [htmlhint](https://www.npmjs.com/package/htmlhint)

pablopunk:
@petrjasek that sounds better

pablopunk:
I'm merging this as it's a high priority quick fix and we should make a task for htmlhint

PR #2351, opened 2018-06-06
Editor3 crashes when clicking on annotation after reloading the page with article already opened (SDESK 2991)

tomaskikutis:
SDESK-2991

The issue was that metadata was accessed before being initialized. It was probably also a race condition, because I couldn't always reproduce it and it explains why it worked when freshly opening the item vs refreshing the page with the item already shown.

### DIFF
--- a/scripts/core/editor3/components/annotations/AnnotationInput.jsx
+++ b/scripts/core/editor3/components/annotations/AnnotationInput.jsx
@@ -188,12 +188,8 @@ const mapStateToProps = (state) => ({
 });
 
 const AnnotationInputBodyWithDependenciesLoaded = connectPromiseResults(() => ({
-    annotationTypes: new Promise((resolve) => {
-        ng.get('metadata').initialize()
-            .then(() => {
-                resolve(ng.get('metadata').values.annotation_types);
-            });
-    }),
+    annotationTypes: ng.get('metadata').initialize()
+        .then(() => ng.get('metadata').values.annotation_types),
 }))(AnnotationInputBody);
 
 export const AnnotationInput = connect(mapStateToProps, {
--- a/scripts/core/editor3/components/annotations/AnnotationPopup.jsx
+++ b/scripts/core/editor3/components/annotations/AnnotationPopup.jsx
@@ -81,12 +81,8 @@ Annotation.propTypes = {
 };
 
 const AnnotationWithDependenciesLoaded = connectPromiseResults(() => ({
-    annotationTypes: new Promise((resolve) => {
-        ng.get('metadata').initialize()
-            .then(() => {
-                resolve(ng.get('metadata').values.annotation_types);
-            });
-    }),
+    annotationTypes: ng.get('metadata').initialize()
+        .then(() => ng.get('metadata').values.annotation_types),
 }))(Annotation);
 
 export const AnnotationPopup = connect(


scripts/core/editor3/components/annotations/AnnotationInput.jsx
```diff
@@ -183,13 +179,23 @@ AnnotationInputBody.propTypes = {
...
 };
 
 const mapStateToProps = (state) => ({
     language: state.item.language,
     spellcheckerEnabled: state.spellcheckerEnabled,
 });
 
+const AnnotationInputBodyWithDependenciesLoaded = connectPromiseResults(() => ({
+    annotationTypes: new Promise((resolve) => {
+        ng.get('metadata').initialize()
```
ioanpocol: IMO it is better to add annotation types on redux store.

tomaskikutis: From the editor's point of view, it's like a config parameter and can't be modified, so I'd rather just treat it as a dependency.

ioanpocol: I think that the dependency on 'ng' all over the editor3 code is not good.
IMO it is better to keep it in store even if it is not changed.

petrjasek: this should be a promise already, wouldn't this work?
```
annotationTypes: ng.get('metadata').initialize().then(() => ng.get('metadata').values.annotation_types)
```

tomaskikutis: Yep, I guess it would. Will commit a fix

ioanpocol: For testing I think it is better to have a better control on those dependencies, to be able to init them on tests. Also gettext and modal can be sent as params to Editor3 component.

tomaskikutis: For testing, you can easily export the component without dependencies connected and you can pass any prop you want instead of that dependency.

Regarding `gettext` and `modal`, you wouldn't like passing the props 10 levels down. It creates pseudo-coupling and is a pain to refactor. We aren't importing all the modules in the root component and passing them all the way down to grand-grand-children, why should we do it to other dependencies, even if they aren't JS modules?

ioanpocol: Whatever, this dependency on ng is not good. Probably we can create an adapter and put in that adapter all calls to ng; in this way we will have only one place to change anything that is related to ng.

ioanpocol: Adapter means to apply classic adapter pattern; ng is an external dependency that appear all over the code.

tomaskikutis: I've read on the pattern, but I don't see what problem would it solve. According to wiki, adapter is used to make two incompatible things work together or be able to swap the implementation. For swapping the implementation we can modify the original service `ng` refers to and I'm not aware of any incompatibilities we would need an adapter for.

Could reply on this?
> Also, what do you see wrong with ng? It's already "the one place" to access services so it doesn't make sense to me to abstract it more.

ioanpocol: I think I already did:
>ng is an external dependency that appear all over the code.
... and it looks like a hack so it is better if we isolate it.
But if you don't agree with it, never mind. It is a too small issue to spend so much time on it.

tomaskikutis: I agree that `ng` is a bit hacky because of being tightly coupled with angular and requiring it to be loaded to use any services, but that aside it doesn't seem to cause problems. Do you have any other issues in mind?

The reason I'm asking you questions is that we still don't have a clear way of sharing code between React and Angular and I'm thinking about different ways how that can be implemented.

You mentioned some ideas which don't seem right to me, so I wanted to understand why do you think this way.

I've been [working](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/helpers/ReactRenderAsync.jsx) on Angular > React integration(at least for services) and I'm quite happy with the result which allows connecting any angular service as react prop, or any promise to prop.

One further improvement I'm thinking about is loading all async child dependencies before rendering the parent, so the child components can be treated as synchronous which makes it easier to work with and it doesn't blink while rendering. We could use something like https://github.com/ctrlplusb/react-tree-walker I'm not sure about this one, but it looks interesting.

What do you think?

@pablopunk @pavlovicnemanja  @petrjasek feel free to join the discussion

PR #2352, opened 2018-06-06
[VOCABULARY] added error message when a required field is missing

jerome-poisson:
SDESK-2923

### DIFF
--- a/scripts/apps/vocabularies/controllers/VocabularyEditController.js
+++ b/scripts/apps/vocabularies/controllers/VocabularyEditController.js
@@ -3,6 +3,7 @@ import _ from 'lodash';
 VocabularyEditController.$inject = [
     '$scope',
     'gettext',
+    '$interpolate',
     'notify',
     'api',
     'vocabularies',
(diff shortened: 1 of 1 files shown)


scripts/apps/vocabularies/controllers/VocabularyEditController.js
```diff
@@ -28,6 +29,11 @@ export function VocabularyEditController($scope, gettext, notify, api, vocabular
             if (angular.isDefined(response.data._issues['validator exception'])) {
                 notify.error(gettext('Error: ' +
                                      response.data._issues['validator exception']));
+            } else if (response.data._issues.error.required_field) {
+                let params = response.data._issues.params;
+
+                notify.error($interpolate(gettext(
+                    'Required {{field}} in item {{item}}'))({field: params.field, item: params.item}));
```
tomaskikutis:
```
`How about ${native_es6_interpolation} ?`
```

petrjasek: not sure if that works with gettext string extraction

tomaskikutis: Didn't see that. I think it might not.

PR #2353, opened 2018-06-06
chore(core): Update version of UI framework

pavlovicnemanja:


petrjasek:
can you remove that `dump.rdb` file?

pavlovicnemanja:
@petrjasek sorry on that

PR #2367, opened 2018-06-12
Media Contacts refactoring

fritzSF:
Media contacts moved to the new photo grid instead of sone to be deprecated 'mgrid'. 
Plus: style changes in vocabularies and search tags.


fritzSF:
@sjunaid Please check if the changes I made will affect the use of contacts in Planning, the structure changed and some 'keys' where removed.

sjunaid:
:+1:, LGTM, works in planning.

PR #2369, opened 2018-06-12
SDESK-3023 [INGEST] Ingest source creation: Configure restriction for feed parser dropdown per feeding service

ride90:
Depends on [server side pr](https://github.com/superdesk/superdesk-core/pull/1302)


ride90:
Instance for testing https://sdpr-2909.test.superdesk.org

PR #2370, opened 2018-06-12
fix(editor3): Fix reset highlight data on regular paste on editor3(SDESK-3027, SDESK-2942)

ioanpocol:


scripts/core/editor3/components/handlePastedText.js
```diff
@@ -121,22 +123,20 @@ function processPastedHtml(props, html) {
     );
 
     const selectionAfterInsert = nextEditorState.getSelection();
-
-    nextEditorState = EditorState.set(nextEditorState, {allowUndo: false});
```
tomaskikutis: Does undo/redo work properly after removing this?

ioanpocol: yes

PR #2371, opened 2018-06-12
Display commented text on unresolved comments inside widget

pablopunk:
We don't save commented text on each comment before they are resolved
(as the highlighted text can change in the editor) so these comments
were lacking the "commented text" property in the view.

Now we search inside the editor content for highlighted text for each
comment. It was a bit tricky to do as we sometimes have orphan comments
(their text was deleted but they were not) that could appear on the
sidebar widget but not on the editor.

SDFID-441

### DIFF
--- a/scripts/apps/authoring/track-changes/inline-comments.js
+++ b/scripts/apps/authoring/track-changes/inline-comments.js
@@ -46,19 +46,17 @@ function getCommentsFromField(customFields, resolved = true) {
     if (resolved) {
         return (obj) => ({
             fieldName: getLabelForFieldId(getFieldId(obj.contentKey), customFields),
-            comments:
-        getCustomDataFromEditorRawState(
-            obj[fieldsMetaKeys.draftjsState],
-            editor3DataKeys.RESOLVED_COMMENTS_HISTORY
-        ) || [],
+            comments: getCustomDataFromEditorRawState(
+                obj[fieldsMetaKeys.draftjsState],
+                editor3DataKeys.RESOLVED_COMMENTS_HISTORY
+            ) || [],
         });
     } else {
         return (obj) => {
-            const highlightsObject =
-        getCustomDataFromEditorRawState(
-            obj[fieldsMetaKeys.draftjsState],
-            editor3DataKeys.MULTIPLE_HIGHLIGHTS
-        ).highlightsData || {};
+            const highlightsObject = getCustomDataFromEditorRawState(
+                obj[fieldsMetaKeys.draftjsState],
+                editor3DataKeys.MULTIPLE_HIGHLIGHTS
+            ).highlightsData || {};
 
             for (const id in highlightsObject) {
                 // Add id to highlight so we can retrieve data from the state
--- a/scripts/apps/authoring/track-changes/inline-comments.js
+++ b/scripts/apps/authoring/track-changes/inline-comments.js
@@ -102,7 +102,9 @@ function InlineCommentsCtrl($scope, userList, metadata, content) {
                 fieldName: getLabelForFieldId(getFieldId(contentKey)),
                 comments: comments,
             };
-        });
+        })
+            .filter((obj) => obj.comments.length > 0);
+
 
         const allComments = []
             .concat(...resolvedComments.map((obj) => obj.comments))
--- a/scripts/apps/authoring/track-changes/inline-comments.js
+++ b/scripts/apps/authoring/track-changes/inline-comments.js
@@ -13,6 +13,7 @@ import {
     getRangeAndTextForStyleInRawState,
 } from 'core/editor3/helpers/highlights';
 import {highlightsConfig} from 'core/editor3/highlightsConfig';
+import {getCustomMetadata} from 'core/editor3/helpers/editor3CustomData';
 
 function getAllUserIdsFromComments(comments) {
     const users = [];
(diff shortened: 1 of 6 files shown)


scripts/apps/authoring/track-changes/inline-comments.js
```diff
@@ -86,6 +111,48 @@ function InlineCommentsCtrl($scope, userList, metadata, content) {
             return;
         }
 
+        unresolvedComments = unresolvedComments.map(({fieldId, comments, ...rest}) => {
```
tomaskikutis: I see you didn't take my advice to check out how `generateAnnotations` is implemented and decided to write it from scratch :grin: `generateAnnotations` calls `getCustomMetadata` which already has all the crawling the `ContentState` for unique stylenames implemented. Actually you would need to refactor it and extract annotation specific stuff to a different function, but otherwise it's 90% ready. I think it'd be best to drop this block of code and retrieve comments from content using `getCustomMetadata`.

scripts/core/editor3/helpers/highlights.js
```diff
@@ -845,6 +844,24 @@ export function getRangeAndTextForStyle(editorState, style) {
     };
 }
 
+export function getRangeAndTextForStyleInRawState(rawEditorState, highlightId) {
```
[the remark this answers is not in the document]
pablopunk: I tried to use an actual editor state by creating one from scratch based on the raw state. It didn't work because our functions try to get the selection (which doesn't exist on the created state). Otherwise I would use the one we already have (`getRangeAndTextForStyle`)

scripts/apps/authoring/track-changes/views/inline-comments-widget.html
```diff
@@ -26,7 +26,7 @@ <h4 class="label label--large">{{commentObj.fieldName}}</h4>
                             </div>
                         </li>
                     </ul>
-                    <div ng-show="!!item.data.commentedText" class="commented-text">
+                    <div ng-if="!!item.data.commentedText" class="commented-text">
```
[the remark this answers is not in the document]
pablopunk: you're right, it should always be populated after this PR

scripts/apps/authoring/track-changes/inline-comments.js
```diff
@@ -38,19 +46,36 @@ function getCommentsFromField(customFields, resolved = true) {
...
-        return (obj) => ({
-            fieldName: getLabelForFieldId(getFieldId(obj.contentKey), customFields),
-            comments: Object.values(getCustomDataFromEditorRawState(
-                obj[fieldsMetaKeys.draftjsState],
-                editor3DataKeys.MULTIPLE_HIGHLIGHTS
-            ).highlightsData || {}).filter((h) => h.type === 'COMMENT'),
-        });
+        return (obj) => {
+            const highlightsObject =
+        getCustomDataFromEditorRawState(
```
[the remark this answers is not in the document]
pablopunk: That was probably a plugin I have for [prettier](https://github.com/prettier/prettier). I don't know why it would do that but both prettier and eslint consider it as "good" indentation.

scripts/apps/authoring/track-changes/inline-comments.js
```diff
@@ -38,19 +46,36 @@ function getCommentsFromField(customFields, resolved = true) {
...
+            obj[fieldsMetaKeys.draftjsState],
+            editor3DataKeys.MULTIPLE_HIGHLIGHTS
+        ).highlightsData || {};
+
+            for (const id in highlightsObject) {
+                // Add id to highlight so we can retrieve data from the state
+                highlightsObject[id].highlightId = id;
```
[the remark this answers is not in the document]
pablopunk: I use it down below to retrieve the highlightedText for all comments

scripts/apps/authoring/track-changes/inline-comments.js
```diff
@@ -38,19 +46,36 @@ function getCommentsFromField(customFields, resolved = true) {
     if (resolved) {
```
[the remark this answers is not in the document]
pablopunk: Agree

scripts/apps/authoring/track-changes/inline-comments.js
pablopunk: I did try to do it but as I explained above it didn't work for me.

tomaskikutis: Don't make any changes for a while now, I'll try to fork your PR tomorrow and see if I can make it work using `getCustomMetadata`

tomaskikutis: I've refactored the code to get comments from text the same way as annotations https://github.com/tomaskikutis/superdesk-client-core/commit/944a5e032c85b8343495ef422686bd7dc7fa0c30

Merge this, and this PR is almost good. One more thing I'd like you to look into is if you can remove `commentedText` from `highlight.data`. It's better to have consistent data structures and not modify them if you need an additional property. It's better to add that additional property outside of a common data structure
```
{
    "highlight": getHighlight(),
    "comment-specific-properties": {
        "commentedText": "abc"
    }
}
```

pablopunk: Okay, thanks @tomaskikutis. About the commentedText, I just copied the [structure we have for resolved comments](https://github.com/superdesk/superdesk-client-core/blob/20a5ee27b2c2218ea048f978cb54ac128a125fa3/scripts/core/editor3/components/comments/CommentPopup.jsx#L99). If I change this one I should change both.

tomaskikutis: Let's keep it then. Let's also keep it immutable and not modify the original highlight object. I think I already did that after forking this PR.

pablopunk:
@tomaskikutis yeah I usually rebase too, but this time I couldn't solve the conflicts of the rebase for some reason, maybe I had something wrong with my git. So I did reset and merge, and that got it working for some reason.

PR #2373, opened 2018-06-13
fix(editor3): change bg color for remove suggestions spaces to red

petrjasek:
SDFID-443

### DIFF
--- a/scripts/core/editor3/components/invisibles/SpaceComponent.jsx
+++ b/scripts/core/editor3/components/invisibles/SpaceComponent.jsx
@@ -1,8 +1,22 @@
 import React from 'react';
+import {get} from 'lodash';
 import PropTypes from 'prop-types';
 
 export const SpaceComponent = (props) => {
     const spanProps = {...props};
+    const styleSet = get(props, 'children.0.props.styleSet');
+    let className = 'space-component';
+
+    if (styleSet) {
+        const isDeleteSuggestion = styleSet.some((style) => style.startsWith('DELETE_SUGGESTION'));
+        const isAddSuggestion = styleSet.some((style) => style.startsWith('ADD_SUGGESTION'));
+
+        if (isDeleteSuggestion) {
+            className += ' delete-suggestion';
+        } else if (isAddSuggestion) {
+            className += ' add-suggestion';
+        }
+    }
 
     // Prevent unknown properties warning
     delete spanProps.contentState;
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/styles.scss
```diff
@@ -526,6 +526,10 @@ $editor-styleButton-size: 3rem;
 	[style^="color"] { // used for suggestions
 		background-color: rgba(0, 180, 0, 0.2);
 	}
+
+	[style^="color: rgb(255"] { // used for delete suggestions
```
tomaskikutis: That's a bit dirty, is it hard to add a class name?

petrjasek: it's added via highlights, afaik there is no way to add class for this

tomaskikutis: And it totally depends on `SpaceDecorator`. Speaking of `SpaceDecorator`, you can edit its renderer - `SpaceComponent` and add classes for suggestions. Shouldn't be too hard :)

scripts/core/editor3/components/invisibles/SpaceComponent.jsx
```diff
@@ -1,16 +1,29 @@
...
 import PropTypes from 'prop-types';
 
 export const SpaceComponent = (props) => {
     const spanProps = {...props};
+    const styleSet = get(props, 'children.0.props.styleSet');
+    let className = 'space-component';
+
+    if (styleSet) {
+        styleSet.forEach((style) => {
+            if (style.startsWith('DELETE_SUGGESTION')) {
```
tomaskikutis: Don't you need to check if there is already the same class added?

petrjasek: what do you mean?

tomaskikutis: You are adding the className in the loop, which means you might add the same class multiple times. I think it might be better to avoid `forEach` altogether here

```
const deleteSuggestionExists = styleSet.filter((style) => style.startsWith('DELETE_SUGGESTION'));
const addSuggestionExists = styleSet.filter((style) => style.startsWith('ADD_SUGGESTION'));

if (deleteSuggestionExists) {
    className += ' delete-suggestion';
} else if (addSuggestionExists) {
    className += ' add-suggestion';
}
```

petrjasek: I see, I don't think there can be more than one of these suggestions at a time, but it won't hurt to be sure

PR #2377, opened 2018-06-14
fix(preview): fix missing expand icon in preview

petrjasek:
fixing markup issue and forcing template reload via webpack

SDESK-3018


scripts/apps/archive/directives/MediaPreview.js
```diff
@@ -2,7 +2,7 @@ MediaPreview.$inject = ['api', '$rootScope', 'desks', 'superdesk'];
 
 export function MediaPreview(api, $rootScope, desks, superdesk) {
     return {
-        templateUrl: 'scripts/apps/archive/views/preview.html',
+        template: require('../views/preview.html'),
```
tomaskikutis: Why that is needed, doesn't template re-render on its own?

petrjasek: it is to avoid using any cached version

PR #2379, opened 2018-06-14
Cannot edit annotations (SDESK 3044)

tomaskikutis:
SDESK-3044


pablopunk:
How come this was working before?

tomaskikutis:
I made a new argument to be required in recent weeks, so it wasn't working since.

PR #2385, opened 2018-06-15
SDESK-2483 [USER PROFILE] Add new values to author's profile

ride90:


ride90:
Instance for testing with server side included https://sdpr-2914.test.superdesk.org/

scripts/apps/users/views/edit-form.html
```diff
@@ -221,6 +221,50 @@ <h5>{{ user.username}}</h5>
                 </div>
             </div>
 
+            <div sd-info-item class="form__row" ng-if="privileges.users">
+                <div class="sd-line-input">
+                    <label class="sd-line-input__label" for="facebook">{{ 'Facebook'|translate }}</label>
```
petrjasek: you can use `::` for translations, eg. `{{ :: 'Facebook' | translate }}` - it will only eval this once and not on every digest

scripts/apps/users/views/edit-form.html
```diff
@@ -221,6 +221,50 @@ <h5>{{ user.username}}</h5>
...
+                              ng-disabled="!user.is_author"
+                              sd-auto-height></textarea>
+                    <div class="sd-line-input__hint">{{ 'e.g. superdeskman from https://www.facebook.com/superdeskman'|translate }}</div>
+                </div>
+            </div>
+
+            <div sd-info-item class="form__row" ng-if="privileges.users">
+                <div class="sd-line-input">
+                    <label class="sd-line-input__label" for="instagram">{{ 'Instagram'|translate }}</label>
+                    <textarea id="instagram"
```
petrjasek: I think all those could be inputs, it's all one line values right?

ride90:
@petrjasek pr was updated.

PR #2387, opened 2018-06-18
Not possible to edit video embed code (SDFID 436)

tomaskikutis:
SDFID-436
DEPENDS ON https://github.com/superdesk/superdesk-core/pull/1306


scripts/core/helpers/generic.js
```diff
@@ -0,0 +1,20 @@
+// https://gist.github.com/jed/982883
+export function uuid(a) {
```
petrjasek: I would call this file rather `helpers/uuid.js`, generic doesn't say much about what is in there

tomaskikutis: I was planning to dedicate the file to stateless functions without context, but it may be a better idea to just name it `uuid` and we'll create some folders for discoverability when there are too much files.

PR #2391, opened 2018-06-19
On switching the language from user s profile the hamburger menu still retains the previously set language unless relogin happens (SDESK 3057)

tomaskikutis:
SDESK-3057

### DIFF
--- a/scripts/apps/archive/views/archived-kill.html
+++ b/scripts/apps/archive/views/archived-kill.html
@@ -1,7 +1,7 @@
 <div sd-modal data-model="item">
     <div class="modal__header">
         <a href="" class="close" ng-click="cancel()"><i class="icon-close-small"></i></a>
-        <h3 class="modal__heading">{{ :: action === 'kill' && 'Kill Item' || 'Takedown Item' | translate}}</h3>
+        <h3 class="modal__heading">{{action === 'kill' && 'Kill Item' || 'Takedown Item' | translate}}</h3>
     </div>
 
     <div class="modal__body">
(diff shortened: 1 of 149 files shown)


MarkLark86:
How does this go in terms of the performance of the site? Removing one way binding will add an Angular watch for each value, increasing the digest times each cycle.

tomaskikutis:
I'll ask QA to test the performance. Most of the big lists use React, so I hope that there will not be too many watches at any given time.

tomaskikutis:
There's little data on the master for QA to test, so I've checked myself on ansa test instance. I didn't notice any slowdown. A quick profiling didn't show clear differences between flamegraphs either.

pablopunk:
If you're changing one-way bindings that means it was happening for all strings and not only the menu? Or it's just for consistency?

IMO if you change the language it should be "acceptable" to logout and login again, I'm guessing it's not a thing you do every day. But that's more of a decision for the product owner.

PS: I like your naming for the branch 😁

tomaskikutis:
> If you're changing one-way bindings that means it was happening for all strings and not only the menu? Or it's just for consistency?

The only known place is the menu. Petr said one way bindings were put "just in case" not that there were any performance problems they solved. As we can see from the existence of the PR, they can cause issues and if removing them doesn't affect performance, it seems to me that's it's worth to get rid of them and remove a potencial source of bugs.

petrjasek:
actually it shouldn't be problem outside the menu, at least not within ng-view because those parts will get reloaded when you will navigate from the user preferences, but the menu stays there hidden all the time, so it would be enough to fix it there.

tomaskikutis:
I thought about the risks again and decided to go with a simpler approach of simply reloading the page. Not because of performance issues, but mostly due to the possibility that there might be places where business logic implicitly relies upon data not updating.

Will continue here https://github.com/superdesk/superdesk-client-core/pull/2404

PR #2406, opened 2018-06-21
fix(editor3): Add support for undo on table editing

ioanpocol:


scripts/core/editor3/components/Editor3Component.jsx
```diff
@@ -428,7 +428,7 @@ export class Editor3Component extends React.Component {
 
                             const selectionRect = getVisibleSelectionRect(window);
 
-                            if (selectionRect != null) {
+                            if (this.editorNode.dataset != null && selectionRect != null) {
```
tomaskikutis: Better check if `this.editorNode != null`, because if it isn't, DOM node will always have `dataset`

scripts/core/editor3/components/tables/TableCell.jsx
```diff
@@ -30,13 +30,35 @@ export class TableCell extends Component {
...
+        }
+
         if (e.ctrlKey && e.key === 'l') {
             return 'toggle-link';
         }
 
         return getDefaultKeyBinding(e);
     }
 
+    componentWillReceiveProps(nextProps) {
```
tomaskikutis: `componentWillReceiveProps` is deprecated, use `UNSAFE_componentWillReceiveProps` if you really need it.

ioanpocol: UNSAFE_componentWillReceiveProps is available in 16.3. We are still using 16.2
There are another 14 places where componentWillReceiveProps is used so I think we have to fix all of them in a single PR.

PR #2407, opened 2018-06-21
[AUTHORING] [VOCABULARIES] added a field to set width in pixels of th…

jerome-poisson:
…e popup

when creating a new controlled vocabulary, a new field is now available
to set size in pixels of the popup window. If this field is empty,
default value is used.

SDANSA-172


tomaskikutis:
products-width-of-popup

If that is the problem, how about allowing multi-line text instead of custom width?

jerome-poisson:
@tomaskikutis this is not what is requested by ANSA, but I'll check with @karelpetrak .

jerome-poisson:
@tomaskikutis just checked with @karelpetrak , your proposal makes sense but we stick with width in pixels:

> @karelpetrak afaik ANSA wants to save space and have it "nice" - i.e. one line per item
> and nobody has such long texts but them (yet)
> @jerome-poisson so we stick with the width in pixels ?
> @karelpetrak yes, at least for now

PR #2413, opened 2018-06-26
Rtl support in the editor (SDANSA 166)

tomaskikutis:
SDANSA-166


petrjasek:
shouldn't we just import Draft.css ?

tomaskikutis:
I saw some unrelated styling and decided to only include text direction, but after checking again I see that it's only a border and a few positioning, so it might be better to import it and overwrite if needed. Will update the PR.

tomaskikutis:
@petrjasek, I've addressed your comment in https://github.com/superdesk/superdesk-client-core/pull/2421

PR #2414, opened 2018-06-26
Correctly assign default options to fix cropping issues

pablopunk:
We had some issues already because we assign default options with `option || true` which is fine for objects but not for booleans.

This commit fixes all the errors related to those default options while cropping, which were causing the error of the "Edit button" not working at all for body images

SDESK-3100

### DIFF
--- a/scripts/apps/authoring/authoring/services/RenditionsService.js
+++ b/scripts/apps/authoring/authoring/services/RenditionsService.js
@@ -75,13 +75,11 @@ export function RenditionsService(metadata, $q, api, superdesk, _) {
             }
 
             const cropOptions = { // default options
-                ...{
-                    isNew: true,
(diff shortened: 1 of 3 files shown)


scripts/apps/authoring/authoring/services/RenditionsService.js
```diff
@@ -74,17 +74,28 @@ export function RenditionsService(metadata, $q, api, superdesk, _) {
                 withRatio = self.renditions;
             }
 
+            const cropOptions = { // default options
+                ...{
```
tomaskikutis: Why do you need to use the spread operator here instead of listing the properties?

pablopunk: @tomaskikutis you're right, my bad

PR #2417, opened 2018-06-26
fix(translation): Fixed spike translation issue and issue with selecting in typeahead

pavlovicnemanja:


petrjasek:
you have to add `, 'Confirm')` to https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/archive/tests/archive.spec.js#L286 and https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/archive/tests/archive.spec.js#L305

petrjasek:
and https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/archive/tests/archive.spec.js#L265

PR #2420, opened 2018-06-26
fix(editor3): Apply Tansa spellchecker on media description

ioanpocol:


scripts/core/editor3/helpers/tansa.js
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * @param {EditorState} editorState
+ * @param {ContentState} contentState
+ * @param {Block} block
+ * @param {String} text
+ * @returns {ContentState}
+ */
+function updateText(editorState, content, block, newText) {
+    const text = block.getText();
+    const diffMatchPatch = new DiffMatchPatch();
```
pablopunk: I'm not sure how DiffMatchPatch works but shouldn't we create only one object outside of the function instead of one object per function call? It seems more efficient

scripts/core/editor3/helpers/tansa.js
```diff
@@ -0,0 +1,201 @@
...
+import DiffMatchPatch from 'diff-match-patch';
+
+/**
+ * Create a custom html for Tansa
+ *
+ * @param {EditorState} editorState
+ * @returns {String}
+ */
+export function getTansaHtml(editorState) {
+    const content = editorState.getCurrentContent();
```
pablopunk: How can this work if you're calling the function like this inside getHtmlForTansa: `getTansaHtml({ state })`, wouldn't it throw an error about `getCurrentContent` not being defined?

ioanpocol: On getHtmlForTansa is called getTansaHtml with EditorState as parameter: getTansaHtml(editorState).

pablopunk: Yes but you're doing:
```js
const editorState = {state};

getTansaHtml(editorState);
```

 so I guess it's like

```js
getTansaHtml({state})
```

right?

ioanpocol: Yes, :), the {} were not in the right place

PR #2425, opened 2018-06-29
fix(editor3): Fix adding insert suggestion on superscript text

ioanpocol:


scripts/core/editor3/helpers/highlights.js
```diff
@@ -88,6 +88,10 @@ function setHighlightsState(editorState, hightlightsState) {
     return setCustomDataForEditor(editorState, editor3DataKeys.MULTIPLE_HIGHLIGHTS, hightlightsState);
 }
 
+export function isHighlightType(styleName) {
+    return styleName.lastIndexOf('-') !== -1;
```
tomaskikutis: Can we use `highlightTypeValid` from `core/scripts/core/editor3/helpers/highlights.js` ?

PR #2426, opened 2018-06-29
feat(templates): Added check if template is set as default on desk

pavlovicnemanja:
Check before deleting template if template is assigned as default on desk settings

### DIFF
--- a/scripts/apps/templates/directives/TemplatesDirective.js
+++ b/scripts/apps/templates/directives/TemplatesDirective.js
@@ -269,11 +269,25 @@ export function TemplatesDirective(gettext, notify, api, templates, modal, desks
             });
 
             function setupContentType(type) {
-                $scope.schema = content.schema(type, $scope.item.type);
-                $scope.editor = content.editor(type, $scope.item.type);
+                if (type) {
+                    $scope.schema = content.schema(type, $scope.item.type);
+                    $scope.editor = content.editor(type, $scope.item.type);
+                }
             }
 
             $scope.remove = function(template) {
+                const desks = _.filter($scope.desks._items, (desk) =>
+                    desk.default_content_template === template._id);
+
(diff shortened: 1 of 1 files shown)


scripts/apps/templates/directives/TemplatesDirective.js
```diff
@@ -269,11 +269,29 @@ export function TemplatesDirective(gettext, notify, api, templates, modal, desks
...
+                }
             }
 
             $scope.remove = function(template) {
+                let deskTemplates = _.filter($scope.desks._items, (desk) =>
+                        desk.default_content_template === template._id),
+                    deskNames = ' ';
+
+                if (!_.isEmpty(deskTemplates)) {
+                    _.forEach(deskTemplates, (desk, index) => {
```
petrjasek: I would prefer functional style here:
```js
const deskNames = deskTemplates.map((desk) => desk.name).join(', ');
```
and then when concatenating the strings:
```js
` ${deskNames}.` +
```

scripts/apps/templates/directives/TemplatesDirective.js
```diff
@@ -269,11 +269,29 @@ export function TemplatesDirective(gettext, notify, api, templates, modal, desks
...
-                $scope.schema = content.schema(type, $scope.item.type);
-                $scope.editor = content.editor(type, $scope.item.type);
+                if (type) {
+                    $scope.schema = content.schema(type, $scope.item.type);
+                    $scope.editor = content.editor(type, $scope.item.type);
+                }
             }
 
             $scope.remove = function(template) {
+                let deskTemplates = _.filter($scope.desks._items, (desk) =>
```
petrjasek: this should be imo called desks and it should be a constant:
```js
const desks = ...
```

pavlovicnemanja:
Why go simple when we can do it in difficult way? :)

PR #2429, opened 2018-07-02
Preview API

tomaskikutis:
SDESK-3095
SDESK-3096
SDESK-3097
DEPENDS ON: https://github.com/superdesk/superdesk-core/pull/1316

### DIFF
--- a/scripts/apps/authoring/authoring/previewModal.jsx
+++ b/scripts/apps/authoring/authoring/previewModal.jsx
@@ -6,23 +6,38 @@ import {ModalHeader} from 'core/ui/components/Modal/ModalHeader';
 import {ModalBody} from 'core/ui/components/Modal/ModalBody';
 import {ModalFooter} from 'core/ui/components/Modal/ModalFooter';
 
+function getFormattedDocument(url) {
+    return new Promise((resolve) => {
+        var xhr = new XMLHttpRequest();
+
+        xhr.addEventListener('load', function() {
+            resolve({
+                fomattedDocument: this.responseText,
+                documentContentType: this.getResponseHeader('content-type'),
+            });
+        });
+
+        xhr.open('GET', url);
+        xhr.send();
+    });
+}
+
 export class PreviewModal extends React.Component {
     openPreviewForItem(subscriberId, format, endpointUrl) {
-        const {api, documentId, gettext} = this.props;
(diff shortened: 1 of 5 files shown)


scripts/apps/authoring/authoring/previewModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+export class PreviewModal extends React.Component {
+    openPreviewForItem(subscriberId, format, endpointUrl) {
+        const {urls, documentId, gettext} = this.props;
+
+        const nextWindow = window.open();
+
+        nextWindow.document.body.innerText = gettext('Loading preview');
+
+        const url = urls.item('format-document-for-preview')
+            + `?subscriber_id=${subscriberId}&formatter=${format}&document_id=${documentId}`;
```
petrjasek: I think it should send full item, otherwise you would have to save before you can preview

tomaskikutis: I don't have a full item nearby. There is `scope.item`, but it has missing properties compared to the object in the database. Let's require saving for now. I've put a message asking to save it in case previewing of an unsaved item is attempted.

scripts/apps/authoring/authoring/previewModal.jsx
```diff
@@ -0,0 +1,115 @@
...
+import PropTypes from 'prop-types';
+
+import {Modal} from 'core/ui/components/Modal/Modal';
+import {ModalHeader} from 'core/ui/components/Modal/ModalHeader';
+import {ModalBody} from 'core/ui/components/Modal/ModalBody';
+import {ModalFooter} from 'core/ui/components/Modal/ModalFooter';
+
+function getFormattedDocument(url) {
+    return new Promise((resolve) => {
+        var xhr = new XMLHttpRequest();
```
MarkLark86: Is there a reason why `XMLHttpRequest` is used instead of the newer Fetch APIs (available since Chrome v42, with a polyfill option for older browsers)?

tomaskikutis: I don't remember the exact issue I had, but I tried using `fetch` and felt like it was a too low-level API for what would have been simpler with XHR.

tomaskikutis:
@petrjasek Unless you have more feedback, please approve it for merging.

PR #2438, opened 2018-07-04
Extend the Drag&Drop UI to support files on the editor

pablopunk:
We already had the feature that shows the drop area when dragging existing media files inside the editor. Now that features works with external files too, so when you drag a file from your computer, or an existing one in superdesk, it will also show the drop area.


pablopunk:
Hey @petrjasek take a look at this PR whenever you have a chance, I basically removed the lines to check if the drag&drop is coming from moving images within the editor, so now it works also with external files. But, I'm not sure if this will affect something else.

PR #2441, opened 2018-07-06
Fix styling issue on tables inside custom text fields

pablopunk:
## Before

before

## After

after


pablopunk:
Hey @fritzSF or @pavlovicnemanja take a look at this PR because I removed some css from the editor preview that might affect some styling I'm not aware of

PR #2449, opened 2018-07-19
[SETTINGS] [METADATA] restrict "popup width" field to categories and …

jerome-poisson:
…controlled vocabularies

This patch check that vocabulary.field_type is null, excluding custom
text, date, media and embed fields.

SDESK-3127


jerome-poisson:
I'm waiting for a new release of superdesk to pass the test before merging, please don't merge until it's done, thanks.

PR #2450, opened 2018-07-19
fix(filter conditions): For fields show display name instead of field…

ioanpocol:
… (SDESK-2749)

### DIFF
--- a/scripts/apps/workspace/helpers/getLabelForFieldId.js
+++ b/scripts/apps/workspace/helpers/getLabelForFieldId.js
@@ -1,17 +1,16 @@
 import {GET_LABEL_MAP} from '../content/constants';
 import ng from 'core/services/ng';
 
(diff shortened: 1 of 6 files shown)


scripts/apps/workspace/helpers/getLabelForFieldId.js
```diff
@@ -1,32 +1,31 @@
 import {GET_LABEL_MAP} from '../content/constants';
```
tomaskikutis: The method is called `getLabelNameResolver`, but as I can see it now not only gets names of article field labels, but all kinds of vocabularies. It'd be useful to rename it to reflect that.

tomaskikutis: I disagree, but you go ahead with the merge and I'll rename it later.

ioanpocol: I think we have to rename it when we use it for something else than to get the label for a field.

PR #2451, opened 2018-07-20
fix(multiedit): Activate medatada edit only for media files (SDESK-3162)

ioanpocol:


scripts/apps/search/controllers/MultiActionBarController.js
```diff
@@ -116,6 +116,16 @@ export function MultiActionBarController(
         multi.reset();
     };
 
+    this.canEditMetadata = function() {
```
tomaskikutis: Consider using `Array.every` and `Array.includes`, it looks more semantic and clean.

```
this.canEditMetadata = () => multi.getItems()
    .every(() => !item.lock_user && ['picture', 'video', 'audio'].includes(item.type));
```

PR #2459, opened 2018-07-24
feat(ingest): (SDESK-3117) When only events type selected on ingest, on 'go to ite…

ioanpocol:
…ms' show events view with events only and current source set

### DIFF
--- a/scripts/apps/ingest/directives/IngestSourcesContent.js
+++ b/scripts/apps/ingest/directives/IngestSourcesContent.js
@@ -413,8 +413,25 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                     });
             };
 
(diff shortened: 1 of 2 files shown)


scripts/apps/ingest/directives/IngestSourcesContent.js
```diff
@@ -413,8 +413,25 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                     });
             };
 
-            $scope.gotoIngest = function(source) {
-                $location.path('/search').search({repo: 'ingest', source: angular.toJson([source])});
+            $scope.gotoIngest = function(provider) {
+                const contentTypes = provider.content_types;
+
+                // the list contains allways 'preformatted' item
```
tomaskikutis: Typo in "always"

PR #2460, opened 2018-07-25
Sequence of tabbing the fields in authoring is broken and incompatible with order defined in cp (SDESK 3164)

tomaskikutis:
SDESK-3164

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ItemAssociationDirective.js
+++ b/scripts/apps/authoring/authoring/directives/ItemAssociationDirective.js
@@ -26,6 +26,7 @@ export function ItemAssociationDirective(renditions) {
             rel: '=',
             item: '=',
             editable: '<',
+            tabindex: '<',
             allowPicture: '<',
             allowVideo: '<',
             allowAudio: '<',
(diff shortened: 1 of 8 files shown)


pavlovicnemanja:
we shouldn't use master of superdesk-ui-framework, sometimes things are broken on superdesk-ui-framework for some other instances that use superdesk-ui-framework, and this way we are avoiding breaking things.

tomaskikutis:
How about using #master and fixing to a specific commit when doing releases?

tomaskikutis:
@petrjasek Can you lock `ui-framework` version to current commit when releasing?

petrjasek:
I think it should get versioned and then we can use semantic versioning and have something like `^1.1.0` in client core. but doing it when releasing might make it more complex if the current version wouldn't work and I would have to look for older version which works with recent core.

tomaskikutis:
I'm not sure if by versioning you mean slapping a tag on a commit or npm release. I'd suggest a tag since it's simpler. I think it also makes sense to just increment the version when releasing so if it breaks, you can just take the previous version which is known to be stable due to being in production for a while.

What do you think? Do you want me to implement it?

petrjasek:
I was thinking about npm releases, so we would add to package.json:
```json
"superdesk-ui-framework": "^1.0.0",
```
and then if we release ui-framework semantically we wouldn't have to change anything until there is a bc break release. not sure if this can be done with tags

PR #2466, opened 2018-07-26
Strip html tags from headline

pablopunk:
Whenever we make changes to the headline and then we preview the item, the headline will have <p> tags around it. If we reload the page, the problem is gone (until you make changes on the headline again)

I used the same function we use to strip the tags on save so the preview is consistent with the saved item.

SDESK-2337

### DIFF
--- a/scripts/apps/authoring/authoring/helpers.js
+++ b/scripts/apps/authoring/authoring/helpers.js
@@ -120,6 +120,8 @@ export function stripHtml(item) {
             item[key] = stripHtmlRaw(item[key]);
         }
     });
+
+    return item;
 }
 
 /**
--- a/scripts/core/editor3/reducers/editor3.jsx
+++ b/scripts/core/editor3/reducers/editor3.jsx
@@ -81,7 +81,9 @@ export const onChange = (state, newState, force = false) => {
     let contentChanged = state.editorState.getCurrentContent() !== newState.getCurrentContent();
 
     if (contentChanged || force) {
-        state.onChangeValue(editorState.getCurrentContent(), {plainText: !state.showToolbar});
+        const plainText = (state.editorFormat || []).length === 0;
+
+        state.onChangeValue(editorState.getCurrentContent(), {plainText});
     }
 
     if (force) {
--- a/scripts/core/editor3/reducers/editor3.jsx
+++ b/scripts/core/editor3/reducers/editor3.jsx
@@ -1,6 +1,7 @@
 import {RichUtils, EditorState, AtomicBlockUtils, SelectionState} from 'draft-js';
 import {setTansaHtml} from '../helpers/tansa';
 import {addMedia} from './toolbar';
+import {isEditorPlainText} from '../store';
 
 /**
  * @description Contains the list of editor related reducers.
@@ -81,7 +82,7 @@ export const onChange = (state, newState, force = false) => {
     let contentChanged = state.editorState.getCurrentContent() !== newState.getCurrentContent();
 
     if (contentChanged || force) {
-        const plainText = (state.editorFormat || []).length === 0;
+        const plainText = isEditorPlainText(state);
 
         state.onChangeValue(editorState.getCurrentContent(), {plainText});
     }
--- a/scripts/core/editor3/reducers/editor3.jsx
+++ b/scripts/core/editor3/reducers/editor3.jsx
@@ -81,7 +81,7 @@ export const onChange = (state, newState, force = false) => {
     let contentChanged = state.editorState.getCurrentContent() !== newState.getCurrentContent();
 
     if (contentChanged || force) {
-        state.onChangeValue(editorState.getCurrentContent());
+        state.onChangeValue(editorState.getCurrentContent(), {plainText: !state.showToolbar});
     }
 
     if (force) {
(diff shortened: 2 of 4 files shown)


tomaskikutis:
I think we should do a proper fix on this one. I guess it's not only for the headline, but the same issue might exist on all editor3 fields. What we should do IMO is check the conditions when editor3 toolbar is hidden, and ensure that the same config outputs plain text instead of HTML. We can get using `contentState.getPlainText` without any need for stripping tags.

pablopunk:
@tomaskikutis but it's not really happening for any other field, take 'abstract' as an example

tomaskikutis:
Nonetheless, there's a bug that text-only field is outputting HTML.

tomaskikutis:
Since you pushed changes, I assume you are done, but my previous statement is still true:

> Nonetheless, there's a bug that text-only field is outputting HTML.

It may not be visible in preview due to processing you are doing(stripping tags), but editor3 still outputting HTML when it shouldn't. Can we use the same setting which determines if toolbar should be shown and output using `contentState.getPlainText` in that case, so no HTML is being outputted?

tomaskikutis:
> You mean that the html shouldn't be saved at all right?
Exactly :)

pablopunk:
@tomaskikutis how about what I just commited

scripts/core/editor3/reducers/editor3.jsx
```diff
@@ -81,7 +81,7 @@ export const onChange = (state, newState, force = false) => {
     let contentChanged = state.editorState.getCurrentContent() !== newState.getCurrentContent();
 
     if (contentChanged || force) {
-        state.onChangeValue(editorState.getCurrentContent());
+        state.onChangeValue(editorState.getCurrentContent(), {plainText: !state.showToolbar});
```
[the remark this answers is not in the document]
pablopunk: That makes sense, I guess I can just rename `showToolbar` everywhere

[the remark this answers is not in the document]
pablopunk: Well not exactly, `showToolbar` would still exist as an input for Editor3 component, but coming from the logic of `plainText === true`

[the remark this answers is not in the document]
pablopunk: hmmm I realized that we already set `showToolbar` based on `editorFormat` array. If the array is empty it's basically plainText, so I could use that logic that's already implemented.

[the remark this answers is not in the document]
tomaskikutis: I've just taken a look at the code and it looks like we can rename `showToolbar` to `plainText`. As you noted, it is an input for Editor3 component, but we can rename that too because it's used only when plaintext is required. `const showToolbar = !props.singleLine && (props.editorFormat || []).length > 0;`

scripts/apps/authoring/authoring/helpers.js
```diff
@@ -120,6 +120,8 @@ export function stripHtml(item) {
             item[key] = stripHtmlRaw(item[key]);
         }
     });
+
```
tomaskikutis: Looks like this function is no longer relevant since we no longer output HTML for the headline?

scripts/core/editor3/reducers/editor3.jsx
```diff
@@ -81,7 +81,9 @@ export const onChange = (state, newState, force = false) => {
     let contentChanged = state.editorState.getCurrentContent() !== newState.getCurrentContent();
 
     if (contentChanged || force) {
-        state.onChangeValue(editorState.getCurrentContent());
+        const plainText = (state.editorFormat || []).length === 0;
```
tomaskikutis: I might confused you with my suggestions, but in this case it's better to use `plainText = showToolbar === false`, because otherwise you are duplicating the logic from here https://github.com/superdesk/superdesk-client-core/blob/787408ceadd538fb26d1694edc5d5bd144b8e3e4/scripts/core/editor3/store/index.js#L44

You also don't take `singleLine` into account. Let's keep the logic in one place by using `plainText = showToolbar === false`. Once you have the logic in one place you can fix the naming and remove `showToolbar` everywhere replacing it with `plainText === false`. If you want, you can skip the second part and I can do it sometime.

scripts/apps/authoring/authoring/helpers.js
pablopunk: It's not but I would keep the `return` statement as it's an expected behaviour for a function. Maybe this should be a separate commit. Also we should not modifiy the input but create a new item instead.

scripts/core/editor3/reducers/editor3.jsx
pablopunk: I agree with having the logic in one place but I still think I shouldn't go back to `plainText = showToolbar`, as you said we might want to hide the toolbar at some point even on formatted editors. I think it would be better to export `!props.singleLine && (props.editorFormat || []).length > 0;` as `isEditorPlainText` function or something similar and use that.

scripts/apps/authoring/authoring/helpers.js
tomaskikutis: Judging from the fact that `fields` is hardcoded to `['headline']` I think this function can be removed altogether.

scripts/core/editor3/reducers/editor3.jsx
tomaskikutis: I agree, let's do that :)

PR #2472, opened 2018-07-31
fix(content profile): Fix the set of required on subject field (SDESK…

ioanpocol:
…-3193)


tomaskikutis:
It isn't obvious from looking at the code what was the issue. Could you explain it briefly?

ioanpocol:
When we edit a content profile we have to use the changed profile returned by content.getTypeMetadata.
There is this complication that subject store other fields and on edit profile we change the profile to edit the fields saved to subject like it is an normal one and on save/update on backend we create the right profile.
@petrjasek Since the custom fields are saved in extra key, probably it will be a good idea to eliminate this complication.

tomaskikutis:
I'm sorry, but I hardly understand anything, could you try to rephrase?

ioanpocol:
Some tests are failing if I set $scope.editing to empty.

PR #2474, opened 2018-07-31
Move block to drop area after dragging new image to body

pablopunk:
SDESK-3116


scripts/core/editor3/reducers/editor3.jsx
```diff
@@ -267,6 +270,10 @@ function moveBlock(state, {block, dest, insertionMode}) {
         insertionMode
     );
 
+    if (returnState) {
```
tomaskikutis: It's much simpler when a return type is always the same. If you need a different one, create another function and put common functionality in a `base` function.

return-type-refactoring

scripts/core/editor3/reducers/toolbar.jsx
```diff
@@ -143,13 +144,14 @@ const removeFormat = (state) => {
  * @ngdoc method
  * @name insertMedia
  * @param {Array} files List of media files to be inserted into document.
+ * @param {String} moveKey Block key
  * @description Inserts a list of media files into the document.
  */
-const insertMedia = (state, files = []) => {
-    var {editorState} = state;
+const insertMedia = (state, {files = [], moveKey = null}) => {
```
tomaskikutis: Would be nicer if it were obvious from naming if the `moveKey` is the key of an item you are moving or the key of the target you are moving the item to.

tomaskikutis:
Yeah, I've noticed that the feature is buggy and doesn't always drop to the right place. It might be a regression since I remember it working fine previously.

I suggest you inspect the previous implementation which I think was done by @petrjasek  and see if you aren't missing anything.

pablopunk:
AFAIK he only did the drag&drop to move existing media to another block

scripts/core/editor3/components/Editor3Component.jsx
```diff
@@ -46,10 +45,26 @@ const VALID_MEDIA_TYPES = [
...
 
+/**
+    * @ngdoc method
+    * @name Editor3#allowItem
+    * @param {Object} e Event
+    * @param {Array} props
+    * @returns {Boolean} Returns true if the item is permitted.
+    * @description Check if the editor accept images and if current item is valid media.
+    */
+export function allowItem(e, props) {
```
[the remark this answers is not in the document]
pablopunk: Agree. I should have renamed it when I exported the function. Will do now.

scripts/core/editor3/reducers/editor3.jsx
pablopunk: Agree

scripts/core/editor3/reducers/toolbar.jsx
pablopunk: yep you're right, `targetBlockKey` would be better

PR #2484, opened 2018-08-07
[SDESK-964] Add configuration for socket output

marwoodandrew:
The config for the AAP socket output transmission mechanism (in the master branch this has been moved to the AAP repo)


marwoodandrew:
@petrjasek Seems as the test server is not getting built, I'm not sure how I can fix this!

marwoodandrew:
Not a winner, whatever version of the superdesk server we are trying to run is to new!

PR #2486, opened 2018-08-07
Add functionality to edit metadata for single files

pablopunk:
SDESK-3204

I used the same logic we use for cropping single images but with videos/audio, opening the metadata tab.

~~There's still one issue that I'm getting (already on master). I can edit a file but if I do it twice I get this error:~~

⤵️ Fixed in 856a803c36357d48fa69de96af442e01901f2a27

captura de pantalla 2018-08-07 a las 14 42 34


pablopunk:
@petrjasek but I didn't write the function we have, is that missing for crop also? Is that the bug?

petrjasek:
I see, yep it could be a bug

pablopunk:
could it be that we are storing the new `_etag` in the `origItem` instead of the `scope.item`? @petrjasek check my last commit

petrjasek:
it looks ok, but you might also need to update that `origItem._etag` in order to be able to save additional changes like headline update after editing metadata

PR #2491, opened 2018-08-10
Typescript integration

tomaskikutis:
SDESK-3226


scripts/tests.js
```diff
@@ -9,6 +9,6 @@ import Adapter from 'enzyme-adapter-react-16';
 
 Enzyme.configure({adapter: new Adapter()});
 
-var testsContext = require.context('.', true, /.spec.[t|j]sx?$/);
+var testsContext = require.context('.', true, /.spec.(js|ts|tsx)$/);
```
pablopunk: This means we won't use `.jsx` any more right?

tomaskikutis: Yes, we'll no longer use `.jsx` - the extension, but will continue using `jsx` syntax in `.tsx` files.

petrjasek:
would it be possible to avoid changes in existing files? it makes it harder to backport bugfixes to older versions

tomaskikutis:
I could keep the existing files unchanged, but we wouldn't get any type checking then.

I left most of `*.js` files untouched and only changed `*.jsx` which allows them to be type-checked. All other files are just bundled for now and not type-checked. The plan is to merge this first, so we don't do too much at once.

If you look at the diff, it's mostly <5 lines per file, so it shouldn't be too hard to backport.

scripts/apps/search/components/actions-menu/Label.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {closeActionsMenu} from '../../helpers';
 
-export default function Label(props) {
+export const Label:React.StatelessComponent<any> = (props) => {
```
pablopunk: not very important but can't we still `export default const Label...` ?

tomaskikutis: That might be important, seems like I removed default export. I'll fix that on Thursday. Please keep an eye on similar situations. Most of the changes should be simply adding types and not changing any logic. There are a few exceptions, but it should be obvious there.

pablopunk: Okay, also remember to take `import {Label}` back to `import Label`

scripts/core/ui/components/Form/SelectTagInput/index.tsx
```diff
@@ -124,7 +128,7 @@ export class SelectTagInput extends React.Component {
                     <LineInput noMargin={true} invalid={invalid} required={required}>
                         <Label text={label} />
 
-                        <div className="sd-line-input__input" readOnly={readOnly}>
```
pablopunk: why are we removing `readOnly`?

scripts/core/ui/components/Toggle/index.tsx
```diff
@@ -10,7 +10,7 @@ import './style.scss';
  * @name Toggle
  * @description Toggle component used to togle check/uncheck status
  */
-export default function Toggle({value, onChange, readOnly, onFocus, className}) {
```
pablopunk: another default here

scripts/globals.d.ts
```diff
@@ -0,0 +1,15 @@
+declare const $ : any; // jquery
```
pablopunk: should we add `Window` to the list of globals?

scripts/core/ui/components/Form/SelectTagInput/index.tsx
tomaskikutis: `readOnly` is an attribute for inputs and `div` is not an input. TypeScript flagged it an invalid property which it is.

petrjasek:
what would be the idea regarding prop-types vs component interfaces?

webpack.config.js
```diff
@@ -79,20 +79,15 @@ module.exports = function makeConfig(grunt) {
                 // ensure that react is loaded only once (3rd party apps can load more...)
                 react: path.resolve('./node_modules/react'),
             },
-            extensions: ['.js', '.jsx'],
+            extensions: ['.js', '.ts', '.tsx'],
```
petrjasek: I think we might need to keep jsx working for superdesk-planning

mdhaman: Guys be lenient allow both. Are we planning to migrate angular to 2+?

tomaskikutis: I forgot the apps are using the same config, will fix this. @mdhaman there are no plans for angular 2+ as far as I know.

tomaskikutis:
> what would be the idea regarding prop-types vs component interfaces?

TypeScript checks are compile-time instead of run-time checks by prop-types, so I don't see any reasons to keep prop-types for long. I'm only keeping them now to save time and minimize changes in this PR.

petrjasek:
I'm just wondering if it can check at compile time that react-redux connect mapping state/dispatch to props.

tomaskikutis:
Looks like it's [possible](https://spin.atomicobject.com/2017/04/20/typesafe-container-components/).

package.json
```diff
@@ -108,8 +108,6 @@
     "btoa": "^1.1.2",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
-    "eslint": "^5.3.0",
```
pablopunk: Shouldn't we still have `eslint` as a dependency as we are using it on `npm run lint`?

tomaskikutis: I've added a commit comment

> use eslint from superdesk-code-style

Eslint is already a dependency in `superdesk-code-style` repo, which is a dependency for this one, so eslint is already there. Not sure if this is a good practice, just leaving it as it were before this PR.

pablopunk: oh okay, I thought we had eslint before

PR #2492, opened 2018-08-13
fix(edit image): After media edit update etag for media items (SDESK-…

ioanpocol:
…3209)


scripts/apps/authoring/authoring/services/AuthoringService.js
```diff
@@ -290,6 +290,10 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
             delete diff._etag;
         }
 
+        if (origItem.type === 'picture' && item._etag != null) {
```
petrjasek: imo this could use some comment why it's needed

PR #2497, opened 2018-08-15
SDESK-3265 Wrong html conversion during saving a story.

ride90:
When text has multiply inline styles, editor converts editor state to html with mistakes.
Example:
screen shot 2018-08-16 at 11 58 31

Before:
```
<p>___<i>ITALIC <b>BOLD-ITALIC </i>BOLD</b>___ </p>
<p>___<i>ITALIC <b>BOLD-ITALIC </i></b>___</p>
<p>---<s>STRIKE <b>BOLD-STRIKE <u>BOLD-STRIKE-UNDERLINE</s> </b>UNDERLINE</u></p>
<p>--- X<sub>22<b>22</b>2</sub>----</p>
```

Now:
```
<p>___<i>ITALIC <b>BOLD-ITALIC </b></i><b>BOLD</b>___ </p>
<p>___<i>ITALIC <b>BOLD-ITALIC </b></i>___</p>
<p>---<s>STRIKE <b>BOLD-STRIKE <u>BOLD-STRIKE-UNDERLINE</u></b></s><u> UNDERLINE</u></p>
<p>--- X<sub>22<b>22</b>2</sub>----</p>
```

### DIFF
--- a/scripts/core/editor3/html/to-html/BlockInlineStyleWrapper.js
+++ b/scripts/core/editor3/html/to-html/BlockInlineStyleWrapper.js
@@ -87,8 +87,14 @@ export class BlockInlineStyleWrapper {
             return '';
         }
 
-        const noLongerApplied = this.activeStyles
+        let noLongerApplied = this.activeStyles
             .filter((s) => styles.toArray().indexOf(s) === -1);
+
+        if (noLongerApplied.length !== 0) {
+            const index = this.activeStyles.indexOf(noLongerApplied[0]);
+
+            noLongerApplied = this.activeStyles.slice(index).reverse();
+        }
 
         return noLongerApplied.map((style) => {
             const tag = this.getTag(style, 'close');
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/html/to-html/BlockInlineStyleWrapper.js
```diff
@@ -87,9 +87,15 @@ export class BlockInlineStyleWrapper {
             return '';
         }
 
-        const noLongerApplied = this.activeStyles
+        let noLongerApplied = this.activeStyles
             .filter((s) => styles.toArray().indexOf(s) === -1);
 
+        if (noLongerApplied.length !== 0) {
+            const index = this.activeStyles.indexOf(noLongerApplied[0]);
```
tomaskikutis: Could you explain how this works and what does it fix?

ride90: Let's say that we have next `activeStyles` ["ITALIC", "BOLD", "UNDERLINED"] and next character is not <b>bold</b> anymore, next expression will return style which is no longer applied:
https://github.com/superdesk/superdesk-client-core/blob/9072076e47b7294c3bd5291e55e406e09877fb59/scripts/core/editor3/html/to-html/BlockInlineStyleWrapper.js#L90-L91
so in our case it's `"BOLD"`, after (return statement) based on this data we will return `</b>` which is wrong, because we need to close `</u>` (children) before `</b>`
That's why, we take index of no longer applied tag and take slice in reverse direction, because we need to close all child tags which are inside `noLongerApplied`

ride90: @tomaskikutis try to log `noLongerApplied` variable before `return` and try to comment my fix, you will see the difference.

PR #2498, opened 2018-08-16
[SDESK-3215] User can set a default planning agenda, if planning is i…

marwoodandrew:
…nstalled

### DIFF
--- a/scripts/apps/users/views/user-preferences.html
+++ b/scripts/apps/users/views/user-preferences.html
@@ -117,7 +117,7 @@
                     </div>
                 </div>
             </li>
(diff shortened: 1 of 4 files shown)


scripts/apps/users/views/user-preferences.html
```diff
@@ -117,7 +117,7 @@
                     </div>
                 </div>
             </li>
-            <li class="fieldset terms-box" ng-hide="privileges['planning'] !== 1">
+            <li class="fieldset terms-box" ng-hide="!planning_installed">
```
petrjasek: you can do the check via features which should be set on root scope, so `ng-if="features.agenda"` would check if the agenda resource is available from the api

PR #2500, opened 2018-08-16
[AUTHORING] update item.associations on "save"

jerome-poisson:
When user push the save button, $scope.item is not updated correctly,
resulting in validation issues when publishing ($scope.item is used in the
HTTP PATCH request).

This patch fixes this.
SDESK-3274


jerome-poisson:
I'm not sure this is the right way to fix the issue, so if a client dev can check it.
To summarize: in editor's scope, there are 2 copies of item (`origItem` and `item`), and `item` has associations set to an empty object (`{}`) by default.

When use press `save` button, the associations is not updated for `item`, which is later user when user wants to `publish` (it's used in the HTTP PATCH request), resulting in incorrect validation.

PR #2502, opened 2018-08-22
Duplicated description field is presented when uploading a media to authoring item (SDESK 3271)

tomaskikutis:
SDESK-3271

`archive_description` is read-only, but AAP still wants to see it in the metadata editor for reference, so I'm only hiding it when uploading.

### DIFF
--- a/scripts/apps/archive/controllers/UploadController.js
+++ b/scripts/apps/archive/controllers/UploadController.js
@@ -14,7 +14,6 @@ export function UploadController($scope, $q, upload, api, archiveService, sessio
     $scope.allowPicture = !($scope.locals && $scope.locals.data && $scope.locals.data.allowPicture === false);
     $scope.allowVideo = !($scope.locals && $scope.locals.data && $scope.locals.data.allowVideo === false);
     $scope.allowAudio = !($scope.locals && $scope.locals.data && $scope.locals.data.allowAudio === false);
-
     $scope.validator = _.omit(deployConfig.getSync('validator_media_metadata'), ['archive_description']);
 
     var uploadFile = function(item) {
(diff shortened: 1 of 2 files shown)


mdhaman:
@tomas instead of removing the field can you rename the label for the field to `Caption` at this line no https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/media/views/media-metadata-editor-directive.html#L23 
Then you won't have duplicate fields and fields will be:
`Caption` -> `description_text`
`Description` -> `archive_description`
 (see https://github.com/superdesk/superdesk-client-core/blob/1.8/scripts/apps/archive/views/upload.html#L45)

tomas:
Please mention the right person (@tomaskikutis). Thanks!

tomaskikutis:
@mdhaman as you yourself mentioned `archive_description` is a read-only field which is coming from an archive, so it would be a bug to allow entering it. I'll rename it to caption though.

mdhaman:
@tomaskikutis after our discussion on slack, I checked the comments/description on ticket https://dev.sourcefabric.org/browse/SDESK-854 and found that the  `archive_description` is read-only field in the `media editor` after the media is associated with the story. Hence, I asked  you to remove while "Uploading Media". 

Many changes have been made recently for uploading media/associated media/media editor and I just want similar behavior when AAP upgrades to latest version of Superdesk. IMO, we don't remove the `archive_description` field and just rename the field.

mdhaman:
@tomas thanks for directing to the correct person.

tomas:
@mdhaman OK, but _please_ don't mention me again (you just did!) :)

tomaskikutis:
@mdhaman I'm not removing it, only hiding it on upload. Do you want to keep it on upload as well?

mdhaman:
As per the ticket https://dev.sourcefabric.org/browse/SDESK-854 it seems that hiding is not required.  You have already done rename and its sufficient. 

If you really think that it should be hidden on upload then I will leave that decision with you and @karelpetrak / vince.

PR #2504, opened 2018-08-22
transpile typescript files, not javascript

petrjasek:


tasks/options/webpack.js
```diff
@@ -7,7 +7,7 @@ module.exports = function(grunt) {
     config.devtool = grunt.option('webpack-devtool') || 'cheap-source-map';
 
     config.module.rules = config.module.rules.map((rule) => {
-        if (rule.loader === 'ts-loader') {
+        if (rule.loader === 'ts-loader' && rule.test.test('index.ts')) {
```
tomaskikutis: You don't need this check, let Webpack set `rule.options.transpileOnly = true` for all `ts-loader`s

webpack.config.js
```diff
@@ -85,13 +85,21 @@ module.exports = function makeConfig(grunt) {
         module: {
             rules: [
                 {
-                    test: /\.[j|t]sx?$/,
+                    test: /\.tsx?$/,
```
tomaskikutis: I think it's better to use `test: /\.(ts|tsx)$/,`. It's the same thing, but easier to read

tasks/options/webpack.js
petrjasek: right

PR #2505, opened 2018-08-23
Hide edit/crop tabs when editing single video/audio item

pablopunk:
SDESK-3204

When editing a single media item, the "edit metadata" button will only show the Details tab for video and audio files

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ArticleEditDirective.js
+++ b/scripts/apps/authoring/authoring/directives/ArticleEditDirective.js
@@ -285,11 +285,6 @@ export function ArticleEditDirective(
                     scope.mediaLoading = true;
 
                     const defaultTab = crop ? 'crop' : 'view';
-                    let hideTabs = [];
-
-                    if (scope.item.type !== 'picture') {
-                        hideTabs = ['image-edit', 'crop'];
-                    }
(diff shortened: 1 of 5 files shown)


scripts/apps/authoring/authoring/directives/ArticleEditDirective.js
```diff
@@ -285,9 +285,23 @@ export function ArticleEditDirective(
...
                     const defaultTab = crop ? 'crop' : 'view';
+                    let hideTabs = [];
 
-                    return renditions.crop(scope.item,
-                        {isNew: false, editable: true, isAssociated: false, defaultTab: defaultTab, showMetadata: true})
+                    if (scope.item.type !== 'picture') {
+                        hideTabs = ['image-edit', 'crop'];
+                    }
+
+                    return renditions.crop(
```
tomaskikutis: There are multiple places where `renditions.crop` is called and I assume the rules which tabs to hide are generic. Wouldn't it be better to put the logic inside the `crop` method instead of using the same parameters in x places?

pablopunk: You're right, pushing the fix now.

PR #2507, opened 2018-08-28
[SDESK-3296] Add a content type filter to publish queue page

marwoodandrew:


scripts/apps/publish/views/publish-queue.html
```diff
@@ -57,6 +57,19 @@
                             </li>
                         </ul>
                     </div>
+                    <span class="lab" translate>Content Type:</span>
+                    <div class="dropdown" dropdown>
+                        <button class="dropdown__toggle" dropdown__toggle>
+                            {{ selectedFilterContentType.name }}
+                            <span ng-if="!selectedFilterContentType">none</span>
```
petrjasek: there should be translate too

marwoodandrew: Yep, on all of them I guess.

PR #2510, opened 2018-08-31
fix(authoring): For image item update etag after apply crops(SDESK-3209)

ioanpocol:


scripts/apps/authoring/authoring/services/AuthoringService.js
```diff
@@ -299,6 +299,7 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
 
         if (_.size(diff) > 0) {
             return api.save('archive', origItem, diff).then((_item) => {
+                item._etag = _item._etag;
```
tomaskikutis: Shouldn't etag updating be done inside `api.save`? Are there any cases where we wouldn't want to update etag when saving?

petrjasek: afaik it's done there, but we're not passing `item` there but `origItem`

tomaskikutis: Why are we passing `origItem` instead of `api.save('archive', item, diff)`?

ioanpocol: The problem is that we return origItem instead of _item, the call of api.save it's ok.

ioanpocol: I think a merge is needed not a deep copy. Anyway the item and _item are the same except the fields updated on backend line etag.

tomaskikutis: If they are the same, you don't need to merge it, just throw the old one away and use the one from the server. I think you don't even need to deep-copy it since it came from the server, isn't used anywhere else and won't cause any other objects to change when updating a property - which is what deep-copying is for.

Please do that, since it abstracts the internal stuff away and will nicely accept any future changes.

ioanpocol: origItem is returned and not item.

ioanpocol: I don't know, but it don't solve the issue. I changed the code to update the etag only for images.
This update is related to the following line: https://github.com/superdesk/superdesk-client-core/blob/92d2b77ca6cdf3b38ae79fd2cff4c455b8575ec3/scripts/apps/authoring/authoring/services/AuthoringService.js#L294.
I think no other updates should be performed on item.

ioanpocol: The _etag update is needed only to be used later on https://github.com/superdesk/superdesk-client-core/blob/92d2b77ca6cdf3b38ae79fd2cff4c455b8575ec3/scripts/apps/authoring/authoring/services/AuthoringService.js#L294. Any other complication is not necessary.

ioanpocol: item = _item is not working and already I told you that a merge is required for that.
A complication is to do something that you are not sure that is ever needed.
The problem appear only for pictures and only for this case I update the etag.

tomaskikutis: > item = _item is not working

Alright then, that's a real reason :)

> The problem appear only for pictures and only for this case I update the etag.

I would update it unconditionally, there is no reason it shouldn't work other types of items.

PR #2516, opened 2018-09-04
feat(products): Show the product id when the product is edited (SDESK…

ioanpocol:
…-3334)

### DIFF
--- a/scripts/apps/products/views/products-config-modal.html
+++ b/scripts/apps/products/views/products-config-modal.html
@@ -24,8 +24,13 @@
                         <div class="field">
                             <i class="field-info pull-right" translate>* mandatory fields</i>
                         </div>
-                        <div >
-                            <label for="product-name" translate>Product name</label><i class="field-info pull-right" translate>*</i>
+                        <div class="field" ng-if="product._id">
+                            <label for="product-id" translate>Product ID</label>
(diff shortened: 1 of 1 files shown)


scripts/apps/products/views/products-config-modal.html
```diff
@@ -24,16 +24,21 @@ <h3 class="modal__heading" ng-show="product.edit._id" translate>Edit "{{product.
...
+                            <label for="product-name" translate>Product name<i class="field-info pull-right" translate>*</i></label>
                             <input id="product-name" type="text" class="fullwidth-input" ng-model="product.edit.name" required>
                         </div>
                         <div class="field">
                             <label translate>Product description</label>
                             <textarea class="fullwidth-input" ng-model="product.edit.description"></textarea>
                         </div>
                         <div class="field">
-                            <label translate>Product Type</label><i class="field-info pull-right" translate>*</i>
+                            <label translate>Product Type<i class="field-info pull-right" translate>*</i></label>
```
tomaskikutis: You don't need `translate` in `<i class="field-info pull-right" translate>`

ioanpocol: Yes, it was already there. I will delete it.

PR #2526, opened 2018-09-10
feat(editor3): Add support for abbreviations (SDESK-2319)

ioanpocol:
SDESK-2319

### DIFF
--- a/scripts/core/editor3/reducers/editor3.tsx
+++ b/scripts/core/editor3/reducers/editor3.tsx
@@ -139,6 +139,11 @@ const applyAbbreviations = (state) => {
     const content = editorState.getCurrentContent();
     const block = content.getBlockForKey(selection.getStartKey());
     const word = getAbbreviationText(block, selection.getStartOffset());
+
+    if (word == null) {
+        return state;
+    }
+
     const keys = Object.keys(abbreviations);
     const pattern = '\\b(' + keys.map((item) => escapeRegExp(item)).join('|') + ')(\\*)';
     const found = word.text.match(new RegExp(pattern, 'g'));
@@ -159,14 +164,18 @@ const applyAbbreviations = (state) => {
  * @param {Object} block
  * @param {Integer} offset
  * @return {String} returns text that can contain abbreviation
- * @description If no current selection from current position extract text that is
- * delimited on both sides by space or round brackets
+ * @description From current position extract text that is
+ * delimited on both sides by space, round brackets or delete suggestion
  */
 const getAbbreviationText = (block, offset) => {
     const text = block.getText();
     const length = block.getLength();
     let start = offset;
     let end = offset;
+
+    if (!text.includes('*')) {
+        return null;
+    }
 
     while (start > 0 && text[start - 1] !== ' ' && text[start - 1] !== '(' && text[start - 1] !== ')') {
         const inlineStyles = block.getInlineStyleAt(start - 1);
@@ -188,7 +197,13 @@ const getAbbreviationText = (block, offset) => {
         end++;
     }
 
-    return {text: text.substring(start, end), offset: start};
+    const textWithAbbreviation = text.substring(start, end);
+
+    if (!textWithAbbreviation.includes('*')) {
+        return null;
+    }
+
+    return {text: textWithAbbreviation, offset: start};
 };
 
 /**
--- a/scripts/core/editor3/reducers/editor3.tsx
+++ b/scripts/core/editor3/reducers/editor3.tsx
@@ -2,6 +2,8 @@ import {RichUtils, EditorState, AtomicBlockUtils, SelectionState} from 'draft-js
 import {setTansaHtml} from '../helpers/tansa';
 import {addMedia} from './toolbar';
 import {isEditorPlainText} from '../store';
+import {replaceWord} from './spellchecker';
+import {DELETE_SUGGESTION} from '../highlightsConfig';
 
 /**
  * @description Contains the list of editor related reducers.
@@ -18,6 +20,8 @@ const editor3 = (state = {}, action) => {
         return onTab(state, action.payload);
     case 'EDITOR_FORCE_UPDATE':
         return forceUpdate(state);
+    case 'EDITOR_SET_ABBREVIATIONS':
+        return setAbbreviations(state, action.payload);
     case 'EDITOR_DRAG_DROP':
         return dragDrop(state, action.payload);
     case 'EDITOR_SET_CELL':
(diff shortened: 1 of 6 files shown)


scripts/core/editor3/reducers/editor3.tsx
```diff
@@ -89,18 +93,116 @@ export const onChange = (state, newState, force = false) => {
...
+    return state;
+};
+
+/**
+ * @ngdoc method
+ * @name getAbbreviationText
+ * @param {Object} block
+ * @param {Integer} offset
+ * @return {String} returns text that can contain abbreviation
+ * @description If no current selection from current position extract text that is
```
tomaskikutis: > If no current selection from current position

Can you clarify this comment? Do you mean if the selection doesn't exist? Is this `draft-js` or browser-selection?

scripts/core/editor3/reducers/editor3.tsx
```diff
@@ -89,18 +93,116 @@ export const onChange = (state, newState, force = false) => {
...
+ * delimited on both sides by space or round brackets
+ */
+const getAbbreviationText = (block, offset) => {
+    const text = block.getText();
+    const length = block.getLength();
+    let start = offset;
+    let end = offset;
+
+    while (start > 0 && text[start - 1] !== ' ' && text[start - 1] !== '(' && text[start - 1] !== ')') {
```
tomaskikutis: Why two loops? Couldn't the logic be simpler here. Wouldn't the following work? (pseudocode)

```
for abbreviation in abbreviations
    for match in text.match(abbreviationRegexr)
          if isDeleteSuggestion(block, matchOffset) === false
              replaceWord(...)
```

scripts/core/editor3/store/index.js
```diff
@@ -65,12 +65,17 @@ export default function createEditorStore(props, isReact = false) {
...
         invisibles: false,
         svc: props.svc,
+        abbreviations: {},
     }, applyMiddleware(thunk));
 
 
     // after we have the dictionary, force update the editor to highlight typos
     dict.finally(() => store.dispatch(forceUpdate()));
 
+    spellcheck.getAbbreviationsDict().then((abbreviations) => {
```
tomaskikutis: We shouldn't mix asynchronous and synchronous code(`createEditorStore` is synchronous). It'd be better to first load `abbreviations` from server and only then call `createEditorStore` with `abbreviations` as an argument. It would be in the initial state and the extra dispatch wouldn't be required.

scripts/core/editor3/reducers/editor3.tsx
ioanpocol: From current position if no selection try to extend on left and right until there is a space, round brackets or delete suggestion.
On editor2 the match is called only once and regex contains all abbreviations. I suppose that this is faster than calling regex for every abbreviation.

scripts/core/editor3/reducers/editor3.tsx
ioanpocol: It is about draft-js selection, but looks that this check is not present in code. I will add it.

scripts/core/editor3/reducers/editor3.tsx
tomaskikutis: We could check `if (blockText.includes('*'))` before running the regexp. Should be fast enough to run on the entire document and ignore the browser selection which would make the code simpler.

scripts/core/editor3/store/index.js
ioanpocol: "Doesn't work properly on pasting: "App*" gets pasted and only after clicking the word, it's replaced with "Application"."
@tomaskikutis  I checked and this works well.


"Cursor is at the wrong position after typing "*" and getting the word autocompleted."
Currently the cursor remains at the same position. 
@karelpetrak Should we move the cursor after the last character added? 

"FYI @karelpetrak, abbreviations don't work in comments, tables, and inline metadata(image title). Fixing it in comments should not take much, but implementing it in tables or metadata labels would take around a week each."

@tomaskikutis I checked and on comments the abbreviations works well.

tomaskikutis: Yeah, I've seen that the same thing is used for dictionaries which I would refactor as well.

> I think serialising all those loads will just slow the editor3 initialisation and will not add any benefit.

The benefit is avoiding the state when editor3 is enabled, but abbreviations not working because the request for abbreviations hasn't completed yet. That's rare for sure, but I like correctness ;)

To speed the initialization we could have all editor3 instances initialize locked, and when you focus the editor, the requests for dictionaries/abbreviations would start and the editor would unlock when they finish. This would be both correct and fast.

Anyway, this is minor stuff, feel free to leave this as is.

scripts/core/editor3/reducers/editor3.tsx
ioanpocol: To run something on entire document is a bad idea.
We can check for '*' on the text that possibly contains an abbreviation before to run regexp.

scripts/core/editor3/store/index.js
ioanpocol: I think the initialisation speed is much important than an improbable problem that can appear because of this.

tomaskikutis: > "Doesn't work properly on pasting: "App*" gets pasted and only after clicking the word, it's replaced with "Application"."

> I checked and this works well.

abbr-paste

scripts/core/editor3/reducers/editor3.tsx
tomaskikutis: > To run something on entire document is a bad idea.

I meant checking the whole document using `if (blockText.includes('*'))` first. It would be fast this way :)

scripts/core/editor3/store/index.js
tomaskikutis: > I checked and on comments the abbreviations works well.

I was talking about the contents of the comment, not the highlighted text.

ioanpocol: @karelpetrak I think on content of the comment it is not needed the abbreviations.
"Doesn't work properly on pasting: "App*" gets pasted and only after clicking the word, it's replaced with "Application"."
As I said on pasting "App*" it works well. On pasting a larger text that contains abbreviations will need more work. @karelpetrak Should we do this?

karelpetrak: @tomaskikutis @ioanpocol Sorry for belated answers. Here they are:
>Doesn't work properly on pasting: "App*" gets pasted and only after clicking the word, it's replaced with "Application".
The abbreviations feature is a typing assistant, I would not expect the pasted text with "App*" to be replaced without user interaction.

>abbreviations don't work in comments, tables, and inline metadata(image title). Fixing it in comments should not take much, but implementing it in tables or metadata labels would take around a week each.
This is OK for now. We may add it for those fields later.

>"Cursor is at the wrong position after typing "*" and getting the word autocompleted."
Currently the cursor remains at the same position. Should we move the cursor after the last character added?
Yes, I would expect the cursor to jump at the end of the auto-completed word so that I can continue typing.

ioanpocol: >Yes, I would expect the cursor to jump at the end of the auto-completed word so that I can continue typing.

This is already done

PR #2533, opened 2018-09-12
Allow custom fields and sorting to implement SDFID-490

pablopunk:
Basically it allows us to use `sd-search-results` with custom rendering of fields that are not part of sd-core (like the new `compliantlifetime` field) on `superdesk-fi`.

Also, `SortService` is a bit more flexible, allowing sorts of custom fields.

*All the `public` code I think was caused by a `tslint` plugin I use but I guess it's right to keep them

### DIFF
--- a/scripts/apps/search/components/Item.tsx
+++ b/scripts/apps/search/components/Item.tsx
@@ -27,9 +27,8 @@ import {closeActionsMenu} from '../helpers';
  * Item component
  */
 export class Item extends React.Component<any, any> {
-    static propTypes: any;
-    static defaultProps: any;
-
+    public static propTypes: any;
+    public static defaultProps: any;
 
     constructor(props) {
         super(props);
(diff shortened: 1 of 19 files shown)


scripts/apps/search/directives/SearchFilters.js
```diff
@@ -81,6 +81,23 @@ export const getDateFilters = (gettext) => [
...
+                key: 'Next month',
+                label: gettext('Month'),
+            },
+            {
+                key: 'Next 3 months',
+                label: gettext('3 Months'),
+            }
+        ],
+        isEnabled: () => false,
+    }
```
pablopunk: @tomaskikutis I added a filter here to use in getSearch, but it's not ideal as it's not used in superdesk-core, only in superdesk-fi. Any thoughts on this?

tomaskikutis: `getDateFilters` is now responsible for 3 things: rendering available filters, displaying active filters at the top via tag service, and generating elasticsearch query from the URL. [image](https://user-images.githubusercontent.com/2076953/45466869-31ae9c80-b725-11e8-8b9f-36c929095445.png)

If I'm correct, you only need the last part: generating elasticsearch query. I suggest you refactor the search service(in client-core) and instead of having elasticsearch filters be generated from predefined list, you can implement life-cycle hooks for registering and de-registering functions which take a specific value in search field and add the elasticsearch filter to `facetrange` defined in search service. [here's a small example](https://user-images.githubusercontent.com/2076953/45467025-2d36b380-b726-11e8-9c7c-976947c36c9e.png)

`registerSearchTermHandlers` would be called both, by client-core search page and from superdesk-fi compliance review page.

scripts/apps/search/components/Item.tsx
```diff
@@ -46,25 +45,25 @@ export class Item extends React.Component<any, any> {
         this.openAuthoringView = this.openAuthoringView.bind(this);
     }
 
-    componentWillUnmount() {
+    public componentWillUnmount() {
```
tomaskikutis: I think you should disable ts-lint rule which asks for this for now.

tslint.json
```diff
@@ -16,7 +16,15 @@
         "variable-name": false,
         "forin": false,
         "member-ordering": false,
-        "only-arrow-functions": false
+        "only-arrow-functions": false,
+        "trailing-comma": [
+            true,
+            {
+                "multiline": {
+                    "functions": "never"
```
tomaskikutis: I think we can keep it `always` for functions as well. The reason it's not set in eslint is due to js not supporting it.

PR #2537, opened 2018-09-13
feat(package): Move a package with all package items (SDFID-372)

ioanpocol:


scripts/apps/authoring/authoring/directives/SendItem.js
```diff
@@ -176,9 +181,9 @@ export function SendItem($q, api, desks, notify, authoringWorkspace,
                 }
             };
 
-            scope.send = function(open) {
+            scope.send = function(open, allPackageItems) {
```
tomaskikutis: Rename `allPackageItems` to `moveAllPackageItems` or `sendAllPackageItems`. Unless there's a verb in front, the name, it implies that the argument is not a bool, but a collection of items.

scripts/apps/authoring/authoring/directives/SendItem.js
```diff
@@ -85,6 +86,10 @@ export function SendItem($q, api, desks, notify, authoringWorkspace,
                 }
 
                 scope.isActive = !!item;
+
+                if (scope.config == null) {
+                    scope.isPackage = item != null && item.type === 'composite';
```
tomaskikutis: We do we need two configs for this?
> `scope.isPackage = config != null && config.isPackage;`
> `scope.isPackage = item != null && item.type === 'composite';`

scripts/apps/authoring/views/send-item.html
```diff
@@ -101,6 +101,14 @@
                     {{!canSendItem() ? 'fetch' : 'send' | translate}}
                 </button>
             </li>
+            <li class="full-width" ng-if="showSendButtonAndDestination() && isPackage">
+                <button class="btn btn--large btn--primary btn--expanded"
+                        ng-click="send(false, true)"
+                        ng-disabled="disableSendButton()"
+                        >
+                    {{'send all' | translate}}
```
tomaskikutis: Use `<span translate>send all</span>`. When translating using the filter, strings don't get registered for translation by grunt task.

scripts/apps/ingest/services/SendService.js
```diff
@@ -174,6 +174,7 @@ export function SendService(desks, api, $q, notify, $injector, multi, $rootScope
         self.config = $q.defer();
         self.config.action = action;
         self.config.itemIds = _.map(items, '_id');
+        self.config.isPackage = _.some(items, (item) => item.type === 'composite');
```
tomaskikutis: Use native [`Array.prototype.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

scripts/apps/authoring/authoring/directives/SendItem.js
ioanpocol: Yes, there are two cases, the operation is for a single item or for a list of items

scripts/apps/ingest/services/SendService.js
ioanpocol: All over the file is used lodash so I think we can keep it for consistency

tomaskikutis: In a lot of cases, it's used for what no native methods are available like `cloneDeep`, `isEmpty`, `get`, `trim`. It's good that `lodash` is used there, but let's not use it for things that are already in the language. We'll never get rid of old stuff if we keep adding it. I've fixed that for you :)

ioanpocol: Please don't do it again. You just restarted my tests :)
There are many places in the application where we use lodash even if now there are native functions. For example there is a map call in the same file.
Probably a better approach is to change all those appearance and not to mixt them.

tomaskikutis: > Please don't do it again. You just restarted my tests :)

I would like to do it more actually(commit to repos I'm reviewing, not restart your tests ;D) It's up to you of course, but I think it'd be more time efficient to resolve PRs when reviewers add commits for renames and other small stuff. What do you think?

> Probably a better approach is to change all those appearance and not to mixt them.

Yep, it's better to change it all, but they don't all have the same APIs and it's not always easy to figure out which types are used. For example `_.includes(scope.config.itemIds, data.item)`. I've no idea if `itemIds` is a delimited string or an aray or if `data.item` is string or an object. Since it's not some complex architectural pattern, but a simple pure function, I prefer mixing them so the amount of the old ones decreases and it's easier to remove them completely at some point.

ioanpocol: >I would like to do it more actually(commit to repos I'm reviewing, not restart your tests ;D) It's up to you of course, but I think it'd be more time efficient to resolve PRs when reviewers add commits for renames and other small stuff. What do you think?

I think it is better to just add comments.

PR #2542, opened 2018-09-17
Date fields are not working in searches (SDESK 3350)

tomaskikutis:
SDESK-3350


scripts/apps/search/directives/DateFilters.ts
```diff
@@ -0,0 +1,157 @@
+interface IDateRange {
+    key: string;
+    label: string;
+    elasticSearchDateRange: {
+        lte?: string,
+        gte?: string,
```
petrjasek: seems like gte could be required

tomaskikutis: According to how it's used now, it can be required, but I wanted the interface to reflect at least the subset of the elastic spec

PR #2545, opened 2018-09-18
Don't use internal clipboard when pasting from the same editor

pablopunk:
I still couldn't reproduce QA's error but this might fix it whenever you paste tables within the same editor


tomaskikutis:
Did you try using the same environment as QA? Can QA reproduce it again?

pablopunk:
@tomaskikutis yep, it is working for me. QA can reproduce it and there's a video about it on JIRA's task.

tomaskikutis:
I remember there was a case when I wasn't able to reproduce the issue even when using windows when QA was also using windows. The issue was with a specific article. Try testing on the same item, with the same OS.

PR #2546, opened 2018-09-18
fix(authoring): call externalsource action on drop to featured media

petrjasek:
SDESK-3392


mdhaman:
Hi @petrjasek, as part of this PR can you move the `item-association--loading` from https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/views/item-association.html#L14 to https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/views/item-association.html#L8

If the `class` is applied at the `figure` element we don't see the progress indicator.

petrjasek:
@mdhaman I've added it to button and also kept it on figure, is that fine?

PR #2548, opened 2018-09-19
linting fixes

tomaskikutis:


package.json
```diff
@@ -133,7 +133,7 @@
     "debug-unit-tests": "karma start karma.conf.js --browsers=Chrome",
     "unit": "grunt unit",
     "build": "grunt build",
-    "lint": "tsc --noEmit && tslint -c tslint.json 'scripts/**/*.ts' && eslint --ext .js --ext .jsx --ext .tsx scripts spec tasks *.js",
+    "lint": "tsc --noEmit && tslint -c tslint.json 'scripts/**/*.{ts,tsx}' && eslint --ext .js --ext .jsx --ext .tsx scripts spec tasks *.js",
     "lint-fix": "tsc --noEmit --version && tslint -c tslint.json 'scripts/**/*.ts' --fix && eslint --ext .js --ext .jsx --ext .tsx scripts spec tasks *.js --fix",
```
pablopunk: `lint-fix` should also get the new pattern `{ts, tsx}`

petrjasek: seems like we lint tsx with both tsc and eslint, is that by design?

tomaskikutis: `tsc` is not doing any linting, only verifying the types and `tslint` takes care of code style/formatting.

petrjasek: yep right I mean tslint and eslint both check `.tsx`

tomaskikutis: I haven't checked, but I'm sure at least some eslint rules won't have a matching tslint rule so I just lint using both.

PR #2550, opened 2018-09-19
fix(image): On image edit, on image transformations preserve text edi…

ioanpocol:
…table metadata values (SDESK-3362)

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.js
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.js
@@ -203,6 +203,19 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
         return false;
     }
 
+    function extractEditableMetadata(metadata) {
+        return _.pick(metadata, [
+            'headline',
+            'description_text',
+            'archive_description',
+            'alt_text',
+            'byline',
+            'copyrightholder',
+            'usageterms',
+            'copyrightnotice',
+        ]);
+    }
+
     /**
     * @ngdoc method
     * @name ChangeImageController#saveAreaOfInterest
@@ -238,7 +251,9 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
                 $scope.data.isDirty = true;
                 return api.save('picture_renditions', {item: result.item, no_custom_crops: true}).then((item) => {
                     $scope.data.item.renditions = item.renditions;
-                    $scope.data.metadata = $scope.data.item;
+                    const editableMetadata = extractEditableMetadata($scope.data.metadata);
+
+                    $scope.data.metadata = angular.extend($scope.data.item, editableMetadata);
                     $scope.data.poi = {x: 0.5, y: 0.5};
                     $rootScope.$broadcast('poiUpdate', $scope.data.poi);
                 });
@@ -328,7 +343,9 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
         }}).then((result) => {
             $scope.data.item.renditions = result.renditions;
             $scope.data.item._etag = result._etag;
-            $scope.data.metadata = $scope.data.item;
+            const editableMetadata = extractEditableMetadata($scope.data.metadata);
+
+            $scope.data.metadata = angular.extend($scope.data.item, editableMetadata);
             $scope.controls = angular.copy(DEFAULT_CONTROLS);
             $scope.data.isDirty = true;
             $scope.loaderForMediaEdit = false;
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.js
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.js
@@ -32,6 +32,17 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
         isDirty: false,
     };
 
+    const EDITABLE_METADATA = [
+        'headline',
+        'description_text',
+        'archive_description',
+        'alt_text',
+        'byline',
+        'copyrightholder',
+        'usageterms',
+        'copyrightnotice',
+    ];
+
     $scope.controls = angular.copy(DEFAULT_CONTROLS);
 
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/controllers/ChangeImageController.js
```diff
@@ -328,7 +343,9 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
         }}).then((result) => {
             $scope.data.item.renditions = result.renditions;
             $scope.data.item._etag = result._etag;
-            $scope.data.metadata = $scope.data.item;
+            const editableMetadata = extractEditableMetadata($scope.data.metadata);
+
+            $scope.data.metadata = angular.extend($scope.data.item, editableMetadata);
```
tomaskikutis: Is `angular.extend` different than native `Object.assign` or `const three = {...one, ...two}` ?

scripts/apps/authoring/authoring/controllers/ChangeImageController.js
```diff
@@ -203,6 +203,19 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
         return false;
     }
 
+    function extractEditableMetadata(metadata) {
+        return _.pick(metadata, [
+            'headline',
```
tomaskikutis: Would be great to implement this without hardcoding the field names.

ioanpocol: any suggestion?

scripts/apps/authoring/authoring/controllers/ChangeImageController.js
ioanpocol: yes, angular.extend changes first parameter and return it

tomaskikutis: > yes, angular.extend changes first parameter and return it

That's how `Object.assign` works.

scripts/apps/authoring/authoring/controllers/ChangeImageController.js
tomaskikutis: It depends on how the saving workflow works. It could work as follows:
1. tab 1, editing something
2. clicking another tab triggers the save > new tabs is opened with latest values

If we don't want to save on every tab change, then I guess hardcoding the fields is alright since they are also probably hardcoded in the view as well. In such case when the same fields are hardcoded in multiple places, it would be helpful to leave a comment pointing to all places so it's easy to add another field and know all other places where it should be added as well.

scripts/apps/authoring/authoring/controllers/ChangeImageController.js
ioanpocol: I think angular.extend goes deep on merging and Object.assign is not going deep

ioanpocol: yes, actually deep merge is not needed because I update only text values
I will change to  Object.assign

scripts/apps/authoring/authoring/controllers/ChangeImageController.js
ioanpocol: I created a constant with all those values.

scripts/apps/authoring/authoring/controllers/ChangeImageController.js
```diff
@@ -32,6 +32,17 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
         isDirty: false,
     };
 
+    const EDITABLE_METADATA = [
```
tomaskikutis: Great! Could you also add a comment including the name of the constant to `scripts/apps/authoring/media/views/media-metadata-editor-directive.html`?

tomaskikutis: > The are a lot of directives that uses different fields

We'd have a bug if I added a new field to `media-metadata-editor-directive.html`, but not to the constants. I think that's related enough to add a comment. I'm not asking to refactor `media-metadata-editor-directive.html` to use a repeater over the constants, only to add a comment.

I can add it myself if you changed your mind regarding me pushing to your repository :)

ioanpocol: On all directives you have to keep controller and the view in synch when you change them.
In a way or other the name of the fields are hardcoded on both controller and view.
Again it is about to be pragmatic and spent time in what matters. 

>I can add it myself if you changed your mind regarding me pushing to your repository :)
It is a bad solution to solve opinion differences.

tomaskikutis: Here's why I think it does matter. Take a look at this [scheme](https://user-images.githubusercontent.com/2076953/45873205-88a61880-bd9a-11e8-9184-889bbca2b9a3.png). If someone is working on "Controller 2" and needs to change "directive 2", it's not realistic that the person will read the code of all controllers which's views use "directive 2". Even if he does, he might not notice the dependency, that's why I think it's helpful to add a comment about it.

ioanpocol: I think the developer follow the action(function call) from view and always will find the right controller where the function is.
The call graph in superdesk is complicated in many cases and we didn't used any kind of documentation for this. 
I think this is a good subject to talk about on next sprint retrospective.

tomaskikutis: > I think the developer follow the action(function call) from view and always will find the right controller where the function is.

I totally agree that if you see a function call in a directive, it's not hard to find the controller where the function is defined. The hard thing is checking all views using a directive and their corresponding controllers to find an _implicit_ dependency which is not obvious.

For example there is no easy way to find out without checking all related code, that if you add a field in say `media-metadata-editor-directive.html`, you should also add it to the constant you defined. The proper solution would be using a repeater in the directive instead of hardcoding the fields, but that might take more time than we would like, that's why I suggested the commenting "solution".

Yeah, we can talk about it in the retrospective :)

PR #2552, opened 2018-09-21
Show assignment details like in the preview pane instead of task details in open articles (SDESK 3414)

tomaskikutis:
SDESK-3414

DEPENDS ON: https://github.com/superdesk/superdesk-planning/pull/1010

### DIFF
--- a/scripts/apps/search/views/item-preview.html
+++ b/scripts/apps/search/views/item-preview.html
@@ -19,6 +19,9 @@
             <li class="nav-tabs__tab" ng-show="isMediaUsed" ng-class="{'nav-tabs__tab--active': vm.current_tab === 'used'}">
                 <button class="nav-tabs__link" ng-click="vm.current_tab = 'used'" translate>Used</button>
             </li>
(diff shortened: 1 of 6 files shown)


scripts/apps/search/views/item-preview.html
```diff
@@ -19,9 +19,6 @@
             <li class="nav-tabs__tab" ng-show="isMediaUsed" ng-class="{'nav-tabs__tab--active': vm.current_tab === 'used'}">
                 <button class="nav-tabs__link" ng-click="vm.current_tab = 'used'" translate>Used</button>
             </li>
-            <li class="nav-tabs__tab" ng-show="isAssigned" ng-class="{'nav-tabs__tab--active': vm.current_tab === 'assignment'}">
```
mdhaman: Hi @tomaskikutis , @fritzSF why are we removing the assignment tab from the `preview`? It would be good to have assignment preview in the preview pane as well as metadata pane (for open articles).

tomaskikutis: My bad, I shouldn't have removed it from the preview. I'll update the PR.

PR #2558, opened 2018-09-25
Sorting is not working in personal (SDESK 3438)

tomaskikutis:
SDESK-3438

### DIFF
--- a/scripts/apps/monitoring/directives/MonitoringGroup.js
+++ b/scripts/apps/monitoring/directives/MonitoringGroup.js
@@ -208,12 +208,18 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
             scope.$on('refresh:list', (event, group) => {
                 var _viewType = event.currentScope.viewType || '';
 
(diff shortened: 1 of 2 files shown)


scripts/apps/monitoring/directives/MonitoringGroup.js
```diff
@@ -208,12 +208,18 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
...
                 var _viewType = event.currentScope.viewType || '';
 
-                if (group && group._id === scope.group._id || !group &&
-                        _.includes(['highlights',
-                            'spiked',
-                            'single_monitoring',
-                            'monitoring'],
-                        _viewType)) {
+                if (
+                    group
```
petrjasek: imo this condition could be a function, would be more readable

tomaskikutis: Would variable work? Fixed in https://github.com/superdesk/superdesk-client-core/pull/2562

PR #2559, opened 2018-09-26
[SDESK-3317](fix): Routing schedule can be set to all day or specific time period.

mdhaman:


scripts/apps/ingest/views/settings/ingest-routing-schedule.html
```diff
@@ -4,11 +4,15 @@
 </div>
 
 <div class="field">
+    <sd-check ng-model="rule.schedule._allDay" label-position="left">All Day</sd-check>
```
petrjasek: translate missing

PR #2560, opened 2018-09-26
feat(autorefresh): Add setting to always autorefresh the monitoring a…

ioanpocol:
…nd search lists (SDNTB-520)

### DIFF
--- a/scripts/apps/monitoring/directives/MonitoringGroup.js
+++ b/scripts/apps/monitoring/directives/MonitoringGroup.js
@@ -24,9 +24,9 @@ import _ from 'lodash';
  *   A directive that generates group/section on stages of a desk or saved search.
  */
 MonitoringGroup.$inject = ['cards', 'api', 'authoringWorkspace', '$timeout', 'superdesk', 'session',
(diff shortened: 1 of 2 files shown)


scripts/apps/monitoring/directives/MonitoringGroup.js
```diff
@@ -412,6 +412,10 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
                 }
 
                 return apiquery(criteria, true).then((items) => {
+                    if (config.features.autorefresh) {
```
petrjasek: I think we should give it some more specific name, in settings it's hard to guess what is autorefreshed

ioanpocol: autorefreshLists?

petrjasek: yep, or monitoringAutorefresh, but it's not only in monitoring right?

ioanpocol: is it in search too

petrjasek: maybe autorefreshContent then?

PR #2565, opened 2018-09-26
Caption and description mix up while uploading an image (SDESK 3355)

tomaskikutis:
SDESK-3355

### DIFF
--- a/scripts/apps/authoring/media/views/media-metadata-editor-directive.html
+++ b/scripts/apps/authoring/media/views/media-metadata-editor-directive.html
@@ -20,7 +20,7 @@
         data-boxed="boxed"
         data-textarea="true"
         data-required="validator.description_text.required"
-        data-label="{{:: 'Caption' | translate}}"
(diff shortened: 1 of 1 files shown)


tomaskikutis:
@mdhaman let me know if this naming works for you. See Vince's [comment](https://dev.sourcefabric.org/browse/SDESK-3355?focusedCommentId=108772&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-108772)

> When an image is uploaded, it should be given a description - which lasts for all time - NOT a caption, which is text written for a specific, one-time usage of the image. 
> Descriptions remain in the image repo for all time as the definitive version of what is depicted in a full image.
> Captions never get written back to the image repo; they merely reflect the use of an image, or part of it, at the time of its publication/broadcast for that instance only.

Related PRs: https://github.com/superdesk/superdesk-client-core/pull/2508 https://github.com/superdesk/superdesk-client-core/pull/2502

PR #2573, opened 2018-10-02
fix(editor3): fix editor3 editing performance

petrjasek:
SDFID-502


tomaskikutis:
:+1: Don't forget to `git blame` around the changed code and assign related issues like `SDESK-2886` for QA to re-test on your branch before merging

ioanpocol:
Also SDESK-3456 can be related

PR #2576, opened 2018-10-02
Refactor compliant filters with elastic date ranges

pablopunk:
Not a task, just something that was missing on a recent refactor

### DIFF
--- a/scripts/apps/search/directives/DateFilters.ts
+++ b/scripts/apps/search/directives/DateFilters.ts
@@ -7,6 +7,24 @@ interface IDateRange {
     };
 }
 
+const nextMonthFilter: IDateRange = {
+    key: 'next_month',
+    label: gettext('Next Month'),
+    elasticSearchDateRange: {
+        lte: 'now+1M/M',
+        gte: 'now+1M/M',
+    },
+};
+
+const next3MonthsFilter: IDateRange = {
+    key: 'next_3_months',
+    label: gettext('Next 3 Months'),
+    elasticSearchDateRange: {
+        lte: 'now+3M/M',
+        gte: 'now+3M/M',
+    },
+};
+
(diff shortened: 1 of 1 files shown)


scripts/apps/search/directives/DateFilters.ts
```diff
@@ -7,6 +7,24 @@ interface IDateRange {
...
+    },
+};
+
+const next3MonthsFilter: IDateRange = {
+    key: 'next_3_months',
+    label: gettext('Next 3 Months'),
+    elasticSearchDateRange: {
+        lte: 'now+3M/M',
+        gte: 'now+3M/M',
+    },
```
tomaskikutis: If today is 2nd Oct, should next 3 months mean 2nd October to 31st December inclusive or November 1st - January 1st inclusive or something different?

pablopunk: it should be from today (Oct 2nd) to January 2nd inclusive

petrjasek: I think you should round it to day then `/d` otherwise it would be till the end of month `/M`

pablopunk: You're right. I'm also thinking I need to remove `gte` as we also need old articles to be marked as "overdue"

scripts/apps/search/directives/DateFilters.ts
```diff
@@ -51,6 +67,8 @@ const last8hoursFilter: IDateRange = {
 };
 
 export const dateRangesByKey: Dictionary<string, IDateRange> = {
+    before_1_month: before1MonthFilter,
```
tomaskikutis: `before_next_month` and `before_3_months_ahead` sound better to me.

PR #2578, opened 2018-10-03
fix(editor3): Fix the revert to previous version on editor3 (SDESK-3460)

ioanpocol:


scripts/core/editor3/directive.js
```diff
@@ -172,6 +180,25 @@ class Editor3Directive {
             store.dispatch(changeEditorState(editorState));
         });
 
+        // bind the directive refreshTrigger attribute bi-directionally between Angular and Redux.
+        $scope.$watch('vm.refreshTrigger', (val, old) => {
+            const draftjsRawState = getFieldMetadata(
```
tomaskikutis: Can we use `getInitialContent` here?

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -727,6 +728,7 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
              * Revert item to given version
              */
             $scope.revert = function(version) {
+                $scope.refreshTrigger = !$scope.refreshTrigger;
```
tomaskikutis: I think it's better to use an integer which increments when you need to update the state. When booleans are used I was intuitively thinking that update happens after boolean is set to `true` which is not the case.

PR #2581, opened 2018-10-04
fixed: Date filters expected behaviour

GyanP:
https://dev.sourcefabric.org/browse/SDFID-501

### DIFF
--- a/spec/search_spec.js
+++ b/spec/search_spec.js
@@ -299,9 +299,9 @@ describe('search', () => {
         scheduleFilter.click();
 
         element.all(by.buttonText('Last 24 Hours')).then(function(items) {
-          expect(items.length).toBe(4);
-          scrollToView(items[3])
-          items[3].click();
+            expect(items.length).toBe(4);
+            scrollToView(items[3]);
+            items[3].click();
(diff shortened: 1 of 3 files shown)


petrjasek:
can you pls rebase the PR?

GyanP:
rebase is done, I am looking e2e tests issue.

spec/search_spec.js
```diff
@@ -298,11 +298,12 @@ describe('search', () => {
         scrollToView(scheduleFilter);
         scheduleFilter.click();
 
-        var scheduleDay = element(by.buttonText('Last 24 Hours'));
+        element.all(by.buttonText('Last 24 Hours')).then((items) => {
+            expect(items.length).toBe(4);
+            scrollToView(items[3]);
+            items[3].click();
```
tomaskikutis: That's a very error-prone approach to click the 3rd item. Tests will break if you swap the order of filters or add another filter while it is easy way to avoid it. To do that first get the group of filters you are testing which is probably 'Date scheduled' and then search for 'Last 24 Hours' in that group. There will only be one match this way.

PR #2583, opened 2018-10-05
Display translated items (sdfid 489)

tomaskikutis:
SDFID-489
DEPENDS ON: https://github.com/superdesk/superdesk-core/pull/1381
DEPENDS ON: SDESK-3487


scripts/apps/authoring/translations/translationsWidget.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+    render() {
+        const {$filter, gettextCatalog, authoringWorkspace, datetime} = this.props;
+        const {translations, references} = this.state;
+
+        if (translations == null) {
+            return null;
+        }
+
+        const sortOriginalFirst = (a: IArticle) => a.translated_from == null ? -1 : 1;
```
petrjasek: in theory you could sort by firstcreated that should make the order stable for other translations

scripts/apps/authoring/views/authoring-header.html
```diff
@@ -4,49 +4,68 @@
...
         <div>
-            <label translate>SOURCE</label>
-            <div id="item-source" class="data">{{ item.source}}</div>
+            <span class="authoring-header__label" translate>SOURCE</span>
+            <div class="authoring-header__value" id="item-source" >{{ item.source}}</div>
         </div>
-        <div ng-if="relatedItems._items.length">
-            <label translate>RELATED</label>
-            <div class="data"><a href="" ng-click="activateWidget()">{{relatedItems._items.length}}</a></div>
+        <div ng-if="relatedItems != null && Array.isArray(relatedItems._items) && relatedItems._items.length > 0">
```
petrjasek: angular should handle all those error states so checking for `.length` being `> 0` or just `true` value is enough and more readable imo

scripts/apps/authoring/views/authoring-header.html
```diff
@@ -4,49 +4,68 @@
...
             <span class="state-label updated" ng-if="item.rewritten_by" translate>Updated</span>
         </div>
         <div ng-if="item.broadcast.master_id">
-            <label ng-class="{'broadcast-status': item.broadcast.status !== ''}" title="{{item.broadcast.status| translate}}"><span translate>MASTER</span><b ng-if="item.broadcast.status"> !</b></label>
-            <div class="data"><a id="preview-master" ng-click="previewMasterStory()" title="{{ :: 'Preview master story' | translate }}" ng-app=""class="open-item">
+            <span
+                class="authoring-header__label"
+                ng-class="{'broadcast-status': item.broadcast.status !== ''}"
+                title="{{item.broadcast.status| translate}}"
+                style="color: #e51c23;">
```
petrjasek: that's like some magic number to me, class would be better

scripts/core/datetime/relativeDate.tsx
```diff
@@ -0,0 +1,28 @@
...
+        const reldate = date.fromNow();
+
+        return (
+            <time dateTime={datetimeIso} title={title}>{reldate}</time>
+        );
+    }
+}
+
+angular.module('superdesk.core.datetime.relativeDate', [])
+    .component('relativeDate', reactToAngular1(RelativeDate, ['datetime']));
```
petrjasek: it's better to keep the `sd` prefix for components to avoid conflicts with external libs

scripts/apps/authoring/views/authoring-header.html
```diff
@@ -4,49 +4,68 @@
...
         <div id="item-correction" ng-if="item.correction_sequence">
-            <label translate>UPDATE</label>
-            <div class="data">{{item.correction_sequence}}</div>
+            <span class="authoring-header__label" translate>UPDATE</span>
+            <div class="authoring-header__value">{{item.correction_sequence}}</div>
         </div>
-        <div>
-            <span class="state-label missing-link" ng-if="missing_link" translate>Missing Link</span>
+        <div ng-if="missing_link || item.flags.marked_for_legal || item.flags.marked_for_sms || item.flags.marked_for_not_publication || item.rewritten_by">
+            <span class="state-label missing-link" ng-if="" translate>Missing Link</span>
```
mdhaman: it should be `ng-if="missing_link"`

scripts/apps/authoring/translations/translationsWidget.tsx
tomaskikutis: `firstcreated` is the same for all translations.

petrjasek: that's weird, imo it should be different

tomaskikutis: Created `SDESK-3487` for this.

PR #2587, opened 2018-10-08
feat(upload): show all fields from picture profile in upload form

petrjasek:
it will also add fields defined in validator for now, it could be
obsolete but keeping it for backward compatibility.

uses https://github.com/superdesk/superdesk-core/pull/1383

SDANSA-197


petrjasek:
@mdhaman yep it's using content profile for pictures, but atm it's only possible to [configure it via settings](https://github.com/superdesk/superdesk/blob/ansa/server/settings.py#L97-L268) and not UI.

PR #2588, opened 2018-10-08
fix(editor3): Fix preview versions and revert to initial fetched vers…

ioanpocol:
…ion editor3 (SDESK-3460)


scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -720,6 +720,7 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
              * Preview different version of an item
              */
             $scope.preview = function(version) {
+                $scope.refreshTrigger = $scope.refreshTrigger + 1;
```
petrjasek: it's like `$scope.refreshTrigger++;`

PR #2598, opened 2018-10-11
refactor logger to be a stateless singleton

tomaskikutis:
Makes it easier to use the sentry logger without having to load it as a service and pass it down 20 levels deep.


scripts/core/services/logger.js
```diff
@@ -30,15 +33,4 @@ export class Logger {
...
- * @module superdesk.core.services
- * @name loger
- *
- * @description
- * logger service
- *
- * Sends logs to sentry servers
- */
-export default angular.module('superdesk.core.services.logger', [])
-    .service('logger', ['config', Logger]);
```
petrjasek: looks good, but I would keep the angular service where you can set logger as constant or something so it won't break any external apps which might be using it (like planning, publisher)

PR #2616, opened 2018-10-25
bring back config.js

petrjasek:
it seems to be required by deployment scripts


pablopunk:
I also noticed an error on my local server throwing a 404 for this file every time I visit the page

PR #2619, opened 2018-10-29
Fix SMS and Legal Filters in search

marwoodandrew:


marwoodandrew:
A change in the elasticsearch version had broken it!

PR #2621, opened 2018-10-29
fix(ansa): allow custom search params via including template

petrjasek:


scripts/apps/search/directives/SearchParameters.js
```diff
@@ -403,6 +404,10 @@ export function SearchParameters($location, asset, tags, metadata, common, desks
                     delete scope.fields[type];
                 }
             };
+
+            scope.getTemplate = (providerType) => providerType && `search-panel-${providerType}.html`;
```
tomaskikutis: I'd like to avoid type coercion and compare explicitly instead: `providerType != null`. Using single equals help to check for `undefined` as well which is kind of type coercion as well, but a clearer one IMO. We can use lodash's `isNil` or make our own helper if `!= null` doesn't work.

I think it'd be a good idea to avoid abusing the `&&` operator. When I see logical operator I expect boolean result. I know it's a js hack, but I prefer being explicit `providerType == null ? false : search-panel-${providerType}.html`. Anyway, I don't think we need the check in the first place since it's already checked in the template.

scripts/apps/search/directives/SearchParameters.js
```diff
@@ -403,6 +404,10 @@ export function SearchParameters($location, asset, tags, metadata, common, desks
                     delete scope.fields[type];
                 }
             };
+
+            scope.getTemplate = (providerType) => `search-panel-${providerType}.html`;
+
+            scope.hasTemplate = (providerType) => providerType && !!$templateCache.get(scope.getTemplate(providerType));
```
tomaskikutis: Could you apply the same changes here as well? I mean comparing explicitly and not using type coercion

petrjasek: :ok:

tomaskikutis: Here's a fix for type coercion `!!$templateCache.get(scope.getTemplate(providerType))` -> `$templateCache.get(scope.getTemplate(providerType)) != null`

Thanks :)

petrjasek:
have to merge this now, that failing test is not related

PR #2627, opened 2018-10-30
Separate article header and article content configuration in content profiles (SDESK 3409)

tomaskikutis:
SDESK-3409

### DIFF
--- a/scripts/core/helpers/typescript-helpers.ts
+++ b/scripts/core/helpers/typescript-helpers.ts
@@ -1,3 +1,7 @@
 export function nameof<T>(name: keyof T) {
     return name;
 }
+
+export function assertNever(x: never): never {
+    throw new Error('Unexpected object: ' + x);
+}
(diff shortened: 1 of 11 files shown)


tomaskikutis:
Ignore the failing test, it's not related to this PR. It's an issue of API returning desk members in random order. Will be fixed in SDESK-3566

scripts/core/helpers/typescript-helpers.ts
```diff
@@ -1,3 +1,7 @@
 export function nameof<T>(name: keyof T) {
     return name;
 }
+
+export function assertNever(x: never): never {
```
pablopunk: what is this function for?

tomaskikutis: It makes exhaustive switch possible.

```
type sectionToRender = 'header' | 'content';

function handleIt (section: sectionToRender) {
    switch(section) {
        case 'header':
            // ...
            break;
        case 'content':
            // ...
            break;
        default:
            assertNever(section)
    }
}
```

If you add another possible value to `sectionToRender` type, the code won't compile until you handle it in all switch statements which switch on that type.

pablopunk: Cool

PR #2648, opened 2018-11-13
Uploading media to sd master is not working (SDESK 3610)

tomaskikutis:
SDESK-3610
DEPENDS ON: https://github.com/superdesk/superdesk-ui-framework/pull/284

### DIFF
--- a/scripts/apps/monitoring/views/monitoring-view.html
+++ b/scripts/apps/monitoring/views/monitoring-view.html
@@ -21,7 +21,7 @@
             <button class="dropdown__toggle navbtn navbtn--text-only" dropdown-toggle>{{ :: 'Monitoring' | translate }} {{workspace.name}} <span class="dropdown__caret"></span></button>
             <ul class="dropdown__menu scrollable">
                 <li><div class="dropdown__menu-label">{{ :: 'Stages' | translate }}</div></li>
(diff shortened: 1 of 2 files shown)


scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -21,7 +21,7 @@
             <button class="dropdown__toggle navbtn navbtn--text-only" dropdown-toggle>{{ :: 'Monitoring' | translate }} {{workspace.name}} <span class="dropdown__caret"></span></button>
             <ul class="dropdown__menu scrollable">
                 <li><div class="dropdown__menu-label">{{ :: 'Stages' | translate }}</div></li>
-                <li ng-repeat="card in aggregate.cards" ng-class="{active: monitoring.singleGroup.">
+                <li ng-repeat="card in aggregate.cards" ng-class="{active: monitoring.singleGroup != null}">
```
tomaskikutis: @petrjasek you added an invalid expression in the previous commit. Check out if my fix is appropriate for what you intended.

petrjasek: I think this ng-class bit shouldn't be there at all

tomaskikutis: Ok, I'll remove it then

PR #2649, opened 2018-11-14
feat(vocabulary): Add the option to set 'single value', 'multi value'…

ioanpocol:
… and 'do not show' opyions for selection type (SDESK-3551)

### DIFF
--- a/scripts/apps/vocabularies/constants.ts
+++ b/scripts/apps/vocabularies/constants.ts
@@ -15,6 +15,21 @@ export const MEDIA_TYPES = {
 
 export const MEDIA_TYPE_KEYS = Object.keys(MEDIA_TYPES).map((type) => MEDIA_TYPES[type].id);
 
+export const VOCABULARY_SELECTION_TYPES = {
+    SINGLE_SELECTION: {
+        id: 'single selection',
+        label: gettext('Single selection')
+    },
+    MULTIPLE_SELECTION: {
+        id: 'multi selection',
+        label: gettext('Multi selection')
+    },
+    DO_NOT_SHOW: {
+        id: 'do not show',
+        label: gettext('Do not show')
+    },
+};
+
 export const DEFAULT_SCHEMA = {
     name: {},
     qcode: {},
--- a/scripts/apps/vocabularies/views/vocabulary-config-modal.html
+++ b/scripts/apps/vocabularies/views/vocabulary-config-modal.html
@@ -55,9 +55,7 @@
                 <div class="sd-line-input" ng-if="matchFieldTypeToTab('vocabularies', vocabulary.field_type)">
                     <label for="cv_selection_type" class="sd-line-input__label" translate>Show to users as:</label>
                     <select class="input-medium" ng-model="vocabulary.selection_type">
-                        <option value="single selection" translate>single selection</option>
-                        <option value="multi selection" translate>multi selection</option>
-                        <option value="do not show" translate>do not show</option>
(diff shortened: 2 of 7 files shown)


scripts/apps/vocabularies/views/vocabulary-config-modal.html
```diff
@@ -52,6 +52,15 @@ <h3 class="modal__heading" ng-show="!vocabulary._id"><span translate>Create</spa
                          ng-show="vocabularyForm.$error.maxlength" translate>The helper text must not exceed 120 characters.</div>
                 </div>
 
+                <div class="sd-line-input" ng-if="matchFieldTypeToTab('vocabularies', vocabulary.field_type)">
+                    <label for="cv_selection_type" class="sd-line-input__label" translate>Show to users as:</label>
+                    <select class="input-medium" ng-model="vocabulary.selection_type">
+                        <option value="single selection" translate>single selection</option>
```
tomaskikutis: Create a constant with selection types. Iterate it in this select and use it for comparison in other places.

ioanpocol: Done

scripts/apps/vocabularies/views/vocabulary-config-modal.html
```diff
@@ -52,6 +52,13 @@ <h3 class="modal__heading" ng-show="!vocabulary._id"><span translate>Create</spa
                          ng-show="vocabularyForm.$error.maxlength" translate>The helper text must not exceed 120 characters.</div>
                 </div>
 
+                <div class="sd-line-input" ng-if="matchFieldTypeToTab('vocabularies', vocabulary.field_type)">
+                    <label for="cv_selection_type" class="sd-line-input__label" translate>Show to users as:</label>
```
tomaskikutis: `for` attribute only makes sense if there's another element with a corresponding ID. In this case, `select` element should have the id `cv_selection_type`

scripts/apps/vocabularies/constants.ts
```diff
@@ -15,6 +15,21 @@ export const MEDIA_TYPES = {
...
+export const VOCABULARY_SELECTION_TYPES = {
+    SINGLE_SELECTION: {
+        id: 'single selection',
+        label: gettext('Single selection')
+    },
+    MULTIPLE_SELECTION: {
+        id: 'multi selection',
+        label: gettext('Multi selection')
+    },
+    DO_NOT_SHOW: {
```
tomaskikutis: I can't find any usages of this. Is it needed?

ioanpocol: Yes, there is a task about it:
https://dev.sourcefabric.org/browse/SDESK-3625

tomaskikutis: Are going to implement this in another PR or should it be working right now?

ioanpocol: The task is still in TODO, so the fix will be in a different PR.

scripts/apps/vocabularies/views/vocabulary-config-modal.html
ioanpocol: Done

tomaskikutis: Please also fix the issue I mentioned in the review comment

ioanpocol:
>scripts/core/notification/notification.js:118 - Put data.vocabulary outside of gettext since the translation will never get matched if it's inside. I know you haven't changed this part, but let's fix it anyway.

Done

PR #2653, opened 2018-11-15
fix(metadata): For media or related item cases activate save button o…

ioanpocol:
…nly when at least one content types is selected (SDESK-3564)

### DIFF
--- a/scripts/apps/authoring/views/article-edit.html
+++ b/scripts/apps/authoring/views/article-edit.html
@@ -658,7 +658,7 @@
 
 
 <div class="field" ng-repeat="field in fields track by field._id"
-     ng-if="field.field_type === 'related_content'"
+     ng-if="field.field_type === mediaTypes.related_content.id"
      sd-width="{{editor[field._id].sdWidth || 'full'}}"
      sd-validation-error="error[field._id.toLowerCase()]"
      order="{{editor[field._id].order}}"
--- a/scripts/apps/authoring/views/article-edit.html
+++ b/scripts/apps/authoring/views/article-edit.html
@@ -658,7 +658,7 @@
 
 
 <div class="field" ng-repeat="field in fields track by field._id"
-     ng-if="field.field_type === 'related-content'"
+     ng-if="field.field_type === 'related_content'"
      sd-width="{{editor[field._id].sdWidth || 'full'}}"
      sd-validation-error="error[field._id.toLowerCase()]"
      order="{{editor[field._id].order}}"
--- a/scripts/apps/authoring/views/article-edit.html
+++ b/scripts/apps/authoring/views/article-edit.html
@@ -658,7 +658,7 @@
 
 
 <div class="field" ng-repeat="field in fields track by field._id"
-     ng-if="field.field_type === mediaTypes.related_content.id"
+     ng-if="field.field_type === mediaTypes.RELATED_CONTENT.id"
      sd-width="{{editor[field._id].sdWidth || 'full'}}"
      sd-validation-error="error[field._id.toLowerCase()]"
      order="{{editor[field._id].order}}"
--- a/scripts/apps/vocabularies/constants.ts
+++ b/scripts/apps/vocabularies/constants.ts
@@ -1,11 +1,15 @@
+// Adding the following because planning webpack when compiled for test cases
+// won't be aware of gettext.
+const gettext = _.get(window, 'gettext', (text) => text);
+
 export const MEDIA_TYPES = {
     GALLERY: {
         id: 'media',
-        label: 'Media gallery'
+        label: gettext('Media gallery')
     },
     RELATED_CONTENT: {
         id: 'related_content',
-        label: 'Related items'
+        label: gettext('Related items')
     },
 };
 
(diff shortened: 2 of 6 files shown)


scripts/apps/vocabularies/views/vocabulary-config-modal.html
```diff
@@ -92,7 +93,7 @@ <h3 class="modal__heading" ng-show="!vocabulary._id"><span translate>Create</spa
                             <div class="form__row-item form__row-item--no-grow">
                                 <select class="input-medium" ng-model="vocabulary.field_type">
                                     <option value="media" translate>Media gallery</option>
-                                    <option value="related-content" translate>Related items</option>
+                                    <option value="related_content" translate>Related items</option>
```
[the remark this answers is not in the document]
ioanpocol: Done

scripts/apps/authoring/views/article-edit.html
```diff
@@ -658,7 +658,7 @@
 
 
 <div class="field" ng-repeat="field in fields track by field._id"
-     ng-if="field.field_type === 'related-content'"
+     ng-if="field.field_type === 'related_content'"
```
tomaskikutis: Still hardcoded here. I expect `ng-if="field.field_type === MEDIA_TYPES.relatedContent"` or similar

scripts/apps/vocabularies/controllers/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const MEDIA_TYPES = {
+    'media': 'Media gallery',
+    'related_content': 'Related items'
```
tomaskikutis:
```
export const MEDIA_TYPES = {
    GALLERY: {
        id: 'GALLERY',
        label: gettext('Media gallery')
    },
    RELATED_CONTENT: {
        id: 'RELATED_CONTENT',
        label: gettext('Related items')
    },
};
```

I think defining it in this way would be better since you could access the id by referring to the object. Like `MEDIA_TYPES.GALLERY.id`. Otherwise, you would have to hardcode the id in the usage. It will not be fun to grep 'media' and expect to find usages.

scripts/apps/authoring/views/article-edit.html
ioanpocol: Done

scripts/apps/vocabularies/controllers/constants.ts
ioanpocol: Done

scripts/apps/vocabularies/constants.ts
```diff
@@ -0,0 +1,18 @@
+export const MEDIA_TYPES = {
+    media: {
```
tomaskikutis: It sounds silly having a media type called `media`. Can you rename it to `gallery`? You also need to wrap the labels in `gettext` - You can use the global `gettext` which will at least mark the translations. We will make it work in the upcoming refactoring of translations.

scripts/apps/vocabularies/constants.ts
```diff
@@ -0,0 +1,18 @@
...
+        id: 'media',
+        label: 'Media gallery'
+    },
+    related_content: {
+        id: 'related_content',
+        label: 'Related items'
+    },
+};
+
+export const MEDIA_TYPE_KEYS = Object.keys(MEDIA_TYPES);
```
tomaskikutis: I think you should use IDs, since keys are simply labels which should be freely refactorable.

scripts/apps/vocabularies/constants.ts
ioanpocol: I prefer to let it called media in order to minimise the data migrations.
The text is already translated on html files.

scripts/apps/vocabularies/constants.ts
ioanpocol: Yeah, a lot of over-complications, ....
Let's keep as it is, it is not needed to have refactorable keys.

tomaskikutis:
Did I mention that it would be nice if you didn't rebase the commits and I could track changes easily and not have to re-read the entire thing on every commit? :grimacing:

ioanpocol:
I think it is better to keep all those style changes in one commit, for sure it will be easily to be read :)

scripts/apps/vocabularies/constants.ts
tomaskikutis: Which naming do you think gives more information about the action `addItem(MEDIA_TYPES.media)` or `addItem(MEDIA_TYPES.gallery)`

`MEDIA_TYPES.media` is just bad naming. It doesn't say which type of media it is. You refused to rename it, because you don't want to write migration scripts which is fair. I then suggest a method without the need to write migration scripts, and you now say you don't need the ability to rename things? :sweat:

tomaskikutis:
You can add "#STYLE" or any other tag to the commit message so it's possible to review them together and separately so it works for both use cases.

scripts/apps/vocabularies/constants.ts
```diff
@@ -0,0 +1,18 @@
+export const MEDIA_TYPES = {
+    GALLERY: {
+        id: 'media',
+        label: 'Media gallery'
```
tomaskikutis: Sorry, forgot this one. You said you are already translating this in the templates, which is fine, but it won't get picked up to be sent for translation by the grunt script if you don't wrap it in `gettext`

ioanpocol: done

PR #2655, opened 2018-11-16
Added search to ingest management page in Settings

GyanP:
https://dev.sourcefabric.org/browse/SDESK-3552

### DIFF
--- a/scripts/apps/ingest/directives/IngestSourcesContent.js
+++ b/scripts/apps/ingest/directives/IngestSourcesContent.js
@@ -107,11 +107,14 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                     if (orTerms !== null) {
                         criteria.where = JSON.stringify(orTerms);
                     }
+                    const searchTerm = $scope.searchQuery;
 
                     return api.ingestProviders.query(criteria)
                         .then((result) => {
-                            $scope.providers = result;
-                            $scope.searchPage = null;
+                            if ($scope.searchQuery === searchTerm) {
+                                $scope.providers = result;
+                                $scope.searchPage = null;
+                            }
                         });
                 }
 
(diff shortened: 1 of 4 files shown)


scripts/apps/ingest/views/settings/ingest-sources-content.html
```diff
@@ -10,7 +11,7 @@
 
     <div class="sd-page__content">
         <ul class="pills-list provider-list">
-            <li ng-repeat="provider in providers._items" class="clearfix">
+            <li ng-repeat="provider in providers._items | filter:query" class="clearfix">
```
tomaskikutis: There's pagination in the ingest-sources page, so it should search all pages.

tomaskikutis: You don't need `| filter:query` anymore if you are filtering using the API. Do you need to change the markup at all in this case?

scripts/apps/ingest/directives/IngestSourcesContent.js
```diff
@@ -79,14 +79,37 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
...
+                        orTerms = {$or: [
+                            {name: {
+                                $regex: $scope.searchQuery,
+                                $options: '-i'},
+                            },
+                        ]};
+                    }
+
+                    if (orTerms !== null) {
+                        criteria.page = 1;
```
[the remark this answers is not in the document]
GyanP: > How does one get to the second page?

you can go to the second page simply like you were going earlier.
`}
                     if (orTerms !== null) {
                        criteria.page = 1;`

If you are on the second page and start searching for some content and it is on the first page then
the criteria.page should be set to 1. So, that data searching starts from the first page always.

scripts/apps/ingest/views/settings/ingest-sources-content.html
GyanP: > You don't need `| filter:query` anymore if you are filtering using the API. Do you need to change the markup at all in this case?

I didn't change any markup.
Design breaks are there because the branch was not updated with the master.

scripts/apps/ingest/directives/IngestSourcesContent.js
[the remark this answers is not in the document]
tomaskikutis: I tested it with the page size of 3 items and pagination after filtering didn't work for me. I might have done something wrong, so check it yourself.

PR #2666, opened 2018-11-21
fix(package move): Add more checks for contextual menu in the case of…

ioanpocol:
…  published items moved along with parent package (SDFID-372)

### DIFF
--- a/scripts/apps/authoring/authoring/services/AuthoringService.js
+++ b/scripts/apps/authoring/authoring/services/AuthoringService.js
@@ -418,7 +418,9 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         if (self.isPublished(currentItem)) {
             // if not the last published version
             if (angular.isDefined(item.archive_item) &&
-                item._current_version !== item.archive_item._current_version && item.operation !== 'move') {
+                item._current_version !== item.archive_item._current_version &&
+                (item.operation == null || item.operation !== 'move') &&
+                (item.archive_item.operation == null || item.archive_item.operation !== 'move')) {
                 return angular.extend({}, helpers.DEFAULT_ACTIONS);
             }
 
--- a/scripts/apps/authoring/authoring/services/AuthoringService.js
+++ b/scripts/apps/authoring/authoring/services/AuthoringService.js
@@ -417,9 +417,7 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         // item is published state - corrected, published, scheduled, killed
         if (self.isPublished(currentItem)) {
             // if not the last published version
-            if (angular.isDefined(item.archive_item) &&
-                item._current_version !== item.archive_item._current_version &&
-                item.operation !== 'move' && item.archive_item.operation !== 'move') {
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/services/AuthoringService.js
```diff
@@ -418,7 +418,9 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         if (self.isPublished(currentItem)) {
             // if not the last published version
             if (angular.isDefined(item.archive_item) &&
-                item._current_version !== item.archive_item._current_version && item.operation !== 'move') {
+                item._current_version !== item.archive_item._current_version &&
+                (item.operation == null || item.operation !== 'move') &&
```
tomaskikutis: Why do you compare to `null`? Is there a case where `(item.operation == null || item.operation !== 'move')` and `(item.operation !== 'move')` produce different results?

ioanpocol: changed

tomaskikutis: Thanks. I'm wondering why the checks have to be here in the first place. If I understand correctly, the function `itemActions` should return which actions are available on the item which in turn are displayed [here](https://user-images.githubusercontent.com/2076953/48822035-39617180-ed64-11e8-8d6b-2952c56307e1.png). What does it have to do with the operation?

ioanpocol: Somehow the condition tried to detect if it is the last published version for item.
Seams that there is already a field for this, 'last_published_version' so I used it.

scripts/apps/authoring/authoring/services/AuthoringService.js
```diff
@@ -417,9 +417,7 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         // item is published state - corrected, published, scheduled, killed
         if (self.isPublished(currentItem)) {
             // if not the last published version
-            if (angular.isDefined(item.archive_item) &&
-                item._current_version !== item.archive_item._current_version &&
-                item.operation !== 'move' && item.archive_item.operation !== 'move') {
+            if (item.last_published_version == false) {
```
tomaskikutis: It should be three equals in the comparison.

ioanpocol: done

tomaskikutis: Looks good now. I'll approve it once tests pass :)

petrjasek: just wondering - is it better to use `=== false` instead of `! item.last_published_version` ?

tomaskikutis: I prefer comparing to `false` since it the allows me to infer that the value is boolean and not, for example, an integer. If it was a property of a TypeScript object, I wouldn't care either way.

ioanpocol: Any conclusion on this?

PR #2672, opened 2018-11-23
fix(media): fix editing of vocabularies in media editor

petrjasek:
SDANSA-213

### DIFF
--- a/scripts/apps/authoring/media/MediaMetadataEditorDirective.ts
+++ b/scripts/apps/authoring/media/MediaMetadataEditorDirective.ts
@@ -1,10 +1,16 @@
 import {max, sortBy, get} from 'lodash';
 import {getLabelNameResolver} from 'apps/workspace/helpers/getLabelForFieldId';
 
-MediaMetadataEditorDirective.$inject = ['metadata', 'deployConfig', '$q'];
-export default function MediaMetadataEditorDirective(metadata, deployConfig, $q) {
+MediaMetadataEditorDirective.$inject = ['metadata', 'deployConfig', 'features', '$q'];
+export default function MediaMetadataEditorDirective(metadata, deployConfig, features, $q) {
     function getCV(field) {
-        return metadata.cvs.find((cv) => !cv.field_type && (cv._id === field || cv.schema_field === field));
+        const cv = metadata.cvs.find((cv) => !cv.field_type && (cv._id === field || cv.schema_field === field));
+
+        if (cv == null && field === 'subject') {
+            return {items: metadata.values.subjectcodes}; // fallback for built in subjectcodes
+        }
+
+        return cv;
     }
 
     return {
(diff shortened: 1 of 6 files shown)


scripts/apps/search/MultiImageEdit.ts
```diff
@@ -44,7 +44,7 @@ export function MultiImageEditController(
         const updatedImages = imagesOriginal.map((image) => {
             const existingImage = $scope.images.find(({_id}) => _id === image._id);
 
-            return existingImage != null ? existingImage : {...image, ...$scope.metadata};
+            return existingImage != null ? existingImage : {...image, ...pick($scope.metadata, identity)};
```
tomaskikutis: According to docs, the second argument for `pick` is `[paths] (...(string|string[]))` and you are passing a function. How does this work?

scripts/apps/authoring/media/MediaMetadataEditorDirective.ts
```diff
@@ -1,10 +1,16 @@
...
 
-MediaMetadataEditorDirective.$inject = ['metadata', 'deployConfig', '$q'];
-export default function MediaMetadataEditorDirective(metadata, deployConfig, $q) {
+MediaMetadataEditorDirective.$inject = ['metadata', 'deployConfig', 'features', '$q'];
+export default function MediaMetadataEditorDirective(metadata, deployConfig, features, $q) {
     function getCV(field) {
-        return metadata.cvs.find((cv) => !cv.field_type && (cv._id === field || cv.schema_field === field));
+        const cv = metadata.cvs.find((cv) => !cv.field_type && (cv._id === field || cv.schema_field === field));
+
+        if (cv == null && field === 'subject') {
```
tomaskikutis: It'd be useful to leave a comment here about what's going on. Why is the custom vocabulary exceptional instead of being included in `metadata.cvs`?

scripts/apps/search/MultiImageEdit.ts
petrjasek: yep weird, I think it uses pickBy if you pass a function, undocumented feature. will fix that

scripts/apps/search/MultiImageEdit.ts
```diff
@@ -44,7 +44,7 @@ export function MultiImageEditController(
         const updatedImages = imagesOriginal.map((image) => {
             const existingImage = $scope.images.find(({_id}) => _id === image._id);
 
-            return existingImage != null ? existingImage : {...image, ...$scope.metadata};
+            return existingImage != null ? existingImage : {...image, ...pickBy($scope.metadata)};
```
tomaskikutis: Are you picking everything? Do you need to pick at all then?

petrjasek: I'm picking everything that's not empty

tomaskikutis: Can you then use `(item) => item != null` predicate. It's sneaky that lodash is silently ignoring `null`.

petrjasek: added explicit check

PR #2674, opened 2018-11-26
Make validation of "point of interest" configurable

tomaskikutis:
SDESK-3656

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
@@ -107,36 +107,8 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
     * modified crop information, point of interest and metadata changes.
     */
     $scope.done = function() {
-        /* Throw an exception if PoI is outside of a crop */
-        function poiIsInsideEachCrop() {
-            const originalImage = $scope.data.metadata.renditions.original;
-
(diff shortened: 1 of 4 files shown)


scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -107,36 +107,8 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
     * modified crop information, point of interest and metadata changes.
     */
     $scope.done = function() {
```
mdhaman: @tomaskikutis Why these check are removed? This is going break our workflow.

Checks have been added so that user don't just update the metadata and publish the item. Without these checks the users get the impression that they have to just update the metadata tab and publish the item and this caused wrong crops going out to subscriber. These checks at least ensures that users have to "click" on the crops tab and see the crops and save the item.

mdhaman: Can't access the discussion. 

Configuration is fine for us.

superdesk.config.js
```diff
@@ -15,6 +15,7 @@ module.exports = function(grunt) {
             savedSearch: {
                 subscriptions: true,
             },
+            validatePointOfInterestForImages: false,
```
petrjasek: I think we should set it to `true` so we don't change the default behaviour

PR #2677, opened 2018-11-27
fixed: Content type filter is empty in publish queue on sdp-master

GyanP:
https://dev.sourcefabric.org/browse/SDESK-3655

### DIFF
--- a/scripts/apps/publish/controllers/PublishQueueController.js
+++ b/scripts/apps/publish/controllers/PublishQueueController.js
@@ -51,8 +51,24 @@ export function PublishQueueController($scope, subscribersService, api, $q, noti
     }));
 
     promises.push(vocabularies.getVocabulary('type').then((items) => {
-        $scope.contentTypes = items.items;
+        $scope.contentTypes = removeDuplicatesByKey(items.items, 'qcode');
     }));
+
+    /* Remove the duplicate items from the array */
+    function removeDuplicatesByKey(collection, keyname) {
+        var output = [],
+            keys = [];
+
+        _.forEach(collection, (item) => {
+            var key = item[keyname];
+
+            if (keys.indexOf(key) === -1) {
+                keys.push(key);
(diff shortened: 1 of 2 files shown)


tomaskikutis:
Are duplicates really the problem? Shouldn't `vocabularies.getVocabulary('type')` already return unique items?

GyanP:
> Are duplicates really the problem? Shouldn't `vocabularies.getVocabulary('type')` already return unique items?

Yeah, `vocabularies.getVocabulary('type')` returns duplicate data. Also, you can check it on sdp-master, when you go to `publish queue`, you'll get that error in the browser console.

tomaskikutis:
I've just checked the [database of sdp-master](https://user-images.githubusercontent.com/2076953/49140356-c6746f80-f2fc-11e8-8839-8a52596e65aa.png). The vocabulary is full of duplicates. I think the database needs to be fixed and nothing needs to be done on the client.

What do you think @petrjasek @jerome-poisson @ioanpocol ?

petrjasek:
yep that db is definitely in wrong state, wondering how it could happened given that it's unmanageable

tomaskikutis:
I've created SDESK-3683 to fix the database.

Gyan, can you close this PR and take the ticket out of the sprint? We'll resolve it once the database is fixed.

GyanP:
> I've created SDESK-3683 to fix the database.
> 
> Gyan, can you close this PR and take the ticket out of the sprint? We'll resolve it once the database is fixed.

Ok, I will close the PR. And about the ticket, should I set the status to open? Which sprint should I move this ticket to?

tomaskikutis:
Yep, set the status to open. Just remove it from the current sprint so it's not in any sprint.

tomaskikutis:
Sorry, clicked "close" instead of "comment". Well, we were about to close it anyway :smile:

PR #2678, opened 2018-11-27
fix(upload): fix iptc metadata loading on image upload

petrjasek:
- and make sure the thumbnails are correct
- and allow upload to be triggered without any editing

SDESK-3665


scripts/apps/search/views/multi-image-edit.html
```diff
@@ -29,8 +29,8 @@ <h3 class="modal__heading modal__heading--grow" translate>Upload media</h3>
 
                         <div class="sd-grid-item sd-grid-item--with-click"
                              ng-click="handleItemClick($event, image)"
-                             ng-repeat="image in images track by image._id"
```
tomaskikutis: Why did you replace tracking by `image._id` to `$index`?

petrjasek: that slipped through, I was debugging something and made that change, will revert

PR #2679, opened 2018-11-27
fix(editor3): we have to call iframely load for new widgets

petrjasek:
in order to make it work with react

https://iframely.com/docs/reactjs

### DIFF
--- a/scripts/core/editor3/components/embeds/EmbedBlock.tsx
+++ b/scripts/core/editor3/components/embeds/EmbedBlock.tsx
@@ -8,6 +8,8 @@ import ng from 'core/services/ng';
 import {loadIframelyEmbedJs} from './loadIframely';
 import {debounce} from 'lodash';
 
+// debounce to avoid multiple widget load calls on initial load
+// when it gets executed for every embed block
 const loadIframely = debounce(loadIframelyEmbedJs, 100);
 
 /**
--- a/scripts/core/editor3/components/embeds/EmbedBlock.tsx
+++ b/scripts/core/editor3/components/embeds/EmbedBlock.tsx
@@ -5,6 +5,10 @@ import {connect} from 'react-redux';
 import {QumuWidget, isQumuWidget} from './QumuWidget';
 import * as actions from '../../actions';
 import ng from 'core/services/ng';
+import {loadIframelyEmbedJs} from './loadIframely';
+import {debounce} from 'lodash';
+
+const loadIframely = debounce(loadIframelyEmbedJs, 100);
 
 /**
  * @ngdoc React
@@ -112,6 +116,14 @@ export class EmbedBlockComponent extends React.Component<any, any> {
         const {block, removeBlock} = this.props;
 
         removeBlock(block.getKey());
+    }
+
+    componentDidMount() {
+        const embed = this.data();
+
+        if (embed.data.html.includes('iframe.ly')) {
+            loadIframely();
+        }
     }
 
     render() {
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/components/embeds/EmbedBlock.tsx
```diff
@@ -5,6 +5,10 @@ import {connect} from 'react-redux';
 import {QumuWidget, isQumuWidget} from './QumuWidget';
 import * as actions from '../../actions';
 import ng from 'core/services/ng';
+import {loadIframelyEmbedJs} from './loadIframely';
+import {debounce} from 'lodash';
+
+const loadIframely = debounce(loadIframelyEmbedJs, 100);
```
tomaskikutis: Should it be loaded more than once? If not, you shouldn't use debouncing. Make it `loaded = true` after the function call ignoring that the request is async and let it fail with the request. If we want to implement retrying, we should implement it properly. In this case, I think not implementing all it is fine.

petrjasek: it should be called more than once, every time a new embed block is rendered. but on initial load if you have embeds already it would get called for each, so this just avoids that and calls it once

tomaskikutis: `loadIframelyEmbedJs` adds the script tag to the document. Adding the script tag to the document should only be done once. You should use `lodash.once` instead of `lodash.debounce`. It may as well be done in `core/editor3/components/embeds/loadIframely.tsx` since the script is not removed through the lifetime of the application.

tomaskikutis:
I was checking `scripts/core/editor3/components/embeds/loadIframely.tsx`. I think we should remove the exception for the test since it's not needed if the method is written properly. There is no need to query the DOM for `s`. The following looks better IMO

```
// see https://iframely.com/docs/reactjs and https://gist.github.com/nleush/7e7775c9709eb3bdb6e6
export function loadIframelyEmbedJs() {
    const iframely = window['iframely'] = window['iframely'] || {};
    const widgets = iframely.widgets = iframely.widgets || {};

    if (widgets.load) {
        widgets.load();
    } else {
        const scriptElement = document.createElement('script');

        scriptElement.type = 'text/javascript';
        scriptElement.async = true;
        scriptElement.src = (document.location.protocol === 'https:' ? 'https:' : 'http:') + '//cdn.iframe.ly/embed.js';

        document.body.appendChild(scriptElement);
    }
}
```

scripts/core/editor3/components/embeds/EmbedBlock.tsx
tomaskikutis: I just noticed that you are trying to emulate `once` using `loaded` variable in `loadIframely.tsx`. So yeah, you can move it up and to the top of the function like this
```
let loaded = false;

export function loadIframelyEmbedJs() {
    if (loaded == true) {
        return;
    }

    loaded === true;

    const iframely = window['iframely'] = window['iframely'] || {};
    const widgets = iframely.widgets = iframely.widgets || {};
    // ...
```

and not use debounce/once or remove the variable and use `once`. There is no point of doing both.

petrjasek: yep it only adds the script once, but then it has to call that `widgets.load()` every time it renders a new block

tomaskikutis: You forgot to remove debouncing. Is there a reason why do you rebase commits? I don't like it mostly for having to review the entire thing again. There is not much to review in this PR, but I also don't get notifications from github about new commits when you push it.

petrjasek: this should still be there, otherwise on first load this would call widgets.load more than once

tomaskikutis: I think I now understand why you use debouncing here. When there is only one embed in the editor3, it's only called once. When there are multiple embeds it's called once per embed which seems is what you are trying to prevent. Is that right?

Why can't you call `widgets.load` more than once on the first load? Is anything breaking due to it or you are just saving http requests?

PR #2688, opened 2018-11-30
Related content item count is not working (SDESK 3617)

tomaskikutis:
SDESK-3617


scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -45,13 +47,31 @@ export function RelatedItemsDirective(authoringWorkspace, relationsService) {
                     event.preventDefault();
                     event.stopPropagation();
 
+                    const relatedItemsForCurrentField = Object.keys(scope.item.associations || {})
+                        .filter((key) =>  key.startsWith(scope.field._id) && scope.item.associations[key] != null)
+                        .map((key) => scope.item.associations[key]);
```
ioanpocol: Looks like map call is not needed, below you need only the list of keys.

tomaskikutis: I need to retrieve `_id` at line 73

ioanpocol: yes, true

PR #2693, opened 2018-12-03
fix(related items): Fix the refresh of the icon when related items ar…

ioanpocol:
…e reordered(SDESK-3644)

### DIFF
--- a/scripts/core/directives/FiletypeIconDirective.ts
+++ b/scripts/core/directives/FiletypeIconDirective.ts
@@ -15,13 +15,13 @@ export default angular.module('superdesk.core.directives.filetypeIcon', [])
 
             scope.$watch('item', (item) => {
                 if (item != null) {
-                    initIcon(item);
+                    setIcon(item);
                 } else {
                     oldClass = null;
                 }
             });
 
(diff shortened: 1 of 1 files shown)


scripts/core/directives/FiletypeIconDirective.ts
```diff
@@ -11,10 +11,13 @@ export default angular.module('superdesk.core.directives.filetypeIcon', [])
     .directive('sdFiletypeIcon', ['gettextCatalog', (gettextCatalog) => ({
         scope: {item: '='},
         link: function(scope, element, attrs) {
-            var stopWatch = scope.$watch('item', (item) => {
-                if (item) {
+            var oldClass = null;
+
+            scope.$watch('item', (item) => {
+                if (item != null) {
                     initIcon(item);
```
tomaskikutis: Rename to updateIcon

ioanpocol: Maybe setIcon since is not always an update

tomaskikutis: That works too

PR #2702, opened 2018-12-07
feat(link): Detect local links based on local_domains cv (SDFID-458)

ioanpocol:


scripts/core/editor3/components/links/LinkInput.tsx
```diff
@@ -80,13 +82,19 @@ export class LinkInputComponent extends React.Component<any, any> {
...
         const _changeLinkSuggestion = this.props.changeLinkSuggestion;
 
         if (linkType === linkTypes.href) {
-            link = {href: this.state.url};
+            const url = this.state.url;
+            const isLocalDomain = (localDomains || []).some((item) => url.includes(item.domain));
+
+            link = {
+                href: url,
+                target: isLocalDomain ? '_self' : '_blank',
```
petrjasek: maybe it could be just `null` instead of `'_self'` it's not needed to be there

petrjasek: I've just never seen it used anywhere, it's the default value after all.
and it changes the output, for that reason I would avoid it.

ioanpocol: I changed the code to don't set any target if local_domains cv is not defined.
Also the target is set only for external links when local_domains cv is defined.

scripts/core/editor3/components/links/LinkInput.tsx
```diff
@@ -80,13 +82,19 @@ export class LinkInputComponent extends React.Component<any, any> {
...
         const _changeLinkSuggestion = this.props.changeLinkSuggestion;
 
         if (linkType === linkTypes.href) {
-            link = {href: this.state.url};
+            const url = this.state.url;
+            const isLocalDomain = (localDomains || []).some((item) => url.includes(item.domain));
+
+            link = {href: url};
+            if (!isLocalDomain && localDomains != null) {
+                link.target = '_blank';
```
tomaskikutis: I think it'd be better to use a custom attribute like `data-sdexternal` instead of a `target`. Target is a display preference and we shouldn't couple it with data.

PR #2709, opened 2018-12-10
Display media contact s correctly if it does not have first name or last name (SDESK 3624)

tomaskikutis:
SDESK-3624

### DIFF
--- a/scripts/apps/contacts/components/fields/ContactName.tsx
+++ b/scripts/apps/contacts/components/fields/ContactName.tsx
@@ -5,7 +5,7 @@ const notEmpty = (str) => typeof str === 'string' && str.length > 0;
 
 export const ContactName: React.StatelessComponent<any> = ({item}) => {
     const displayContact = notEmpty(item.first_name) || notEmpty(item.last_name) ?
-        `${notEmpty(item.first_name) ? item.first_name : ''} ${notEmpty(item.last_name) ? item.last_name : ''}`
+        [item.first_name, item.last_name].filter(notEmpty).join(' ')
         : item.organisation;
 
     return (
--- a/scripts/apps/contacts/components/fields/ContactName.tsx
+++ b/scripts/apps/contacts/components/fields/ContactName.tsx
@@ -1,8 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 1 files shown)


scripts/apps/contacts/components/fields/ContactName.tsx
```diff
@@ -1,8 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+const notEmpty = (str) => typeof str === 'string' && str.length > 0;
+
 export const ContactName: React.StatelessComponent<any> = ({item}) => {
-    const displayContact = item.first_name ? `${item.first_name} ${item.last_name}` : item.organisation;
+    const displayContact = notEmpty(item.first_name) || notEmpty(item.last_name) ?
+        `${notEmpty(item.first_name) ? item.first_name : ''} ${notEmpty(item.last_name) ? item.last_name : ''}`
```
petrjasek: imo something like `[item.first_name, item.last_name].filter(notEmpty).join(' '),` would be better here so there is no extra space added, and maybe it's more readable?

scripts/apps/contacts/components/fields/ContactName.tsx
```diff
@@ -1,8 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+const notEmpty = (str) => typeof str === 'string' && str.length > 0;
+
 export const ContactName: React.StatelessComponent<any> = ({item}) => {
-    const displayContact = item.first_name ? `${item.first_name} ${item.last_name}` : item.organisation;
+    const displayContact = notEmpty(item.first_name) || notEmpty(item.last_name) ?
```
ioanpocol: Can be: 

> const displayContact = [item.first_name, item.last_name].filter(Boolean).join(' ') || item.organisation;

?

tomaskikutis: I prefer it to be explicit, and not rely on type castings. When you filter using `Boolean`, the result is the same, but it's not obvious what you are doing. It would be fine either way in this case, but I like keeping it simple as a general practice in the codebase.

ioanpocol: The functional programming is very powerful but sometime not so easy to understand. To me looks that here is unnecessary code but it is your choice, this is something minor.

PR #2710, opened 2018-12-10
[SDESK-3732] Refactor the Send-to pane to introduce more space for th…

fritzSF:
…e Publish section

### DIFF
--- a/scripts/core/ui/views/sd-datepicker.html
+++ b/scripts/core/ui/views/sd-datepicker.html
@@ -1,6 +1,5 @@
 <div class="datepicker-input">
      <input type="text"
-         class="input-small"
          ng-model="dt"
          ng-change="change()"
          sd-datepicker-inner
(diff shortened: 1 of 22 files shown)


fritzSF:
Might need some code optimisation.

scripts/core/ui/views/sd-datepicker.html
```diff
@@ -1,6 +1,6 @@
 <div class="datepicker-input">
      <input type="text"
-         class="input-small"
+         class=""
```
tomaskikutis: Remove empty attribute

fritzSF: UI framework is updated, it should be fine once the PR is merged. Class removed

tomaskikutis:
I've tested again and the UI problems are still there. I don't think anything will change if you merge it to master. The latest release of UI framework is from 20 days ago, that might be the problem.

fritzSF:
Tomas, please check the _Work in progress_ label on this PR, it won't be merged yet. The UI framework is updated 2 days ago but probably not deployed in SD. 

The right side of the Publish pane is that way by design as there will be additional content – mostly Publisher related stuff. We will most likely introduce some rules to hide it if publisher isn't integrated.

tomaskikutis:
Sorry, I didn't see the WIP label. To use the latest version of UI framework, a release has to be made on GitHub and version updated in `package.json` - that was done 20 days ago - so make sure you do that if you want to use a later version.

PR #2716, opened 2018-12-12
fix(macros): Apply macros simultaneous on body and other fields and fi…

ioanpocol:
…x the replacing of multiple words on body (SDESK-3654)

### DIFF
--- a/scripts/apps/authoring/macros/macros.js
+++ b/scripts/apps/authoring/macros/macros.js
@@ -159,16 +159,20 @@ function MacrosController($scope, macros, desks, autosave, $rootScope, storage) 
 
         $scope.loading = true;
         return macros.call(macro, item).then((res) => {
-            if (!res.diff) {
-                Object.keys(res.item || {}).forEach((field) => {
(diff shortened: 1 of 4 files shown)


scripts/apps/authoring/macros/macros.js
```diff
@@ -159,16 +159,20 @@ function MacrosController($scope, macros, desks, autosave, $rootScope, storage)
...
+                if (isString(res.item[field]) === false || field === 'body_html') {
+                    return;
+                }
+                ignoreFields.push(field);
+                if (res.item[field] !== item[field]) {
+                    $rootScope.$broadcast('macro:refreshField', field, res.item[field]);
+                }
+            });
+            angular.extend($scope.item, _.omit(res.item, ignoreFields));
+            autosave.save($scope.item, $scope.origItem);
```
petrjasek: maybe it should check if there is anything left in `_.omit(res.item, ignoreFields)` first

petrjasek: ok np

PR #2719, opened 2018-12-13
Fix cursor position after paste+undo+redo

pablopunk:
SDESK-3240

This was only happening when pasted text was not coming from
draft-js (i.e when we were handling pasted text manually).

I managed to reproduce this on a unit test and I fixed
it by looking at draft-js internal method to paste text.


scripts/core/editor3/components/handlePastedText.ts
```diff
@@ -46,7 +46,9 @@ function pasteContentFromOpenEditor(
                     internalClipboard.forEach((b) => blocksArray.push(b));
                     const contentState = ContentState.createFromBlockArray(blocksArray);
 
-                    return insertContentInState(editorState, contentState, onChange, editorFormat);
+                    const editorWithContent = insertContentInState(editorState, contentState, editorFormat);
+
+                    return 'handled';
```
tomaskikutis: If you are returning 'handled', is there a point to do the operations from line 44-49?

tomaskikutis: Sorry, I thought you were returning 'not-handled'. Anyway, it doesn't seem `const editorWithContent` is used anywhere. If it isn't, you don't need the operations?

pablopunk: Forgot to call onChange again, fixed it on latest commit

PR #2722, opened 2018-12-17
fixed: Do not validate media gallery type of related items field on p…

GyanP:
…ublishing (SDESK-3550)
https://dev.sourcefabric.org/browse/SDESK-3550

### DIFF
--- a/scripts/apps/relations/services/RelationsService.ts
+++ b/scripts/apps/relations/services/RelationsService.ts
@@ -3,7 +3,7 @@ import {IArticle} from 'superdesk-interfaces/Article';
 RelationsService.$inject = ['archiveService', 'mediaIdGenerator'];
 
 export function RelationsService(archiveService, mediaIdGenerator) {
-    this.getRelatedItemsWithoutMediaGallery = function(item: IArticle, mediaFields) {
+    this.getRelatedItemsWithoutMediaGallery = function(item: IArticle, mediaFields: Dictionary<string, Array<number>>) {
         if (!item.associations) {
             return [];
         }
--- a/scripts/apps/relations/services/RelationsService.ts
+++ b/scripts/apps/relations/services/RelationsService.ts
@@ -3,7 +3,7 @@ import {IArticle} from 'superdesk-interfaces/Article';
 RelationsService.$inject = ['archiveService', 'mediaIdGenerator'];
 
 export function RelationsService(archiveService, mediaIdGenerator) {
-    this.getRelatedItems = function(item: IArticle, mediaFields: IArticle) {
+    this.getRelatedItems = function(item: IArticle, mediaFields: object) {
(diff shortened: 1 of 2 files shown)


scripts/apps/relations/services/RelationsService.ts
```diff
@@ -3,12 +3,15 @@ import {IArticle} from 'superdesk-interfaces/Article';
...
 
 export function RelationsService(archiveService) {
-    this.getRelatedItems = function(item: IArticle) {
+    this.getRelatedItems = function(item: IArticle, mediaFields: IArticle) {
         if (!item.associations) {
             return [];
         }
 
-        const related = Object.values(item.associations);
+        const relatedWithoutMedia = _.pickBy(item.associations, (value, key) =>
```
ioanpocol: If an image is in media gallery it should not be validated but if it is in related items it should be validated.
I think here the check is only by item type and not by field type (media gallery and related items)

GyanP: fixed the issue

scripts/apps/relations/services/RelationsService.ts
```diff
@@ -3,12 +3,15 @@ import {IArticle} from 'superdesk-interfaces/Article';
 RelationsService.$inject = ['archiveService'];
 
 export function RelationsService(archiveService) {
-    this.getRelatedItems = function(item: IArticle) {
+    this.getRelatedItems = function(item: IArticle, mediaFields: IArticle) {
```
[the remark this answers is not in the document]
GyanP: I have fixed that.

scripts/apps/relations/services/RelationsService.ts
```diff
@@ -1,14 +1,20 @@
 import {IArticle} from 'superdesk-interfaces/Article';
 
-RelationsService.$inject = ['archiveService'];
+RelationsService.$inject = ['archiveService', 'mediaIdGenerator'];
 
-export function RelationsService(archiveService) {
-    this.getRelatedItems = function(item: IArticle) {
+export function RelationsService(archiveService, mediaIdGenerator) {
+    this.getRelatedItems = function(item: IArticle, mediaFields: object) {
```
tomaskikutis: Type annotations are not mandatory, and when unsure it's best to not use them at all. Adding type `object` isn't much more helpful than not adding a type annotation at all since everything in JavaScript is an object - dictionary, array, set and map are all objects. The correct annotation would be `Array<number>`

tomaskikutis: The naming no longer matches the code in my opinion. If the method is called `getRelatedItems`, it should get all of them, including media galleries. If you want to exclude media galleries, rename/create another method like `getRelatedItemsWithoutGalleries` or similar.

PR #2723, opened 2018-12-17
update ui framework

petrjasek:
SDESK-2780

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "sass-loader": "6.0.6",
     "shortid": "2.2.8",
     "style-loader": "0.20.2",
-    "superdesk-ui-framework": "1.5.6",
+    "superdesk-ui-framework": "^1.7.0",
     "ts-loader": "3.5.0",
     "tslint": "5.11.0",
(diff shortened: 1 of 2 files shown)


tomaskikutis:
You won't be able to update the UI framework without commenting out [this line](https://github.com/superdesk/superdesk-ui-framework/blob/master/app/scripts/helpers/react-to-angular-1.jsx#L9) or fixing the test. The line is preventing angular from swallowing the errors that occur in React. It's causing one test to fail which I wasn't able to figure out yet.

petrjasek:
not sure it's about some test, it fails during some bootstrap:
```
[12:07:51] E/direct - Error code: 135
[12:07:51] E/direct - Error message: Could not find update-config.json. Run 'webdriver-manager update' to download binaries.
[12:07:51] E/direct - Error: Could not find update-config.json. Run 'webdriver-manager update' to download binaries
```

tomaskikutis:
The test I'm referring to is not related to currently failing ones. It was a unit test actally and I see they all pass. It might not be a problem after all.

PR #2733, opened 2018-12-20
Subject matter expert review v2 (SDFID-513)

tomaskikutis:
SDFID-513

REQUIRED BY: https://github.com/superdesk/superdesk-fi/pull/213


scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -682,7 +682,9 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
              */
             $scope.close = function() {
                 _closing = true;
-                authoring.close($scope.item, $scope.origItem, $scope.save_enabled()).then(() => {
+
+                // returned promise used by superdesk-fi
```
pablopunk: We could make a simple test just to check if this is returning a promise (or just checking it works properly) instead of writing a comment

scripts/apps/contacts/helpers.js
```diff
@@ -44,7 +44,7 @@ export function renderArea(area, itemProps, props) {
             return fields[field](itemProps);
         }
 
-        console.warn('missing field in list: ' + field);
+        // console.warn('missing field in list: ' + field);
```
pablopunk: I also hate this warning because I get it every day 😁, should we just remove it?

scripts/apps/search/helpers.js
```diff
@@ -147,7 +147,7 @@ export function renderArea(area, itemProps, props, customRender = {}) {
             return fields[field](itemProps);
         }
 
-        console.warn('missing field in list: ' + field);
+        // console.warn('missing field in list: ' + field);
```
pablopunk: :rotating_light:

tomaskikutis:
subject-matter-expert-review

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
tomaskikutis: I'm adding a comment, because of external usage. Someone would see that the code is not used in superdesk-client-core and might remove it. I'd have to add the same comment in the test case, so I might as well keep it here.

styles/sass/variables/_dimensions.scss
```diff
@@ -32,15 +32,11 @@ $sidebar-width: 48px;
...
-// Scratchpad
-$scratchpad-closed: 0px;
-$scratchpad-content-height: 0px;
-
 // Article edit screen
 $authoring-opened-articles: 30px;
 
 //Layout
-$bottom-view-padding: $scratchpad-closed + $authoring-opened-articles + 20px;
+$bottom-view-padding: 0px + $authoring-opened-articles + 20px;
```
petrjasek: adding 0 is doesn't do much here

PR #2737, opened 2018-12-22
fix(preview): Add missing fields on info->metadata(SDESK-3412)

ioanpocol:


scripts/apps/authoring/metadata/views/metadata-widget.html
```diff
@@ -66,6 +66,12 @@
                     {{item.copyrightholder}}
                 </div>
             </li>
+            <li class="basic-list__item" ng-if="item.copyrightnotice">
```
petrjasek: imo there is too much boilerplate, we should introduce some directive like:
```
<li sd-list-item-basic data-label="{{ 'Copyright notice' | translate }}">{{ item.copyrightnotice }}</li>
```

tomaskikutis: How about having an Array<string> of fields and using a repeater to display it. The directive is not necessary this way.

In the future, it would be best in my opinion if we didn't have any field names in the code at all and kept it in a database alongside with required configs instead. When you sprinkle them all over the code it's getting harder to see all exceptional behavior a field can have.

ioanpocol:
rebased

scripts/apps/authoring/metadata/views/metadata-widget.html
ioanpocol: Some of the metadata are editable, some of them contains some logic for 'ng-if' & shown value and the rest of them respect the above template. Unfortunately the order of the metadata items is mixed so it is not possible for example to show the list of all metadata items that respect the above template.

tomaskikutis: > Some of the metadata are editable, some of them contains some logic for 'ng-if' & shown value

Yeah, that's too much for this PR, but regarding the future of fields, configs do solve conditional logic and order of items.

scripts/apps/authoring/metadata/views/metadata-list-item.html
```diff
@@ -0,0 +1,6 @@
+<li class="basic-list__item">
+    <label class="basic-list__item-label">{{label}}</label>
+    <div class="basic-list__item-data">
+        {{value}}
```
petrjasek: imo it would make sense to use transclude here so you could also use it for things like:
```
<li sd-metadata-list-item data-label="Updated">
    <sd-reldate-complex datetime="item.versioncreated"></sd-reldate-complex>
</li>
```

ioanpocol:
rebased

PR #2743, opened 2018-12-31
fixed: 'failed to generate image crops' message is displayed when add…

GyanP:
…ing an image to media gallery (SDESK-3780)
https://dev.sourcefabric.org/browse/SDESK-3780

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
@@ -70,7 +70,7 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
     $scope.metadata = {
         isDirty: false
     };
-    
+
     $scope.crops = {
         isDirty: false
     };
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -101,8 +101,16 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
         }
     };
 
+    $scope.handleNoImageCrops = function(_origCropsData) {
```
tomaskikutis: There is not much use of having the function if you are only using it once and it depends on the closure values. Can you inline it?

GyanP: You mean inline it, somewhere in the template?

tomaskikutis: By inline, I mean to replace the function call with the contents of the function and remove the function then.

GyanP: fixed that.

PR #2749, opened 2019-01-04
fix(translate): Replace gettextCatalog.getString calls to calls to ge…

ioanpocol:
…ttext from core/ui/components/utils (SDESK-3556)

PR #2751, opened 2019-01-04
fix(users): update user etag after password change

petrjasek:
SDESK-3779

requires https://github.com/superdesk/superdesk-core/pull/1469

### DIFF
--- a/scripts/apps/users/directives/ChangePasswordDirective.ts
+++ b/scripts/apps/users/directives/ChangePasswordDirective.ts
@@ -18,6 +18,7 @@ export function ChangePasswordDirective(usersService, notify, gettext) {
                         scope.oldPasswordInvalid = false;
                         notify.success(gettext('The password has been changed.'), 3000);
                         scope.show.change_password = false;
+                        scope.user._etag = response._etag;
                     }, (response) => {
                         scope.oldPasswordInvalid = true;
                     });


ioanpocol:
The fix to e2e test is here: https://github.com/superdesk/superdesk-client-core/pull/2749/commits/22e2658eed28d05f4fbca1be68baa6fcae9eeb57

scripts/apps/users/directives/ChangePasswordDirective.ts
```diff
@@ -18,6 +18,7 @@ export function ChangePasswordDirective(usersService, notify, gettext) {
                         scope.oldPasswordInvalid = false;
                         notify.success(gettext('The password has been changed.'), 3000);
                         scope.show.change_password = false;
+                        scope.user._etag = response._etag;
```
tomaskikutis: How about updating the entire user like `scope.user = reponse`? There is no reason to mention etags explicitly since they are like a background service and not specific to any view/controller.

We should get them out of all views IMO. If we simpy overwritten the entire object received from the server etags and all other attributes we might wish to update on the server would work out of the box.

petrjasek: it's not using an api returning user object, it's specific api to change password, and only thing in response that is useful here is the etag

tomaskikutis: Is it possible to make the endpoint return the object being modified?

petrjasek: it is, but this wouldn't work in the form because that would override any local changes

tomaskikutis: Requiring to save/discard changes before being able to change the password solve this.

tomaskikutis: You do get more code and worse UX, but that's a tradeoff for better code. You can keep it as is if you see fit. We can address all etags at once in the future.

PR #2752, opened 2019-01-04
fixed: cursor is placed at the end of the string when continuing typi…

GyanP:
…ng in qcode value
https://dev.sourcefabric.org/browse/SDESK-3771


tomaskikutis:
Why was the cursor appearing at the wrong place? You've added the code to change the cursor position again, which may work, but it would be better to prevent it from appearing the wrong position in the first place.

GyanP:
On change, the value gets updated and the component gets rerendered and since the value get changed it's the expected behavior of the input element to position the cursor at the end.

PR #2755, opened 2019-01-07
Implemented a filter to see ingest sources by 'All/Active/Closed' sta…

GyanP:
…tus(SDNTB-543)
https://dev.sourcefabric.org/browse/SDNTB-543

### DIFF
--- a/scripts/apps/ingest/directives/IngestSourcesContent.js
+++ b/scripts/apps/ingest/directives/IngestSourcesContent.js
@@ -79,6 +79,12 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                     13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24];
                 $scope.ingestExpiry = deployConfig.getSync('ingest_expiry_minutes');
 
+                $scope.selectedStatusFilter = 'Active';
+                $scope.filterIngestSources = function(option) {
+                    $scope.selectedStatusFilter = option;
+                    fetchProviders();
+                };
+
                 $scope.search = function(query) {
                     if (query) {
                         $scope.searchPage = 1;
@@ -100,6 +106,14 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                             {name: {
                                 $regex: $scope.searchQuery,
                                 $options: '-i'},
+                            },
+                        ]};
+                    } else if ($scope.selectedStatusFilter !== 'All') {
+                        const status = $scope.selectedStatusFilter !== 'Active';
+
+                        orTerms = {$or: [
+                            {
+                                is_closed: status,
                             },
                         ]};
                     }
--- a/scripts/apps/ingest/directives/IngestSourcesContent.js
+++ b/scripts/apps/ingest/directives/IngestSourcesContent.js
@@ -79,9 +79,15 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                     13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24];
                 $scope.ingestExpiry = deployConfig.getSync('ingest_expiry_minutes');
 
-                $scope.selectedStatusFilter = 'Active';
-                $scope.filterIngestSources = function(option) {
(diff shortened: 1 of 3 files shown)


scripts/apps/ingest/views/settings/ingest-sources-content.html
```diff
@@ -6,6 +6,17 @@
                 <input id="search-input" type="text" placeholder="{{ :: 'Search' | translate }}" ng-model="query" ng-model-options="{debounce: 500}" ng-change="search(query)">
             </div>
         </div>
+        <div class="dropdown" dropdown>
+            <button class="dropdown__toggle navbtn navbtn--text-only" dropdown__toggle translate>
+                Status: {{selectedStatusFilter}}
+                <span class="dropdown__caret"></span>
+            </button>
+            <ul class="dropdown__menu">
+                <li ng-repeat="option in ['Active', 'All', 'Closed'] track by $index" >
```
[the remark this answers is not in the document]
GyanP: fixed that.

GyanP:
> Yes. It works for me. I mean the text is in English because no translations exist yet, but the parameter is getting replaced.
> 
> peek 2019-01-09 13-47

Yes, parameter getting replaced but I tried to translate it using ntb-master

GyanP:
> Yes. It works for me. I mean the text is in English because no translations exist yet, but the parameter is getting replaced.
> 
> peek 2019-01-09 13-47

Sorry, clicked on close and comment by mistake.
Yes, params getting replaced but not getting translated 
translate
But if I write like this `<span translate translate-params-filter="activeStatusFilter | translate">Status: {{ filter }}</span>` it works fine. But it's of no use I guess.

scripts/apps/ingest/directives/IngestSourcesContent.js
```diff
@@ -79,6 +79,18 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
...
 
+                $scope.statusFilters = [
+                    {label: gettext('Active'), id: 'active'},
+                    {label: gettext('All'), id: 'all'},
+                    {label: gettext('Closed'), id: 'closed'},
+                ];
+
+                $scope.activeStatusFilter = $scope.statusFilters[0];
+                $scope.filterIngestSources = function(id) {
+                    $scope.activeStatusFilter = _.find($scope.statusFilters, (item) => item.id === id);
```
tomaskikutis: Use [`Array.prototype.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead of lodash

scripts/apps/ingest/views/settings/ingest-sources-content.html
```diff
@@ -6,6 +6,18 @@
...
         </div>
+        <div class="dropdown" dropdown>
+            <button id="status_filter_button" class="dropdown__toggle navbtn navbtn--text-only" dropdown__toggle>
+                <span translate translate-params-filter="activeStatusFilter.label">Status: {{filter}}</span>
+                <span class="dropdown__caret"></span>
+            </button>
+            <ul class="dropdown__menu">
+                <li ng-repeat="filter in statusFilters track by filter.id" >
+                    <!-- dynamic id only for testing purpose --->
+                    <button ng-click="filterIngestSources(filter.id)" id="status--{{::filter.label}}-filter">{{filter.label | translate}}</button>
```
tomaskikutis: Use filter ID here instead of a label `id="status--{{::filter.label}}-filter"`

scripts/apps/ingest/directives/IngestSourcesContent.js
```diff
@@ -102,6 +114,12 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                                 $options: '-i'},
                             },
                         ]};
+                    } else if ($scope.activeStatusFilter.id !== 'all') {
```
tomaskikutis: It should be possible to filter ingest sources by name and status at the same time. It should also work well with pagination.

GyanP: If we want to filter sources by name then by default status should be "active" or "all"?

PR #2759, opened 2019-01-10
As a user i want to add urls to a story as extra fields (SDBELGA 22)

tomaskikutis:
SDBELGA-22


ioanpocol:
I think the custom url fields should be added on preview pages too

PR #2761, opened 2019-01-11
fixed: Warning for 'unpublished related items' thrown for scanpix med…

GyanP:
…ia on publishing
https://dev.sourcefabric.org/browse/SDNTB-546

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.js
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.js
@@ -86,6 +86,7 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
         link: function($scope, elem, attrs) {
             var _closing;
             var mediaFields = {};
+            var relatedItemFields = []; // To store the _id of related_content
 
             const UNIQUE_NAME_ERROR = gettext('Error: Unique Name is not unique.');
             const MEDIA_TYPES = ['video', 'picture', 'audio'];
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -1026,6 +1027,9 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
                 var parts = mediaIdGenerator.getFieldParts(fieldId);
                 var field = _.find($scope.fields, (field) => field._id === parts[0]);
 
+                if (field && field.field_type === 'related_content' && relatedItemFields.indexOf(field._id) === -1) {
```
tomaskikutis: The code you added is not related to checking if the field `isMediaField` which is what this function is for. Another reason this code should not be here is because it creates unnecessary coupling. For your code in another file(`RelationsService.ts`) to work properly, `isMediaField` has to be called exactly as it is now.

Undo changes to this file and fix it in `RelationsService.ts` instead.

GyanP: fixed it in `RelationsService.ts`

scripts/apps/relations/services/RelationsService.ts
```diff
@@ -3,15 +3,15 @@ import {IArticle} from 'superdesk-interfaces/Article';
...
 export function RelationsService(archiveService, mediaIdGenerator) {
-    this.getRelatedItemsWithoutMediaGallery = function(item: IArticle, mediaFields: Dictionary<string, Array<number>>) {
+    this.getRelatedItemsWithoutMediaGallery = function(item: IArticle, fields) {
         if (!item.associations) {
             return [];
         }
-
         const relatedWithoutMedia = _.pickBy(item.associations, (value, key) => {
             var parts = mediaIdGenerator.getFieldParts(key);
+            var field = _.find(fields, (field) => field._id === parts[0]);
```
tomaskikutis: Do not use lodash when native solutions are available. Use `Array.prototype.find`.

PR #2769, opened 2019-01-15
feat(target): Moved target entity to Publish pane

pavlovicnemanja:


mdhaman:
@petrjasek this pr should be merged with #2737 in 1.28 as the #2737 removes the target entity.

PR #2774, opened 2019-01-17
(SDESK-3793) fix on creating a vocabulary having the same ID as previ…

GyanP:
…ously deleted one. No errors displayed to the user.
required by https://github.com/superdesk/superdesk-core/pull/1481

### DIFF
--- a/scripts/apps/vocabularies/views/vocabulary-config-modal.html
+++ b/scripts/apps/vocabularies/views/vocabulary-config-modal.html
@@ -18,7 +18,7 @@
                                       ng-readonly="!!vocabulary._links"
                                       required>
                     <div class="sd-line-input__message"
-                         ng-show="issues._id.unique" translate>ID must be unique.</div>
+                         ng-show="issues._id.unique || issues._id.deleted" translate>ID must be unique.</div>
                     <div class="sd-line-input__message"
                          ng-show="issues._id.conflict" translate>ID conflicts with system field.</div>
                     <div class="sd-line-input__message"


scripts/apps/vocabularies/views/vocabulary-config-modal.html
```diff
@@ -18,7 +18,7 @@ <h3 class="modal__heading" ng-show="!vocabulary._id"><span translate>Create</spa
                                       ng-readonly="!!vocabulary._links"
                                       required>
                     <div class="sd-line-input__message"
-                         ng-show="issues._id.unique" translate>ID must be unique.</div>
+                         ng-show="issues._id.unique || issues._id.deleted" translate>ID must be unique.</div>
```
tomaskikutis: Use error messages returned by API. `<div ng-repeat="issue in issues track by $index" class="sd-line-input__message">{{issue.message}}</div>`

tomaskikutis: Do the same for a few fields below the one you added.

GyanP: Actually, the message in the API response is of the form `{"unique":1}` or `{"deleted":1}` or `{"conflict":1}`. So I think it won't be a good idea to display them as it is.

tomaskikutis: Translation is not supported on the back-end yet, so we don't have a choice in this case. If it was, I would definitely want error messages returned from the server.

I've created `SDESK-3834` to enable translations on the server.

tomaskikutis:
By the way, why can't the vocabulary be created with the same id which was removed in the past? It's no longer there or is it?

petrjasek:
it is still on backend, we use soft delete, but no idea why

tomaskikutis:
Found some docs on the issue page of `SDESK-2057` which introduced the change.

> IMPORTANT: If technically it is not convenient to delete a content field because it's been previously used and we want to keep data consistency, it would be Ok to follow a similar approach than for users. It means, the remove option wouldn't really delete the fields but rather mark them as removed. The field would remain in the database but it wouldn't be available for usage any longer.

PR #2776, opened 2019-01-17
fix(authoring): fix duplicate authoring header cvs

petrjasek:
add function to get all custom header cvs for profile editor

SDNTB-550

reverts https://github.com/superdesk/superdesk-client-core/pull/2766

### DIFF
--- a/scripts/apps/authoring/metadata/metadata.js
+++ b/scripts/apps/authoring/metadata/metadata.js
@@ -1222,15 +1222,35 @@ function MetadataService(api, subscribersService, config, vocabularies, $rootSco
                     return true;
                 } else {
                     cv.terms = cv.items;
-                    return qcodes.length === 0 || qcodes.some((qcode) => !!cvService[qcode]);
+                    return qcodes.some((qcode) => !!cvService[qcode]);
                 }
             }));
         },
+        /**
+         * Get custom vocabularies for selected services, used in authoring
+         *
+         * @param {Array} qcodes - selected service qcodes
+         * @param {Object} editor
+         * @param {Object} schema
+         * @return {Array}
+         */
         getCustomVocabulariesForArticleHeader: function(qcodes, editor, schema) {
             return this.getFilteredCustomVocabularies(qcodes)
                 .then(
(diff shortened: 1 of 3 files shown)


scripts/apps/authoring/metadata/metadata.js
```diff
@@ -1199,16 +1199,21 @@ function MetadataService(api, subscribersService, config, vocabularies, $rootSco
...
             }));
         },
         getCustomVocabulariesForArticleHeader: function(qcodes, editor, schema) {
             return this.getFilteredCustomVocabularies(qcodes)
                 .then(
-                    (cvs) => cvs.filter((cv) => cv.items && cv.items.length && (editor[cv._id] || schema[cv._id]))
+                    (cvs) => cvs.filter((cv) => cv.terms.length && (editor[cv._id] || schema[cv._id]))
                 );
         },
+        getAllCustomVocabulariesForArticleHeader: function(editor, schema) {
```
tomaskikutis: Could you add a comment on how `getAllCustomVocabulariesForArticleHeader` and `getCustomVocabulariesForArticleHeader` differ and where are they used?

petrjasek:
@tomaskikutis this is it

scripts/apps/authoring/metadata/metadata.js
tomaskikutis: What was the problem for ntb, why weren't the content profiles working?

petrjasek: that's not actually fixing those content profiles not working, it's only fixing issue with some cvs displayed in content tab

tomaskikutis: What was then the issue with cvs in the content tab?

tomaskikutis:
I tried to run this PR with ntb master API and the situation with content profiles seems to be the same - I can't add anything and see an error from the API.

500 api/content_types/59105a86177e5b00e602fe2c?edit=true

petrjasek:
that's right, I didn't push a fix for that yet, it's a problem with init data

PR #2777, opened 2019-01-17
fixed: Changes in ingest channel article types are not activating the…

GyanP:
… save button (SDESK-3825)

### DIFF
--- a/scripts/apps/ingest/directives/IngestSourcesContent.js
+++ b/scripts/apps/ingest/directives/IngestSourcesContent.js
@@ -505,6 +505,9 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                  * @param {string} fileType
                  */
                 $scope.addOrRemoveFileType = function(fileType) {
+                    let modalBodyContainer = angular.element('.modal__body');
+
+                    modalBodyContainer.scope().editForm.$setDirty();
                     if (!$scope.provider.content_types) {
                         $scope.provider.content_types = [];
                     }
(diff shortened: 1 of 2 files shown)


scripts/apps/ingest/views/settings/ingest-sources-content.html
```diff
@@ -109,7 +109,7 @@ <h3 class="modal__heading" translate ng-hide="provider._id" translate>Add New So
                             <a  href=""
                                 ng-repeat="fileType in fileTypes"
                                 title="{{fileType.type}}"
-                                ng-click="addOrRemoveFileType(fileType.type)"
+                                ng-click="addOrRemoveFileType(fileType.type); editForm.$setDirty()"
```
tomaskikutis: Set it dirty in the function itself instead of where it's called so if it's called in another place, you don't need to remember to set dirty as well.

GyanP: fixed it.

scripts/apps/ingest/directives/IngestSourcesContent.js
```diff
@@ -505,6 +505,9 @@ export function IngestSourcesContent(ingestSources, gettext, notify, api, $locat
                  * @param {string} fileType
                  */
                 $scope.addOrRemoveFileType = function(fileType) {
+                    let modalBodyContainer = angular.element('.modal__body');
```
tomaskikutis: Don't query the DOM. Pass the form as a parameter instead.

GyanP: done

PR #2778, opened 2019-01-17
fix(authoring): fix media not saving without editing poi

petrjasek:
when poi is not defined, set it to center of the picture
so when user clicks on done it will preserve it.

SDESK-3780

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/AssociationController.ts
+++ b/scripts/apps/authoring/authoring/controllers/AssociationController.ts
@@ -210,8 +210,6 @@ export function AssociationController(config, send, api, $q, superdesk,
             return renditions.crop(item, cropOptions)
                 .then((rendition) => {
                     self.updateItemAssociation(scope, rendition, options.customRel, callback);
-                }, () => {
-                    notify.error(gettext('Failed to generate picture crops.'));
                 })
                 .finally(() => {
                     scope.loading = false;
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
@@ -177,7 +177,6 @@ export function ChangeImageController($scope, gettext, notify, _, api, $rootScop
 
         $scope.crops.isDirty = false;
         $scope.data.isDirty = true;
-        $scope.portraitPOIWarning = false;
     };
 
     /**
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
@@ -28,9 +28,9 @@ export function validateMediaFieldsThrows(validator, metadata) {
     }
 }
 
-ChangeImageController.$inject = ['$scope', 'gettext', 'notify', 'modal', 'lodash', 'api', '$rootScope',
+ChangeImageController.$inject = ['$scope', 'gettext', 'notify', 'lodash', 'api', '$rootScope',
     'deployConfig', '$q', 'config'];
-export function ChangeImageController($scope, gettext, notify, modal, _, api, $rootScope, deployConfig, $q, config) {
+export function ChangeImageController($scope, gettext, notify, _, api, $rootScope, deployConfig, $q, config) {
     $scope.data = $scope.locals.data;
     $scope.data.cropData = {};
     $scope.validator = deployConfig.getSync('validator_media_metadata');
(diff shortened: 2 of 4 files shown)


scripts/apps/authoring/authoring/controllers/AssociationController.ts
```diff
@@ -210,8 +210,6 @@ export function AssociationController(config, send, api, $q, superdesk,
             return renditions.crop(item, cropOptions)
                 .then((rendition) => {
                     self.updateItemAssociation(scope, rendition, options.customRel, callback);
-                }, () => {
```
MarkLark86: `renditions.crop` performs an `api.save('picture_crop'...)` on the item. If the api returns an error, should this error be displayed to the user (interpreting the api error response)?

scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -489,4 +490,10 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
             }
         });
     };
+
+    // init poi if not set
+    if (!$scope.data.poi) {
+        $scope.data.poi = {x: 0.5, y: 0.5};
```
mdhaman: @petrjasek @karelpetrak setting poi to 0.5 x 0.5 and saving crops is a potential for users making mistake. If we default the poi to 0.5 for portrait images then the default crops generated will be wrong.  

As per the current media editor change, the user could directly go to metadata tab make changes click "Save" and "Done" button. If we don't enforce users to select the poi then they are bound to make mistakes. This would be fine if system does not need crops.

scripts/apps/authoring/authoring/controllers/AssociationController.ts
petrjasek: yep, will add the notification to the api call

scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
petrjasek: what about setting it only for landspace pictures then and display some warning for portraits that it must be picked?

tomaskikutis: There will still be mistakes with landscape pictures if it has a face on the top/side for [example](https://ichef.bbci.co.uk/images/ic/1200x675/p06rs7fd.jpg).

I agree with @mdhaman - if the client doesn't need crops, let's not crop at all and if he does, let's put validation somewhere. It can be on upload, on publishing or in some other point, but publishing without validation at all doesn't sound good.

petrjasek: ok will reuse that validate crop config then and only preserve the default when it's off

petrjasek: makes sense?

tomaskikutis: If I understand correctly you want to push default crops for instances where `validatePointOfInterestForImages` is false and keep everything else the same?

```
// init poi if not set
    if (!$scope.data.poi) {
```

Shouldn't this if also have a check on `validatePointOfInterestForImages`? We shouldn't set defaults where `validatePointOfInterestForImages === true`.

When I look back at `validatePointOfInterestForImages` config, it seems that it shouldn't even exist and instead it may be better to have a config called `enableImageCropping` which would ensure that POI is selected and if not set, we wouldn't set POI at all. I don't think we should use default POI.

scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -173,9 +173,11 @@ export function ChangeImageController($scope, gettext, notify, modal, _, api, $r
         // update crop and poi data in `item`
         angular.extend($scope.data.item, $scope.data.metadata);
         $scope.data.item.poi = $scope.data.poi;
+        $scope.data.metadata.poi = $scope.data.poi;
 
         $scope.crops.isDirty = false;
         $scope.data.isDirty = true;
+        $scope.portraitPOIWarning = false;
```
tomaskikutis: Is this no longer used?

scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
petrjasek: that `data.poi` must be set always otherwise poi editor doesn't work, that was always there but before it was initialized outside the controller using it

PR #2779, opened 2019-01-17
Custom script tags on index.html

pablopunk:
SDESK-3781

We have a custom script on `index.html` only if `qumu` feature is enabled. As this is specific to fidelity, I created `features.scriptTags` field to inject custom scripts on superdesk.

In repos like `superdesk-fi` we can put:

```js
features: {
   ...,
   scriptTags: [ 'https://video.fidelity.tv/widgets/1/application.js' ]
}
```

Thoughts on this? @hlmnrmr @petrjasek  @tomaskikutis

### DIFF
--- a/index.html.js
+++ b/index.html.js
@@ -1,7 +1,4 @@
-module.exports = function buildIndex(features) {
-    const scriptTags = features.scriptTags || [];
-
-    return `
+const buildIndex = ({scriptTags}) => `
         <!doctype html>
         <html class="no-js">
           <head>
@@ -20,4 +17,5 @@ module.exports = function buildIndex(features) {
           </body>
         </html>
     `;
-};
+
+module.exports = buildIndex;
--- a/index.html.js
+++ b/index.html.js
@@ -1,4 +1,6 @@
-module.exports = function buildIndex(o) {
+module.exports = function buildIndex(features) {
+    const scriptTags = features.scriptTags || [];
+
     return `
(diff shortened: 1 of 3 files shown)


petrjasek:
makes sense, but wouldn't add it into features, more like have it on its own, features are more like feature flags to set something on/off

pablopunk:
@petrjasek that makes sense, I'll update the PR. BTW why do we also have an index.html (no .js) ?

petrjasek:
that's the one which is generated, it's not on github

tomaskikutis:
Sorry for the late notice, but I just realized that the scope of the task is a bit off. This script isn't a dependence of Fidelity, but of a qumu video plugin which is a thing Fidelity uses.

pablopunk:
@tomaskikutis then it should still need to be on fidelity's repo right?

tomaskikutis:
It doesn't help much extracting only loading of the lib, but not all components related to qumu which are still in the core. I'd close this for now and come back to it later when we have superdesk extension API and can extract the entire qumu properly together with this script.

tomaskikutis:
Sorry for two comments, I wanted to add a space and clicked ctrl+enter which submitted a comment. I think it would go to some kind of superdesk-plugins repo, and it would be initialized and configured from fidelity's repo

pablopunk:
I'll close this PR and the ticket for now then

PR #2784, opened 2019-01-21
As a user i want to manage the content of the rhs toolbar in a content profile (SDESK 3832)

tomaskikutis:
SDESK-3832

DEPENDS ON: https://github.com/superdesk/superdesk-core/pull/1488


test-server/requirements.txt
```diff
@@ -1,3 +1,3 @@
 gunicorn==19.7.1
 honcho==1.0.1
-git+git://github.com/superdesk/superdesk-core.git@master#egg=Superdesk-Core
+git+git://github.com/tomaskikutis/superdesk-core.git@bd1978b#egg=Superdesk-Core
```
petrjasek: I think you can revert this now

scripts/apps/workspace/content/views/profile-settings.html
```diff
@@ -90,7 +90,7 @@ <h3 class="modal__heading">{{ :: 'Editing' | translate }} "{{ editing.form.label
     </div>
     <div class="modal__body">
         <form name="editForm">
-            <fieldset>
+            <fieldset ng-init="setNgForm(editForm)">
```
petrjasek: not sure this is needed, the form should be available on scope as `editForm`

tomaskikutis: Accessing it from scope [isn't working](https://user-images.githubusercontent.com/2076953/51896550-044d4d00-23ad-11e9-8002-39c7450e452e.png)

scripts/apps/workspace/content/components/WidgetsConfig.tsx
```diff
@@ -0,0 +1,79 @@
...
+                        <li className="sd-list-item" key={i}>
+                            <span className="sd-list-item__column">
+                                <input
+                                    type="checkbox"
+                                    checked={isWidgetVisibleForContentProfile(this.state.widgetsConfig, widget._id)}
+                                    onChange={(e) => this.handleChange(widget._id, e.target.checked)}
+                                />
+                            </span>
+                            <span className="sd-list-item__column">
+                                {widget.label}
```
pablopunk: Does this label need to be translated?

scripts/apps/workspace/content/views/profile-settings.html
petrjasek: :ok:

scripts/apps/workspace/content/components/WidgetsConfig.tsx
tomaskikutis: It's translated in the widget definitions https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/comments/comments.ts#L136

We can't translate dynamic values. We attempt to do it currently, but that's a bad practice. When translating using dynamic values there is no string to extract for translation, so it only works if there is a proper prevous usage . `gettext` should only be called with inline strings.

pablopunk: @tomaskikutis got it

PR #2785, opened 2019-01-22
[SDESK-3837] If gettext is passed null, return empty string

MarkLark86:


mdhaman:
@petrjasek this needs to be in version v1.28.1

tomaskikutis:
@petrjasek Why do we need a tag "merged"? Doesn't the default picture by GitHub not work?
screenshot_6

petrjasek:
I use this when cherry-picking something for a bugfix release, so I know it's on 1.28 branch

PR #2787, opened 2019-01-22
fixed: Item not updated after adding second related item (SDESK-3816)

GyanP:


GyanP:
> Could you explain how is removing image renditions related with "related items"? You added the code which has no indication of it and would be very hard to maintain in the future.
> 
> I've also talked with Oleg who created this function about changing how it works(**SDESK-3854**), so renditions are removed when you remove the image and not on save as it is now. After the change, this PR might not be necessary, but I'm interested why is it now.

This is the error in the console when trying to add or remove a related item.
error

tomaskikutis:
Got it, so it doesn't have any "related items" specific logic. Could you rewrite the condition to `if updateRenditions == null || origRenditions == null) { return;}`. Omiting what are you comparing against and using the negation at the same time makes it harder to understand.

GyanP:
fixed

scripts/apps/authoring/authoring/helpers.js
```diff
@@ -231,7 +231,7 @@ export function cutoffPreviousRenditions(update, origItem) {
         try {
             updateRenditions = update.associations[key].renditions;
             origRenditions = origItem.associations[key].renditions;
-            if (!(updateRenditions && origRenditions)) {
+            if (!updateRenditions || !origRenditions) {
```
tomaskikutis: Can you use `if(updateRenditions == null || origRenditions == null)` instead of `if (!updateRenditions || !origRenditions)`?

GyanP: It's value is sometimes undefined and sometimes null.

tomaskikutis: Double equals covers for it.
screenshot

GyanP: done

PR #2790, opened 2019-01-22
As a user i want to quickly fetch content without having to open it (SDBELGA 27)

tomaskikutis:
SDBELGA-27


scripts/apps/search/components/ItemList.tsx
```diff
@@ -85,7 +85,7 @@ export class ItemList extends React.Component<any, any> {
         this.setState({narrow: setNarrow});
     }
 
-    select(item, event?) {
```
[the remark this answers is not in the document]
tomaskikutis: It was optional and I made it mandatory. The reason it was optional was that someone was not able to pass it and was working around that with a timeout which I removed. If I tried not to pass it when calling the function the ts compiler would complain. Where do you see the truthiness check?

scripts/core/activity/activity.ts
```diff
@@ -512,12 +504,13 @@ angular.module('superdesk.core.activity', [
     .run(['$rootScope', 'superdesk', function($rootScope, superdesk) {
         $rootScope.superdesk = superdesk; // add superdesk reference so we can use constants in templates
 
-        $rootScope.intent = function() {
-            return superdesk.intent(...arguments);
+        $rootScope.intent = function(...args) {
+            //
```
petrjasek: I expect something missing here

petrjasek: yep

PR #2792, opened 2019-01-23
run tslint on travis

petrjasek:


.travis.yml
```diff
@@ -15,4 +15,5 @@ cache:
 
 script:
     - npm run test
+    - npm run tslint
```
tomaskikutis: I think you should use `npm run lint` instead so eslint and other tools we might add run too.

petrjasek: eventually I would like to, but it's failing on external libs and not sure atm how to fix that

tomaskikutis: What does it say?

petrjasek: eg.:
```
node_modules/@types/react/index.d.ts:2321:19 - error TS2320: Interface 'IntrinsicClassAttributes<T>' cannot simultaneously extend types 'ClassAttributes<T>' and 'ClassAttributes<T>'.
  Named property 'ref' of types 'ClassAttributes<T>' and 'ClassAttributes<T>' are not identical.

2321         interface IntrinsicClassAttributes<T> extends React.ClassAttributes<T> { 
```e

tomaskikutis: Can't see what the source line of the error is. Does it originate from our code or something from node_modules?

petrjasek: `node_modules/@types/react/index.d.ts:2321:19`

tomaskikutis: That's strange, it looks like it's type checking the definitions themselves and not the code. Is there an easy way to get that error locally?

petrjasek: I just run `tsc --noEmit` in client core repo

PR #2793, opened 2019-01-23
fix(profiles): fix anpa category label

petrjasek:
- we use lang override to change this, but if there is no anpa
it doesn't work

SDNTB-550


tomaskikutis:
Could we instead keep it called category and change their lang override to override from "Category" to something different?

petrjasek:
they use category too as custom vocabulary, so I have to avoid that conflict.

PR #2794, opened 2019-01-24
When moving preview from right to left the top part is partly visible (SDESK 3849)

tomaskikutis:
SDESK-3849


tomaskikutis:
No, we don't. We just put whatever to fix the current situation and repeat when we find a contradiction :D

On a more positive side, a lot of the indexes were required to counter the fact that a lot of elements were being positioned absolutely including the toolbars which I refactored. They are no longer positioned absolutely and I may not need that high of an index. I'll check it out :)

PR #2797, opened 2019-01-25
[SDESK-3855] feat(content-profile): Fixed issue with ordering on content profiles

pavlovicnemanja:


tomaskikutis:
What was the issue with tabs?

karelpetrak:
> What was the issue with tabs?

Hi @tomaskikutis it's all described in the SDESK-3855 ticket.

tomaskikutis:
I meant to ask what went wrong in the code that caused the issue.

pavlovicnemanja:
@tomaskikutis Issue was that header and content items had same tabindex, so when you press tab you will get on first header item, after that on first content item, tab again, and now you have second header item and after that second content item. This code is adding identation for order in content items, number of header items + new order in content items.

PR #2798, opened 2019-01-25
(hotfix)When moving preview from right to left the top part is partly visible (sdesk 3849)

tomaskikutis:
SDESK-3858


MarkLark86:
@tomaskikutis This causes the date/time pickers to be hidden in the publish panel. Please revise the z-index values of `.main-section` and `zindexSlidePane`. I believe `9999` and `10001` are too high.

@fritzSF Any thoughts on the usage of z-index and any standards that should be adhered (such as what types of components should be within what z-index range)?

PR #2799, opened 2019-01-28
Enabling/Disabling spell-checker if no language dictionary (SDESK-3473)

GyanP:


scripts/core/spellcheck/spellcheck.js
```diff
@@ -525,9 +531,17 @@ function SpellcheckMenuController($rootScope, editorResolver, spellcheck) {
...
-        self.isAuto = status && !useTansaProofing();
-        render();
+    spellcheck.getDictionary($scope.item.language).then((dict) => {
+        self.isActiveDictionary = !!dict.length;
+
+        if (!self.isActiveDictionary) {
+            spellcheck.setSpellcheckerStatus(self.isActiveDictionary);
+        }
+
+        spellcheck.getSpellcheckerStatus().then((status) => {
```
petrjasek: is there a reason to call this part when there is no active dictionary?

GyanP: Not actually. I missed the return statement.

PR #2803, opened 2019-01-29
Hide login form if Single Sign on is enabled

pablopunk:
SDFID-357

ezgif com-video-to-gif

### DIFF
--- a/scripts/core/auth/login-modal-directive.ts
+++ b/scripts/core/auth/login-modal-directive.ts
@@ -19,6 +19,7 @@ angular.module('superdesk.core.auth.login', []).directive('sdLoginModal', [
             link: function(scope, element, attrs) {
                 scope.features = features;
                 scope.changePassword = false;
+                scope.showOnlySaml = false;
 
                 deployConfig.all({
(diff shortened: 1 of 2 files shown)


tomaskikutis:
I don't like how you mix standard login and SAML into the same page and put ifs where you need them. That's not sustainable in the long run where you have >4 flags and can no longer understand which go together and which conflict and how many possible states there can be. It will not be an issue right now, but that's how it starts. I would like you to extract it to be separate pages since there is almost zero overlap in functionality besides except that the pages link to each other.

pablopunk:
@tomaskikutis that's the plan for the future, we need to modify the client-core to allow the index template to be overriden and then create the custom view on the client's repo only

tomaskikutis:
Even if you can't override the index template right now, you should still be able to split it into 2 directives or even better React components. Isn't that right?

pablopunk:
That would be too much work that will be deleted in the near future. I'll do the following: I'm closing this PR and creating a ticket to allow client-core's index to be overriden (which is a small change actually) and take care of the new view in the client's repo

PR #2815, opened 2019-02-04
fixed: Cannot pick a colour for urgency and priority CVs

GyanP:
SDESK-3906

### DIFF
--- a/scripts/apps/vocabularies/components/ItemsTableComponent.tsx
+++ b/scripts/apps/vocabularies/components/ItemsTableComponent.tsx
@@ -52,7 +52,7 @@ export default class ItemsTableComponent extends React.Component<any, any> {
     }
 
     setCaretPosition(ctrl, pos) {
-        if (ctrl.setSelectionRange) {
+        if (ctrl.setSelectionRange && pos !== null) {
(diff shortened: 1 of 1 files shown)


scripts/apps/vocabularies/components/ItemsTableComponent.tsx
```diff
@@ -52,7 +52,7 @@ export default class ItemsTableComponent extends React.Component<any, any> {
     }
 
     setCaretPosition(ctrl, pos) {
-        if (ctrl.setSelectionRange) {
+        if (ctrl.setSelectionRange && pos !== null) {
```
tomaskikutis: Use `!=` nstead of `!==` when comparing with `null` to also catch undefined. What is pos and why was it `null`?

GyanP: `pos` is cursor position so it is null in case of color.

tomaskikutis: How is setting caret position for a table related with color picker? It looks to me that the function shouldn't be called in the first place.

PR #2816, opened 2019-02-05
[SDESK-3902] mailto href should not open blank tab.

mdhaman:


tomaskikutis:
I think it would be better to handle only `http:` and `https:` protocols without having to hardcode very exception:

```
function isLinkExternal(href) {
    try {
        const url = new URL(href);

        if (url.protocol !== 'http:' && url.protocol !== 'https:') {
            return false;
        }

        return url.host !== window.location.host;
    } catch (e) {
        // will throw if string is not a valid link
        return false;
    }
}
```

PR #2817, opened 2019-02-05
fix(notifications): Fix the translation for notifications(SDESK-3810)

ioanpocol:


scripts/core/menu/notifications/notifications.js
```diff
@@ -66,6 +68,7 @@ function UserNotificationsService(
 
                     item._unread = !isRead(recipients, identity._id, true);
                     this.unread += item._unread ? 1 : 0;
+                    item.message = gettext(item.message, item.data);
```
mdhaman: @ioanpocol you might need this change for the `ActivityMessageService` as well.  see https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/stream/services/index.ts#L10

PR #2818, opened 2019-02-06
feat(localisation): Add support for plural on translation(SDESK-3821)

ioanpocol:
Example:

import {gettextCatalog} from 'core/ui/components/utils';

console.log(
gettextCatalog.getPlural(2, "One {{size}} Bird", "{{$count}} {{size}}
Birds", {size: 'large'})
);

Console:
2 large Birds

console.log(
gettextCatalog.getPlural(1, "One {{size}} Bird", "{{$count}} {{size}}
Birds", {size: 'large'})
);

Console:
One large Bird

superdesk.po
msgid "One {{size}} Bird"
msgid_plural "{{$count}} {{size}} Birds"
msgstr[0] ""
msgstr[1] ""

### DIFF
--- a/scripts/core/ui/components/utils.ts
+++ b/scripts/core/ui/components/utils.ts
@@ -40,12 +40,14 @@ export const gettext = (text, params = null) => {
 
 /**
  * @ngdoc method
- * @name gettextCatalog
+ * @name gettextPlural
+ * @param {Number} count
  * @param {String} text
+ * @param {String} pluralText
  * @param {Object} params
- * @description Used angular gettext service for displaying localised text on Browser
+ * @description Used angular gettext service for displaying plural localised text on Browser
  */
-export const gettextCatalog = (text, params = null) => {
+const gettextPlural = (count, text, pluralText, params) => {
     if (!text) {
         return '';
     }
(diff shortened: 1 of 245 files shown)


scripts/core/ui/components/utils.ts
```diff
@@ -40,27 +40,32 @@ export const gettext = (text, params = null) => {
...
-
-        return params ? injector.get('$interpolate')(translated)(params) : translated;
+        return injector.get('gettextCatalog').getPlural(count, text, pluralText, params);
     }
 
     return text;
 };
 
+export const gettextCatalog = {
+    getString: gettext,
```
tomaskikutis: Why are you exporting `getString`? Aren't simple strings translated with the new helper you recently made in utils?

ioanpocol: The original gettextCatalog support both gettextCatalog.getPlural and gettextCatalog.getString for text extraction so I added it also on this proxy object.

ioanpocol:
>We should move `gettextPlural` out of the `ui/components` folder. It would also be good to move out `gettext` too.

I think both of them should be on the same file

scripts/core/ui/components/utils.ts
tomaskikutis: Can you remove it from here? `gettext` already does all we need regarding singular translations. It would only confuse people having two methods doing basically the same thing.

scripts/core/utils.ts
```diff
@@ -32,3 +32,54 @@ export const getSuperdeskType = (event, supportExternalFiles = true) =>
...
+/**
+ * @ngdoc method
+ * @name gettextPlural
+ * @param {Number} count
+ * @param {String} text
+ * @param {String} pluralText
+ * @param {Object} params
+ * @description Used angular gettext service for displaying plural localised text on Browser
+ */
+const gettextPlural = (count, text, pluralText, params) => {
```
tomaskikutis: `params` argument shouldn't be required

ioanpocol: OK, I can add default value:
const gettextPlural = (count, text, pluralText, params={}) => {

PR #2820, opened 2019-02-07
(SDESK-3910) Fixed: Errors when saving an image after cropping and no…

GyanP:
… further actions available.

Requires https://github.com/superdesk/superdesk-core/pull/1520

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
@@ -328,14 +328,19 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, deploy
                     media: result._id,
                 });
                 $scope.data.isDirty = true;
-                return api.save('picture_renditions', {item: result.item, no_custom_crops: true}).then((item) => {
-                    $scope.data.item.renditions = item.renditions;
-                    const editableMetadata = extractEditableMetadata($scope.data.metadata);
+                return api.save('picture_renditions', {
+                    item: result.item,
+                    no_custom_crops: true,
+                    is_article_edit_media: $scope.data.isArticleEditMedia != null
+                })
+                    .then((item) => {
+                        $scope.data.item.renditions = item.renditions;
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -328,14 +328,32 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, deploy
...
-                    $scope.data.metadata = Object.assign($scope.data.item, editableMetadata);
-                    $scope.data.poi = {x: 0.5, y: 0.5};
-                    $rootScope.$broadcast('poiUpdate', $scope.data.poi);
-                });
+                return api.save('picture_renditions', {
+                    item: result.item,
+                    no_custom_crops: true,
+                })
+                    .then((item) => {
+                        /* if the image is in editor body or as feature media
```
tomaskikutis: When you change the image in any way(metadata or renditions) while in the article edit view, it must never alter the original image. Why do you want to update the original?

GyanP: If you crop an image which is a feature media or in the editor body,  save it and then change the brightness you will see the original image is brought back.

tomaskikutis: What is the reason the previous image is brought back?

GyanP: And also if we rotate or change the brightness of feature media we apply the changes on the original image. Why we avoid it only in case of cropping the image?

tomaskikutis: > And also if we rotate or change the brightness of feature media we apply the changes on the original image

That needs to be fixed too then to not modify the original.

GyanP: Should I do it on the same ticket or create a new one?

tomaskikutis: You can do it on this one.

PR #2824, opened 2019-02-07
fix(upload): upload files sequentially

petrjasek:
when all uploads are started at once it sometimes fails.

SDANSA-223

### DIFF
--- a/scripts/apps/archive/controllers/UploadController.ts
+++ b/scripts/apps/archive/controllers/UploadController.ts
@@ -2,6 +2,7 @@ import EXIF from 'exif-js';
 import _ from 'lodash';
 import {getDataUrl} from 'core/upload/image-preview-directive';
 import {gettext} from 'core/utils';
+import {isEmpty} from 'lodash';
 
 /* eslint-disable complexity */
 
@@ -10,6 +11,15 @@ const getExifData = (file) => new Promise((resolve) => {
         resolve(this);
     });
 });
+
+function serializePromises(promiseCreators: Array<() => Promise<any>>): Promise<Array<any>> {
+    let promise = Promise.resolve();
+
+    return Promise.all(promiseCreators.map((promiseCreator) => {
+        promise = promise.then(promiseCreator);
+        return promise;
+    }));
+}
 
 UploadController.$inject = [
     '$scope',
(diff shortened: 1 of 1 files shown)


scripts/apps/archive/controllers/UploadController.ts
```diff
@@ -258,18 +259,23 @@ export function UploadController(
...
-        var promises = [];
+        if (isEmpty($scope.items)) {
+            return Promise.resolve();
+        }
+
+        let promise = Promise.resolve();
 
-        _.each($scope.items, (item) => {
+        // upload items in sequence, and resolve when all are done
+        return Promise.all($scope.items.map((item) => {
```
tomaskikutis: Why do you need `Promise.all` if you want it to be sequential. Wouldn't this work?

```
var p = Promise.resolve();

$scope.items.forEach((item) => {
    p.then(() => uploadFile(item));
});

return p;
```

petrjasek: it's just to return list of uploaded files at the end, and not just the last one

tomaskikutis: Could you instead implement and use `promiseAllSerial`? You can check some examples [here](https://decembersoft.com/posts/promises-in-serial-with-array-reduce/).

Wouldn't it be better to fix async upload in the first place? What's wrong with it?

petrjasek: there is no bug in our code, but if you upload many images at once sometimes it fails without any obvious reason, post request is rejected before reaching server

petrjasek: not sure why do that promiseAllSerial, it's doing basically the same thing, but imo less readable because they use reduce to make a list from a list, instead of using a map..

tomaskikutis: I'm not saying you should implement it as on that website. I'd only want the function to be a standalone function and have the same interface as `Promise.all` meaning it would take an array of promises and when resolved, would return an array with the results of the promises.

scripts/apps/archive/controllers/UploadController.ts
```diff
@@ -258,20 +259,30 @@ export function UploadController(
...
-        if (promises.length) {
-            return $q.all(promises);
-        }
-        return $q.when();
+
+            return () => Promise.resolve(item);
+        }));
     };
 
+    function serializePromises(callbacks) {
```
tomaskikutis: rename `callbacks` to `promiseCreators` or something else more exact. I'd also add typings:

```
function serializePromises<T>(promiseCreators: Array<() => Promise<T>>): Promise<Array<T>> {
    let promise = Promise.resolve(null);

    return Promise.all(promiseCreators.map((callback) => {
        promise = promise.then(callback);
        return promise;
    }));
}
```

PR #2829, opened 2019-02-08
Fixed: Opening an item by pasting the URL does not show all actions. …

GyanP:
…(SDESK-3919)

### DIFF
--- a/scripts/apps/desks/services/DesksFactory.js
+++ b/scripts/apps/desks/services/DesksFactory.js
@@ -37,7 +37,7 @@ export function DesksFactory($q, api, preferencesService, userList, notify,
                 let pg = page;
                 let extended = items.concat(result._items);
 
-                if (result._links.next) {
+                if (result._links && result._links.next) {
                     pg++;
                     return _fetchAll(endpoint, parent, pg, extended);
                 }
@@ -180,7 +180,7 @@ export function DesksFactory($q, api, preferencesService, userList, notify,
                 return $q.when(self.userDesks);
             }
 
-            return $q.all(this.fetchCurrentDeskId()) // make sure there will be current desk
+            return $q.when().then(this.fetchCurrentDeskId()) // make sure there will be current desk
                 .then(angular.bind(session, session.getIdentity))
                 .then(angular.bind(this, this.fetchUserDesks))
                 .then(angular.bind(this, function(desks) {
--- a/scripts/apps/desks/services/DesksFactory.js
+++ b/scripts/apps/desks/services/DesksFactory.js
@@ -180,7 +180,7 @@ export function DesksFactory($q, api, preferencesService, userList, notify,
                 return $q.when(self.userDesks);
             }
 
-            return this.fetchCurrentDeskId() // make sure there will be current desk
+            return $q.all(this.fetchCurrentDeskId()) // make sure there will be current desk
                 .then(angular.bind(session, session.getIdentity))
                 .then(angular.bind(this, this.fetchUserDesks))
                 .then(angular.bind(this, function(desks) {
--- a/scripts/apps/desks/services/DesksFactory.js
+++ b/scripts/apps/desks/services/DesksFactory.js
@@ -180,7 +180,7 @@ export function DesksFactory($q, api, preferencesService, userList, notify,
                 return $q.when(self.userDesks);
             }
 
-            return $q.when().then(this.fetchCurrentDeskId()) // make sure there will be current desk
+            return this.fetchCurrentDeskId() // make sure there will be current desk
                 .then(angular.bind(session, session.getIdentity))
                 .then(angular.bind(this, this.fetchUserDesks))
                 .then(angular.bind(this, function(desks) {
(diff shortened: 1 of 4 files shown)


scripts/apps/desks/services/DesksFactory.js
```diff
@@ -180,7 +180,7 @@ export function DesksFactory($q, api, preferencesService, userList, notify,
                 return $q.when(self.userDesks);
             }
 
-            return this.fetchCurrentDeskId() // make sure there will be current desk
+            return $q.all(this.fetchCurrentDeskId()) // make sure there will be current desk
```
petrjasek: I guess this should be `$q.when`

GyanP: yes, I too thought the same, but it does not solve the issue so I used `$q.all`.

scripts/apps/desks/services/DesksFactory.js
```diff
@@ -180,7 +180,7 @@ export function DesksFactory($q, api, preferencesService, userList, notify,
                 return $q.when(self.userDesks);
             }
 
-            return this.fetchCurrentDeskId() // make sure there will be current desk
+            return $q.when().then(this.fetchCurrentDeskId()) // make sure there will be current desk
```
tomaskikutis: Why is `$q.when().then` necessary? Doesn't `this.fetchCurrentDeskId()` already return a promise?

GyanP: Yep, it does return the promise. I used it to get the promise resolved immediately, as on pasting the url or refreshing the page `userDesks` were not available.

tomaskikutis: I don't get it. If the result of the promise is not available yet, you can't do anything, but ensure that the calling function can wait for the promise.

If you resolve it immediately aren't you resolving it with an empty value?

GyanP: Not resolving with empty values. I just saw it [here](https://github.com/superdesk/superdesk-client-core/blob/2691fb450e25cf7eda1e38b2663ed3fc55868f6b/scripts/apps/desks/services/DesksFactory.js#L130) and used it and worked for me.

tomaskikutis: I wanted to check the issue myself but couldn't reproduce it. Can you reproduce it consistently yourself? If so, how?

GyanP: Use sd-master as the server. Open any published item. Put a debugger here https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/authoring/services/AuthoringService.js#L570 and refresh the page. You will see `self.userDesks = []`.

tomaskikutis: You are not solving it at the right place Gyan.

Let's look at how the issue happens. As you said `self.userDesks` is empty at AuthoringService.js:570 (line numbers according to master `15ab2131`)

Let's investigate where `self.userDesks` is populated - there is only one additional usage in the file at line 45.

```
desks.fetchCurrentUserDesks().then((desksList) => {
    self.userDesks = desksList;
});
```

`self.userDesks` is populated asynchronously, but allows synchronous access which means race conditions will happen and is exactly the issue behind the ticket.

To solve this properly we must ensure that data which is fetched asynchronously can't be accessed synchronously. There are two ways to achieve that:
1. Make all functions who access `self.userDesks` asynchronous too, so they can wait for the promise to resolve.
2. Remove `self.userDesks` from the file entirely and ensure it is passed as an argument to the functions which required it. This way you could keep methods synchronous.

GyanP: @tomaskikutis fixed it.

tomaskikutis: > To solve this properly we must ensure that data which is fetched asynchronously can't be accessed synchronously.

The underlying issue is still there. If you decided to go with option 2 and pass `userDesks` as an argument, you need to remove `self.userDesks` from the file entirely and pass `userDesks` to all other functions which depend on it.

GyanP: `authoring.itemActions` function is called at more than 50 places across 15 files. So, every time the `self.userDesks` is required, that is why we need to keep it there.

tomaskikutis: Let's fix it in the 50 places then. That's why I provided 2 options: to make the methods themselves async or provide an argument. If you prefer I can do it.

GyanP: "If you prefer I can do it" 
I didn't get you?

tomaskikutis: If you want I can take over this task from you and update the usages in 50+ places you mentioned.

GyanP: Ok if you wish you can take it over.

GyanP:
> I gave a shot at solving in properly, but it requires too many functions to be converted from sync to async and is very likely to break a lot of stuff. I think going with your version is the best for now. Can you take a look at the e2e tests?

Ok, I will

PR #2832, opened 2019-02-11
attachments in editor content section

petrjasek:


scripts/apps/authoring/attachments/AttachmentsEditorDirective.ts
```diff
@@ -0,0 +1,5 @@
+
+export default function AttachmentsEditorDirective() {
```
tomaskikutis: Is this supposed to be an empty directive for a React component? If so, you can use `reactToAngular1`

scripts/apps/authoring/attachments/actions.ts
```diff
@@ -0,0 +1,89 @@
+import {gettext} from 'core/utils';
+
+export const INIT_ATTACHMENTS = 'INIT_ATTACHMENTS';
+export function initAttachments(item) {
+    return (dispatch, geState, {deployConfig, attachments}) => {
```
tomaskikutis: Named arguments would help. Where is `{deployConfig, attachments}` coming from? I thought only `dispatch, geState` was available in action creators.

scripts/apps/authoring/attachments/actions.ts
```diff
@@ -0,0 +1,89 @@
...
+        const payload = {
+            maxSize: deployConfig.getSync('attachments_max_size'),
+            maxFiles: deployConfig.getSync('attachments_max_files'),
+            files: [],
+        };
+
+        attachments.byItem(item)
+            .then((files) => {
+                payload.files = files;
```
tomaskikutis: How about initializing the state object and dispatching right here. Mutating objects or using `finally` is not needed then.

```
attachments.byItem(item)
    .then((files) => {
        dispatch({type: INIT_ATTACHMENTS, payload: {
            maxSize: deployConfig.getSync('attachments_max_size'),
            maxFiles: deployConfig.getSync('attachments_max_files'),
            files: files,
        }});
    });
```

scripts/apps/authoring/attachments/actions.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export const ADD_ATTACHMENTS = 'ADD_ATTACHMENTS';
+export function selectFiles(files) {
+    return (dispatch, getState, {$scope, notify, superdesk}) => {
+        const state = getState();
+
+        if (Array.isArray(files) && files.length > 0 && !state.editor.isLocked) {
+            if (files.length + state.attachments.files.length > state.attachments.maxFiles) {
+                notify.error(gettext('Sorry, too many files selected.'));
```
tomaskikutis: Ioan already merged translating plurals, so you can make the error messages more informative.

```
gettextCatalog.getPlural(
    state.attachments.maxFiles,
    "Too many files selected. Only one file is allowed.",
    "Too many files selected. Only {{$count}} files are allowed."
)
```

scripts/apps/authoring/attachments/actions.ts
```diff
@@ -0,0 +1,89 @@
...
+
+            const bigFiles = files.filter((file) => file.size > state.attachments.maxSize);
+
+            if (bigFiles.length) {
+                notify.error(gettext('Sorry, but some files are too big.'));
+                return;
+            }
+
+            superdesk
+                .intent('upload', 'attachments', files)
```
tomaskikutis: Could the intent thing be removed if redux is used? Which android developer added the intents here anyway? :D

scripts/apps/authoring/attachments/actions.ts
```diff
@@ -0,0 +1,89 @@
...
+    };
+}
+
+export function closeEdit() {
+    return {type: EDIT_ATTACHMENT, payload: null};
+}
+
+export function download(file) {
+    return (dispatch, getState, {urls, $window}) => {
+        $window.open(urls.media(file.media, 'attachments'), '_blank');
```
tomaskikutis: Why are you using `$window` here instead of native `window`?

scripts/apps/authoring/attachments/attachments.js
```diff
@@ -1,97 +1,67 @@
...
-    }
-
-    selectFiles(files) {
-        if (Array.isArray(files) && files.length > 0 && !this.isLocked) {
-            if (files.length + this.$scope.files.length > this.maxFiles) {
-                this.notify.error(gettext('Sorry, too many files selected.'));
-                return;
+    constructor($scope) {
+        // connect
+        Object.assign(this, mapDispatchToCtrl($scope.store.dispatch));
```
tomaskikutis: It isn't very clear how this works. Could you instead use a var like
```
const initialState = Object.assign(this, mapDispatchToCtrl($scope.store.dispatch));
```
and use that `initialState` where you need. I would also advise not to modify the original object by passing an empty object as a first argument to `Object.assign`. That way it's more clear what are the limits of influence of that variable.

scripts/apps/authoring/attachments/reducer.ts
```diff
@@ -0,0 +1,32 @@
...
+export function attachments(state = initialState, action) {
+    switch (action.type) {
+    case INIT_ATTACHMENTS:
+        return {...action.payload};
+    case ADD_ATTACHMENTS:
+        return {...state, files: state.files.concat(action.payload)};
+    case REMOVE_ATTACHMENT:
+        return {...state, files: state.files.filter((f) => f !== action.payload)};
+    case EDIT_ATTACHMENT:
+        return {...state, file: action.payload, diff: action.payload != null ? Object.create(action.payload) : null};
```
tomaskikutis: Why are you using `Object.create` in ` Object.create(action.payload)`?

scripts/apps/authoring/attachments/actions.ts
petrjasek: this opens a modal, for that I would probably go for another refactoring, will need to create support for that somehow

scripts/apps/authoring/attachments/actions.ts
petrjasek: that's angular thing, it's easier to mock/test it, but it's probably not needed here, just didn't change that part

scripts/apps/authoring/attachments/reducer.ts
petrjasek: that's like a diff, so that object will only have own properties for modified keys

scripts/apps/authoring/attachments/attachments.js
petrjasek: here it's populating controller methods, which are then called from template. it's there to avoid:
```
this.edit = (file) => dispatch(edit(file));
this.close = () => dispatch(close());
```

scripts/apps/authoring/attachments/actions.ts
petrjasek: yep that makes sense, when there is an error it should probably do something else instead of showing empty list

scripts/apps/authoring/attachments/actions.ts
petrjasek: that's from [thunk](https://github.com/reduxjs/redux-thunk#injecting-a-custom-argument)

scripts/apps/authoring/attachments/attachments.js
tomaskikutis: This is way more readable to me. I don't think there's any point of dynamic assignment when it's only used in one place.
```
const {dispatch} = $scope.store;

this.edit = (file) => dispatch(edit(file));
this.close = () => dispatch(close());
// all the other methods
```

scripts/apps/authoring/attachments/actions.ts
tomaskikutis: I think calling it `thunkCustomArgument` and optionally destructing from it in the function body would be more clear.

scripts/apps/authoring/attachments/reducer.ts
tomaskikutis: I'm not sure I get that. When you do `Object.create` you simply create a copy of an object. Is there more to it? How is it related to diffs?

petrjasek: yep so you can use new object to access old properties, it will get it via inheritance, but once you change it there will be an own property created on the diff one
```
const original = {foo: 1};
const diff = Object.create(original);

diff.foo
> 1

diff.hasOwnProperty('foo')
> false

diff.foo = 2;

diff.hasOwnProperty('foo')
> true
```
so when we use it later for patch request it's only sending own properties of diff, so only parts that were changed.

scripts/apps/authoring/attachments/attachments.js
petrjasek: I wanted to make some utils do handle it, something like in [redux-react](https://react-redux.js.org/api/connect#dispatch)

scripts/apps/authoring/attachments/reducer.ts
tomaskikutis: That's a neat trick, I haven't seen diffing done this way before. As you can see from me not understanding this right away, it would be good instead of calling `object.create` to call `prepareDiff` which would call `object.create` except that it would be documented in the function body how the diffing works.

Why are we calculating the diffs in the first place? Don't etags protect from conficting writes? Do etags work on document level or it takes the fields into account too?

petrjasek: that's not about etags and conflicting writes, but to reduce payloads and not to send data which were not changed.

tomaskikutis: I'm all for saving the data, but I'd like it to be abstracted away properly. Can it be done in a way that reducer and the whole application except the function which doing the diffing is not required to know about it? If so, we should do it.

scripts/apps/authoring/attachments/AttachmentsEditorDirective.ts
petrjasek: it renders a template

scripts/apps/authoring/attachments/actions.ts
petrjasek: this whole thing with `return (dispatch, getState)` is from thunk, and it's a pattern we use in superdesk-planning as well, I would rather stick to it

scripts/apps/authoring/attachments/reducer.ts
petrjasek: at the end I got rid of it, handling those changes as component state

tomaskikutis:
You've added a few commits after removing wip label. Are you finished with this now?

petrjasek:
yep

scripts/apps/authoring/attachments/reducer.ts
```diff
@@ -0,0 +1,38 @@
...
+    case INIT_ATTACHMENTS:
+        return {...action.payload};
+    case ADD_ATTACHMENTS:
+        return {...state, files: state.files.concat(action.payload)};
+    case REMOVE_ATTACHMENT:
+        return {...state, files: state.files.filter((f) => f !== action.payload)};
+    case EDIT_ATTACHMENT:
+        return {
+            ...state,
+            edit: action.payload,
```
tomaskikutis: I would like naming to be improved:
EDIT_ATTACHMENT -> ENABLE_EDIT_MODE_FOR_ATTACHEMENT
SAVE_ATTACHMENT -> UPDATE_ATTACHEMENT

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
```diff
@@ -1144,6 +1148,30 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
...
             macros.setupShortcuts($scope);
             initEmbedFieldsValidation();
+
+            // init redux
+            const initialState = {
+                isLocked: $scope.isLocked(),
+                isLockedByMe: $scope.isLockedByMe(),
+            };
+
+            function editor(state = initialState) {
```
tomaskikutis: What does this do?

scripts/core/ui/utils.tsx
```diff
@@ -0,0 +1,22 @@
...
+import {render, unmountComponentAtNode} from 'react-dom';
+
+interface IStoreScope extends ng.IScope {
+    store: Store<{}>;
+}
+
+/**
+ * Render react component and connect it to redux store
+ */
+export function renderComponent(component: React.ReactElement<any>, scope: IStoreScope, elem: Array<Element>) {
```
tomaskikutis: Doesn't `reactToAngular1` do the same thing? You could wrap the initial component in the `<Provider>` and pass store as a prop.

tomaskikutis:
Could you also change *.js files you changed to *.ts? I think it's better to rename it in one commit and fix errors in another to preserve history.

scripts/apps/authoring/attachments/AttachmentsEditorModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+    update(key: string, val: string) {
+        const updates = {};
+
+        updates[key] = val;
+        this.setState(updates);
+    }
+
+    render() {
+        return (
+            <React.Fragment>
```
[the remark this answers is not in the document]
petrjasek: it doesn't work otherwise, you have to return single element or array, fragment is to avoid that top level div or something

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
petrjasek: that's initial editor store, I have to for state namespacing, `isLocked` shouldn't be part of attachments state

scripts/apps/authoring/attachments/AttachmentsList.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+import {Item, Column, Row, ActionMenu} from 'core/ui/components/List';
+
+interface IFile {
+    _id: string;
+    title: string;
+    mimetype: string;
+    filename: string;
+    description: string;
+    media: {
```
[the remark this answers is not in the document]
petrjasek: no, it's internal media metadata, length there is file length

scripts/apps/authoring/attachments/actions.ts
```diff
@@ -0,0 +1,84 @@
...
+                });
+        }
+    };
+}
+
+export const REMOVE_ATTACHMENT = 'REMOVE_ATTACHMENT';
+export function removeFile(file) {
+    return (dispatch, getState, {$scope}) => {
+        dispatch({type: REMOVE_ATTACHMENT, payload: file});
+        updateItem(getState(), $scope);
```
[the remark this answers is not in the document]
petrjasek: I will think about that, not every store change should call autosave, so there must be some logic and not sure yet where to put it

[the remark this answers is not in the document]
tomaskikutis: You don't have it to update on every store change. You can put conditionals to check the objects you care about.

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
tomaskikutis: What do you mean by **initial** editor reducer? Does the reducer change over time?

scripts/apps/authoring/attachments/AttachmentsList.tsx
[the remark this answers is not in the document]
tomaskikutis: Is the length property native to the object(like it's native for all arrays) or set manually? Why does it only have a length property and no others?

scripts/apps/authoring/attachments/attachments.ts
```diff
@@ -0,0 +1,162 @@
...
+            return api.remove(file);
+        }
+
+        /**
+         * Get attachment by id
+         *
+         * @param {String} id
+         * @return {Promise}
+         */
+        byId(id) {
```
tomaskikutis: Is the method used anywhere? I didn't see it in current changes. Does the cache live through application lifecycle and isn't specific to any view? How are cached items updated when someone else changes them?

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
petrjasek: I mean it should get extended as we port more features to react/redux from angular

scripts/apps/authoring/attachments/actions.ts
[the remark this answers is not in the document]
petrjasek: I think it's better explicit as part of the action

scripts/apps/authoring/authoring/directives/AuthoringDirective.js
tomaskikutis: So it doesn't do anything now, right? I would remove it that case, or at least move it to a separate file and document why is it empty.

petrjasek: it's used by attachments component, it just doesn't change

scripts/apps/authoring/attachments/actions.ts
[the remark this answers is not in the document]
tomaskikutis: I'm all for explicitness, but it seems more like unnecessary coupling. You could use the subscription and define the pieces of data that should trigger autosave in an equally explicit way.

scripts/apps/authoring/attachments/AttachmentsList.tsx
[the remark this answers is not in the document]
petrjasek: it's backend media object, we get it from api

scripts/apps/authoring/attachments/actions.ts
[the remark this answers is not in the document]
petrjasek: will give it a shot

scripts/apps/authoring/attachments/AttachmentsList.tsx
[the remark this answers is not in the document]
tomaskikutis: What's the schema for it in the back-end?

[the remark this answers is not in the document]
petrjasek: I've added core interface for that

scripts/apps/authoring/attachments/attachments.ts
petrjasek: I think it's used in editor when rendering attachment used as link

scripts/apps/authoring/attachments/actions.ts
```diff
@@ -0,0 +1,76 @@
...
+        if (Array.isArray(files) && files.length > 0 && !state.editor.isLocked) {
+            if (files.length + state.attachments.files.length > state.attachments.maxFiles) {
+                notify.error(gettext('Sorry, too many files selected.'));
+                return;
+            }
+
+            const bigFiles = files.filter((file) => file.size > state.attachments.maxSize);
+
+            if (bigFiles.length) {
+                notify.error(gettext('Sorry, but some files are too big.'));
```
tomaskikutis: Include maximum file size and offending file names in the message.

scripts/apps/authoring/attachments/actions.ts
```diff
@@ -35,7 +36,13 @@ export function selectFiles(files: Array<File>) {
                 .map((file) => file.name);
 
             if (filenames.length) {
-                notify.error(gettext('Sorry, but files "{{ filenames }}" are too big.', {filenames: filenames.join(', ')}));
+                notify.error(gettext(
+                    'Sorry, but some files "{{ filenames }}" are bigger than limit ({{ limit }}).',
```
tomaskikutis: I think it's better to remove the spaces between double curly brackets and variable names so it's more clear for the translators that it's a placeholder and must not be translated.

scripts/apps/authoring/attachments/actions.ts
```diff
@@ -19,7 +20,7 @@ export function initAttachments(item) {
 
 export const ADD_ATTACHMENTS = 'ADD_ATTACHMENTS';
 export function selectFiles(files: Array<File>) {
-    return (dispatch, getState, {notify, superdesk}) => {
+    return (dispatch, getState, {notify, superdesk, $filter}) => {
```
tomaskikutis: Is `$filter` used anywhere?

scripts/core/ui/utils.tsx
petrjasek: not sure how to do it, it works with `reactToAngular1(Provider, ['store'])` but not sure how to pass that children component

tomaskikutis: Instead of doing `reactToAngular1(Provider, ['store'])`, do `reactToAngular1(MyCustomComponent, ['customProps'])` and use `Provider` inside `MyCustomComponent`. You would do so for each individual component. `reactToAngular1` is more like `AttachmentsListDirective` which isn't doing much more than making it possible to use a React component as a directive.

petrjasek: you can give it a shot, I can't make it work

tomaskikutis: https://github.com/tomaskikutis/superdesk-client-core/tree/pr-2832

Tests pass and UI seems to work, so I think you can merge it from my branch.

scripts/apps/authoring/attachments/actions.ts
petrjasek: not anymore

PR #2833, opened 2019-02-12
feat(macros): For editor3 add support to change html_body on macros (…

ioanpocol:
…SDANSA-222)

### DIFF
--- a/scripts/core/editor3/helpers/tansa.ts
+++ b/scripts/core/editor3/helpers/tansa.ts
@@ -41,12 +41,14 @@ export function getTansaHtml(editorState) {
 
 /**
  * Update the editor with the changes performed by Tansa
+ * If the simpleReplace is true try to preserve the existing inline styles and entities
  *
  * @param {EditorState} editorState
  * @param {String} html
+ * @param {String} simpleReplace
  * @returns {EditorState}
  */
-export function setTansaHtml(editorState, html) {
+export function setTansaHtml(editorState, html, simpleReplace) {
     let content = editorState.getCurrentContent();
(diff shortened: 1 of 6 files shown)


scripts/core/editor3/helpers/tansa.ts
```diff
@@ -157,30 +157,124 @@ function updateText(editorState, content, block, newText, diffMatchPatch) {
...
+function insertText(editorState, content, block, offset, text) {
+    const selection = createSelectionForBlock(editorState, block, offset);
+    const newContent = Modifier.insertText(content, selection, text);
+
+    offset += text.length;
+    return {newContent, offset};
+}
+
+/**
+ * Replate text with newText and preserve the existing inline styles and entities
```
petrjasek: I wouldn't try that, it's error prone imo, dropping styles and entities should be fine for now

ioanpocol: It should be ok since I replaced the text character by character and I preserved the inline styles and entity.

petrjasek: but the words order can be different too

ioanpocol: For uppercase and lowercase no and probably for other cases. For translation yes.

ioanpocol: Maybe we can set on macro if we want to preserve or not the style and entity. On translation I think it is not possible to preserve them.

petrjasek: ah right, I was talking about translation, uppercase/lowercase should be fine

PR #2834, opened 2019-02-12
Fixed: Fields are not being added to the selected position in CPs (SD…

GyanP:
…ESK-3908)

### DIFF
--- a/scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
+++ b/scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
@@ -232,7 +232,7 @@ export function ContentProfileSchemaEditor(content, metadata, vocabularies, noti
                  */
                 scope.toggle = (schema, order, position) => {
                     if (scope.model.editor[schema.key].enabled) {
-                        throw new Error('Unexpected behaviour: Item arlready added.');
+                        throw new Error('Unexpected behaviour: Item already added.');
                     }
 
                     if (scope.model.editor[schema.key]) {
--- a/scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
+++ b/scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
@@ -234,9 +234,11 @@ export function ContentProfileSchemaEditor(content, metadata, vocabularies, noti
                     let newOrder = order;
 
                     if (scope.model.editor[schema.key].enabled) {
-                        notify.error(gettext(
-                            `This item is already added in ${scope.model.editor[schema.key].section} fields.`
-                        ));
+                        notify.error(
+                            gettext(
+                                `This item is already added in ${scope.model.editor[schema.key].section} fields.`
+                            ),
+                        );
                         return;
                     }
                     if (scope.model.editor[schema.key]) {
--- a/scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
+++ b/scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
@@ -5,6 +5,7 @@ import {getTypeForFieldId} from '../../helpers/getTypeForFieldId';
 import {IArticle} from 'superdesk-interfaces/Article';
 import {IVocabulary} from 'superdesk-interfaces/Vocabulary';
 import {assertNever} from 'core/helpers/typescript-helpers';
+import {gettext} from 'core/utils';
 
 interface IScope extends ng.IScope {
     model: any;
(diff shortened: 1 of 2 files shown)


scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
```diff
@@ -229,6 +231,14 @@ export function ContentProfileSchemaEditor(content, metadata, vocabularies) {
                  * @param {String} id the key of the field to toggle.
                  */
                 scope.toggle = (schema, order, position) => {
+                    let newOrder = order;
+
+                    if (scope.model.editor[schema.key].enabled) {
+                        notify.error(
+                            gettext(`This item is already added in ${scope.model.editor[schema.key].section} fields.`),
```
ioanpocol: please use the . following format for gettext:
>gettext('This item is already added in {{section}} fields.', {section: scope.model.editor[schema.key].section});

GyanP: @ioanpocol  fixed that.

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
```diff
@@ -238,7 +247,17 @@ export function ContentProfileSchemaEditor(content, metadata, vocabularies) {
                         scope.model.schema[schema.key] = {enabled: true};
                     }
 
-                    schemaKeys.splice(position === 'before' ? order - 1 : order + 1, 0, schema.key);
+                    if (scope.sectionToRender === 'content') {
```
tomaskikutis: Why is it important which section it is? Why does the header section work correctly without modification?

GyanP: We have different `schemaKeys` for `header fields` and `content fields` and for `content fields` the `order = no. of header fields + index in schemaKeys for content fields`. So, if header fields are `[h1, h2]` and content fields are `[c1, c2]` and I want to add a new content field before `c2` then  `order = 4, position = 'before'` and `schemaKeys.length = 2` for content fields. So it will push the `c3` at the end of schemaKeys of content fields like`[c1, c2, c3]` but expected result should be like this` [c1, c3, c2]`. And such is not the case with header fields as both `order` as well as length of schemaKeys starts from 1.

tomaskikutis: The way we are handling the order looks messy and harder to understand on every modification. Do you think you could refactor it so the same code could work on both groups with no exceptional handling?

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
```diff
@@ -229,6 +231,13 @@ export function ContentProfileSchemaEditor(content, metadata, vocabularies) {
                  * @param {String} id the key of the field to toggle.
                  */
                 scope.toggle = (schema, order, position) => {
+                    let newOrder = order;
+
+                    if (scope.model.editor[schema.key].enabled) {
+                        notify.error(gettext('This item is already added in {{section}} fields.',
```
tomaskikutis: How can the error happen? Shouldn't the dropdown only contain the values user is allowed to select?

GyanP: yep, the dropdown should contain the allowed values only.

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
GyanP: I will try to do that.

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
tomaskikutis: So this should never happen because it's already validated before displaying the dropdown. In such cases, it's better to simply throw an error to indicate to other developers that this not expected behavior and not bother the translators with translating texts which will probably never be used.

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
```diff
@@ -231,13 +231,10 @@ export function ContentProfileSchemaEditor(content, metadata, vocabularies, noti
                  * @param {String} id the key of the field to toggle.
                  */
                 scope.toggle = (schema, order, position) => {
-                    let newOrder = order;
-
                     if (scope.model.editor[schema.key].enabled) {
-                        notify.error(gettext('This item is already added in {{section}} fields.',
-                            {section: scope.model.editor[schema.key].section}));
-                        return;
+                        throw new Error('Unexpected behaviour: Item arlready added.');
```
tomaskikutis: typo in word already

PR #2835, opened 2019-02-12
SDFID-357 make login overridable

pablopunk:
To allow the login modal to be overriden by templateCache

Required by https://github.com/superdesk/superdesk-fi/pull/234

### DIFF
--- a/scripts/core/auth/login-modal-directive.ts
+++ b/scripts/core/auth/login-modal-directive.ts
@@ -16,7 +16,7 @@ angular.module('superdesk.core.auth.login', []).directive('sdLoginModal', [
     function(session, auth, features, config, deployConfig, usersService, notify, $route) {
         return {
             replace: true,
-            template: require('./login-modal.html'),
+            templateUrl: 'scripts/core/auth/login-modal.html',
             link: function(scope, element, attrs) {
                 scope.features = features;
                 scope.changePassword = false;
--- a/scripts/core/auth/login-modal-directive.ts
+++ b/scripts/core/auth/login-modal-directive.ts
@@ -16,6 +16,7 @@ angular.module('superdesk.core.auth.login', []).directive('sdLoginModal', [
     function(session, auth, features, config, deployConfig, usersService, notify, $route) {
         return {
             replace: true,
+            // login template can be overriden (like on superdesk-fi)
             templateUrl: 'scripts/core/auth/login-modal.html',
             link: function(scope, element, attrs) {
                 scope.features = features;


scripts/core/auth/login-modal-directive.ts
```diff
@@ -16,7 +16,7 @@ angular.module('superdesk.core.auth.login', []).directive('sdLoginModal', [
     function(session, auth, features, config, deployConfig, usersService, notify, $route) {
         return {
             replace: true,
-            template: require('./login-modal.html'),
+            templateUrl: 'scripts/core/auth/login-modal.html',
```
tomaskikutis: I'd like the value to be received from our extensibility/plugin system, but since we don't have one yet, I think it's best to add this to `superdesk.config.js`.

pablopunk: What would you put inside config? The path to the template?

tomaskikutis: yep. `loginTemplateUrl: 'scripts/core/auth/login-modal.html'`

pablopunk: I don't know if it would be a good idea to start declaring template paths on config. And if it is I would rather not put it in the core. just on client repos. So it would be `templateUrl: config.templateUrl || 'scripts/core...'`

tomaskikutis: As we don't have an extension system, I don't see a better place than the config. It makes it obvious that this part is extensible which wouldn't be clear otherwise.

petrjasek: makes sense to make it explicit

pablopunk: @petrjasek in the code or in the config file?

Anyway, I tried the config option and it's not working because even if I put the path there, this directive won't find the path on the outter repo

tomaskikutis: If it doesn't find the path, you can pass the entire template string instead by using `template` instead `templateUrl` in the directive.

pablopunk: @tomaskikutis template doesn't find the path either as it needs a `require()`

tomaskikutis: How about using `require` in the config?
`loginTemplate: require('scripts/core/auth/login-modal.html')`

pablopunk: @tomaskikutis doesn't work because Grunt fails to start: `SyntaxError: Unexpected token <` it's probably trying to load the html as a module and of course it fails

tomaskikutis: Keep it as is then and add a comment that the path is overwritten in fi repo.

PR #2837, opened 2019-02-14
fix(photo-editor): Fixed bugs with media fields and media editor

pavlovicnemanja:
- SDESK-3736 - can't access and edit metadata for video or audio in media_gallery
- SDESK-3380 - error is thrown on sending correction for an item containing images with adjusted brightness in custom media field
- SDANSA-225 - Photo editing bug

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ArticleEditDirective.js
+++ b/scripts/apps/authoring/authoring/directives/ArticleEditDirective.js
@@ -281,14 +281,16 @@ export function ArticleEditDirective(
                  * @description Opens the Change Image Controller to modify the image metadata.
                  */
                 scope.editMedia = (defaultTab = 'view') => {
-                    let hideTabs = [];
+                    let showTabs = [];
 
                     scope.mediaLoading = true;
 
-                    if (scope.item.type === 'video' || scope.item.type === 'audio') {
-                        hideTabs = ['image-edit', 'crop'];
-                    } else if (!scope.metadata.crop_sizes) {
-                        hideTabs = ['crop'];
+                    if (scope.item.type === 'picture' && scope.metadata.crop_sizes) {
+                        showTabs = ['view', 'image-edit', 'crop'];
+                    } else if (scope.item.type === 'picture' && !scope.metadata.crop_sizes) {
+                        showTabs = ['view', 'image-edit'];
+                    } else {
+                        showTabs = ['view'];
                     }
 
                     return renditions.crop(
@@ -298,7 +300,7 @@ export function ArticleEditDirective(
                             editable: true,
                             isAssociated: false,
                             defaultTab: defaultTab,
-                            hideTabs: hideTabs,
+                            tabs: showTabs,
                             showMetadata: true,
                         }
                     )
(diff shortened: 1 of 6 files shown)


scripts/apps/authoring/authoring/directives/ArticleEditDirective.js
```diff
@@ -280,17 +280,25 @@ export function ArticleEditDirective(
...
                  * @description Opens the Change Image Controller to modify the image metadata.
                  */
-                scope.editMedia = (defaultTab = 'view', hideTabs = []) => {
+                scope.editMedia = (defaultTab = 'view') => {
+                    let hideTabs = [];
+
                     scope.mediaLoading = true;
 
+                    if (scope.item.type === 'video' || scope.item.type === 'audio') {
+                        hideTabs = ['image-edit', 'crop'];
```
tomaskikutis: Use direct checks:
```
if (scope.item.type === 'image') {
    showTabs = ['image-edit', 'crop']
```

When using negative checks, future changes break stuff easily. If I added a new type say 'powerpoint slides' cropping and image-edit would enable automatically which wouldn't be the case if using direct checks.

scripts/apps/authoring/authoring/directives/ArticleEditDirective.js
```diff
@@ -280,17 +280,27 @@ export function ArticleEditDirective(
...
                  */
-                scope.editMedia = (defaultTab = 'view', hideTabs = []) => {
+                scope.editMedia = (defaultTab = 'view') => {
+                    let showTabs = [];
+
                     scope.mediaLoading = true;
 
+                    if (scope.item.type === 'picture') {
+                        showTabs = ['view', 'image-edit', 'crop'];
+                    } else if (scope.item.type === 'picture' && !scope.metadata.crop_sizes) {
```
ioanpocol: This branch will be never reached

scripts/apps/authoring/authoring/services/AuthoringService.js
```diff
@@ -230,6 +230,11 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
     this.publish = function publish(orig, diff, action = 'publish') {
         let extDiff = helpers.extendItem({}, diff);
 
+        // if current document is image and it has been changed on 'media edit' we have to update the etag
+        if (orig.type === 'picture') {
```
tomaskikutis: Why is this conditional? Shouldn't etag work the same for every type?

pavlovicnemanja: because differentiation of etags can be only if there is some change on image with media editor. You can't change brightness and contrast on video (for now) and audio files

tomaskikutis: If I understand correctly it would work fine even without the conditional. It's only that other types of files can't take advantage of it yet. Could you remove the conditional and add a comment describing the case when it's necessary?

pavlovicnemanja: here you are

pavlovicnemanja: I needed to add check if etag exists on diff, because there are some cases that etag is not send through diff.

PR #2839, opened 2019-02-15
[SDESK-3915] Allow clearing the publish schedule on save and publish

marwoodandrew:


marwoodandrew:
depends on superdesk/superdesk-core#1534

PR #2841, opened 2019-02-18
fix(monitoring): avoid click action on double click

petrjasek:
SDESK-3931

### DIFF
--- a/scripts/apps/search/components/Item.tsx
+++ b/scripts/apps/search/components/Item.tsx
@@ -68,7 +68,7 @@ export class Item extends React.Component<IProps, IState> {
 
     readonly state = {hover: false, actioning: false, isActionMenuOpen: false, showNested: false};
 
-    clickTimeout: NodeJS.Timeout;
+    clickTimeout: number;
 
     constructor(props) {
         super(props);
@@ -140,7 +140,7 @@ export class Item extends React.Component<IProps, IState> {
     select(event) {
         if (!this.props.item.gone && !this.clickTimeout) {
             event.persist(); // make event available in timeout callback
-            this.clickTimeout = setTimeout(() => {
+            this.clickTimeout = window.setTimeout(() => {
                 this.clickTimeout = null;
                 this.props.onSelect(this.props.item, event);
             }, CLICK_TIMEOUT);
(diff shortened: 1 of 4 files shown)


scripts/apps/search/components/Item.tsx
```diff
@@ -134,8 +138,14 @@ export class Item extends React.Component<IProps, IState> {
     }
 
     select(event) {
-        if (!this.props.item.gone) {
-            this.props.onSelect(this.props.item, event);
+        const {$timeout} = this.props.svc;
```
tomaskikutis: Why are you using $timeout in a React component? Doesn't `setTimeout` work?

scripts/apps/search/components/Item.tsx
```diff
@@ -134,8 +138,14 @@ export class Item extends React.Component<IProps, IState> {
...
+
+        if (!this.props.item.gone && !this.clickTimeout) {
```
tomaskikutis: What is `item.gone`?

scripts/apps/search/components/Item.tsx
petrjasek: because protractor knows it should wait for $timeouts

scripts/apps/search/components/Item.tsx
tomaskikutis: Shouldn't the React component unmount if it's spiked/sent to another stage?

petrjasek: it's visible but grayed out so other users can see that the item was there but is not there anymore

scripts/apps/search/components/Item.tsx
tomaskikutis: Can you tell him about `setTimeout` too? :D Jokes aside, you can wait for an element in tests. The code shouldn't be coupled with test frameworks. Good locators should be all that's needed.

scripts/apps/search/components/Item.tsx
tomaskikutis: Got it.

scripts/apps/search/components/Item.tsx
tomaskikutis: Can you replace it with `setTimeout` and make the test wait for the element?

PR #2842, opened 2019-02-19
fixed: the Save button gets activated but the item is not added to me…

GyanP:
…dia_gallery due to count limitation (SDESK-3964)

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
+++ b/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
@@ -30,6 +30,7 @@ export function ItemCarouselDirective($timeout, notify) {
             save: '&',
             onchange: '&',
             maxUploads: '=',
+            field: '=',
         },
         transclude: true,
         templateUrl: 'scripts/apps/authoring/views/item-carousel.html',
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/directives/ItemCarouselDirective.js
```diff
@@ -129,6 +130,33 @@ export function ItemCarouselDirective($timeout, notify) {
...
+                        notify.error(
+                            gettext(
+                                'Media item was not added, because the field reached the limit of allowed media items.'
+                            )
+                        );
+                        return;
+                    }
+
+                    if (itemAlreadyAddedAsMediaGallery) {
+                        notify.error('This item is already added as media gallery.');
```
petrjasek: what about adding only missing files and not just stopping?

petrjasek: it would make sense to show some warning if all files were added already and there is no file left

GyanP: > what about adding only missing files and not just stopping?

sorry, I didn't get you.

petrjasek: my bad, didn't realise it deals with only 1 file, you can ignore those comments

GyanP: > it would make sense to show some warning if all files were added already and there is no file left

@petrjasek you mean something like this?
related_media

petrjasek: you can ignore my comments, I didn't realise it's only handling one file at a time

PR #2843, opened 2019-02-20
fix(macro): Fix the apply of macros for editor2 (SDESK-4002)

ioanpocol:


scripts/apps/authoring/macros/macros.js
```diff
@@ -191,6 +191,7 @@ function MacrosController($scope, macros, desks, autosave, $rootScope, storage,
             }
 
             if (isEditor3 || res.diff == null) {
+                $rootScope.macro_dirty = true;
```
ioanpocol: @petrjasek 
Any idea how to set dirty for the current item?

petrjasek: what about `$scope.autosave()` ?

ioanpocol: Nemanja proposed first a more generic solution to use broadcast.
Please review it.

petrjasek: imo using autosave like all other components is better

ioanpocol: OK, it is better autosave then. Please review again.

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -26,6 +26,7 @@ export function AuthoringTopbarDirective(TranslationService, privileges, authori
             scope.saveTopbar = function() {
                 scope.saveDisabled = true;
                 return scope.save(scope.item)
+                    .then((item) => angular.extend(scope.item, item))
```
petrjasek: why is this needed here?

ioanpocol: to be sure that after save the item has the same value as origItem

petrjasek: shouldn't it be part of that `scope.save` method then?

ioanpocol: OK, moved on save

PR #2845, opened 2019-02-20
feat(mmetadata): Added ability to copy&paste metadata

pavlovicnemanja:
SDANSA-224 - Save / Restore metadata into photo uploader

### DIFF
--- a/scripts/apps/search/MultiImageEdit.ts
+++ b/scripts/apps/search/MultiImageEdit.ts
@@ -195,7 +195,13 @@ export function MultiImageEditDirective(asset, $sce) {
         controller: MultiImageEditController,
         templateUrl: asset.templateUrl('apps/search/views/multi-image-edit.html'),
         link: function(scope) {
+            const METADATA_ITEMS = 'metadata:items';
+            const METADATA_AUTOFILL = 'metadata:autofill';
+
+            scope.metadataFromStorage = !!localStorage.getItem(METADATA_ITEMS);
+
             scope.trustAsHtml = $sce.trustAsHtml;
+            scope.autoFill = localStorage.getItem(METADATA_AUTOFILL) === 'true';
 
             scope.handleItemClick = function(event, image) {
                 if (event.target != null && event.target.classList.contains('icon-close-small')) {
@@ -204,6 +210,32 @@ export function MultiImageEditDirective(asset, $sce) {
                     scope.selectImage(image);
                 }
             };
+
+            scope.copyMetadata = (metadata) => {
+                scope.metadataFromStorage = true;
+                localStorage.setItem(METADATA_ITEMS, JSON.stringify(metadata));
+            };
+
+            scope.toggleAutofill = () => {
+                scope.autoFill = !scope.autoFill;
+                localStorage.setItem(METADATA_AUTOFILL, scope.autoFill);
+            };
+
+            scope.pasteMetadata = () => {
+                scope.metadata = JSON.parse(localStorage.getItem(METADATA_ITEMS));
+                _.forEach(scope.metadata, (metadata, key) => {
+                    scope.onChange(key);
+                });
+            };
+
+            scope.clearSavedMetadata = () => {
+                localStorage.removeItem(METADATA_ITEMS);
+                scope.metadataFromStorage = false;
+            };
+
+            if (scope.autoFill) {
+                scope.pasteMetadata();
+            }
         },
     };
 }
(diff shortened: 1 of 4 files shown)


scripts/apps/search/views/multi-image-edit.html
```diff
@@ -87,6 +87,14 @@ <h3 class="sd-slide-in-panel__heading" translate>Metadata</h3>
                         ></div>
                     </div>
                 </div>
+                <div class="sd-slide-in-panel__footer">
```
petrjasek: it's missing translate annotations

scripts/apps/search/views/multi-image-edit.html
```diff
@@ -87,6 +87,14 @@ <h3 class="sd-slide-in-panel__heading" translate>Metadata</h3>
...
+                    <div class="btn-group--flex sd-margin-b--1">
+                        <button type="button" class="btn btn--hollow btn--primary btn--expanded" ng-click="copyMetadata(metadata)" ng-disabled="!isDirty()" translate>Copy metadata</button>
+                        <button type="button" class="btn btn--hollow btn--expanded" ng-click="pasteMetadata()" ng-disabled="!metadataFromStorage" translate>Paste metadata</button>
+                    </div>
+                    <sd-check ng-model="autoFill" ng-click="toggleAutofill()">Autofill</sd-check>
+                    <a ng-click="clearSavedMetadata()" class="text-link text-link--white pull-right" ng-show="metadataFromStorage">Clear saved metadata</a>
```
petrjasek: there is missing translate

scripts/apps/search/MultiImageEdit.ts
```diff
@@ -196,6 +196,8 @@ export function MultiImageEditDirective(asset, $sce) {
         templateUrl: asset.templateUrl('apps/search/views/multi-image-edit.html'),
         link: function(scope) {
             scope.trustAsHtml = $sce.trustAsHtml;
+            scope.metadataFromStorage = localStorage.getItem('metadata:items');
```
petrjasek: would be good to define constants for those storage keys

scripts/apps/search/MultiImageEdit.ts
```diff
@@ -204,6 +206,32 @@ export function MultiImageEditDirective(asset, $sce) {
...
+                localStorage.setItem('metadata:items', JSON.stringify(metadata));
+            };
+
+            scope.toggleAutofill = () => {
+                scope.autoFill = !scope.autoFill;
+                localStorage.setItem('metadata:autofill', scope.autoFill);
+            };
+
+            scope.pasteMetadata = () => {
+                scope.metadata = JSON.parse(localStorage.getItem('metadata:items'));
```
petrjasek: you have `scope.metadataFromStorage` defined earlier, why not use it here? or why load it before?

scripts/apps/search/MultiImageEdit.ts
```diff
@@ -204,6 +206,32 @@ export function MultiImageEditDirective(asset, $sce) {
...
+                _.forEach(scope.metadata, (metadata, key) => {
+                    scope.onChange(key);
+                });
+            };
+
+            scope.clearSavedMetadata = () => {
+                localStorage.removeItem('metadata:items');
+                delete scope.metadataFromStorage;
```
petrjasek: seems like you only use this to indicate if it's empty, imo would be better to populate it like:
```
scope.metadataFromStorage = !!localStorage.getItem('metadata:items');
```
and here you can just set it to `false`, so it doesn't compare the whole metadata json just to check it's empty

pavlovicnemanja: done 🙂

scripts/apps/search/MultiImageEdit.ts
pavlovicnemanja: I changed scope.metadataFromStorage to be just boolean, so now I use it just for checks

scripts/apps/search/MultiImageEdit.ts
pavlovicnemanja: Where should I define them? Should I put it on top like ```CONST metadataItems = "metadata:item"```?

petrjasek: top of file or just top of link, doesn't matter. and should  be more like:
```
const METADATA_ITEMS = 'metadata:items';
const METADATA_AUTOFILL = 'metadata:autofill';
```

pavlovicnemanja: done 😄

PR #2848, opened 2019-02-21
fix(editor3): fix html generated for media blocks

petrjasek:
to be compatible with editor2

SDESK-3856


tomaskikutis:
Is the new format fine with Fidelity?

hlmnrmr:
They never replied because there is nobody who knows the answer on their side. Let's merge this and I will follow up on how this affects Fidelity by deploying these changes to test envs.

tomaskikutis:
We should at least tweak the HTML on their front-end using browser dev tools to see if their design doesn't break.

PR #2850, opened 2019-02-21
sync item associations on client

petrjasek:
atm it's done on server on save, but if user publishes without
saving first associations data is missing or incomplete

SDESK-4015

### DIFF
--- a/scripts/apps/authoring/authoring/authoring.spec.js
+++ b/scripts/apps/authoring/authoring/authoring.spec.js
@@ -86,6 +86,76 @@ describe('authoring', () => {
         expect(context.$rootScope.$applyAsync).toHaveBeenCalled();
     }));
 
+    it('synchronizes associations field from editor state', inject(() => {
+        const entity1 = {guid: 'entity1', renditions: {original: {href: 'foo'}}};
+        const entity2 = {guid: 'entity2', renditions: {original: {href: 'bar'}}};
+
+        const contentStateRaw = {
+            blocks: [
+                {
+                    key: 'aaaa1', text: 'test', type: 'unstyled', depth: 0, inlineStyleRanges: [], entityRanges: [],
+                },
+                {
+                    key: 'aaaa2', text: ' ', type: 'atomic', depth: 0, inlineStyleRanges: [], entityRanges: [
+                        {
+                            offset: 0,
+                            length: 1,
+                            key: 0,
+                        },
+                    ],
+                },
+                {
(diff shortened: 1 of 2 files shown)


tomaskikutis:
How is it possible to publish without saving first?

karelpetrak:
@tomaskikutis I think what Petr means is when you do not explicitly save (hit the save button) but proceed to publish (and the item is saved as a part of the "publish" workflow).

tomaskikutis:
How about running the same code no matter if saving happens via save button or as a part of publishing workflow?

petrjasek:
@tomaskikutis yep that makes sense, but not sure how it's related to this fix. imo it's better to send correct data all the time

tomaskikutis:
I do agree, that it's better to send correct data all the time, but the function which returns bad results should be fixed instead of having an extra function which fixes the bad output of the first one. Or isn't that the case?

petrjasek:
that part is done on server, it was syncing the data only on save, now that part is gone

tomaskikutis:
Which line on the server is doing this? Are you saying that if I take this PR and remove the sync function, all associations including images, gallery, related media won't appear in items' `associations` field?

petrjasek:
https://github.com/superdesk/superdesk-core/blob/master/apps/archive/archive.py#L94-L114

it will only appear there on save

tomaskikutis:
If the code which saves the item is also updating the associations, why can't we run the same code(which does the saving) before publishing? It would update the associations and we'd avoid duplicating code between client/server.

It might require a bit more work, but it's much more sustainable since every future change in update logic would automatically get applied when saving.

petrjasek:
the idea is to remove that code from server once this PR is merged

tomaskikutis:
That makes more sense now. Do you know why [update_associations](https://github.com/superdesk/superdesk-core/blob/master/apps/archive/archive.py#L94-L114) function exists in the first place? Why individual components can't take care of updating the field. I don't mind if it's done the same way across the app, but I see some components like related items writing to `item.associations` directly.

@ioanpocol I see you implemented [update_associations](https://github.com/superdesk/superdesk-core/blob/master/apps/archive/archive.py#L94-L114), do you remember what was the case? I can't find the ticket reference in [the PR](https://github.com/superdesk/superdesk-core/pull/1081).

ioanpocol:
All images added to editor3 should also appear on associations.  So update_associations function process editor3 state and update the associations.  I don't know why this has been implemented on . backend and not on frontend.

tomaskikutis:
Thanks, Ioan. I now see that we don't update associations when images or embeds change in editor3 and leave it up to the synchronizer function.

PR #2852, opened 2019-02-22
fix(modal): fix empty confirm modal

petrjasek:
when using translations

SDESK-3965


spec/errors_spec.js
```diff
@@ -0,0 +1,20 @@
+
+var modal = require('./helpers/modal');
+var prefs = require('./helpers/user_prefs');
+var authoring = require('./helpers/authoring');
+var monitoring = require('./helpers/monitoring');
+
+describe('errors', () => {
+    it('SDESK-3965: render modals with translated specs', () => {
```
tomaskikutis: Write a clearer name for the test, for example `it('displays default language strings in the modals when no translations are available')`

spec/errors_spec.js
```diff
@@ -0,0 +1,20 @@
...
+        prefs.navigateTo();
+        prefs.setLang('English (UK)');
+        prefs.save();
+        monitoring.openMonitoring();
+        monitoring.createFromDeskTemplate();
```
tomaskikutis: Can this be a unit test so it can be run only on the modal and wouldn't depend on monitoring or desk templates?

spec/helpers/modal.js
```diff
@@ -0,0 +1,10 @@
+
+const header = element(by.className('modal__header'));
+const body = element(by.className('modal__body'));
+
+module.exports = {
```
tomaskikutis: Are these used?

scripts/core/services/modalService.spec.ts
```diff
@@ -0,0 +1,17 @@
+describe('modal service', () => {
+    beforeEach(window.module('superdesk.core.services.modal'));
+
+    beforeEach(inject(($templateCache) => {
+        $templateCache.put('scripts/core/views/confirmation-modal.html', require('./../views/confirmation-modal.html'));
+    }));
+
+    it('does not translate modal header/body', inject((modal, gettextCatalog, $rootScope) => {
```
tomaskikutis: Can you use the test name I suggested? Reading this one it looks like it only applies to modals and that it is supposed to never translate text, when in fact it only applies to missing translations.

petrjasek: but that's the idea, it should never translate it in the component, because that prevents rendering. it should be translated before like when calling that modal service:
```
modal.confirm(gettext('foo'), gettext('bar'));
```
and not expect modal to handle translation

tomaskikutis: Oh, I get it now. You are testing the removal of ` | translate`. If there are cases not worth writing tests for this is the one in my opinion. The only way this can fail is if someone adds `| translate` back. There are probably hundreds of places where adding a translate filter would break it the same way while the test is only addressing this one. I think the test should be removed.

I would suggest modifying the translate filter instead so it returns the original string if no translations are found, but using the `| translate` filter is a bad practice^1 in the first place and we are slowly getting rid of it anyway. At least I ask to remove it if I see it in the PRs.

I was thinking about writing a simple "regex linter" which would count the occurrences of `| translate` and would fail if more are found than in the previous commit(or simply a hardcoded value) making adding more impossible.

^1 it's used for translating variables which shouldn't be done since the translations can't be extracted that way.

petrjasek: that's a regression test, because it's not so obvious it would break anything if it's added again

tomaskikutis: > it should never translate it in the component, because that prevents rendering. it should be translated before like when calling that modal service

Could you copy this comment to the code so it's easier to understand? To clear it up, I believe it doesn't prevent rendering, but renders an empty string.

I've also modified a grep-lint feature I had in the stash for a while, so it shouldn't be that easy to add use translate filters. https://github.com/superdesk/superdesk-client-core/pull/2863

spec/helpers/modal.js
petrjasek: not anymore

PR #2856, opened 2019-02-26
Add a description text to a URL field (SDBELGA-63)

GyanP:
https://dev.sourcefabric.org/browse/SDBELGA-63

### DIFF
--- a/scripts/apps/archive/views/preview.html
+++ b/scripts/apps/archive/views/preview.html
@@ -139,7 +139,9 @@
               <div class="text body-text" ng-if="field.field_type === 'text'" sd-html-preview="selected.preview.extra[field._id]"></div>
               
               <ul class="text body-text" ng-if="field.field_type === 'urls'">
-                <li ng-repeat="url in selected.preview.extra[field._id] track by $index">{{url}}</li>
+                <li ng-repeat="item in selected.preview.extra[field._id] track by $index">{{item.url}}<br>
+                  <p class="nav-space description-text">{{item.description}}</p>
+                </li>
               </ul>
             </div>
 
--- a/scripts/apps/archive/views/preview.html
+++ b/scripts/apps/archive/views/preview.html
@@ -140,7 +140,7 @@
               
               <ul class="text body-text" ng-if="field.field_type === 'urls'">
                 <li ng-repeat="item in selected.preview.extra[field._id] track by $index">{{item.url}}<br>
-                  <p class="nav-space description-text">{{item.description}}</p>
+                  <p class="nav-space description-text">{{item.description | translate}}</p>
                 </li>
               </ul>
             </div>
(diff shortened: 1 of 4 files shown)


scripts/apps/authoring/authoring/article-url-fields.tsx
```diff
@@ -31,18 +32,17 @@ export class ArticleUrlFields extends React.Component<IProps, IState> {
...
     addUrl() {
         this.setState({
-            urls: this.state.urls.concat('https://'),
+            urls: this.state.urls.concat({url: 'https://', description: ''}),
         }, () => {
             this.props.onChange(this.props.fieldId, this.state.urls);
         });
     }
-    handleChange(index, event) {
+    handleChange(index, key, event) {
```
tomaskikutis: `field` instead of `key` would sound clearer to me. You could also add a typing `field: keyof IUrl`

scripts/apps/authoring/authoring/article-url-fields.tsx
```diff
@@ -31,18 +32,17 @@ export class ArticleUrlFields extends React.Component<IProps, IState> {
...
         const nextUrls = this.state.urls.map((currentValue, i) => {
             if (i === index) {
-                return event.target.value;
-            } else {
-                return currentValue;
+                currentValue[key] = event.target.value;
```
tomaskikutis: Use immutable updates when changing React state.

```
const nextUrls = this.state.urls.map((currentValue, i) => {
    if (i === index) {
        return {...currentValue, [key]: event.target.value};
    } else {
        return currentValue;
    }
});
```

scripts/apps/authoring/authoring/article-url-fields.tsx
```diff
@@ -52,23 +52,35 @@ export class ArticleUrlFields extends React.Component<IProps, IState> {
...
+                                onChange={this.handleChange.bind(this, i, 'url')} />
+                            <button disabled={!editable} type="button" onClick={() => this.removeUrl(i)}>
+                                <i className="icon-remove-sign" style={{display: 'block'}} />
+                            </button>
+                        </div>
+                        <textarea type="text"
+                            style={{marginTop: 0.5 + 'em'}}
+                            disabled={!editable}
+                            className="sd-editor__default-input"
+                            placeholder="Description"
```
tomaskikutis: translate the string

GyanP: done

scripts/apps/archive/views/preview.html
```diff
@@ -140,7 +140,7 @@
               
               <ul class="text body-text" ng-if="field.field_type === 'urls'">
                 <li ng-repeat="item in selected.preview.extra[field._id] track by $index">{{item.url}}<br>
-                  <p class="nav-space description-text">{{item.description}}</p>
```
tomaskikutis: You can't, translate dynamic values. This is user-entered content, so you don't need to translate it in the first place.

GyanP: done

tomaskikutis: You don't need to comment "done". I get an email when you push a commit :)

PR #2859, opened 2019-02-28
feat(search): Add support for search & replace with tables on editor3…

ioanpocol:
… (SDESK-4017, SDESK-3888, SDANSA-222)

### DIFF
--- a/scripts/core/editor3/reducers/find-replace.tsx
+++ b/scripts/core/editor3/reducers/find-replace.tsx
@@ -189,13 +189,12 @@ const render = (state) => {
     let changedContent = false;
 
     let {content, editorState} = clearHighlights(es.getCurrentContent(), es);
-    let newContent = content;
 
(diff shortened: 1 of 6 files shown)


scripts/core/editor3/reducers/find-replace.tsx
```diff
@@ -174,22 +186,27 @@ const render = (state) => {
...
+    let newContent = content;
 
     if (!pattern) {
         return {...state, editorState};
     }
 
-    const changedContent = forEachMatch(content, pattern, caseSensitive, (i, selection) => {
-        content = Modifier.applyInlineStyle(
-            content,
+    newContent = forEachMatch(content, pattern, caseSensitive, (i, selection, block, _newContent) => {
```
pablopunk: I'd put here `const newContent` and remove the above `let newContent = content` as that value is not being used at all

PR #2861, opened 2019-03-01
fixed: Templates which are not associated to any desk show up in all …

GyanP:
…of them (SDESK-4044)

### DIFF
--- a/scripts/apps/templates/directives/TemplateSelectDirective.ts
+++ b/scripts/apps/templates/directives/TemplateSelectDirective.ts
@@ -43,10 +43,10 @@ export function TemplateSelectDirective(api, desks, session, templates, notify) 
                             scope.publicTemplates = [];
                             scope.privateTemplates = [];
                             result._items.forEach((template) => {
-                                if (template.is_public !== false) {
-                                    scope.publicTemplates.push(template);
-                                } else {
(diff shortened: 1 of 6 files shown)


scripts/apps/templates/directives/TemplateSelectDirective.ts
```diff
@@ -43,10 +43,10 @@ export function TemplateSelectDirective(api, desks, session, templates, notify)
                             scope.publicTemplates = [];
                             scope.privateTemplates = [];
                             result._items.forEach((template) => {
-                                if (template.is_public !== false) {
-                                    scope.publicTemplates.push(template);
-                                } else {
+                                if (template.is_public === false) {
                                     scope.privateTemplates.push(template);
+                                } else if (template.template_desks && template.template_desks.length > 0) {
```
tomaskikutis: What is this condition for?

GyanP: If the template is not assigned to any desk, then it shouldn't be shown when creating a new item.
That's what the ticket is all about.

tomaskikutis: Got it, thanks. I didn't notice you were already getting the templates that are only accessible to the desk & user.

PR #2862, opened 2019-03-04
feat(search): Extend search service for custom queries

pablopunk:
SDESK-3929

Extend search service to allow custom elasticsearch queries

### DIFF
--- a/scripts/apps/search/services/SearchService.js
+++ b/scripts/apps/search/services/SearchService.js
@@ -509,7 +509,7 @@ export function SearchService($location, config, session, multi,
         this.filter({not: {term: {package_type: 'takes'}}});
 
         // inject custom elastic search queries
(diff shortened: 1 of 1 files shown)


scripts/apps/search/services/SearchService.js
```diff
@@ -508,6 +508,13 @@ export function SearchService($location, config, session, multi,
         // this is needed for archived collection
         this.filter({not: {term: {package_type: 'takes'}}});
 
+        // inject custom elastic search queries
+        if (typeof params.customQueries !== 'undefined') {
```
tomaskikutis: I would use `!= null` (with single `=`) which checks against both `null` and `undefined`

MarkLark86: The `typeof` operator does not return null. `typeof null === 'object'`

PR #2865, opened 2019-03-06
feat(monitoring): add associated item type icons to list

petrjasek:
SDBELGA-36

### DIFF
--- a/scripts/apps/search/components/fields/associatedItems.tsx
+++ b/scripts/apps/search/components/fields/associatedItems.tsx
@@ -1,0 +1,34 @@
+import React from 'react';
+import {IArticle} from 'superdesk-interfaces/Article';
+
+const TYPES_MAPPING = {
+    picture: 'photo',
+    graphic: 'graphic',
+    video: 'video',
+    audio: 'audio',
+};
+
+/**
+ * Render icon with count for each item type present in associations.
+ */
+export function associatedItems({item}: {item: IArticle}) {
+    const associations = Object.values(item.associations || {});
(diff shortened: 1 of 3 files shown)


scripts/apps/search/components/fields/associatedItems.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import {IArticle} from 'superdesk-interfaces/Article';
+
+const TYPES_MAPPING = {
```
tomaskikutis: Use clearer naming like `TypeToIconMap`. It's not clear now what does it map to.

scripts/apps/search/components/fields/associatedItems.tsx
```diff
@@ -1,33 +1,33 @@
...
 /**
  * Render icon with count for each item type present in associations.
  */
 export function associatedItems({item}: {item: IArticle}) {
     const associations = Object.values(item.associations || {});
-    const icons = Object.keys(TYPES_MAPPING).map((type) => ({
+    const icons = Object.keys(TYPES_TO_ICONS).map((type) => ({
         type: type,
-        icon: TYPES_MAPPING[type],
+        icon: TYPES_TO_ICONS[type],
```
tomaskikutis: Will it work given the entire classname? Only the part after `icon-` used to be passed before.

petrjasek: yep the keys didn't change, just values

tomaskikutis: Right, I didn't notice that `icon` value is not passed anywhere outside and only used in this file.

PR #2870, opened 2019-03-11
feat(content profile): As a user I want to have the slugline in the c…

GyanP:
…ontent article area (SDBELGA-65)


scripts/apps/authoring/views/authoring-header.html
```diff
@@ -74,7 +74,7 @@
             <div class="authoring-header__item"
                  sd-validation-error="error.slugline"
                  data-required="schema.slugline.required"
-                 ng-if="schema.slugline"
+                 ng-if="schema.slugline && editor['slugline'].section === 'header'"
```
petrjasek: imo would be better like `&& editor['slugline'].section !== 'content'` to be backward compatible

PR #2875, opened 2019-03-12
change(all): Move all remaining JS files to Typescript

pablopunk:
SDESK-3559

So, this is a huge PR so I'll explain here the basic issues I had to solve:

- All ES6 classes needed implicit attributes declaration in order to access them like `this.key` so there are **a lot** of `key: any;` after class declaration. Ideally we would declare the real type here instead of any. I did it for obvious cases but there are just too many cases to take care of in this PR, so this should be updated by any developer whenever they use an attribute that can be type checked.
- This was also the case for variables declared like `var o = {a: 1}` and then appending an item `o.b = 2`. This was fixed by `var o: any = {a: 1}`.
- The last issue were also happening in some function parameters for example in `.map(item => item.name)` name was not recognized, so it had to be declared as `(item: any)`.
- Some variables that could not be declared as any (for example overriding a JS global object like RegExp) were causing the error `RegExp.customFunc: invalid property 'customFunc' in...`. In these few cases there's a simple workaround `RegExp['customFunc'] = ...`.
- I updated draft-js types module as it had a bug related to a ContentState method.
- I added types for webpack-env as they were needed to avoid errors like `require.context: no key 'context' on type NodeRequire`.
- I disabled lib check on tsconfig as it was throwing a lot of errors inside `node_modules/` after installing with yarn (not with npm, for some reason).
- A lot of optional parameters were declared as optional (`?`).
- A lot of functions were using more parameters than required, usually something like `func(foo, bar, true)` when the function declaration was only expecting 2 parameters. This was fixed by removing those parameters that probably were there in the past and were not updated when refactoring the function.
- Lots of `import _ from 'lodash'` and `import moment from 'moment'`
- Allowed `json` and `html` to be imported as a module with `import`.
- Added some more global variables for tests like `afterEach`.


scripts/apps/archive/controllers/BaseListController.ts
```diff
@@ -55,12 +62,13 @@ export class BaseListController {
...
     }
 
-    fetchItems(criteria) {
+    fetchItems(criteria, next?) {
         console.warn('No API defined.');
     }
 
     fetchItem(id) {
         console.warn('no api defined');
+        return new Promise((r) => r());
```
petrjasek: that's like `return Promise.resolve()`; which is imo more readable

pablopunk: @petrjasek yeah agree, I didn't know you could do that actually

scripts/core/filters/index.ts
```diff
@@ -1,9 +1,7 @@
+import _ from 'lodash';
 import {stripHtmlTags} from '../utils';
 
 export default angular.module('superdesk.core.filters', [])
-    .filter('any', () => function(data, key) {
```
tomaskikutis: Why did you remove this?

scripts/core/helpers/uuid.ts
```diff
@@ -8,13 +9,13 @@ export function uuid(a) {
         >> a / 4 // 8 to 11
         ).toString(16) // in hexadecimal
         : ( // or otherwise a concatenated string:
-            [1e7] + // 10000000 +
```
tomaskikutis: Did you really mean to change Array<number> to a string?

scripts/core/services/preferencesService.ts
```diff
@@ -89,7 +89,7 @@ export default angular.module('superdesk.core.preferences', ['superdesk.core.not
                         if ('Notification' in window && Notification.permission !== 'denied') {
                             Notification.requestPermission((permission) => {
                                 if (permission === 'granted') {
-                                    new Notification(msg);
+                                    return new Notification(msg);
```
tomaskikutis: Why have you added the return?

scripts/globals.d.ts
```diff
@@ -37,8 +44,14 @@ interface Window {
     webkitURL: any;
     superdeskConfig: any;
     module: any;
+    RunTansaProofing: any;
+    iframely: any;
 }
 
+// Allow importing json/html files
+declare module "*.json";
```
tomaskikutis: Where is that used? I don't think importing json/html was supported.

scripts/apps/archive/controllers/ArchiveListController.ts
```diff
@@ -64,7 +64,7 @@ export class ArchiveListController extends BaseListController {
             $scope.loading = true;
             criteria.aggregations = 1;
             criteria.es_highlight = search.getElasticHighlight();
-            resource.query(criteria).then((items) => {
+            return resource.query(criteria).then((items) => {
```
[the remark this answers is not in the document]
pablopunk: I think I did this to prevent some `tsc` error but I can't see it anymore, so I'll revert it

scripts/core/editor3/helpers/insertAtomicBlockWithoutEmptyLines.ts
```diff
@@ -88,7 +88,7 @@ function insertAtomicBlockWithoutEmptyLines(editorState, entityKey, character) {
 
     var withAtomicBlock = Modifier.replaceWithFragment(asAtomicBlock, target.selectionState, fragment);
 
-    var newContent = withAtomicBlock.merge({
+    var newContent: ContentState|any = withAtomicBlock.merge({
```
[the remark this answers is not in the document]
pablopunk: for some reason, the method `merge` returns a Map instead of a ContentState but it does act like a ContentState (see next lines, it is being used as a cs)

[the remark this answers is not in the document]
pablopunk: so, the compiler didn't allow the map to be assigned to a content state and it didn't recognize the following methods used

scripts/core/filters/index.ts
pablopunk: It was using `_.any` which is deprecated and I search for usages of the filter "any" in superdesk and couldn't find one

scripts/core/helpers/uuid.ts
pablopunk: Yeah it was actually doing `[1e7] + -1e3` which is basically converting both to string and concatenating them. The result was `'10000000-1000'`, the same as `` `${1e7}` + -1e3``. I changed it because the compiler was complaining about types.

scripts/core/services/preferencesService.ts
pablopunk: It was complaining about the value not being used, so I "used" it. It's not complaining now for some reason so I'll change it back.

scripts/globals.d.ts
pablopunk: For example here https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/html/tests/from-html-snapshot.spec.jsx#L1-L2

scripts/core/services/preferencesService.ts
tomaskikutis: Yeah, it may look like it to a static analyzer, because you are creating a class instance and not using it. It's not obvious it's doing side effects in the constructor.

scripts/globals.d.ts
tomaskikutis: Yeah, I added that test, but it's been disabled for ages. Importing json/html is pretty handy for snapshot testing, so let's keep it.

scripts/core/services/preferencesService.ts
pablopunk: I was wrong, it was `tslint` complaining about it, I put it back

tomaskikutis: I think adding a ignore for tslint is better. It's harder to change code when there are returns which's values aren't being used but you still need to check all the places calling it if you are refactoring something.

scripts/core/editor3/helpers/insertAtomicBlockWithoutEmptyLines.ts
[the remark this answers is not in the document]
tomaskikutis: You can do `withAtomicBlock.merge() as ContentState` and the compiler won't complain.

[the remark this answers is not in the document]
tomaskikutis: Merge returns a map because it's a method of immutablejs not draft-js, but I think it's compatible.

package.json
```diff
@@ -47,8 +47,10 @@
     "angular-resource": "1.6.9",
     "angular-route": "1.6.9",
     "angular-vs-repeat": "1.1.7",
+    "body-parser": "1.18.3",
```
tomaskikutis: I forgot to mention this in previous comments. Can you remove 'body-parser' and 'cors' from package.json? They aren't really used anywhere but in a test server which is itself is only for reference, so you can comment the entire file out if typescript is complaining about these dependencies.

pablopunk: okay I'll just put the example in javascript instead as it is not using any typescript feature

PR #2876, opened 2019-03-12
fix(fields): fix missing related content items as associations

petrjasek:
those are stored as ids only, so it must be fetched to get the type

SDBELGA-73

### DIFF
--- a/scripts/apps/search/components/fields/AssociatedItemsList.tsx
+++ b/scripts/apps/search/components/fields/AssociatedItemsList.tsx
@@ -1,0 +1,58 @@
+import React from 'react';
+import {IArticle} from 'superdesk-interfaces/Article';
+
+const TYPES_TO_ICONS = {
+    picture: 'icon-photo',
+    graphic: 'icon-graphic',
+    video: 'icon-video',
+    audio: 'icon-audio',
+};
+
+interface IProps {
+    svc: any;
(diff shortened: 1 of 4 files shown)


scripts/apps/search/components/fields/AssociatedItemsList.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+interface IState {
+    associations: Array<IArticle>;
+}
+
+/**
+ * Render icon with count for each item type present in associations.
+ */
+export class AssociatedItemsList extends React.Component<IProps, IState> {
+    readonly state = {associations: []};
```
tomaskikutis: That's an interesting way to initialize state. :+1:

scripts/apps/workspace/content/services/ContentService.js
```diff
@@ -336,6 +336,20 @@ export function ContentService(api, superdesk, templates, desks, packages, archi
 
     this.getCustomFields = getCustomFields;
 
+    this.fetchAssociations = (item) => {
+        const associations = item.associations || {};
+        const keys = Object.keys(associations);
+
+        return Promise.all(keys.map((key) => {
+            // there is only _id, maybe _type for related items
```
tomaskikutis: It should be _id only.

PR #2880, opened 2019-03-14
[SDANSA-240] feat(metadata): Created directive for copy&paste metadata

pavlovicnemanja:


scripts/apps/authoring/media/MediaCopyMetadataDirective.ts
```diff
@@ -0,0 +1,30 @@
+MediaCopyMetadataDirective.$inject = [];
+export default function MediaCopyMetadataDirective() {
```
tomaskikutis: Make it into a React component and use `reactToAngular1` to display it inside angular

pavlovicnemanja: Important for Ansa to be merged and deployed, don't want to do twice same job. We can refactor this with all other components we create. I will develop new directives as React Components, but this is already done and I really don't want to go back with it.

PR #2885, opened 2019-03-18
fixed: Canceling changes in media metadata does not reset the usage t…

GyanP:
…erms field. (SDESK-4079)

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
@@ -217,16 +217,7 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, deploy
     * @description
     */
     $scope.cancelMetadataChanges = () => {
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -217,6 +217,16 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, deploy
     * @description
     */
     $scope.cancelMetadataChanges = () => {
+        const selectedUsageTerms = [];
```
tomaskikutis: The function which is canceling changes shouldn't need to know about specific fields. Changes should be kept in one object and resetting that object should be enough. Otherwise, it will continuously cause bugs as fields change.

petrjasek:
hey @GyanP you don't have to add that merged label, I only use it when backporting something to older release

PR #2886, opened 2019-03-19
fixed: When editing raw search, if the query is erased totally then t…

GyanP:
…he field disappears. (SDESK-4098)

### DIFF
--- a/scripts/apps/search/directives/SaveSearch.ts
+++ b/scripts/apps/search/directives/SaveSearch.ts
@@ -39,6 +39,12 @@ export function SaveSearch($location, asset, api, session, notify, $rootScope) {
                 scope.edit = null;
                 scope.activateSearchPane = false;
             };
+            scope.isValid = function(edit) {
+                if (edit.filter.query.raw == null) {
+                    return !!edit.name;
+                }
+                return edit.filter.query && edit.filter.query.raw !== '' && edit.name;
+            };
 
             scope.clear = function() {
                 scope.editingSearch = false;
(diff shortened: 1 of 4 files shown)


scripts/apps/search/directives/SaveSearch.ts
```diff
@@ -39,6 +39,12 @@ export function SaveSearch($location, asset, api, session, notify, $rootScope) {
                 scope.edit = null;
                 scope.activateSearchPane = false;
             };
+            scope.isValid = function(edit) {
+                if (edit.filter.query.raw == null) {
+                    return !!edit.name;
```
tomaskikutis: It's easier to understand when you are more explicit. I assume the reason you are trying to convert a string to bool is to check if the string is empty. `typeof edit.name === 'string' && edit.name.length > 0` does exactly that and doesn't hide your intention :) The same applies to a line below.

scripts/apps/search/views/save-search-dialog.html
```diff
@@ -1,6 +1,6 @@
 <!-- <form name="viewsForm"> -->
     <fieldset>
-        <div class="field" ng-if="edit.filter.query.raw">
+        <div class="field" ng-hide="{{edit.filter.query.raw === undefined}}">
```
tomaskikutis: `== null` covers for both `null` and `undefined`. Use `== null` if there is no special meaning when the value is null.

GyanP: I have to hide it in case of undefined only, that's why I didn't use `== null`

scripts/apps/search/directives/SaveSearch.ts
```diff
@@ -39,6 +39,13 @@ export function SaveSearch($location, asset, api, session, notify, $rootScope) {
                 scope.edit = null;
                 scope.activateSearchPane = false;
             };
+            scope.isValid = function(edit) {
+                if (edit.filter.query.raw == null) {
+                    return typeof edit.name === 'string' && edit.name.length > 0;
+                }
+                return edit.filter.query && typeof edit.filter.query.raw === 'string'
```
petrjasek: imo it could use some helper func for the non empty string check, it's used 3x already

PR #2887, opened 2019-03-19
fix(shortcuts): Changed behaviour of shortcuts for widgets

pavlovicnemanja:


scripts/apps/authoring/widgets/views/authoring-widgets.html
```diff
@@ -11,8 +11,8 @@
                     ng-click="activate(widget)"
                     ng-disabled="isWidgetLocked(widget)"
                     id="{{ :: widget._id }}"
-                    sd-tooltip="{{ :: widget.label | translate}} (ctrl+alt+{{widget.order}})" flow="left"
-                    sd-hotkey="ctrl+alt+{{widget.order}}"
+                    sd-tooltip="{{ :: widget.label | translate}} ({{generateHotkey(widget.order)}})" flow="left"
```
petrjasek: can you use rather `$index + 1` instead of `widget.order` ?
it would be better if we set orders with some space like: 10, 20, 30... for internal widgets so we could position something in between on client instances

pavlovicnemanja: Yep, for sure!

PR #2888, opened 2019-03-19
Feat(template): As a user I want to save values in custom text fields…

GyanP:
… in a template. (SDBELGA-77)

### DIFF
--- a/scripts/core/editor3/directive.js
+++ b/scripts/core/editor3/directive.js
@@ -160,6 +160,7 @@ class Editor3Directive {
         this.disableSpellchecker = this.disableSpellchecker || false;
         this.bindToValue = this.bindToValue || false;
         this.tabindex = this.tabindex || 0;
+        this.refreshTrigger = this.refreshTrigger || 0;
         this.showTitle = this.showTitle || false;
         this.$rootScope = $rootScope;
         this.$scope = $scope;
--- a/scripts/core/editor3/helpers/fieldsMeta.ts
+++ b/scripts/core/editor3/helpers/fieldsMeta.ts
@@ -39,8 +39,10 @@ export function setFieldMetadata(item, fieldKey, contentKey, contentValue) {
         throw new Error('Invalid key');
     }
 
-    item[META_FIELD_NAME][fieldKey] = {
-        ...item[META_FIELD_NAME][fieldKey],
-        [contentKey]: wrapper.wrap(contentValue),
-    };
+    if (item[META_FIELD_NAME]) {
+        item[META_FIELD_NAME][fieldKey] = {
+            ...item[META_FIELD_NAME][fieldKey],
+            [contentKey]: wrapper.wrap(contentValue),
+        };
+    }
 }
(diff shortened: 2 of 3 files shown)


scripts/core/editor3/helpers/fieldsMeta.ts
```diff
@@ -39,8 +39,10 @@ export function setFieldMetadata(item, fieldKey, contentKey, contentValue) {
         throw new Error('Invalid key');
     }
 
-    item[META_FIELD_NAME][fieldKey] = {
-        ...item[META_FIELD_NAME][fieldKey],
-        [contentKey]: wrapper.wrap(contentValue),
-    };
+    if (item[META_FIELD_NAME]) {
```
tomaskikutis: Why do you need the `if`? Set a default empty object if you really need to introduce a conditional.

scripts/core/editor3/directive.js
```diff
@@ -160,6 +160,7 @@ class Editor3Directive {
         this.disableSpellchecker = this.disableSpellchecker || false;
         this.bindToValue = this.bindToValue || false;
         this.tabindex = this.tabindex || 0;
+        this.refreshTrigger = this.refreshTrigger || 0;
```
tomaskikutis: Why do you need this?

GyanP: When we edit the template the value of `refreshTrigger` is undefined.

scripts/core/editor3/helpers/fieldsMeta.ts
GyanP: `item` is the one we are currently working on, so if the `META_FIELD_NAME` is not in `item`, it throws the error "cannot read property `fieldKey` of undefined".

tomaskikutis: :+1: add an else and initialize empty object then.

GyanP: Which variable?

tomaskikutis: Before your change `item[META_FIELD_NAME][fieldKey]` used to **always** be an object. If you don't have data to populate it, make it an empty object instead of it being `undefined`.

GyanP: It's not because of my changes if you go on to edit the editor3 templates in sd-master and type something in headline or bodyline field you will get that error and save button won't get enabled.
I don't know what's wrong but I didn't see that `META_FIELD_NAME=fields_meta` in any of templates.

tomaskikutis: I've looked at it again and I think we should set metadata even if `item[META_FIELD_NAME]` is null. It simply needs to be initialized.

```
export function setFieldMetadata(item, fieldKey, contentKey, contentValue) {
    if (Object.keys(fieldsMetaKeys).includes(contentKey) === false) {
        throw new Error('Invalid key');
    }

    if (item[META_FIELD_NAME] == null) {
        item[META_FIELD_NAME] = {};
    }

    item[META_FIELD_NAME][fieldKey] = {
        ...item[META_FIELD_NAME][fieldKey],
        [contentKey]: wrapper.wrap(contentValue),
    };
}
```

PR #2889, opened 2019-03-19
fix(monitoring): update swimlane markup to fix scrolling

petrjasek:
SDESK-4100

### DIFF
--- a/scripts/apps/monitoring/styles/monitoring.scss
+++ b/scripts/apps/monitoring/styles/monitoring.scss
@@ -287,18 +287,15 @@
     }
 }
 
-.swimlane2 {
-    display: flex;
-    justify-content: space-between;
+.sd-kanban-list-container {
+    height: 100%;
+}
 
-    > * {
-        flex-basis: 0;
-        flex-grow: 1;
-        min-width: 250px;
-    }
+.sd-kanban-list {
+    overflow: auto;
 
-    > * + * {
-        margin-left: 15px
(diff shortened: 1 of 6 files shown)


scripts/apps/monitoring/styles/monitoring.scss
```diff
@@ -287,18 +287,15 @@
...
-    display: flex;
-    justify-content: space-between;
-
-    > * {
-        flex-basis: 0;
-        flex-grow: 1;
-        min-width: 250px;
-    }
+.sd-kanban-list {
+    position: absolute;
```
tomaskikutis: Why do you need to use absolute positioning? The previous implementation I've done worked fine without it. I think scrollable contents and static labels can be done with flexbox, can't it?

petrjasek: seems like it can

scripts/apps/monitoring/styles/monitoring.scss
```diff
@@ -287,18 +287,15 @@
...
-        flex-basis: 0;
-        flex-grow: 1;
-        min-width: 250px;
-    }
+.sd-kanban-list {
+    overflow: auto;
 
-    > * + * {
-        margin-left: 15px
+    .fix-stage { // there should be no div for .stage, but it's needed for list view
```
tomaskikutis: `display: contents` might also work.

PR #2890, opened 2019-03-19
Add rules for disallowing adding more angularjs templates.

tomaskikutis:
Don't use shell=True in grep-lint.  Quotes aren't escaped properly when it's used.

### DIFF
--- a/grep-lint.py
+++ b/grep-lint.py
@@ -4,7 +4,7 @@ import sys
 def get_command(branch=None):
     arguments_list = ["git", "grep", "-P", rule_regex]
 
-    if(branch is not None):
+    if branch is not None:
         arguments_list.append(branch)
     
     arguments_list.append("--")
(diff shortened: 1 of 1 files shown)


grep-lint.py
```diff
@@ -2,38 +2,83 @@
 import sys
 
 def get_command(branch=None):
-    branch_arg = '' if branch is None else f" {branch}"
+    arguments_list = ["git", "grep", "-P", rule_regex]
 
-    # git grep returns an error code when no matches are found
-    # and subprocess.check_output throws an error when an error code is returned
-    ignore_error = ' || true'
+    if(branch is not None):
```
petrjasek: more pythonic is `if branch is not None:`

grep-lint.py
```diff
@@ -2,38 +2,83 @@
...
-    violations_count_master = len(
-        subprocess.check_output(get_command("master"), shell=True, stderr=subprocess.STDOUT).decode('utf-8').splitlines()
-    )
+    try:
+        violations_count_master = len(
+            subprocess.check_output(get_command("master")).decode('utf-8').splitlines()
+        )
+    except subprocess.CalledProcessError as e:
+        # ignore exception if grep simply didn't find matches
+        if(len(e.output) is not 0):
```
petrjasek: pythonic `if len(e.output) != 0`

PR #2898, opened 2019-03-27
Fix tansa load.

ride90:
This one fixes NTB instance on my local. 
I don't see a point why we are using `setTimeout` here❓ 
A link in comment points to a simple usage of `setTimeout` .

Created NTB instance:
https://ntbpr-310.test.superdesk.org

### DIFF
--- a/scripts/index.ts
+++ b/scripts/index.ts
@@ -9,8 +9,7 @@ import 'external-apps';
 const appConfig = __SUPERDESK_CONFIG__;
 
 if (appConfig.features.useTansaProofing) {
-    // see: http://www.hiddenwebgenius.com/blog/guides/understanding-javascripts-asynchronous-code/
-    setTimeout(() => require('apps/tansa'), 0);
+    require('apps/tansa');
 }
 
 let body = angular.element('body');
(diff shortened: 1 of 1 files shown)


tomaskikutis:
I was looking into the issue as well. If you check the history of the comment, you'll find the related PR. https://github.com/superdesk/superdesk-client-core/pull/1845 Ioan was trying to delay script execution so it doesn't block the UI. When porting to typescript, Pablo modified the line, and added another function which delayed execution even more, so the function fired after DOM has already loaded. The tansa script uses `document.write` which deletes the entire content of the page if called after document load - that's exactly what was happening in our case.

I suggest to simply undo Pablo's change and add a tslint ignore line so it works as before and includes Ioans change to not block the UI.

screenshot_5

ride90:
Hi @tomaskikutis ,
Yes, I saw Nelu's changes, but I don't think that it ever worked as expected. 

I'm not 100% sure, but 
```
setTimeout(require('apps/tansa'), 0);
``` 
is equal to:
```
let res = require('apps/tansa');
setTimeout(res, 0);
```
setTimeout takes a function as a first argument, not a result of it. Pablo's changes made it work as expected :)

PR #2899, opened 2019-03-28
fix find-replace in editor3

petrjasek:
when there are multiple highlights render them all together and not
one by one to avoid partial matches.

SDESK-4017


tomaskikutis:
What was the issue with partial matches? Why is it important if you render matches one by one or all at once? How does that cause partial matches?

petrjasek:
the issue was that if you would have both `$500` and `$50` in the text, first it would highlight `$500` and that's fine, but in the second step it was looking for `$50` and it would highlight it also in `$500`

PR #2901, opened 2019-03-28
fix(profiles): hide field from both sections when used

petrjasek:
SDESK-3932


petrjasek:
replaces https://github.com/superdesk/superdesk-client-core/pull/2871

PR #2903, opened 2019-04-01
Add 'Edit image metadata' action to image items in the monitoring lis…

GyanP:
…t (SDANSA-242)

### DIFF
--- a/scripts/apps/authoring/authoring/index.ts
+++ b/scripts/apps/authoring/authoring/index.ts
@@ -140,6 +140,27 @@ angular.module('superdesk.apps.authoring', [
                     return authoring.itemActions(item).edit;
                 }],
             })
+            .activity('edit.media.metadata', {
+                label: gettext('Edit Media Metadata'),
+                priority: 3,
+                icon: 'edit-line',
+                keyboardShortcut: 'ctrl+alt+i',
+                controller: ['data', 'multiImageEdit', 'authoring', 'api',
+                    function(data, multiImageEdit, authoring, api) {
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/index.ts
```diff
@@ -140,6 +140,27 @@ angular.module('superdesk.apps.authoring', [
                     return authoring.itemActions(item).edit;
                 }],
             })
+            .activity('edit.media.metadata', {
+                label: gettext('Edit Media Metadata'),
+                priority: 3,
+                icon: 'edit-line',
+                keyboardShortcut: 'ctrl+alt+i',
```
petrjasek: what about `ctrl+alt+m` given that it's about media

GyanP: yes, right. But I there is also a `metasearch` so I thought it could be used for that. What do you think?

petrjasek: should be fine, that's ansa specific, so we can change it for them if needed, but afaik there is no shortcut now for metasearch

petrjasek: my bad, seems like there should be a shortcut (SDANSA-233), so not sure what to do now

GyanP: done

GyanP: @petrjasek I discussed it with Karel, and he says that keep `ctrl+alt+m`  for `edit media metadata`.

PR #2906, opened 2019-04-01
use geonames for dateline if available

petrjasek:
SDANSA-229

### DIFF
--- a/scripts/apps/authoring/metadata/PlacesService.ts
+++ b/scripts/apps/authoring/metadata/PlacesService.ts
@@ -1,35 +1,64 @@
 import {omit} from 'lodash';
 
 interface IGeoname {
+    /** name of the place */
     name: string;
-    code: string;
 
     state: string;
-    country: string;
-
     state_code: string;
+
+    country: string;
     country_code: string;
 
+    /** timezone identifier, eg. Europe/Prague */
     tz: string;
-    scheme: string;
+
+    /** geonames id */
+    code: string;
+
+    scheme: 'geonames';
 }
 
 interface ILocated {
-    dateline: string;
-    country_code: string;
-    city_code: string;
-    tz: string;
-    state_code: string;
-    state: string;
+    /** dateline format - list of fields which should be used to identify the place */
+    dateline: 'city' | 'city,state' | 'city,country' | 'city,state,country';
+
     city: string;
-    alt_name: string;
+    state: string;
     country: string;
-    code: string;
+
+    city_code: string;
+    state_code: string;
+    country_code: string;
+
+    /** timezone identifier, eg. Europe/Prague  */
+    tz: string;
+
+    /** scheme identifier */
     scheme: string;
+
+    /** code for place in the scheme */
+    code: string;
 }
 
 export interface IPlacesService {
+
+    /**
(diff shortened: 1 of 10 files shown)


scripts/apps/authoring/metadata/PlacesService.ts
```diff
@@ -0,0 +1,60 @@
+import {omit} from 'lodash';
+
+export interface IPlacesService {
+    search: (name: string, lang: string, format: 'geonames' | 'dateline') => Promise<Array<{}>>;
```
tomaskikutis: The return types aren't clear in the interface. We would also get much better typing support if it were 2 methods instead:

```
searchGeonames: (name: string, lang: string) => Promise<Array<IGeoName>>;
searchDateline: (name: string, lang: string) => Promise<Array<IDateline>>;
```

Specifying the exact return type is easy and you don't need to do checks to determine the type after you call the function.

scripts/apps/authoring/metadata/PlacesService.ts
```diff
@@ -23,25 +52,39 @@ export default function PlacesServiceFactory(api, features, metadata) {
...
-                    format === 'dateline'
-                        ? this.searchCities(name)
-                        : Promise.reject(), // can't handle geonames via cities
-                );
+        /**
+         * Search for dateline
+         *
+         * it will use geonames if available, cities cv otherwise
+         *
+         * @param name
```
tomaskikutis: Why is `@params` included in the comment? The type signature is already in the interface & implementation.

Can you move the comment to the interface? It's easier to find it that way. If one needs to consume the service and not modify it, reading the implementation shouldn't be required.

Specify what is supported as `name` argument. Can it be a street name, a post code, the coordinates?

scripts/apps/authoring/metadata/PlacesService.ts
```diff
@@ -1,13 +1,42 @@
...
+    name: string;
+    code: string;
+
+    state: string;
+    country: string;
+
+    state_code: string;
+    country_code: string;
+
+    tz: string;
```
tomaskikutis: Add comments for unclear fields. These aren't clear to me:

- `IGeoname['name']` - what does it contain? Full adress, coordinates or something else?
- `IGeoname['code']`
- `IGeoname['tz']`
- `IGeoname['scheme']`

- `ILocated['dateline']`
- `ILocated['tz']`
- `ILocated['alt_name']`
- `ILocated['code']`
- `ILocated['scheme']`

scripts/apps/authoring/metadata/PlacesService.ts
```diff
@@ -1,35 +1,64 @@
 import {omit} from 'lodash';
 
 interface IGeoname {
+    /** name of the place */
```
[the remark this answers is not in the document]
petrjasek: it's just name of town like `Prague`, in the language requested, we don't use addresses there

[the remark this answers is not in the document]
tomaskikutis: Are both `IGeoname` and `ILocated` only return city level info and not district/street? It would be good to mention that in the comment above the interface declaration.

PR #2908, opened 2019-04-01
fixed: can't edit templates with custom text fields in the settings

GyanP:
SDBELGA-77

### DIFF
--- a/scripts/core/editor3/store/index.ts
+++ b/scripts/core/editor3/store/index.ts
@@ -137,9 +137,11 @@ export function onChange(contentState, {plainText = false} = {}) {
     let objectToUpdate = pathToValueArray.length < 2 ?
         this.item :
         pathToValueArray.slice(0, -1).reduce((obj, pathSegment) => {
-            const nextObj = obj[pathSegment];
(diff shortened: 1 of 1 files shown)


scripts/core/editor3/store/index.ts
```diff
@@ -137,9 +137,11 @@ export function onChange(contentState, {plainText = false} = {}) {
     let objectToUpdate = pathToValueArray.length < 2 ?
         this.item :
         pathToValueArray.slice(0, -1).reduce((obj, pathSegment) => {
-            const nextObj = obj[pathSegment];
+            if (obj[pathSegment] == null) {
+                obj[pathSegment] = {};
```
tomaskikutis: What's wrong with null? :slightly_smiling_face:

GyanP: https://github.com/superdesk/superdesk-client-core/blob/ee37f7c142ef115938f40ea479fd7e6ab6658fa5/scripts/core/editor3/store/index.ts#L150
gives cannot read [key] of undefined if there is no `extra` field in a template.

PR #2911, opened 2019-04-03
fix(search): fix list jumping up all the time

petrjasek:
it was scrolling on top when it got results, but that scrolling
to top triggers another refresh, and that triggered another update,
and so on.

now it will only scroll on top when refresh btn is clicked and
starts the query later, so scroll handler won't trigger another load.

SDESK-4146

### DIFF
--- a/scripts/apps/monitoring/directives/MonitoringGroup.ts
+++ b/scripts/apps/monitoring/directives/MonitoringGroup.ts
@@ -310,6 +310,10 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
             scope.refreshGroup = function() {
                 monitoring.showRefresh = scope.showRefresh = false;
                 scheduleQuery(null, {force: true});
+                // update scroll position to top when forced refresh
+                if (containerElem[0].scrollTop > 0) {
+                    containerElem[0].scrollTop = 0;
+                }
             };
 
             function updateGroupStyle() {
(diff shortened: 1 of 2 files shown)


scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -310,6 +310,10 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
             scope.refreshGroup = function() {
                 monitoring.showRefresh = scope.showRefresh = false;
                 scheduleQuery(null, {force: true});
+                // update scroll position to top when forced refresh
+                if (containerElem[0].scrollTop > 0) {
```
tomaskikutis: Why is the conditional used instead of always updating it?

scripts/apps/search/directives/SearchResults.ts
```diff
@@ -303,6 +300,9 @@ export function SearchResults(
             scope.refreshList = function() {
                 scope.showRefresh = false;
                 queryItems(null, {force: true});
+                if (containerElem[0].scrollTop > 0) {
```
tomaskikutis: Why is the conditional used instead of always updating it?

scripts/apps/monitoring/directives/MonitoringGroup.ts
petrjasek: to avoid scroll event which is triggered even if you set it from 0 to 0, we trigger refresh when user scrolls to top.

PR #2918, opened 2019-04-08
Fixed(Custom CVs): 'Single selection' option is not reflected in the …

GyanP:
…editor. (SDESK-4107)

### DIFF
--- a/scripts/apps/authoring/metadata/metadata.ts
+++ b/scripts/apps/authoring/metadata/metadata.ts
@@ -357,10 +357,10 @@ function MetaDropdownDirective($filter) {
         },
         templateUrl: 'scripts/apps/authoring/metadata/views/metadata-dropdown.html',
         link: function(scope, elem) {
-            scope.listFields = ['place', 'genre', 'anpa_category', 'subject', 'authors'];
+            scope.multiInputFields = ['place', 'genre', 'anpa_category', 'subject', 'authors'];
 
             scope.select = function(item) {
-                var o = {};
+                var fieldObject = {};
 
                 if (item) {
                     if (scope.cv) {
@@ -368,19 +368,20 @@ function MetaDropdownDirective($filter) {
                         // so that it can be differentiated from another cv inside same parent field(subject).
                         // ex: subject:[{name: "a", qcode: "a", scheme: "new-cv"}]
                         item.scheme = scope.cv._id;
-                        o[scope.field] = scope.item[scope.field].filter((v) => v.scheme !== scope.cv._id);
-                        o[scope.field].push(item);
+                        fieldObject[scope.field] = scope.item[scope.field].filter((v) => v.scheme !== scope.cv._id);
+                        fieldObject[scope.field].push(item);
                     } else {
-                        o[scope.field] = scope.key ? item[scope.key] : [item];
+                        fieldObject[scope.field] = scope.key ? item[scope.key] : [item];
                     }
                 } else if (scope.cv) {
-                    // if there is cv as well as field don't set o[scope.field] to null.
-                    o[scope.field] = scope.item[scope.field].filter((v) => v.scheme !== scope.cv._id);
+                    // if there is cv as well as field don't set fieldObject[scope.field] to null.
+                    // as in the backend it is set to "nullable: false" for ex: subject.
+                    fieldObject[scope.field] = scope.item[scope.field].filter((v) => v.scheme !== scope.cv._id);
                 } else {
-                    o[scope.field] = null;
+                    fieldObject[scope.field] = null;
                 }
 
-                _.extend(scope.item, o);
+                _.extend(scope.item, fieldObject);
                 scope.change({item: scope.item, field: scope.field});
 
                 // retain focus on same dropdown control after selection.
@@ -389,7 +390,7 @@ function MetaDropdownDirective($filter) {
                 });
 
                 if (scope.values) {
-                    scope.selected = scope.values[o[scope.field]] || null;
+                    scope.selected = scope.values[fieldObject[scope.field]] || null;
                 }
             };
 
--- a/scripts/apps/authoring/metadata/metadata.ts
+++ b/scripts/apps/authoring/metadata/metadata.ts
@@ -350,6 +350,7 @@ function MetaDropdownDirective($filter) {
             icon: '@',
             label: '@',
             change: '&',
+            cv: '=',
             key: '@',
             tabindex: '=',
             containingDirective: '@',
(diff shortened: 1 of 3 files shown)


scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -362,7 +363,15 @@ function MetaDropdownDirective($filter) {
                 var o = {};
 
                 if (item) {
-                    o[scope.field] = scope.key ? item[scope.key] : [item];
+                    if (scope.cv) {
```
tomaskikutis: Can you explain what is the code doing? :)

scripts/apps/authoring/metadata/views/metadata-dropdown.html
```diff
@@ -3,10 +3,14 @@
         {{ key && values[item[field]]  && values[item[field]].name ? values[item[field]].name: (item[field].name || item[field] ? item[field].name || item[field] : "none") | translate }}
         <b class="dropdown__caret"></b>
     </button>
-    <button ng-disabled="disabled" class="dropdown__toggle line-input" tabindex="{{tabindex}}" dropdown__toggle ng-if="listFields.indexOf(field) > -1 && !icon">
+    <button ng-disabled="disabled" class="dropdown__toggle line-input" tabindex="{{tabindex}}" dropdown__toggle ng-if="listFields.indexOf(field) > -1 && !icon && !cv">
```
tomaskikutis: It would be useful to comment under which conditions which button is displayed since that's hard to understand.

scripts/apps/authoring/metadata/views/metadata-dropdown.html
```diff
@@ -1,12 +1,15 @@
...
     <button ng-disabled="disabled" class="dropdown__toggle line-input" tabindex="{{tabindex}}" dropdown__toggle ng-if="listFields.indexOf(field) === -1 && !icon">
         {{ key && values[item[field]]  && values[item[field]].name ? values[item[field]].name: (item[field].name || item[field] ? item[field].name || item[field] : "none") | translate }}
         <b class="dropdown__caret"></b>
     </button>
+    <!-- if field is one of 'place', 'genre', 'anpa_category', 'subject', 'authors' and no cv-->
     <button ng-disabled="disabled" class="dropdown__toggle line-input" tabindex="{{tabindex}}" dropdown__toggle ng-if="listFields.indexOf(field) > -1 && !icon && !cv">
         {{ item[field].length > 0 ? item[field][0].name : "none" | translate }}
         <b class="dropdown__caret"></b>
     </button>
+    <!-- if cv stored inside ohter parent field ex: subject:[{name: "a", qcode: "a", scheme: cv._id}]-->
```
tomaskikutis: fix the spelling of **other**

scripts/apps/authoring/metadata/views/metadata-dropdown.html
```diff
@@ -1,12 +1,15 @@
 <div class="dropdown dropdown--noarrow dropdown--dropleft" dropdown sd-metadropdown-focus>
+    <!-- if field is other than 'place', 'genre', 'anpa_category', 'subject', 'authors'-->
```
tomaskikutis: It's better not to list the exact fields, because they might change and get out of sync with this comment. It would be more helpful if you commented on what is exceptional about this group of fields. If I understand correctly from the name `listFields` it probably means that one button is for multi-value inputs and another for single value inputs. It would be useful to rename the variable to `multiInputFields` if that's the case.

scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -364,13 +364,17 @@ function MetaDropdownDirective($filter) {
...
+                        // so that it can be differentiated from another cv inside same parent field(subject).
+                        // ex: subject:[{name: "a", qcode: "a", scheme: "new-cv"}]
                         item.scheme = scope.cv._id;
                         o[scope.field] = scope.item[scope.field].filter((v) => v.scheme !== scope.cv._id);
                         o[scope.field].push(item);
                     } else {
                         o[scope.field] = scope.key ? item[scope.key] : [item];
                     }
                 } else if (scope.cv) {
+                    // if there is cv as well as field don't set o[scope.field] to null.
```
tomaskikutis: What is `o`? I know it's not you who added, but renaming to a more meaningful name would help.

> don't set o[scope.field] to null.

What does setting to null do?

GyanP: some of the fields like `subject` are set as `nullable: false` in the backend.

tomaskikutis: Add it to the comment in the code :)

scripts/apps/authoring/metadata/views/metadata-dropdown.html
```diff
@@ -1,16 +1,16 @@
...
-    <!-- if field is other than 'place', 'genre', 'anpa_category', 'subject', 'authors'-->
-    <button ng-disabled="disabled" class="dropdown__toggle line-input" tabindex="{{tabindex}}" dropdown__toggle ng-if="listFields.indexOf(field) === -1 && !icon">
+    <!-- if field is single input field-->
+    <button ng-disabled="disabled" class="dropdown__toggle line-input" tabindex="{{tabindex}}" dropdown__toggle ng-if="multiInputFields.indexOf(field) === -1 && !icon">
         {{ key && values[item[field]]  && values[item[field]].name ? values[item[field]].name: (item[field].name || item[field] ? item[field].name || item[field] : "none") | translate }}
         <b class="dropdown__caret"></b>
     </button>
-    <!-- if field is one of 'place', 'genre', 'anpa_category', 'subject', 'authors' and no cv-->
-    <button ng-disabled="disabled" class="dropdown__toggle line-input" tabindex="{{tabindex}}" dropdown__toggle ng-if="listFields.indexOf(field) > -1 && !icon && !cv">
+    <!-- if field is multi input field and no cv-->
```
tomaskikutis: > and no cv
is not that clear. I'd prefer "and the data source is not a custom vocabulary"

What other data sources are supported, free text input?

scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -357,30 +357,31 @@ function MetaDropdownDirective($filter) {
         },
         templateUrl: 'scripts/apps/authoring/metadata/views/metadata-dropdown.html',
         link: function(scope, elem) {
-            scope.listFields = ['place', 'genre', 'anpa_category', 'subject', 'authors'];
+            scope.multiInputFields = ['place', 'genre', 'anpa_category', 'subject', 'authors'];
 
             scope.select = function(item) {
-                var o = {};
+                var fieldObject = {};
```
tomaskikutis: What does the object contain/is used for? Could you add types to it?

GyanP: It contains data like this
custom-cv.

scripts/apps/authoring/metadata/views/metadata-dropdown.html
GyanP: It supports custom vocabulary only. But there are some CVS which are stored inside multiInputFields and some create their own field in the item.

scripts/apps/authoring/metadata/metadata.ts
tomaskikutis: You can type it as `fieldObject: {[fieldId: string]: any}` since it's hard to type it more specifically due to the structure.

PR #2922, opened 2019-04-10
fixed: Cannot perform save and close from 'edit media metadata' after…

GyanP:
… pressing save first. (SDESK-4151, SDESK-4152)

### DIFF
--- a/scripts/apps/search/MultiImageEdit.ts
+++ b/scripts/apps/search/MultiImageEdit.ts
@@ -56,14 +56,17 @@ export function MultiImageEditController(
     $scope.isDirty = () => unsavedChangesExist;
 
     $scope.selectImage = (image) => {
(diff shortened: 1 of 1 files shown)


scripts/apps/search/MultiImageEdit.ts
```diff
@@ -56,14 +56,17 @@ export function MultiImageEditController(
...
+        if ($scope.images.length === 1) {
+            $scope.images[0].selected = true;
+        } else {
+            image.selected = !image.selected;
+        }
         updateMetadata();
     };
 
     // wait for images for initial load
     const stopInitWatch = $scope.$watch('images', (images: Array<any>) => {
```
petrjasek: I guess you don't need to store that `stopInitWatch` if it's not used anymore

PR #2925, opened 2019-04-11
feat(search): display selected saved searches as shortcuts

petrjasek:
SDANSA-196

requires https://github.com/superdesk/superdesk-core/pull/1577

### DIFF
--- a/scripts/apps/search/controllers/SearchMenuController.ts
+++ b/scripts/apps/search/controllers/SearchMenuController.ts
@@ -68,15 +68,25 @@ export default function SearchMenuController(
     // init saved searches
     const initSavedSearches = () => {
         savedSearch.getAllSavedSearches().then((savedSearches: Array<ISavedSearch>) => {
-            this.providers = this.providers.concat(savedSearches
+            let providers = [];
+            const shortcuts = savedSearches
                 .filter((search) => search.shortcut && search.is_global)
                 .map((search) => ({
                     _id: search._id,
                     name: search.name,
                     filter: search.filter,
-                })),
-            );
+                }));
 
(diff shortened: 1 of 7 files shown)


scripts/apps/search/controllers/SearchMenuController.ts
```diff
@@ -1,26 +1,61 @@
+import {get, isEqual} from 'lodash';
+import {ISavedSearch} from '../SavedSearch';
```
tomaskikutis: What does the dropdown in the search page contain now? It used to only contain a provider e.g. a source a search to be executed on. Does the dropdown now contain both providers and shortcuts? Do shortcuts work with any provider? I would split them into groups so it's easy to know which ones are providers/shortcuts.

scripts/apps/search/views/save-search-dialog.html
```diff
@@ -13,5 +13,9 @@
             <span id="search_global" sd-switch ng-model="edit.is_global"></span>
             <label class="label--inline" translate>Make global</label>
         </div>
+        <div class="field" ng-if="edit.is_global">
```
tomaskikutis: Doesn't this need the same feature flag check as in https://github.com/superdesk/superdesk-client-core/pull/2925/files#diff-275cb0eb9ee87557fa6c69e095e061deR85 ?

tomaskikutis:
Could you explain how bundling shortcuts with its repo works?

petrjasek:
it's like:
- superdesk
- superdesk shortcut 1
- superdesk shortcut 2
- external source
- external source shortcut 1
- external source shortuct 2

PR #2928, opened 2019-04-16
fixed: All data in authoring is lost if an article is refreshed.

GyanP:
SDESK-4184

### DIFF
--- a/scripts/core/spellcheck/spellcheck.ts
+++ b/scripts/core/spellcheck/spellcheck.ts
@@ -5,7 +5,7 @@ SpellcheckService.$inject = ['$q', 'api', 'dictionaries', '$rootScope', '$locati
 function SpellcheckService($q, api, dictionaries, $rootScope, $location, _, preferencesService) {
     var PREFERENCES_KEY = 'spellchecker:status',
         lang,
-        dict,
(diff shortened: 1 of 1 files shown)


scripts/core/spellcheck/spellcheck.ts
```diff
@@ -24,7 +24,6 @@ function SpellcheckService($q, api, dictionaries, $rootScope, $location, _, pref
     this.setLanguage = function(_lang) {
         if (lang !== _lang) {
             lang = _lang;
-            dict = null;
```
petrjasek: shouldn't it reset dict on language change?

GyanP:
> What was the root cause? How were dictionaries related to losing changes?

Since `dict = null`, it throws error on getting `dict.content` like here https://github.com/superdesk/superdesk-client-core/blob/78a4d735223307ed325acd7c5e7fd7ee8f7a687a/scripts/core/spellcheck/spellcheck.ts#L433.

tomaskikutis:
screenshot_9

Format the titles of the PRs so it doesn't include ellipses. It's easier for anyone reviewing the PR to read and find the corresponding JIRA tickets.

PR #2929, opened 2019-04-16
feat(authoring): add limits service

petrjasek:
put min/max length calculation to a service, which will
allow us to override it in client repo.

SDANSA-247

it should allow custom limits function like in https://github.com/petrjasek/superdesk/commit/15d0ec70d59bd4c852f23f9063e4b56a930c4f66

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
@@ -82,13 +82,14 @@ AuthoringDirective.$inject = [
     'relationsService',
     '$injector',
     'functionPoints',
+    'limits',
 ];
 export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace, notify,
     desks, authoring, api, session, lock, privileges, content, $location,
     referrer, macros, $timeout, $q, modal, archiveService, confirm, reloadService,
     $rootScope, $interpolate, metadata, suggest, config, deployConfig, editorResolver,
     compareVersions, embedService, $sce, mediaIdGenerator, relationsService, $injector,
-    functionPoints: IFunctionPointsService) {
+    functionPoints: IFunctionPointsService, limits) {
     return {
         link: function($scope, elem, attrs) {
             var _closing;
@@ -119,6 +120,7 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
             $scope.mediaFieldVersions = {};
             $scope.refreshTrigger = 0;
             $scope.isPreview = false;
+            $scope.limits = limits;
 
             $scope.$watch('origItem', (newValue, oldValue) => {
                 $scope.itemActions = null;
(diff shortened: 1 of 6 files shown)


tomaskikutis:
If validation info can't be static, I'd prefer to have a client-overwritable function to get the entire schema instead of putting additional logic to overwrite some of its fields. See if can be done easily, if not let's stick with this.

If we are to stick with this, I'd like naming to be more specific: `ArticleFieldLengthLimitService` instead of just `LimitService` which is too generic.

petrjasek:
I can imagine having this `onChange` middleware which would get schema reference and could make some changes on item change, it would probably also need some `onSchemaLoaded` middleware to set it also when you open item before doing any changes. that would avoid yet another specific api, what do you think @tomaskikutis ?

tomaskikutis:
I would like the API to be as stateless, immutable, side-effect free and low-knowledge as possible. That might sound like a lot, but I don't think it would be difficult to implement. By low-knowledge I mean that we should hide as much implementation details as possible and not expose things if we can implement what we need without exposing it.

On the other hand, we should not attempt to hide things which are central to how Superdesk works. It wouldn't make sense to hide desks for example. Sure we should hide the implementation, but we should expose basic properties and actions. I see schema similarly, so I'd rather have `getSchema(item: IArticle): ISchema`, then add statefulness or mutation to other parts of the API.

petrjasek:
it would have to be `getSchema(item: IArticle): Promise<ISchema>` because it must fetch it from server, and I'm not sure how to make it somehow efficient so it won't fetch it on every change. any ideas?

tomaskikutis:
Yep, I forgot the promise. All APIs should be promise based. You never know when you need to call the server or one of the dependencies is async. I think we should call it `IArticleSchema` since we also have an image schema.

Is schema fetched from the server? I thought it was somewhere here `scripts/apps/workspace/content/constants.ts`

Regarding efficiency, I would first fetch and cache the base schema(the one we use currently) at the authoring component level and only then pipe it through the plugins on every change expecting registered functions to respond fast. In this min/max case, it would respond instantly since the "plugin function" would be synchronous despite using the promise API.

petrjasek:
what about [middleware](https://github.com/petrjasek/superdesk-client-core/commit/a8c6d0b13381164231451ae5fafe9a9bf2c1f332) @tomaskikutis ?

tomaskikutis:
Do you mean middleware for this exact function? `getSchema(item: IArticle): Promise<IArticleSchema>`

I was referring to middleware when I said "plugin function". Sorry for the confusion.

Here's how I see it:
article-edit initializes -> call server to get the schema -> cache it article-edit view
when the article changes, pipe the original schema though all middlewares and update the article-edit view to use the returned schema.

petrjasek:
yep that's what it does, but it gets that initial schema and also item as params:
https://github.com/petrjasek/superdesk-client-core/commit/a8c6d0b13381164231451ae5fafe9a9bf2c1f332

PR #2930, opened 2019-04-17
[SDESK-4128] FunctionPoint service for generic function extensions

MarkLark86:
* Provide hook in AuthoringDirective.publishItem to process function extensions after validating and before publishing

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
@@ -5,6 +5,7 @@ import thunk from 'redux-thunk';
 import {gettext} from 'core/utils';
 import {combineReducers, createStore, applyMiddleware} from 'redux';
 import {attachments, initAttachments} from '../../attachments';
+import {IFunctionPointsService} from "apps/extension-points/services/FunctionPoints";
 
 /**
  * @ngdoc directive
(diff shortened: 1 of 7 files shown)


petrjasek:
hi, we were thinking about introducing similar api, but a bit different.
you can find a draft in https://gist.github.com/tomaskikutis/59646455c85a9ce61cea0c4e5665e843
and initial implementation in https://github.com/petrjasek/superdesk-client-core/commit/00e1a6adb594e23054fe1d02cadc72003c1f8e8d
we should decide on one approach and use it everywhere.
cc @tomaskikutis

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -425,7 +429,8 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
 
                 $scope.error = {};
 
-                return checkMediaAssociatedToUpdate()
+                return functionPoints.run('authoring:publish', {orig, item})
```
tomaskikutis: Could you only pass `item`, but not `orig`? Why do you need `orig`?

MarkLark86: `orig` and `item` are passed into the `publishItem` function, so I made the decision that the `FunctionPoint` should provide access to the same arguments from the parent function.

tomaskikutis: Got it. Please don't pass `orig` to `FunctionPoint` then. You can fetch the original from the server if you ever need it.

MarkLark86:
@petrjasek Could we tag this one for v1.28 as well please? We need this for Planning and have v1.28 in production.

PR #2932, opened 2019-04-23
As a user i want to use different spell checkers for different languages (SDBELGA 88)

tomaskikutis:
SDBELGA-88


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "superdesk-core",
-  "version": "1.28.0",
+  "version": "1.29.0-rc.1",
```
petrjasek: I think it should be `1.30.0-dev` now

tomaskikutis: Do you want to merge it now?

petrjasek: not really :wink:

tomaskikutis: I don't think there's a point to touch the version number then. When you decide you want to merge it, you can push an empty version bump commit to master and merge all the PRs which will inherit the new version number if `package.json` hasn't been modified in the PR.

PR #2935, opened 2019-04-24
fix(search): Fix issue saving changes on search after updating subscriptions

pablopunk:
SDESK-4183

Pressing 'Save changes' was causing an error only after user had changed
subscriptions for the current search. The issue was that saving the
subscriptions was changing the `etag` in the server but the current search
on the client was not being udpated.

As you can see in the code I didn't replace the whole object with the
updated one as the user might lose any changes to the current search
that's being edited, so I just udpated the `etag`.

### DIFF
--- a/scripts/apps/search/directives/SavedSearchManageSubscribers.ts
+++ b/scripts/apps/search/directives/SavedSearchManageSubscribers.ts
@@ -227,12 +227,13 @@ export function SavedSearchManageSubscribers(asset, userList, api, modal, desks)
                 modal.confirm(
                     gettext('Are you sure to remove this subscription?'),
                     gettext('Unsubscribe desk'),
-                ).then(() => {
-                    unsubscribeDesk(
+                ).then(async () => {
+                    const updatedSearch = await unsubscribeDesk(
                         scope.savedSearch,
                         desk._id,
                         api,
                     );
+                    scope.onSubscriptionsChange(updatedSearch);
                 });
 
             scope.editUserSubscription = (user: IUser) => {
--- a/scripts/apps/search/directives/SavedSearchManageSubscribers.ts
+++ b/scripts/apps/search/directives/SavedSearchManageSubscribers.ts
@@ -36,7 +36,8 @@ function isDeskSubscription(x: IModel['subscriptionInCreateOrEditMode']): x is I
 
 interface IScope extends IDirectiveScope<IModel> {
     savedSearch: ISavedSearch;
-    manageSubscriptions(active: boolean): void;
+    setIsManagingSubscriptions(active: boolean): void;
+    onSubscriptionsChange(nextValue: ISavedSearch): void;
     unsubscribeUser(user: IUser): Promise<void>;
     unsubscribeDesk(desk: IDesk): Promise<void>;
     editUserSubscription(user: IUser): void;
@@ -60,7 +61,8 @@ export function SavedSearchManageSubscribers(asset, userList, api, modal, desks)
     return {
         scope: {
             savedSearch: '=',
-            manageSubscriptions: '=',
+            setIsManagingSubscriptions: '=',
+            onSubscriptionsChange: '=',
         },
         templateUrl: asset.templateUrl('apps/search/views/saved-search-manage-subscribers.html'),
         link: function(scope: IScope) {
@@ -194,7 +196,11 @@ export function SavedSearchManageSubscribers(asset, userList, api, modal, desks)
                     desk_subscriptions: nextDeskSubscriptions,
                 };
 
-                updateSubscribers(scope.savedSearch, nextSubscribers, api).then(scope.backToList);
+                updateSubscribers(scope.savedSearch, nextSubscribers, api)
+                    .then((newSavedSearch: ISavedSearch) => {
+                        scope.onSubscriptionsChange(newSavedSearch)
+                        scope.backToList()
+                    });
             };
 
             scope.handleIntervalChange = (cronExpression: CronTimeInterval) => {
@@ -279,7 +285,7 @@ export function SavedSearchManageSubscribers(asset, userList, api, modal, desks)
                 () => {
                     // when modal is closed via ESC key, stop managing subscription so it can be started again.
                     if (!scope.wrapper.modalOpen) {
-                        scope.manageSubscriptions(false);
+                        scope.setIsManagingSubscriptions(false);
                     }
                 },
             );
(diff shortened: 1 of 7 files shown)


tomaskikutis:
Did you consider disallowing editing other parts of the saved search while editing subscriptions? That would allow replacing the entire object without having to deal with the etag explicitly.

pablopunk:
@tomaskikutis you can't lock the saved search because you don't know if the user is gonna go to the subscription edit view or not. A solution could be disallowing editing the subscriptions once you edited something on the saved search but that wouldn't be very confusing for the user. Also we could remove the "Save changes" from the edit subscription and just save the subscriptions to the current saved search and use the normal "Save changes" button to push those changes. But this would require integrating that view on the edit search form which is also not very friendly if you just wanna modify subscriptions

tomaskikutis:
> A solution could be disallowing editing the subscriptions once you edited something on the saved search but that wouldn't be very confusing for the user.

I think this one would be best. Making the "manage subscriptions" button disabled when there are unsaved changes. When disabled, a button could have a title indicating the search has to be saved before managing subscriptions.

pablopunk:
Don't you think that would be confusing? At least it would be necessary to include a warning or explanation when you're editing to explain why you can't click on "manage subscriptions". I think the current behaviour is more friendly

tomaskikutis:
> When disabled, a button could have a title indicating the search has to be saved before managing subscriptions.

If there was no title on hover, it might be confusing. With the tooltip explaining how it works, I think it's clear enough.

Managing `etags` in individual UI components is a bad practice because it breaks the boundaries of responsibility. It's also a sign that something is wrong with the implementation. If we want to allow editing subscriptions independently from a saved search, saved searches should be in their separate database collection in which case there wouldn't be any `etag` issues.

pablopunk:
@tomaskikutis it is a sign that something is wrong. What's wrong is that we a have a form with two Save buttons, one for "all fields" and one for subscriptions, and that causes this issues. The solution would be having just one "Save changes" button

tomaskikutis:
> The solution would be having just one "Save changes" button

I'm fine with a solution which would only include one "save" button if we can make it clear that changes weren't saved after closing a modal. Or if we could programmatically press that button to save it. Or save it the same way without faking a button click :)

pablopunk:
I went back to the original solution. Turns out updating the open search to edit does work when updating each field individually. And there are no issues even when edit-search>edit-subscriptions>save

scripts/apps/search/directives/SavedSearchManageSubscribers.ts
```diff
@@ -208,24 +214,26 @@ export function SavedSearchManageSubscribers(asset, userList, api, modal, desks)
                 modal.confirm(
                     gettext('Are you sure to remove this subscription?'),
                     gettext('Unsubscribe user'),
-                ).then(() => {
-                    unsubscribeUser(
+                ).then(async () => {
```
tomaskikutis: Why do you use async/await here?

In order to use async/await, the function you are awaiting has to return a promise. If it returns a promise and you want to perform an action after it resolves, using `then` is all you need.

```
modal.confirm(
    gettext('Are you sure to remove this subscription?'),
    gettext('Unsubscribe user'),
).then(unsubscribeUser(
    scope.savedSearch,
    user._id,
    api,
)).then(() => {
    scope.onSubscriptionsChange(updatedSearch);
});
```

Also, async/await is not used in the codebase yet, and I think the code is more readable if we stick to one way of doing things. We could start using async/await as well if it has significant advantages. Does it? I haven't looked deeply into it myself.

scripts/apps/search/directives/SearchPanel.ts
```diff
@@ -63,16 +60,24 @@ export function SearchPanel($location,
...
-            scope.manageSubscriptions = (nextValue) => {
+            scope.setIsManagingSubscriptions = (nextValue: boolean) => {
                 scope.isManagingSubscriptions = nextValue;
             };
 
+            // called after changing the subscriptions for current search
+            scope.onSubscriptionsChange = (updatedSavedSearch) => {
+                // subscriptions were updated via API so the object has changed
+                for (const key in updatedSavedSearch) {
+                    scope.editingSearch[key] = updatedSavedSearch[key];
```
tomaskikutis: Why don't update the entire object like `scope.editingSearch = updatedSavedSearch`? Is angular not picking up the updates?

scripts/apps/search/directives/SavedSearchManageSubscribers.ts
pablopunk: async/await is just syntactic sugar for promises. I can change it but there's no need to choose between `then` or `async/await`, they can live together. Actually I did have the code you're showing and then changed it to await cause it's simpler IMO but I don't care about it much, both are readable

scripts/apps/search/directives/SearchPanel.ts
pablopunk: For some reason I still don't know, it doesn't. `scope.editingSearch = updatedSavedSearch` doesn't work.

scripts/apps/search/directives/SavedSearchManageSubscribers.ts
tomaskikutis: It is similar to syntactic sugar, but not entirely. If you check the polyfills, it's built on es6 generators. The concept itself of writing async code inline is quite different. I'd say let's stick to promises for now since that's a uniform way of doing async in the codebase. We can add async/await later if we bump into a use case that's hard to do with promises.

scripts/apps/search/directives/SearchPanel.ts
tomaskikutis: I'm pretty sure it's this one https://stackoverflow.com/a/38763654 which seems to be everyone "not understanding scopes" and definitely not an angular bug which is probably hard to solve in the framework and everyone is doing workarounds. :smile:

pablopunk: 🤦 I hate angular... So basically it's [shadowing the parent variable](https://github.com/angular/angular.js/wiki/Understanding-Scopes). I'll fix it

pablopunk: Actually, something else is happening. When I do `scope.editingSearch._etag = `, it's also changing `scope.edit._etag` (which is the etag that matters on the request), but I couldn't find why they are bound together.

pablopunk: This doesn't happen when I change all of the `scope.edit = create(scope.editingSearch)` to `scope.edit = scope.editingSearch`. But I'm not sure why lodash `create` would bind those together.

PR #2937, opened 2019-04-25
Fixed: Edit Ingest source- Critical errors tab is not loaded correctly. (SDESK-4198)

GyanP:


MarkLark86:
@petrjasek @mdhaman This bug is reproducible in v1.28 as well.

PR #2940, opened 2019-04-26
fix(authoring): add missing error indicator on custom required fields

petrjasek:
it was using _id to find out if there is an error, but server sends
display name instead for custom fields.

SDESK-2390


scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
```diff
@@ -395,6 +395,7 @@ export function ArticleEditDirective(
                                 scope.editor = authoring.editor = content.editor(type, scope.item.type);
                                 scope.schema = authoring.schema = content.schema(type, scope.item.type);
                                 scope.fields = content.fields(type);
+                                console.info('schema', scope.schema, scope.editor);
```
MarkLark86: Do we have lint rules for `console.*` methods?

petrjasek: I guess we have it only for `console.log`

PR #2942, opened 2019-04-30
fixed: While editing saved searches some of the fields do not keep and display the search criteria that was originally there. (SDESK-4182)

GyanP:


scripts/apps/search/directives/SearchParameters.ts
```diff
@@ -78,6 +78,14 @@ export function SearchParameters($location, asset, tags, metadata, common, desks
                 scope.search_config = getSearchConfig();
                 scope.lookupCvs = {};
                 scope.params = params.params ? JSON.parse(params.params) : {};
+
+                if (scope.query && !Object.keys(scope.meta).length) {
```
tomaskikutis: Make it more readable. What are you trying to match with this regex? add an example.

There are too many things done in lines 85,86. I can't see where are the boundaries of the functions and which operations take place in which order since everything is that condensed. Use clearer variable names - `query` instead of `t`, `key` and `value` instead of `k`,`v`.

scripts/apps/search/directives/SearchParameters.ts
```diff
@@ -80,10 +80,17 @@ export function SearchParameters($location, asset, tags, metadata, common, desks
                 scope.params = params.params ? JSON.parse(params.params) : {};
 
                 if (scope.query && !Object.keys(scope.meta).length) {
-                    let queryTerms = scope.query.split(/(?!\(.*)\s(?![^(]*?\))/g);
+                    // split scope.query by spaces outside the parentheses
+                    // ex: "slugline:(abc) headline:(xyz)" will become ["slugline:(abc)", "headline:(xyz)"]
+                    const queryTerms = scope.query.split(/(?!\(.*)\s(?![^(]*?\))/g);
```
tomaskikutis: Use the internal data structures instead of trying to parse the same information from the string which is displayed to the user. The format which the information is displayed to the user should be allowed to change without breaking features.

GyanP: Sorry, I didn't get you.

tomaskikutis: Strings like `slugline:(abc) headline:(xyz)` are created from objects with separate fields. You must not parse the string. If you need to read the data, read it from the original objects the string is made from, not from the string itself.

Here's an example of where a similar string is created
> `tags.selectedParameters.push(tag(value + ':(' + name + ')'))`

GyanP: It will work only for those fields which are kind of `cv` like `subject` or `company codes`. Whereas `slugline`, `headline` or `body` are hardcoded strings.

scripts/apps/search/services/TagService.ts
```diff
@@ -54,10 +54,24 @@ export function TagService($location, desks, userList, metadata, search,
         };
     }
 
-    function initSelectedParameters(params) {
+    function getParamObject(paramArray: Array<string>) {
+        const key = paramArray[0].trim();
+        // remove parentheses ex: "(two)" becomes "two"
+        const value = paramArray[1].replace(/[()]/g, '');
```
tomaskikutis: Remove only one opening parenthesis if it's the first char in the string and only one closing parenthesis if it's the last char in the string. Current regex removes all parenthesis in the middle as well.

PR #2943, opened 2019-05-01
fixed: Broken media gallery. (SDESK-4201)

GyanP:
If the media element is not found in the source URL, the `load` and `loadedmetadata` eventListener does not fire the callback. Due to this promise never gets resolved. So, now checking for the errors in source URL, if there are any then call the same callback function.

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
@@ -73,8 +73,10 @@ export function ItemCarouselDirective($timeout, notify) {
                         carousel.trigger('destroy.owl.carousel');
                     }
 
-                    const carouselImages = Array.from(elem.get(0).querySelectorAll(`${carouselContainerSelector} img`));
-                    const carouselAudiosAndVideos = Array.from(
+                    const carouselImages: Array<HTMLImageElement> = Array.from(
+                        elem.get(0).querySelectorAll(`${carouselContainerSelector} img`),
+                    );
+                    const carouselAudiosAndVideos: Array<HTMLAudioElement | HTMLVideoElement> = Array.from(
                         elem.get(0).querySelectorAll(
                             `${carouselContainerSelector} video, ${carouselContainerSelector} audio`,
                         ),
--- a/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
@@ -86,10 +86,9 @@ export function ItemCarouselDirective($timeout, notify) {
                         return;
                     }
 
-                    Promise.all([
-                        waitForMediaToLoad(carouselImages),
-                        waitForMediaToLoad(carouselAudiosAndVideos),
-                    ]).then(initCarousel);
+                    const mediaItems = carouselImages.concat(carouselAudiosAndVideos);
+
+                    waitForMediaToLoad(mediaItems).then(initCarousel);
                 });
             });
             /*
--- a/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
@@ -85,10 +85,8 @@ export function ItemCarouselDirective($timeout, notify) {
                     }
 
                     Promise.all([
-                        waitForMediaToLoad(carouselImages.filter(
-                            (img: HTMLImageElement) => img.complete === false), 'load'),
-                        waitForMediaToLoad(carouselAudiosAndVideos.filter(
-                            (media: HTMLAudioElement | HTMLVideoElement) => media.readyState < 1), 'loadedmetadata'),
+                        waitForMediaToLoad(carouselImages),
+                        waitForMediaToLoad(carouselAudiosAndVideos),
                     ]).then(initCarousel);
                 });
             });
--- a/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
@@ -3,7 +3,7 @@
 import 'owl.carousel';
 import _ from 'lodash';
 import * as ctrl from '../controllers';
-import {waitForImagesToLoad, waitForAudioAndVideoToLoadMetadata} from 'core/helpers/waitForMediaToBeReady';
+import {waitForMediaToLoad} from 'core/helpers/waitForMediaToBeReady';
 import {getSuperdeskType} from 'core/utils';
 import {gettext} from 'core/utils';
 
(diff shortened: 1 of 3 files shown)


tomaskikutis:
If you are using different logic depending on element type, it might be better to do a little refactor and get rid off `waitForImagesToLoad` and `waitForAudioAndVideoToLoadMetadata` and instead handle it in `waitForEventToFireForAllElements`. I'd rename it to `waitForMediaToLoad`.

Replace jsdoc types with typescript types.

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -85,8 +85,10 @@ export function ItemCarouselDirective($timeout, notify) {
                     }
 
                     Promise.all([
-                        waitForImagesToLoad(carouselImages),
-                        waitForAudioAndVideoToLoadMetadata(carouselAudiosAndVideos),
+                        waitForMediaToLoad(carouselImages.filter(
+                            (img: HTMLImageElement) => img.complete === false), 'load'),
```
tomaskikutis: Since you now have element-type specific behavior in `waitForMediaToLoad`, can you also move the logic of filtering the elements and determining the event type there as well?

scripts/core/helpers/waitForMediaToBeReady.ts
```diff
@@ -1,37 +1,34 @@
 /**
- * @param {Array<Node>} elements
- * @param {string} eventName
+ * @param elements An array of img, audio and video elements
+ * @param eventName EventListener's name
  */
-const waitForEventToFireForAllElements = (elements, eventName) => new Promise((resolve) => {
-    let itemsLeftToLoad = elements.length;
+export const waitForMediaToLoad = (elements: Array<any>, eventName: string): Promise<void> => new Promise((resolve) => {
```
tomaskikutis: Don't use `any` type when the actual type is known. It should be `elements: Array<Element>`.

scripts/core/helpers/waitForMediaToBeReady.ts
```diff
@@ -1,9 +1,17 @@
 /**
  * @param elements An array of img, audio and video elements
- * @param eventName EventListener's name
  */
-export const waitForMediaToLoad = (elements: Array<any>, eventName: string): Promise<void> => new Promise((resolve) => {
-    let itemsLeftToLoad: number = elements.length;
+export const waitForMediaToLoad = (elements: Array<Element>): Promise<void> => new Promise((resolve) => {
```
tomaskikutis: TypeScript checks aren't passing since `Element` doesn't have either a properly `complete` nor `readyState`.

I checked it again, and I think the following type signature is more accurate: `elements: Array<HTMLImageElement> | Array<HTMLVideoElement> | Array<HTMLAudioElement>)`

Create helpers like the one below to execute code specific to an element of a particular type.

```
export function isImage(x: Element): x is HTMLImageElement {
    return x.tagName === 'IMG';
}

if (isImage(document.querySelector('...'))) {
    // execute image specific code
}
```

scripts/core/helpers/waitForMediaToBeReady.ts
```diff
@@ -1,9 +1,17 @@
...
+    // EventListener's name
+    let eventName: string = 'load';
```
tomaskikutis: event name depends on element type, so it's more accurate if you don't set it for the entire scope and use inline strings in conditional blocks instead.

scripts/core/helpers/waitForMediaToBeReady.ts
```diff
@@ -1,9 +1,17 @@
...
+
+    const filteredElelemnts = elements.filter((element: Element) => {
+        if (element.tagName === 'IMG') {
+            return element.complete === false;
+        }
+        return element.readyState < 1;
+    });
+    let itemsLeftToLoad: number = filteredElelemnts.length;
```
tomaskikutis: spelling should be `filteredElements`

scripts/core/helpers/waitForMediaToBeReady.ts
```diff
@@ -19,11 +27,12 @@ export const waitForMediaToLoad = (elements: Array<any>, eventName: string): Pro
...
     };
 
-    elements.forEach((element) => {
-        // check for error in image src
+    filteredElelemnts.forEach((element: Element) => {
         if (element.tagName === 'IMG') {
+            // check for error in image src
             element.addEventListener('error', eventHandler, {once: true});
         } else {
+            eventName = 'loadedmetadata';
```
tomaskikutis: instead of else, use `else if(element.tagName === 'IMG' || element.tagName === 'VIDEO') `. It's faster to understand exactly which types of elements the else branch is applied to.

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -85,8 +87,8 @@ export function ItemCarouselDirective($timeout, notify) {
                     }
 
                     Promise.all([
-                        waitForImagesToLoad(carouselImages),
-                        waitForAudioAndVideoToLoadMetadata(carouselAudiosAndVideos),
+                        waitForMediaToLoad(carouselImages),
+                        waitForMediaToLoad(carouselAudiosAndVideos),
```
pablopunk: is there a reason not to call the function once with all the data? Something like:
```js
const allMedia = carouselImages.concat(carouselAudiosAndVideos)
waitForMediaToLoad(allMedia).then(initCarousel)
```

scripts/core/helpers/waitForMediaToBeReady.ts
```diff
@@ -1,37 +1,45 @@
...
+        if (isImage(element)) {
+            // check for error in image src
+            element.addEventListener('error', eventHandler, {once: true});
+        } else if (isAudio(element) || isVideo(element)) {
+            // for audio and video check for the error in source
+            const source = element.getElementsByTagName('source')[0];
+
+            source.addEventListener('error', eventHandler, {once: true});
+        }
+        element.addEventListener(isImage(element) ? 'load' : 'loadedmetadata', eventHandler, {once: true});
```
tomaskikutis: this should be inside `else if (isAudio(element) || isVideo(element))` branch

GyanP: What's wrong with this one? :)

tomaskikutis: Sorry, my bad. I thought the ternary was about audio/video. Since it's about image and audio/video, it makes sense to put it in respective branches.

Your approach is not wrong at the moment but you are applying similar conditions twice which is less clear. In your case, you have 4 if statements, which are almost the same, but not exactly(the ternary uses else, not else if).

It would break easily if another media type was added. It's not likely that we'll need another media type, but it's a good practice not to duplicate if statements and put everything in one place if possible.

It looks clearer if you replace ternary with `if` operator

screenshot_7

PR #2945, opened 2019-05-08
Set 'Allow multiple items' limit  in related content to minimum 2. (SDESK-4237)

GyanP:


scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
```diff
@@ -369,7 +369,7 @@ export function ArticleEditDirective(
                 scope.maxUploads = function(fieldOptions) {
                     if (fieldOptions && fieldOptions.multiple_items &&
                         fieldOptions.multiple_items.enabled) {
-                        return fieldOptions.multiple_items.max_items ? fieldOptions.multiple_items.max_items : 0;
+                        return fieldOptions.multiple_items.max_items ? fieldOptions.multiple_items.max_items : 2;
```
petrjasek: imo it should start with 1, allowing multiple doesn't mean it must be enforced

scripts/apps/vocabularies/views/vocabulary-config-modal.html
```diff
@@ -161,8 +161,14 @@ <h3 class="modal__heading" ng-show="!vocabulary._id"><span translate>Create</spa
                                 <span translate>Max. number of items</span>
                             </div>
                             <div class="form__row-item form__row-item--no-grow">
-                                <input id="max_items" type="number" min="0" max="99" class="input-small line-input"
-                                                                                     ng-model="vocabulary.field_options.multiple_items.max_items" style="width: 60px; text-align: center;">
+                                <input id="max_items" name="max_items" type="number"
+                                        min="2" max="99" class="input-small line-input"
```
petrjasek: you should change it here too `min="2"`

GyanP: I had a discussion with Karel, and he said that we are talking about multiple uploads, so it must start with 2, not 1.

PR #2948, opened 2019-05-09
add initial middleware support to authoring

petrjasek:
required for SDANSA-247


scripts/core/middleware.ts
```diff
@@ -0,0 +1,9 @@
+
+type IMiddlewareFn = (arg0: any) => any;
```
MarkLark86: Should this function type return a promise?
`(arg0: any) => Promise<any>`

scripts/apps/authoring/authoring/index.ts
```diff
@@ -12,6 +12,24 @@ import {ArticleUrlFields} from './article-url-fields';
...
+    item: IArticle;
+    original: IArticle;
+}
+
+export interface IArticleSchemaParams {
+    item: IArticle;
+    schema: IArticleSchema;
+}
+
+type IOnChangeMiddleware = (params: IOnChangeParams) => any;
```
MarkLark86: Should these two function types return a promise?
`(params: IOnChangeParams) => Promise<any>`
`(params: IArticleSchemaParams) => Promise<any>`

petrjasek: I'm not sure yet how to pass that return value and keep the applyMiddleware code generic at the same time, will check it

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -850,12 +853,16 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
             $scope.autosave = function(item, timeout) {
                 $scope.dirty = true;
                 angular.extend($scope.item, item); // make sure all changes are available
+                return coreApplyMiddleware(onChangeMiddleware, {item: $scope.item, original: $scope.origItem}, 'item')
```
tomaskikutis: Could you talk me through how your implementation works? I don't understand why it calls a middleware before AND after autosave. I assume it should fire on change, but you are using autosave as a debouncer?

petrjasek: right autosave is called whenever user makes some changes so I use it to trigger the middleware.

scripts/core/middleware.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function applyMiddleware(middleware: Array<IMiddlewareFn>, params: object, key?: string): Promise<any> {
```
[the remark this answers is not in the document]
petrjasek: yep problem is it's limited to single param which must also be the return value. unless it would be the full state like in redux

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
tomaskikutis: Could you also answer the first part of the question? :slightly_smiling_face:

petrjasek: yep that's because first set of middlewares update the item before it gets stored. second set updates schema based on item changes. so to make sure all middlewares with item changes are done before the schema middlewares get called

tomaskikutis: Are you updating anything in the item before it gets stored or just put the middleware just so it's there for the future?

petrjasek: atm it's there for future, won't need it after all

scripts/superdesk-interfaces/ArticleSchema.ts
```diff
@@ -0,0 +1,9 @@
+
+export interface IArticleSchema {
+    [field: string]: {
+        type: string;
```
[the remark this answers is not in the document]
petrjasek: this is different than item type, it's schema field type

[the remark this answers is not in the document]
tomaskikutis: What type of schemas do we have for `IArticleSchema`?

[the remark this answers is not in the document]
petrjasek: there is `text`, `media`, `date`, `embed` etc

PR #2950, opened 2019-05-10
Client side extensions

tomaskikutis:
SDESK-4261

https://github.com/superdesk/superdesk-core/issues/1585

- ~~generate extension imports on build~~
- ~~activate extensions on application start~~
- ~~allow extensions to use core API~~
- ~~allow extensions to use APIs of other extensions~~
- ~~make build script install dependencies for extensions & compile them ts>js~~
- ~~ensure API types can be used from extensions~~
- ~~can translate(and mark for translation) strings in extensions~~


scripts/core/get-superdesk-api-implementation.ts
```diff
@@ -0,0 +1,49 @@
...
+                    && manifest.superdeskExtension.dependencies.includes(id)
+                ) {
+                    const extensionShallowCopy = {...extension};
+
+                    delete extensionShallowCopy['activate'];
+
+                    return Promise.resolve(extensionShallowCopy);
+                } else {
+                    console.error('Not authorized.');
+                    return Promise.resolve(null);
```
petrjasek: what about `Promise.reject` here? so it would not call any `.then` code if extension is not available

tomaskikutis: Good catch. I guess I forgot I can reject promises.

tasks/generate-extension-imports.js
```diff
@@ -0,0 +1,44 @@
...
+import {IExtensions} from 'superdesk-api';
+
+export const extensions: IExtensions = {};
+`;
+
+const importStatements = [];
+const insertIntoObjectStatements = [];
+
+directories.forEach((extensionName) => {
+    importStatements.push(`import * as ${extensionName} from '../extensions/${extensionName}/dist/extension';`);
```
petrjasek: might be better to read `dist/extension` from package.json file rather than hardcoding it

petrjasek:
would we also allow extensions to live in client repo? or they should be all in this one?

tasks/generate-extension-imports.js
```diff
@@ -0,0 +1,44 @@
+var fs = require('fs');
+var path = require('path');
+var lstatSync = fs.lstatSync;
+
+var directories = fs.readdirSync(path.resolve(`${__dirname}/../scripts/extensions`)).filter(
```
petrjasek: this part is duplicated in both tasks

tomaskikutis:
We will allow extensions on client repositories. In the long run, a client repository should become no more than a collection of extensions and not touch core internals at all as it does now.

The way I plan to implement this now is to add a command to client repositories which would copy extensions from a client repository to `node_modules/superdesk-core/extensions` on build so extensions work the same as if they were in the core.

PR #2952, opened 2019-05-14
[1.28](fix): FunctionPoint 'authoring:publish' not sending through _id and type

MarkLark86:


petrjasek:
@MarkLark86 is the fix also on master?

MarkLark86:
@petrjasek Yes it is.

PR #2953, opened 2019-05-14
Fix(Subscribers): Cannot create 'Content API-only' destination. (SDESK-4257)

GyanP:
SDESK-4194

### DIFF
--- a/scripts/apps/publish/views/subscribers.html
+++ b/scripts/apps/publish/views/subscribers.html
@@ -137,7 +137,6 @@
                                  data-field="products"
                                  data-unique="_id"
                                  data-list="directProducts"
-                                 data-change="enableSave(editForm)"
                                  data-header="true">
                             </div>
                         </div>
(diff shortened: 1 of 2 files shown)


tomaskikutis:
Why `expires` field was triggering changes while other fields weren't?

scripts/apps/publish/views/subscribers.html
```diff
@@ -280,7 +284,7 @@ <h3 class="modal__heading" translate ng-hide="subscriber._id">Add New Subscriber
         <div class="modal__footer">
             <button class="btn" ng-click="cancel()" translate>Cancel</button>
             <button class="btn btn--primary" ng-click="save()"
-                    ng-disabled="editForm.$invalid || (destinations.length === 0 && api_products.length === 0) || (destinations.length === 0 && products.length > 0) || !editForm.$dirty" translate>Save</button>
+                    ng-disabled="editForm.$invalid || !editForm.$dirty" translate>Save</button>
```
tomaskikutis: Remove `$scope.enableSave` method and instead check if current data is different than saved data.

PR #2957, opened 2019-05-14
Fix end to end test

tomaskikutis:
@petrjasek  can you take a look at this test? I noticed that it's breaking the build on master, but couldn't figure out how it's possible it ever passed before. I've tried running it with older code and still got the same result.

The test assumes that there are 2 notifications initially and 3rd gets added when a user is mentioned when in fact there are 3 initially. There are no problems to adjust the numbers, but it would help to find out how come it started failing just now.

### DIFF
--- a/spec/notifications_spec.js
+++ b/spec/notifications_spec.js
@@ -12,7 +12,7 @@ describe('notifications', () => {
         monitoring.openMonitoring();
     });
 
-    it('create a new user mention', () => {
+    fit('create a new user mention', () => {
         expect(monitoring.getTextItem(2, 0)).toBe('item5');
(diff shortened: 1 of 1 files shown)


pavlovicnemanja:
I was also confused with this test yesterday when I was trying localy. How it's suppoused to pass? Maybe we added some additional inhest channel? 3 notifications are about ingest channels added and one is comments.

petrjasek:
seems like it started after https://github.com/superdesk/superdesk-core/pull/1579 - there is a new provider added in prepopulate data, that might be it

tomaskikutis:
Thanks. I didn't realize that test server referenced to core master so my bisecting only client-core only didn't work.

PR #2962, opened 2019-05-16
Fixed: error is thrown on sending correction for an item containing images with adjusted brightness in custom media field. (SDESK-3380)

GyanP:


scripts/apps/authoring/views/item-carousel.html
```diff
@@ -11,132 +11,132 @@
         <a ng-click="navNext()" class="sd-media-carousel__nav-button sd-media-carousel__nav-button--next-inner"></a>
 
         <div class="sd-media-carousel__content owl-carousel">
-            <div class="sd-media-carousel__page" ng-repeat="item in carouselItems">
+            <div class="sd-media-carousel__page" ng-repeat="carouselItem in carouselItems">
```
tomaskikutis: Why did you need to rename this?

scripts/apps/authoring/authoring/controllers/AssociationController.ts
```diff
@@ -159,11 +159,14 @@ export function AssociationController(config, send, api, $q, superdesk,
      * @param {Function} callback to call after save
      */
     this.updateItemAssociation = function(scope, updated, customRel, callback = null, autosave = false) {
-        let data = {}, rel = customRel || scope.rel;
+        let data = {};
+        // if the media is of type media-gallery, update same association-key not the next one
+        // as the scope.rel contains the next association-key of the new item
```
tomaskikutis: > update same association-key not the next one as the scope.rel contains the next association-key of the new item

Why there are multiple keys and multiple items? Could you explain how this works?

scripts/apps/authoring/views/item-carousel.html
GyanP: Whenever we click on `edit image or edit metadata` etc. buttons, we are calling a method like `associations.edit(this, carouselItem[carouselItem.fieldId])`, where `this` is the directive scope.
In the directive scope `scope.item` is the one which contains the media-gallery, slugline, body, etc. But when we do `ng-repeat="item in carouselItems"`, this ng-repeat `item` overwrites the directive scope item. Which means we don't get the expected item on doing `associations.edit(this, carouselItem[carouselItem.fieldId])`. Which causes the errors.

scripts/apps/authoring/authoring/controllers/AssociationController.ts
GyanP: A field of type media-gallery can hold multiple items, so here https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts#L62  the `scope.rel` contains the next key where a media item can be added. Like if I have added one media with 'media--1` key then `scope.rel` will hold the next key as `media--2`.

PR #2964, opened 2019-05-17
fix error display in authoring for custom fields

petrjasek:
with multi-word display name

SDESK-2390


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -459,14 +460,30 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
                                     .replace(/\]/g, '')
                                     .split(',');
 
-                                for (var i = 0; i < modifiedErrors.length; i++) {
-                                    var message = _.trim(modifiedErrors[i]);
-                                    // the message format is 'Field error text' (contains ')
-                                    var field = message.split(' ')[0].substr(1);
+                                getLabelNameResolver().then((getLabelForFieldId) => {
```
tomaskikutis: Why not include the field name in the JSON response from the server so string parsing is not required?

petrjasek:
requires https://github.com/superdesk/superdesk-core/pull/1597

tomaskikutis:
I see some splitting and trimming is still done. What is the format now? Does it return error codes or human-readable text?

petrjasek:
it returns both

tomaskikutis:
If a human-readable error is returned, why do you need to trim/split anything?

petrjasek:
I think it's pseudo human readable

tomaskikutis:
:smile: can you make it fully readable & translatable?

tomaskikutis:
Can you also post a snippet how the format looks like?

petrjasek:
this is the current "human-readable" message: `"[['DATELINE is a required field']]"`
and the new structured info is like:
```
"fields": {
    "dateline": {
        "located": "required field",
        "date": "required field",
        "source": "required field",
        "text": "required field"
    }
}
```

tomaskikutis:
Do we need nested fields in error messages? It would be simpler not to have it. I would also return the entire error messages which could be displayed to the user without no further processing required.

Regarding the old message, I would remove that entirely and use messages from the new structure. What do you think?

```
"fields": {
    "dateline": [
        {
            "error": "Dateline field is required",
        },
        {
            "error": "Dateline source is required",
        }
    ]
}
```

petrjasek:
dateline is probably not good example, but we have also extra dict for custom fields and there we need to know which field has issues, so it's better to keep it more specific imo

tomaskikutis:
It would work if fields from extra were put in the root errors object:

```
"fields": {
    "dateline": [
        {
            "error": "Dateline field is required",
        },
        {
            "error": "Dateline source is required",
        }
    ],
   "custom-field-1": [
       {
            "error": "Dateline field is required",
        },
    ],
}
```

Do you remember why custom fields are put in extra in the first place?

petrjasek:
I think it's to avoid conflicts with other fields, and to be able to define elastic mapping for it

PR #2966, opened 2019-05-21
Annotations library extension

tomaskikutis:


scripts/extensions/annotationsLibrary/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "private": true,
+  "main": "dist/extension.js",
+  "scripts": {
+    "compile": "tsc",
+    "watch": "tsc -watch"
+  },
```
pablopunk: How does it work when you're developing an extension? Does `grunt` watch and compile extensions automatically?

scripts/extensions/annotationsLibrary/src/extension.tsx
```diff
@@ -0,0 +1,48 @@
...
+        {field: 'name', direction: 'ascending'},
+        {name: generateFilterForServer(nameField.type, annotationText)},
+    ).then((res: IRestApiResponse<IKnowledgeBaseItem>) => res._items.length > 0);
+}
+
+var extension: IExtension = {
+    activate: (superdesk: ISuperdesk) => {
+        const {gettext} = superdesk.localization;
+
+        return Promise.resolve({
```
pablopunk: You don't need `Promise.resolve` if you declare the function as `async`, I know that we don't use `async/await` but at least `async` is just very convenient and readable, as you immediately know if it will return a Promise

scripts/extensions/helloWorld/tsconfig.json
```diff
@@ -1,14 +1,28 @@
...
+        "strictPropertyInitialization": true,
+        "strictNullChecks": true,
+        "noImplicitAny": true,
+        "noImplicitReturns": true,
+        "noImplicitThis": true,
+        
+        "forceConsistentCasingInFileNames": true,
+        "noFallthroughCasesInSwitch": true,
+        "noUnusedLocals": true,
+        "noUnusedParameters": true,
```
pablopunk: I'm thinking if we should just make each extension use the root `tsconfig.json`

pablopunk: Ok nevermind, I realized we should remove `['scripts']` and stuff like that from root config, not needed.

scripts/core/superdesk-api.d.ts
```diff
@@ -1,9 +1,41 @@
...
+                ? DeepReadonlyArray<U>
+                : DeepReadonlyObject<T>;
+
+    export type Omit<K, V> = Pick<K, Exclude<keyof K, V>>;
+
+
+
+    // EXTENSIONS
+
+    export interface IExtensionActivationResult {
```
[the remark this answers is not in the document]
tomaskikutis: One of the main things I'm trying to achieve with the extensions is to minimize coupling between implementation details of the core and extensions API. That allows for greater flexibility in changing the implementation details while maintaining the same API.

Imagine I wanted an extension to be able to register menu items. A naive approach would be to give the extension access to the `ul` element, and it would insert DOM nodes to it. That would work, but would also cause some issues:
* screen jumping around / flickering while different extensions insert items at different times
* if we wanted to add sorting of menu items, it wouldn't be straight-forward since we would have DOM nodes, not text. We could try getting `innerText`, but extensions might be using an icon-font for adding icons in front of the text and that would mess up the sorting. On top of that, extensions would be able to insert items at any time which means we would have to listen to DOM changes and re-sort afterward which wouldn't even have been possible a few years ago without recently introduced mutation observer.
* if the menu is closed, an extension wouldn't know that and might be doing HTTP requests which would slow down the app unnecessarily.

The solution for such issues is to limit what the API exposes invert the control - that is to call the extensions ourselves instead of allowing extensions call us. Continuing with the menu example, instead of allowing an extension to insert DOM nodes, I want to allow it to provide a function the core would call when it wants to render the menu for example. The function would return a promise, the core would wait for all results render the menu. We are also not giving it access to the DOM node, but are forcing it to return text which allows us to control the rendering. I want extensions to only be able to provide such functions once, so I have extensions returning them from `activate` function and call it activation result.

Feel free to suggest better naming anywhere in the code.

scripts/core/superdesk-api.d.ts
```diff
@@ -1,9 +1,41 @@
...
+            editor3?: {
+                annotationInputTabs?: Array<IEditor3AnnotationInputTab>;
+            }
+            pages?: Array<IPage>;
+        }
     }
 
-    export interface IExtensionObject {
+    export type IExtension = DeepReadonly<{
+        activate: (superdesk: ISuperdesk) => Promise<IExtensionActivationResult>;
```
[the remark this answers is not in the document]
tomaskikutis: I don't see extensions as having a **value**. Such value doesn't represent the entire extension, but only the functions it registered for the core to call.

I'm not sure about other methods. The plan is to allow extensions to communicate with each other, so other functions might be exported for that or it might not.

scripts/extensions/annotationsLibrary/package.json
tomaskikutis: Grunt only watches the extension output and reloads when it changes, but doesn't know about the link between extension code and output. You have to run both grunt and `npm run watch` inside the extension folder. Extensions have a standalone build process by TypeScript and do not depend on core for anything but extracting translations from `gettext` calls.

scripts/extensions/annotationsLibrary/src/extension.tsx
tomaskikutis: I don't mind the resolve. Async is a new thing and more developers understand promises than async/await. I wouldn't mind switching if there were significant benefits, but I don't see them yet. You'll have to up your selling skills :smile:

scripts/extensions/helloWorld/tsconfig.json
tomaskikutis: The core wouldn't compile under strict flags. We could extend and overwrite core tsconfig, but that might not work well with extensions which will be in another repository so keeping it separate is easiest at the moment.

scripts/core/editor3/components/annotations/AnnotationInput.tsx
```diff
@@ -75,12 +72,24 @@ class AnnotationInputBody extends React.Component<IProps, IState> {
...
+            loaded: false,
         };
 
         this.onSubmit = this.onSubmit.bind(this);
         this.onChange = this.onChange.bind(this);
         this.onSelect = this.onSelect.bind(this);
         this.deleteAnnotation = this.deleteAnnotation.bind(this);
+        this.getSelectionText = this.getSelectionText.bind(this);
+
+        this.annotationInputTabsFromExtensions = flatMap(
```
petrjasek: imo it would be good to have some api to work with extensions, so you don't have to import that generated file everywhere and do the map on it, something like:
```
import extensions from 'core/extensions';

this.annotationInputTabsFromExtensions = extensions.getFlatMap('contributions.editor3.annotationInputTabs');
```
so it will be simpler to write tests for code allowing extensions

tomaskikutis: I agree, but I don't think it's possible to do it in a straight-forward way and also preserve static-typing. It generates a few lines of boring boilerplate, but there's no complexity there, so I'm not worried about it.

scripts/extensions/annotationsLibrary/package.json
pablopunk: okay, good to know 👍

scripts/extensions/annotationsLibrary/src/extension.tsx
pablopunk: I actually learned first async functions than Promise.resolve so maybe I'm biased 😁 But still I think it's way more readable, returning a promise is only a way of saying the function is async. Also, with async you make sure you always have a promise as a result. If the function has several `return`, one of them might be a promise, other one can be a value...

scripts/core/superdesk-api.d.ts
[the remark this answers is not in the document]
pablopunk: yeah I get your point on extensions but I was just confused by the naming, not a big deal tho, I didn't think of any better naming

scripts/extensions/annotationsLibrary/src/extension.tsx
tomaskikutis: > Also, with async you make sure you always have a promise as a result.

TypeScript makes sure of this for me. If you declare that function returns a promise, it won't compile if you try to return a value.

> I think it's way more readable

I'd argue that for most developers it's less readable. For 20 years JavaScript has been around if you saw `return {a: 1}` you could be 99% sure that the function was synchronous. 1% goes to people who don't know what they are doing and write Schrodinger functions which are used in both ways at the same time.

Still not buying it :smile:

pablopunk: Yep typescript mitigates that issue, I was talking about Javascript in general. Anyway use whatever you want 😄 I just think dealing with promises is kind of like doing the work `async/await` abstract you from doing, even tho know it is doing the same thing. Kind of like writing a `class` instead of a `function`

tomaskikutis: It's not about what _I want_, but rather what enables us to write better code. I think we agree that in this case, the main difference is in syntax. I assume people will be more familiar with promises since that's what we currently use and promises have been around for longer. If you convince others I don't mind starting using async/await in the project :slightly_smiling_face:

PR #2973, opened 2019-05-24
fix default edit featured image config

petrjasek:
it should be on by default, only off when it's set to false

### DIFF
--- a/scripts/core/config.ts
+++ b/scripts/core/config.ts
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 
 export function isMediaEditable(config) {
-    return _.get(config, 'features.editFeaturedImage') !== false;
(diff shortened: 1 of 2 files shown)


scripts/core/config.ts
```diff
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 
 export function isMediaEditable(config) {
-    return !!_.get(config, 'features.editFeaturedImage');
+    return _.get(config, 'features.editFeaturedImage', true) !== false;
```
tomaskikutis: Is there a reason you use `!== false` instead of `=== true`?

petrjasek: to make it enabled when it's not set

petrjasek: actually, I'm not sure now

PR #2975, opened 2019-05-28
add api for custom field types

petrjasek:
it allows adding new field types which can be then used in metadata
settings to create new custom fields.

field type has editor and preview component.

SDBELGA-86


petrjasek:
usage https://github.com/superdesk/superdesk-belga/pull/47

scripts/superdesk-interfaces/Vocabulary.ts
```diff
@@ -16,10 +16,11 @@ export interface IVocabulary {
...
     unique_field: string;
     schema: {};
-    field_type: 'text' | 'media' | 'date' | 'embed' | 'related_content';
+    field_type: 'text' | 'media' | 'date' | 'embed' | 'related_content' | 'custom';
     field_options: {};
     init_version: number;
     preffered_items: boolean;
     tags: Array<IVocabularyTag>;
     date_shortcuts: Array<{value: number; term: string; label: string}>;
+    custom_field_type: string;
```
tomaskikutis: Are there any advantages to adding another field instead of converting `field_type` to string and storing custom field types there as well?

scripts/core/ui/components/drop-zone.tsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import classnames from 'classnames';
+
+interface IProps {
+    text: string;
+    allowedTypes: Array<string>;
+    onDrop: ({data: any}) => any;
```
tomaskikutis: `onDrop` should probably return `void`.

scripts/apps/fields/custom-fields-provider.ts
```diff
@@ -0,0 +1,33 @@
...
+    constructor() {
+        this.fields = {};
+    }
+
+    add(fieldId: string, options: IFieldOptions) {
+        this.fields[fieldId] = options;
+    }
+}
+
+export const fields = new CustomFields();
```
tomaskikutis: Accessing fields looks ugly `fields.fields`. What about getting rid of the class?

```
export const fields: fields: {[key: string]: IFieldOptions} = {};
export const addField = (fieldId: string, options: IFieldOptions) => {
    fields[fieldId] = options;
}
```

scripts/apps/authoring/authoring/preview-custom-field.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export class PreviewCustomField extends React.PureComponent<IProps> {
+    render() {
+        const {item, field} = this.props;
+
+        const FieldType = fields.fields[field.custom_field_type];
+
+        if (FieldType == null) {
+            console.warn('unkwnow custom type', field.custom_field_type);
```
tomaskikutis: I would use `logger.warn` so we can see it in sentry.

scripts/superdesk-interfaces/Vocabulary.ts
petrjasek: that's for backward compatibility

scripts/core/ui/components/drop-zone.tsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import classnames from 'classnames';
+
+interface IProps {
+    text: string;
+    allowedTypes: Array<string>;
```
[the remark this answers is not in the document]
petrjasek: yep that would be more flexible, will add it and maybe some helper to work with allowedTypes

scripts/superdesk-interfaces/Vocabulary.ts
tomaskikutis: Could you explain more? What do you want this to be compatible with? Client repositories? Or you want to be able to implement this without having to refactor too much of server/client code. Would you need to refactor a lot if you wanted to avoid extra field and store additional field types in the existing field?

petrjasek: actually it would probably work, not sure what I was thinking about. the only issue might be with the metadata settings UI where each new type has its own tab, and we show/hide some form fields based on it

tomaskikutis: I don't see a reason why shouldn't we be able to generate the metadata tabs too. We can keep `custom_field_type` if it proves to be too much work to have it in the main field, but see if you can do that in a straight-forward way.

scripts/apps/authoring/authoring/preview-custom-field.tsx
```diff
@@ -13,16 +13,14 @@ interface IProps {
 export class PreviewCustomField extends React.PureComponent<IProps> {
     render() {
         const {item, field} = this.props;
-
-        const FieldType = fields.fields[field.custom_field_type];
+        const FieldType = fields[field.custom_field_type];
+        const value = get(item.extra, field._id);
```
tomaskikutis: What does `lodash.get` do differently here than would `const value = item.extra[field._id]`?

petrjasek: it won't fail if extra is not defined, that's about it

scripts/apps/fields/custom-fields-provider.ts
```diff
@@ -18,16 +18,8 @@ interface IFieldOptions {
...
-        this.fields = {};
-    }
-
-    add(fieldId: string, options: IFieldOptions) {
-        this.fields[fieldId] = options;
-    }
-}
-
-export const fields = new CustomFields();
+export const customField = (fieldId: string, options: IFieldOptions) => {
```
tomaskikutis: `addCustomField` sounds more intuitive. I think in most cases having a verb in a function name is desirable.

tomaskikutis:
How is it going? I assume you are still working on this?

petrjasek:
I still have to make it an extension

tomaskikutis:
That's what I thought, just wanted to make sure that we aren't waiting for each other. It would be clearer if you only pushed changes when there is nothing more to address.

scripts/apps/authoring/authoring/index.ts
```diff
@@ -107,6 +109,20 @@ angular.module('superdesk.apps.authoring', [
         ),
     )
 
+    .component('sdAuthoringCustomField',
+        reactToAngular1(
+            AuthoringCustomField,
+            ['item', 'field', 'editable', 'onChange'],
```
tomaskikutis: This doesn't match with component props

```
item: IArticle;
field: IVocabulary;
onChange: (item: Partial<IArticle>, time: number) => any;
readonly: boolean;
```

scripts/apps/authoring/authoring/preview-custom-field.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export class PreviewCustomField extends React.PureComponent<IProps> {
+    render() {
+        const fields = getFields();
+        const {item, field} = this.props;
+        const FieldType = fields[field.custom_field_type];
+        const value = get(item.extra, field._id);
+
+        if (FieldType == null) {
+            console.warn('unkwnow custom type', field);
```
tomaskikutis: use `logger.warn`

scripts/apps/authoring/authoring/preview-custom-field.tsx
```diff
@@ -0,0 +1,35 @@
...
+            return null;
+        }
+
+        if (isEmpty(value)) {
```
tomaskikutis: Isn't `!= null` enough?

scripts/superdesk-interfaces/Vocabulary.ts
tomaskikutis: Have you looked into this?

scripts/apps/authoring/authoring/authoring-custom-field.tsx
```diff
@@ -0,0 +1,47 @@
...
+    onChange: (item: Partial<IArticle>, time: number) => any;
+    readonly: boolean;
+}
+
+export class AuthoringCustomField extends React.PureComponent<IProps> {
+    setValue(value) {
+        const extra = Object.assign({}, this.props.item.extra);
+
+        extra[this.props.field._id] = value || null;
+        this.props.item.extra = extra;
```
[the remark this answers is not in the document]
petrjasek: it's problematic for storage if we add new fields dynamically, also to avoid users overriding some metadata which we need.

scripts/core/superdesk-api.d.ts
```diff
@@ -303,4 +331,25 @@ declare module 'superdesk-api' {
...
+        setValue: (value: any) => void;
+        readOnly: boolean;
+    }
+
+    export interface IPreviewComponentProps {
+        item: any;
+        value: any;
+    }
+
+    export interface ICustomFieldType {
```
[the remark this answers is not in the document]
petrjasek: there can be multiple fields created using same field type, so imo it's better like this.
eg. field type date and fields would be created, updated, published etc.

scripts/superdesk-interfaces/Vocabulary.ts
petrjasek: I'm still not sure it would be better, instead of explicit check if `field_type === 'custom'` I would have to use `field_type not in ('text', 'media', 'urls', 'date', 'embed', 'related_content')` and there would be no type check on `field_type` possible

petrjasek: but I would do some refactoring anyhow, eg those metadata tabs can be generated anyhow

scripts/apps/authoring/authoring/authoring-custom-field.tsx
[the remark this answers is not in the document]
petrjasek: but anyhow that value must be stored somewhere, and imo the component handling it is the place to do it, where else?

scripts/apps/authoring/authoring/preview-custom-field.tsx
petrjasek: no idea what values might be stored there eventually, could be `{}`, `[]`, `''` and that still would probably mean no preview

scripts/apps/authoring/authoring/authoring-custom-field.tsx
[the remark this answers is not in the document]
tomaskikutis: > it's problematic for storage if we add new fields dynamically

Dynamic fields is a problem with pyeve, not anything else, right? What about not adding fields dynamically? Are there situations where it must be added dynamically?

Regarding unintentional overwriting of fields, what if an extension or two client repos want to add the same field to `extra`? It's not that unrealistic as we see with assignments when 2 clients have different implementations. It would be pretty easy to solve by requiring client instances/extensions to statically declare which fields they want to access and we could then verify on build that there are no conflicts.

> but anyhow that value must be stored somewhere, and imo the component handling it is the place to do it, where else?

I didn't get the question. Could you rephrase it? It's not a problem that you set the fields in this component, I was only commenting on `extra` field and not mutating the object,

scripts/superdesk-interfaces/Vocabulary.ts
tomaskikutis: > I'm still not sure it would be better, instead of explicit check if field_type === 'custom' I would have to use field_type not in ('text', 'media', 'urls', 'date', 'embed', 'related_content') and there would be no type check on field_type possible

The idea I have for the future of this is to remove the difference between custom and non-custom fields. It doesn't have to be a separate extension for separate fields, but if it used the same interface, there would be no difference from the code point of view. It's easier to maintain when you need to work against a single interface rather than x types with different logic.

I've implemented such thing in [multi action bar](https://github.com/superdesk/superdesk-client-core/pull/2989/files#diff-80f5f9652545a6cdc07d3240a7a4b60aR32). There is an interface to define an action for multiple items that extensions can implement, but I also moved existing items on the same interface.

scripts/core/superdesk-api.d.ts
```diff
@@ -204,6 +204,29 @@ declare module 'superdesk-api' {
         noBorder?: boolean;
     }
 
+    export interface IGridComponentProps {
+        columns: number;
+        boxed?: boolean;
+        children: React.ReactNodeArray;
```
tomaskikutis: You don't need `children`. https://github.com/superdesk/superdesk-client-core/pull/2975#discussion_r292519518

scripts/superdesk-interfaces/Article.ts
```diff
@@ -1,52 +1 @@
...
-    rewritten_by?: string;
-
-    // might be only used for client-side state
-    created: any;
-    archived: any;
-
-    // TODO: remove this, it's only used for client-side state
-    selected: any;
-}
+export {IArticle} from 'superdesk-api';
```
tomaskikutis: Don't be lazy, update the imports and remove the file :smile:

scripts/superdesk-interfaces/Author.ts
```diff
@@ -1,4 +1 @@
-export interface IAuthor {
-    role: string;
-    parent: string;
-}
+export {IAuthor} from 'superdesk-api';
```
tomaskikutis: remove file

scripts/apps/authoring/authoring/authoring-custom-field.tsx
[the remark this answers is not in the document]
petrjasek: it will be added dynamically, we have UI for users to add new fields, not sure if we can avoid that

scripts/core/superdesk-api.d.ts
petrjasek: it doesn't work if I remove it, typescript would complain about propTypes not matching IProps and lint fails

scripts/apps/authoring/authoring/authoring-custom-field.tsx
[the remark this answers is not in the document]
tomaskikutis: You mean avoid unintentional overwriting of fields? Sure we can. We can keep a list of static fields in `package.json` or other file and validate that between client repositories/extensions and we can have dynamic fields in the DB which we can check against static fields on application start.

Or we can have one namespace for static and one for dynamic fields which was probably the idea behind `extra`, but it doesn't work unless `extra` reserved for dynamic fields only. Is that the case now?

scripts/core/superdesk-api.d.ts
tomaskikutis: Right, I forgot about prop-types. I don't think there's a point to keep those in TypeScript files. Do you see any use for them?

scripts/apps/authoring/authoring/authoring-custom-field.tsx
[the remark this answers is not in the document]
petrjasek: yes `extra` is only for dynamic fields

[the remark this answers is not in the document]
tomaskikutis: Let's keep it then. Can you add the comment in the interface that explains it?

scripts/superdesk-interfaces/Vocabulary.ts
petrjasek: yep that's something I'm also having in mind, all fields should use the same api

scripts/core/superdesk-api.d.ts
petrjasek: prop-types can be still useful for runtime checking if you work with some api data, but personally I don't remember when it would find an issue

tomaskikutis: > useful for runtime checking if you work with some api data

That's a good point since we often don't know what is the shape of the data, especially for the archive collection. We often have a property specified as a generic `dict` in the back-end schema, and put anything we want from the client-side there. I don't think the UI is a good place to solve this, I'd rather put work into making back-end schema more specific and adding back-end validation ensuring that responses respect that schema.

scripts/apps/authoring/authoring/authoring-custom-field.tsx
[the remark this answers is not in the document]
tomaskikutis: We sidetracked into talking about the `extra` field and forgot about mutation. Can you update the `setValue` function so it doesn't mutate props? We shouldn't mutate props of React components in general. We might have to in some dirty angular integration code, but I think you should be able to easily avoid it here.

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -1212,6 +1212,15 @@ export function AuthoringDirective(superdesk, superdeskFlags, authoringWorkspace
...
                     });
             };
+
+            $scope.setCustomValue = (field, value) => {
+                const extra = Object.assign({}, $scope.item.extra);
+
+                extra[field._id] = value || null;
+                $scope.item.extra = extra;
+
+                $scope.autosave($scope.item, 200);
```
tomaskikutis:
```
$scope.autosave({
    ...$scope.item,
    extra: {
        ...$scope.item.extra,
        [field._id]: value || null,
    },
}, 200);
```

I thought you would use immutable update instead of moving the mutation :) Avoiding mutations when easy is beneficial even outside React. I don't mind the move though, it looks a bit nicer that components added via client API don't manipulate extra `directly`.

pablopunk:
I'm reviewing this BTW 🙏

scripts/apps/workspace/content/services/ContentService.ts
```diff
@@ -324,7 +324,7 @@ export function ContentService(api, superdesk, templates, desks, packages, archi
             self._fieldsPromise = api.getAll('vocabularies', {
                 where: {
                     $or: [
-                        {field_type: {$in: ['text', 'date', 'media', 'embed', 'urls']}},
+                        {field_type: {$in: ['text', 'date', 'media', 'embed', 'urls', 'custom']}},
```
pablopunk: I feel like this array should be inside `../constants`, it's sometimes hard to find the different values a field can have, and `constants` seems like a centralized solution we already have

tomaskikutis:
`import { constants } from 'zlib';`

Your import seems wrong.

petrjasek:
yep that was some auto import trying to fix my code, I've removed it

PR #2976, opened 2019-05-28
Grammalecte & Leuven spellcheckers

tomaskikutis:
~~There's a strange HTTP API issue. The `spellchecker` endpoint sometimes returns 404 with the same request payload which returned 200 in the previous request. [see gif](https://user-images.githubusercontent.com/2076953/58504287-88141c00-818a-11e9-8b68-267e866ed08e.gif) It's easily reproducible by creating an article in French and deleting and adding the same letter in editor3~~


scripts/core/editor3/components/spellchecker/default-spellcheckers.tsx
```diff
@@ -33,29 +33,89 @@ function check(str: string): Promise<Array<ISpellcheckWarning>> {
...
+                    };
+
+                    xhr.send(JSON.stringify({spellchecker: spellcheckerName, text: str}));
+                });
+            }),
+            getSuggestions: (str) => ng.getServices(['config', 'session']).then((services: Array<any>) => {
+                const [config, session] = services;
+
+                return new Promise((resolve, reject) => {
+                    const xhr = new XMLHttpRequest();
```
petrjasek: this xhr handling is duplicated in both methods, and imo it shouldn't be in spellchecker anyhow

scripts/core/editor3/components/spellchecker/default-spellcheckers.tsx
```diff
@@ -33,29 +34,59 @@ function check(str: string): Promise<Array<ISpellcheckWarning>> {
...
+        return {
+            check: (str: string) => new Promise((resolve, reject) => {
+                httpRequestJsonLocal<{errors: Array<ISpellcheckWarning>}>({
+                    method: 'POST',
+                    payload: {spellchecker: spellcheckerName, text: str},
+                    path: '/spellchecker',
+                    onSuccess: (json) => {
+                        resolve(json.errors);
+                    },
+                    onError: () => reject(),
```
pablopunk: I think `reject` should include the error so we can `.catch(err)`. This should be enought:

```js
onSuccess: ...,
onError: reject,
```

scripts/core/editor3/components/spellchecker/default-spellcheckers.tsx
```diff
@@ -33,29 +34,59 @@ function check(str: string): Promise<Array<ISpellcheckWarning>> {
...
+            }),
+            getSuggestions: (str) => new Promise((resolve, reject) => {
+                httpRequestJsonLocal<ISpellcheckWarning>({
+                    method: 'POST',
+                    payload: {spellchecker: spellcheckerName, text: str, suggestions: true},
+                    path: '/spellchecker',
+                    onSuccess: (spellcheckerWarning) => {
+                        resolve(spellcheckerWarning.suggestions);
+                    },
+                    onError: () => reject(),
```
pablopunk: Same here

scripts/core/editor3/components/spellchecker/default-spellcheckers.tsx
tomaskikutis: I already log the error in `function httpRequest(options: IHttpRequestOptions)`. If I were to pass an argument to `reject` I should have a use for it, which I don't. Even if we were to display a message to a user, the only relevant thing the message could say is that there was a network error, which is helpful but would be better off handled at the application level, not at feature/component level. I've created `SDESK-4464` to address that.

scripts/core/helpers/xhr.ts
```diff
@@ -0,0 +1,107 @@
...
+
+interface IHttpRequestReturnValue {
+    abort(): void;
+}
+
+export function httpRequest(options: IHttpRequestOptions): IHttpRequestReturnValue {
+    const {method, url, payload, headers, onSuccess, onError} = options;
+    const xhr = new XMLHttpRequest();
+
+    xhr.open(method, url, true);
```
petrjasek: imo using [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) would be better, it's supported in the browsers we care about

tomaskikutis: The interface doesn't expose any details on which API is being used to talk to the network, so changing which API is being used without changing the interface won't change anything.

What benefits do you expect to gain by using `fetch`?

petrjasek: it's cleaner api than xhr, uses promises

tomaskikutis: Cleanliness is [relative](https://gomakethings.com/why-i-still-use-xhr-instead-of-the-fetch-api/) and isn't relevant if you don't interact with it directly as is the case here. 

I've added promise-based versions for all helpers.

petrjasek: imo having less code is better, so if we don't need promise-less version I don't see any reason to have it, nor to use xhr, why would we do that?
it could be:
```
export function httpRequest(options): Promise {
    const {method, url, payload, headers} = options;

    return fetch(url, {
        method,
        headers,
        body: payload != null ? JSON.stringify(payload): null,
    })
        .then((response) => response.ok ? response : Promise.reject(response))
        .catch((reason) => logger.error(...));

};
```

PR #2977, opened 2019-05-29
Build extensions from client repositories

tomaskikutis:
Extensions are now first-class citizens in client repositories. Everything that works in client core will work from client repositories as well, including webpack dev server and live reload :tada: 

Extensions should be placed at `superdesk/client/extensions/myExtension`

The only difference is the path to api typings which will be
```
/// <reference path='../../../../node_modules/superdesk-core/scripts/core/superdesk-api.d.ts' />
```

assuming the line above is located in `superdesk/client/extensions/myExtension/src/typings/refs.d.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "typescript-eslint-parser": "^18.0.0"
   },
   "scripts": {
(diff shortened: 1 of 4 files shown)


petrjasek:
could you add some hello world extension also to superdesk/superdesk repo?

tomaskikutis:
Sure https://github.com/superdesk/superdesk/pull/3092

PR #2978, opened 2019-05-29
[SDESK-4310] [SDESK-4313] Embed and style fixes

fritzSF:


styles/sass/archive-preview.scss
```diff
@@ -384,16 +361,27 @@
     }
 }
 
+.workspace .main-section .content-item-preview {
+    --embed-width: 52rem;
+}
+.workspace.authoring .main-section .content-item-preview {
+    --embed-width: 35.2rem;
+}
+.workspace .main-section.with-preview [sd-item-preview].sd-preview--overlay .content-item-preview {
```
tomaskikutis: Don't use names of directives in selectors like `[sd-item-preview]`.

It looks like we aren't using CSS variables in the codebase yet, can we stick with sass variables?

The entire matching looks fragile in a way that it depends on a lot of things. Doesn't it work setting width to 100% and letting it adjust to the parent container?

fritzSF: It doesn't work that's why it is added. CSS variables are different than sass variables as they are not compiled and they come handy in some cases. We are using them in the framework recently.

fritzSF: As for the [sd-item-preview] selector, I had to use what I have. This part is a bit messy but this selector is used also on other places.

tomaskikutis: Petr mentioned there's a script for controlling embed dimensions. Could you then remove CSS related to width/iframe?

PR #2979, opened 2019-05-30
fix: Changes in subscribers content API tab are not activating the save button. (SDESK-4194)

GyanP:


scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -262,13 +264,15 @@ export function SubscribersDirective(
                 $scope.newDestination = null;
             };
 
-            $scope.$watchCollection('subscriber', (newValue, oldValue) => {
-                if (newValue && oldValue && newValue !== oldValue) {
+            $scope.$watch('subscriber', (newValue, oldValue) => {
+                if (newValue && oldValue && !_.isEqual(newValue, oldValue)) {
```
tomaskikutis: Why isn't it enough to handle this here? Why do you need `enableSave` method?

GyanP: Tokens are stored in different collection and making any changes to them does not affect `subscriber`. So actually, there is no need to enable the save button on token generation. But Karel said that we must enable it from a users point of view.

scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -262,13 +264,15 @@ export function SubscribersDirective(
...
                     $scope.saveEnabled = true;
                 } else {
                     $scope.saveEnabled = false;
                 }
-            });
+            }, true);
```
petrjasek: with deep watch I don't see a need for `isEqual` call, it should be always false

scripts/apps/publish/controllers/SubscriberTokenController.ts
```diff
@@ -18,6 +18,11 @@ export function SubscriberTokenController($scope, api) {
             api.query('subscriber_token', {where: {subscriber: subscriber._id}})
                 .then((response) => {
                     this.tokens = response._items;
+                    $scope.$watch(() => this.tokens, (newVal, oldVal) => {
```
petrjasek: I guess you can watch `this.tokens.length` and avoid that if in the listener

scripts/apps/publish/directives/SubscribersDirective.ts
tomaskikutis: Can you watch both collections and enable save button when any of them changes? I assume that would allow to get rid of `enableSave` method. Or at least only keep it in one place only and not use it in the UI templates.

GyanP: Both collections can't be watched together as one of them is the part of `controller as` syntax. But I have removed the `enableSave` method.

scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -146,13 +146,15 @@ export function SubscribersDirective(
             $scope.saveNewDestination = function() {
                 $scope.destinations.push($scope.newDestination);
                 $scope.newDestination = null;
+                $scope.saveEnabled = true;
```
tomaskikutis: Is there a reason this can't be covered by a `watch`? Save buttons should be enabled/disabled when data changes, not when you explicitly set a variable. Setting a variable might also work for a while, but it's fragile in a way that people need to remember to set the variable. Why not take this complexity away by managing it in a single place?

GyanP: For that, we have to deep watch the destinations collections also, that's why I avoided it.

tomaskikutis: What's wrong with deep watch? Was it too slow?

GyanP: It wasn't too slow. But a deep watch is expensive. I will change it.

tomaskikutis: Test it out, if the expensive operations don't slow the code down, but are required for better code - let's use them, especially when the code is rarely executed. In this case, it will only run when editing subscribers which isn't what users do every day.

PR #2980, opened 2019-05-30
feat(monitoring): Custom sort for a single/multiple group in monitoring

pablopunk:
SDNTB-545

New config option:

```js
// superdesk.config.js
{
  ...
  monitoring: {
    scheduled: {
      sort: [ 'published_schedule' ] // array of fields to sort
    },
    stage: {
      sort: [ 'headline', 'slugline' ]
    },
    // we can keep adding stage types like 'output'
  }
}
```

Needed for https://github.com/superdesk/superdesk-ntb/pull/328

### DIFF
--- a/scripts/apps/monitoring/directives/MonitoringGroup.ts
+++ b/scripts/apps/monitoring/directives/MonitoringGroup.ts
@@ -587,7 +587,7 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
             scope.selectCustomSortOption = function(field: string) {
                 scope.customSortOptionActive = {
                     field,
-                    order: 'desc'
+                    order: 'desc',
                 };
                 queryItems();
             };
--- a/scripts/apps/monitoring/directives/MonitoringGroup.ts
+++ b/scripts/apps/monitoring/directives/MonitoringGroup.ts
@@ -123,12 +123,12 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
 
             const customSort = getCustomSortForGroup(config, scope.group);
 
-            if (customSort) {
+            if (customSort) {
                 scope.customSortOptions = {
                     ...customSort,
                     [DEFAULT_SORT_FIELD]: {label: 'Default'},
                 };
-                scope.customSortOptionActive = { field :DEFAULT_SORT_FIELD, order: 'desc' };
+                scope.customSortOptionActive = {field: DEFAULT_SORT_FIELD, order: 'desc'};
             }
 
             scope.page = 1;
@@ -521,7 +521,8 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
 
                         // custom sort for group (if it exists)
                         if (scope.customSortOptionActive && scope.customSortOptionActive.field !== DEFAULT_SORT_FIELD) {
-                            criteria.source.sort = [{[scope.customSortOptionActive.field]: scope.customSortOptionActive.order}];
+                            criteria.source.sort =
+                                [{[scope.customSortOptionActive.field]: scope.customSortOptionActive.order}];
                         }
 
                         return apiquery(criteria, true)
@@ -568,7 +569,7 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
             }
 
             scope.selectCustomSortOption = function(field: string) {
-                scope.customSortOptionActive = { field, order: 'desc' };
+                scope.customSortOptionActive = {field, order: 'desc'};
                 queryItems();
             };
 
--- a/scripts/apps/monitoring/directives/MonitoringGroup.ts
+++ b/scripts/apps/monitoring/directives/MonitoringGroup.ts
@@ -1,13 +1,13 @@
 /* eslint-disable complexity */
 import _ from 'lodash';
-import getCustomSortForGroup, { GroupSortOptions, DEFAULT_SORT_FIELD } from '../helpers/CustomSortOfGroups'
+import getCustomSortForGroup, {GroupSortOptions, DEFAULT_SORT_FIELD} from '../helpers/CustomSortOfGroups';
 
-export interface StageGroup {
+export type StageGroup = {
     _id?: any;
     type?: string;
     search?: 'ingest';
     max_items?: any;
-}
+};
 
 interface IScope extends ng.IScope {
     customDataSource: {
@@ -123,14 +123,15 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
             ITEM_HEIGHT = search.singleLine ? 29 : 57;
 
             const customSort = getCustomSortForGroup(config, scope.group);
+
             if (customSort) {
                 scope.customSortOptions = {
                     ...customSort,
-                    [DEFAULT_SORT_FIELD]: { label: 'Default' }
+                    [DEFAULT_SORT_FIELD]: {label: 'Default'},
                 };
                 scope.customSortOptionActive = DEFAULT_SORT_FIELD;
                 scope.customSortOrder = 'desc';
-            };
+            }
 
             scope.page = 1;
             scope.fetching = false;
@@ -489,7 +490,6 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
                     multi.reset();
                 }
 
-
                 return (function() {
                     if (scope.customDataSource != null && typeof scope.customDataSource.getItems === 'function') {
                         return scope.customDataSource.getItems(0, PAGE_SIZE);
@@ -523,7 +523,7 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
 
                         // custom sort for group (if it exists)
                         if (scope.customSortOptionActive && scope.customSortOptionActive !== DEFAULT_SORT_FIELD) {
-                            criteria.source.sort = [ { [scope.customSortOptionActive]: scope.customSortOrder } ]
+                            criteria.source.sort = [{[scope.customSortOptionActive]: scope.customSortOrder}];
                         }
 
                         return apiquery(criteria, true)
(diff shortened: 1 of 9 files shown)


scripts/apps/monitoring/views/monitoring-group-header.html
```diff
@@ -24,6 +24,23 @@
         <button ng-if="showRefresh" class="btn btn--primary btn--icon-only-circle btn--small" ng-click="refreshGroup(group)" tooltip="{{:: 'Refresh' | translate}}" tooltip-placement="left">
             <i class="icon-repeat"></i>
         </button>
+        <div class="sortbar" ng-show="customSortOptions">
+            <div class="dropdown dropdown--align-right dropdown--hover" dropdown__toggle>
+                <button id="order_button" class="dropdown__toggle" dropdown__toggle>
+                    <span translate>{{ customSortOptions[customSortOptionActive].label }}</span>
```
tomaskikutis: No need to translate _here_. It's a dynamic value which should come here already translated.

scripts/apps/monitoring/helpers/CustomSortOfGroups.ts
```diff
@@ -0,0 +1,24 @@
+import {get} from 'lodash';
+import {StageGroup} from '../directives/MonitoringGroup';
+
+export const CUSTOM_SORT_SETTING = 'monitoring.customSortOfGroups';
+export const DEFAULT_SORT_FIELD = 'default';
```
tomaskikutis: Doesn't look like a field name. Wouldn't it be better to have `last_updated` or what's the actual default value?

scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -25,21 +33,21 @@ interface IScope extends ng.IScope {
...
-        _id?: any;
-        type?: string;
-        search?: 'ingest';
-        max_items?: any;
-    };
     currentGroup: any;
     fetchNext(i: number): any;
     refreshGroup(): void;
+    customSortOptions: GroupSortOptions;
+    customSortOptionActive?: string;
```
tomaskikutis: How about storing related data in the same field? `selectedSortOption: {field: string; direction: 'asc' | 'desc'}`

scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -114,6 +122,17 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
 
             ITEM_HEIGHT = search.singleLine ? 29 : 57;
 
+            const customSort = getCustomSortForGroup(config, scope.group);
+
+            if (customSort) {
+                scope.customSortOptions = {
+                    ...customSort,
+                    [DEFAULT_SORT_FIELD]: {label: 'Default'},
```
tomaskikutis: Translate label. `{label: gettext('Default')},`

scripts/apps/monitoring/views/monitoring-group-header.html
```diff
@@ -24,6 +24,23 @@
...
+                    <span class="dropdown__caret"></span>
+                </button>
+                <ul class="dropdown__menu">
+                    <li ng-repeat="(field, option) in customSortOptions">
+                        <button ng-click="selectCustomSortOption(field)" translate>{{ option.label }}</button>
```
tomaskikutis: No need to translate here. It's a dynamic value which should come here already translated.

pablopunk:
> `label` should be a function which returns a string, since you can't translate an inline string.
> 
> `'publish_schedule': { label: (gettext) => gettext('Scheduled time' )} // field-to-sort: { ui-label }`
> 
> --
> 
> Does sorting groups other than output work? I know it requires a config, but if you added it, would it work without doing anything else?

What do you mean I can't translate an inline string?

And yes, you can sort other groups than output. If you put something like `'stage': { 'headline': {label} }` you could sort stages by headline while maintaining a global sort.

scripts/apps/monitoring/views/monitoring-group-header.html
pablopunk: Can't we have dynamic values translated with `<div translate>`?

scripts/apps/monitoring/helpers/CustomSortOfGroups.ts
pablopunk: This value just means that if it's selected, the sort will match the global sort (so, no custom sort used, which is the default behaviour)

tomaskikutis:
> What do you mean I can't translate an inline string?

There's a script which extracts translation strings from the code. The extraction only works when strings are directly either in `gettext` calls or in angular template elements marked with `translate` attribute. You can use the following: `gettext('mystring')` or `<div translate>mystring</div>`(only in angular templates).

What you can't do is this:
```
var translationString = 'mystring';
var translated = gettext(translationString);
```

The extraction script isn't smart, probably regex, so it won't be able to track your usages.

pablopunk:
@tomaskikutis so we need to include the function in `superdesk.config.js`?

tomaskikutis:
Nope. Instead of returning a string, you can return a function:

```
'scheduledDeskOutput': {  // type of group to sort
      'publish_schedule': { label: (gettext) => gettext('Scheduled time' )} // field-to-sort: { ui-label }
}
```

You will then have to pass `gettext` function from the client to retrieve the label.

tomaskikutis:
I've read your question incorrectly. Yes, you do need to return the function from the config :) Same as I described.

pablopunk:
@tomaskikutis that seems weird tho, we should be able to just put a string that can then be translated dynamically. Anyway, I'll change it until we can do it that way

tomaskikutis:
We can translate things dynamically, but we need to be able to send strings for translators, which means we need to somehow mark them in the code so they can be extracted by the script.

You could mark the strings themselves like `const toBeTranslated = '__Click here'`; and later call `translate(toBeTranslated)`. You would be then able to regex the strings by double underscore prefix and extract it for sending to translators. The current method is cleared IMO since you do translation and extraction via a single function. Which means when you have a variable, you don't have to worry if you have a key which needs to be translated or a string which is already translated.

scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -126,7 +126,7 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
             if (customSort) {
                 scope.customSortOptions = {
                     ...customSort,
-                    [DEFAULT_SORT_FIELD]: {label: 'Default'},
+                    [DEFAULT_SORT_FIELD]: {label: gettext('Default')},
```
tomaskikutis: This won't work. You need it to be `[DEFAULT_SORT_FIELD]: {label: (gettext) => gettext('Default')},`

tomaskikutis: Sorry, I thought this was a config... It's not, so it's fine.

pablopunk: Yeah right here the `customSort` object has all the labels translated already

pablopunk:
@tomaskikutis but we still have the ability to do so with the translate filter for example, like here:

https://github.com/superdesk/superdesk-client-core/blob/0f636723a844d1a9cb8b2301185a8d6bd0a015c0/scripts/apps/search/views/search-parameters.html#L47

I know we don't use it anymore tho

tomaskikutis:
> but we still have the ability to do so with the translate filter for example, like here:

That's why we don't use it. because it doesn't work :laughing: It translates the string only if the same string has been extracted previously(and sent to transifex) due to using 
> gettext('mystring') or <div translate>mystring</div>(only in angular templates)

scripts/apps/monitoring/helpers/CustomSortOfGroups.ts
tomaskikutis: How about `const SORT_FIELD = null;` instead? It makes it explicit that the field may not be set and avoids the confusion of using a non-existing field name.

pablopunk: But it makes it simpler for the [UI](https://github.com/superdesk/superdesk-client-core/pull/2980/files#diff-4883afa01e282c9d8acc43e8c8a31952R34), because I can just show the fields to sort with this default included, and then the logic is in [the directive](https://github.com/superdesk/superdesk-client-core/pull/2980/files#diff-a65e3fd2ffce9f208245c22a7d5e2639R523)

tomaskikutis: It doesn't look like a lot of work to adjust it to work with `null`. Clean logic is more important in my opinion than saving a few lines of code by using a non-existing field.

scripts/apps/monitoring/views/monitoring-group-header.html
```diff
@@ -27,16 +27,20 @@
...
             <div class="dropdown dropdown--align-right dropdown--hover" dropdown__toggle>
                 <button id="order_button" class="dropdown__toggle" dropdown__toggle>
-                    <span>{{ customSortOptions[customSortOptionActive].label }}</span>
+                    <span ng-if="!customSortOptionActive.field" translate>Default</span>
+                    <span ng-if="customSortOptionActive.field">{{ customSortOptions[customSortOptionActive.field].label }}</span>
                     <span class="dropdown__caret"></span>
                 </button>
                 <ul class="dropdown__menu">
+                    <li>
+                        <button ng-click="selectCustomSortOption(customSortOptionActiveDefault)" translate>Default</button>
```
tomaskikutis: isn't default always `null`? If it is, I'd rather use it inline instead of creating a constant which adds additional lookup step when reading the code, but doesn't add any clarity - inline `null` would be as clear.

scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -520,7 +535,8 @@ export function MonitoringGroup(cards, api, authoringWorkspace, $timeout, superd
                         }
 
                         // custom sort for group (if it exists)
-                        if (scope.customSortOptionActive && scope.customSortOptionActive.field !== DEFAULT_SORT_FIELD) {
+                        if (scope.customSortOptionActive &&
+                            scope.customSortOptionActive.field !== USE_DEFAULT_SORT_FIELD) {
```
tomaskikutis: what about having entire `customSortOptionActive` equal to `null` when a custom sort is not set? Having it set and having a field equal `null` is less clear.

PR #2981, opened 2019-05-31
fixed: Removing a filter statement from a content filter does not enable the save button. (SDESK-4303)

GyanP:


scripts/apps/content-filters/controllers/ManageContentFiltersController.ts
```diff
@@ -235,6 +235,18 @@ export function ManageContentFiltersController($scope, contentFilters, notify, m
...
+            if (newVal && oldVal && !_.isEqual(newVal, oldVal)) {
+                $scope.saveEnabled = true;
+            } else {
+                $scope.saveEnabled = false;
+            }
+            // if there is filter-statement in contentFilter and preview is not available then disable the save button
+            if (newVal && newVal.content_filter.length > 0 && !$scope.preview) {
+                $scope.saveEnabled = false;
+            }
+        }, true);
```
petrjasek: if you use deep watch I think there is no need for calling `isEqual`, it should be never equal if the watch is called

PR #2982, opened 2019-05-31
Fix embeds in preview

tomaskikutis:
SDESK-4310
SDESK-4333


fritzSF:
This did not fix the issue in SDESK-4333

PR #2974, opened 2019-05-24
Create planning extension

tomaskikutis:
DEPENDS ON: https://github.com/superdesk/superdesk-client-core/pull/2966

PR #2989, opened 2019-06-10
Mark for user

tomaskikutis:
SDBELGA-28

~~This PR is brached from https://github.com/superdesk/superdesk-client-core/pull/2974 and contains the same changes. I recommend reviewing the #2974 first and I'll update this PR then.~~


pablopunk:
I see a lot of `browser.sleep()` in the tests. Isn't there a way of waiting for an element to be available with a certain timeout?

scripts/apps/monitoring/MultiActionBarReact.tsx
```diff
@@ -0,0 +1,141 @@
...
+                });
+            });
+        }
+    }
+    render() {
+        if (this.state.actions == null) {
+            return null;
+        }
+
+        const onTrigger = (action: IArticleActionBulkExtended) => {
```
pablopunk: Shouldn't this be a method instead?

tomaskikutis: I don't like having methods if I can avoid it. The issue with methods is that they can be called by any other method in the class and any code outside it that has a reference to the class. That's a lot of stuff to check in order to understand which code the method can be called by. We have TypeScript now and can use "find usages" command which makes it easier, but I still like the simplicity of having a function instead of a class method and being able to immediately understand that no other code can call it.

There is one issue though. Since I defined the function in the render method, it will get recreated every time render runs. Which means child components I pass the function to, won't be able to understand, at least by comparing the reference, that it's the same function, and will trigger a re-render. To avoid this, it's better to define the function outside of the component which I'll do.

tomaskikutis:
> I see a lot of browser.sleep() in the tests. Isn't there a way of waiting for an element to be available within a certain timeout?

It doesn't always help, for example here, the element is always present. I'm thinking about a way to wait for next browser repaint or react re-render which would be a better way to do it.

```
el(['multi-actions-inline', 'Remove']).click();
browser.sleep(100);

expect(content.count()).toBe(0);	        expect(content.count()).toBe(0);
```

I'll merge this now and will address your comments in the other mark-to-user PR.

scripts/apps/monitoring/MultiActionBarReact.tsx
pablopunk: I understand. We could also start to use `private` methods to avoid instance calls

PR #2990, opened 2019-06-11
Fix parser when html had empty characters

pablopunk:
SDBELGA-120

There's an issue when we save empty html (not 0 length, just empty chars like line breaks) in the parser. Basically `draftjs.convertFromHTML` doesn't like that kind of input and returns a null content state instead of empty.

This issue affects templates like *article export* on planning, when the template result is empty.

Don't pay much attention to all the changes I made, I basically did some JSDoc->Typescript refactor wherever I could. The actual code is [this](https://github.com/superdesk/superdesk-client-core/pull/2990/files#diff-d36b15895cebd29622048cb16e7d1530R144) and [the test for it](https://github.com/superdesk/superdesk-client-core/pull/2990/files#diff-526d7b90caac219611fb5bb7ffd2229a)

### DIFF
--- a/scripts/core/editor3/html/from-html/index.ts
+++ b/scripts/core/editor3/html/from-html/index.ts
@@ -168,7 +168,7 @@ class HTMLParser {
         let contentState = initialState;
 
         contentState.getBlocksAsArray().forEach((block) => {
-            block.findEntityRanges((characterMetadata) => Boolean(characterMetadata.getEntity()), (start, _end) => {
+            block.findEntityRanges((characterMetadata) => characterMetadata.getEntity() != null, (start, _end) => {
                 const key = block.getEntityAt(start);
 
                 if (key != null) {
--- a/scripts/core/editor3/html/from-html/index.ts
+++ b/scripts/core/editor3/html/from-html/index.ts
@@ -53,11 +53,10 @@ class HTMLParser {
 
     /**
      * @name HTMLParser#createTree
-     * @param {string} html
      * @description Takes the given HTML, extracts the atomic blocks, and creates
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/html/from-html/index.ts
```diff
@@ -159,21 +163,18 @@ class HTMLParser {
...
-     * @param {ContentState} initialState
-     * @returns {ContentState}
      */
-    processLinks(initialState) {
+    processLinks(initialState: ContentState): ContentState {
         let contentState = initialState;
 
         contentState.getBlocksAsArray().forEach((block) => {
-            block.findEntityRanges((characterMetadata) => characterMetadata.getEntity(), (start, end) => {
+            block.findEntityRanges((characterMetadata) => Boolean(characterMetadata.getEntity()), (start, _end) => {
```
tomaskikutis: Is this `Boolean(characterMetadata.getEntity())` different than `characterMetadata.getEntity() != null`?

pablopunk: I hope not

tomaskikutis: If what you intend to check is whether the value is null/undefined I suggest using `!= null` since it's simple and doesn't require to read a spec to understand all possible outcomes.

pablopunk: Yeah I didn't give it much thought, this was not part of the task, it was for tsc not complaining about it, but it was like this already

tomaskikutis: Can you update it to `characterMetadata.getEntity() != null`?

PR #2991, opened 2019-06-14
Fix support for string fields

pablopunk:


petrjasek:
@pablopunk did you also push this to master branch?

PR #2996, opened 2019-06-19
Fixed: Cannot insert embeds to body. (SDESK-4386)

GyanP:


scripts/core/editor3/components/embeds/EmbedInput.tsx
```diff
@@ -66,7 +66,7 @@ export class EmbedInputComponent extends React.Component<any, any> {
      * the action that embeds the response into the editor.
      */
     processSuccess(data) {
-        this.props.embed(data);
+        this.props.embed(data.html || data);
```
tomaskikutis: How did it work for 10 months since this line was last modified? This looks like a regression. Did you check which commit broke?

GyanP: I checked but couldn't find any commit as such. Either it never worked for normal URLs, it worked only for embed code or the iframely changed the way it returns the data.

tomaskikutis: How did you search for it? Have you tried `git bisect`?

> Either it never worked for normal URLs, it worked only for embed code or the iframely changed the way it returns the data.

It has definitely worked with normal URLs. If you try it out with older code, it still works this way.

GyanP: Ok, so I got it. This is the PR responsible for this regression https://github.com/superdesk/superdesk-client-core/commit/c317cb7ac57c7a83dd46d08d8a88b0a076096894.  And to fix this we need change https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/actions/editor3.tsx#L187 here like this 
`export function processEmbedCode(data) {
    return makeIframesResponsive(data.html);
}`

GyanP: We were sending the whole object instead of HTML.

tomaskikutis: Good job finding it :)

Try to always understand the problem before fixing it, especially when the "fix" is adding a null or a similar check without understanding why the value is not correct. That's especially easy to check with regressions since you can find the exact commit which caused the issue.

In this case, your change would have fixed the display of embeds, but the core issue was still there.

> export function processEmbedCode(data) { return makeIframesResponsive(data.html); }

Not exactly. `processEmbedCode` should return the same type it received. In this snipppet, it receives an object, but returns a string.

Try adding an html embed and a url embed and see what kind of data do you receive [here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/actions/editor3.tsx#L197). I suspect it might be string or object with html attribute. If that's the case, the `processEmbedCode` should look like this:

```
function processEmbedCode(data) {
    if (typeof data === 'string') {
        return makeIframesResponsive(data);
    } else if(typeof data === 'object' && typeof data.html === 'string') {
        return {
            ...data,
            html: makeIframesResponsive(data.html),
        }
    } else {
        logger.error('embed format not recognized');
    }
}
```

PR #2997, opened 2019-06-19
Mark for user  display people content is marked for

tomaskikutis:
SDBELGA-29

The PR is forked from https://github.com/superdesk/superdesk-client-core/pull/2989 so I'd first get it merged before reviewing this one.


tomaskikutis:
No suggestions? :crying_cat_face: :smile:

pablopunk:
@tomaskikutis It was a long reading but no 😁  

I read all of it, maybe it was to much to assimilate... But didn't notice any issue. One question tho, what's the difference between internal and external extensions?

tomaskikutis:
With internal extensions, there's no activate method and extensions can be inserted/removed at any time. This can be prone to errors, so I don't want to allow that on external ones if we can get by without it.

Internal extensions aren't really extensions in a sense that it's not an encapsulated piece of functionality like it is with other extensions. I'm not sure if it's a good idea in the first place. I simply needed a way to intercept a function call and the mechanism to do that was already implemented in the extensions API, so I decided to use that so I don't have to add more code for the same thing.

pablopunk:
Got it. One last thing. Is there anything needed in the backend to have this feature? I can run it locally but nothing happened when I selected an user to mark

tomaskikutis:
Yep, forgot to mention it, you need to add a field in `superdesk/metadata/item.py`
screenshot_4

PR #2998, opened 2019-06-20
Fix duplicated string fields (like language) on image upload

pablopunk:
SDANSA-282

String fields with an existing `cv` where shown twice to the user

### DIFF
--- a/scripts/apps/authoring/media/views/media-metadata-editor-directive.html
+++ b/scripts/apps/authoring/media/views/media-metadata-editor-directive.html
@@ -1,7 +1,7 @@
 <div class="form__row" ng-repeat="field in fields track by field.field">
     <sd-line-input
         class="sd-line-input--no-margin"
-        ng-if="field.type === 'string' || (!field.type && !field.cv)"
+        ng-if="!field.cv && (!field.type || field.type === 'string')"
         ng-model="item[field.field]"
         data-sync="true"
         data-dark="dark"


scripts/apps/authoring/media/views/media-metadata-editor-directive.html
```diff
@@ -1,7 +1,7 @@
 <div class="form__row" ng-repeat="field in fields track by field.field">
     <sd-line-input
         class="sd-line-input--no-margin"
-        ng-if="field.type === 'string' || (!field.type && !field.cv)"
+        ng-if="!field.cv && (!field.type || field.type === 'string')"
```
tomaskikutis: Looks like the issue was caused by not understanding that there are two inputs in the code which should be displayed on a if/else basis, isn't it?

If it is, it would be useful to mark it for the future like [here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/search/views/saved-search-manage-subscribers.html#L122) for example

pablopunk: What do you mean by mark it? Adding a comment on each if?

tomaskikutis: Yep, but only if that `if` actually means `else`. Angular doesn't support `else` in templates, so people just write another `if` which is opposite to the first one, but it's not obvious that it's supposed to be an `else` if the condition is made of multiple expressions.

pablopunk: But it's not really an else, I could explain the if with a comment like "If field doesn't have a CV and it's type is either null or string", but that's like reading the actual if statement

tomaskikutis: I agree, adding such comment wouldn't be helpful. What would be helpful though is if you explained why does it matter whether a field has a CV, what kind of types there are and how it's relevant which one the object has.

It's strange that you say it's not an else. Could you then explain why the same component is rendered twice, but with slightly different parameters?

screenshot_2

For me, it seems that all fields are iterated (`ng-repeat="field in fields track by field.field"`) and if the field is a custom field coming from a controlled vocabulary, one type of input field configuration is rendered and *else*, the component with different configs is rendered. It doesn't seem to me that it's valid that both are rendered in the same iteration, which would make it an if/else case. Am I not understanding something?

pablopunk: Ok I think I know what you mean, but there are not only those two cases (if-else), it's more like a bunch of if-elseif, and I mean [all that file](https://github.com/superdesk/superdesk-client-core/blob/edf71440e0fe1eee5656434c7174bc378c9460f4/scripts/apps/authoring/media/views/media-metadata-editor-directive.html)

* If field doesn't have a CV but it's type is string or null
* Else if field does have a CV but type is not text
* Else if field is 'place' and has autocomplete
* Else if field is not 'place' nor 'keywords' and doesn't have autocomplete
* Else if field type is date and has a CV

If you think the first case is an ELSE for all the other IFs I could put it at the end but still making sure it checks the `ng-if`.

I kind of hate template languages, it would be much more clear in React 😔

tomaskikutis: I didn't think the first one was `else`. I thought the second one was, but yeah, all of them except the first seem to be if-elses.

petrjasek:
@pablopunk and this one?

pablopunk:
@petrjasek I didn't, I guess we should merge 1.30 into master?

PR #3000, opened 2019-06-24
Issues with saving changes in the Desk settings. (SDESK-4325)

GyanP:
requires https://github.com/superdesk/superdesk-ui-framework/pull/365

### DIFF
--- a/scripts/apps/desks/controllers/DeskConfigController.ts
+++ b/scripts/apps/desks/controllers/DeskConfigController.ts
@@ -1,9 +1,11 @@
 import _ from 'lodash';
 import {gettext} from 'core/utils';
 
-DeskConfigController.$inject = ['$scope', '$controller', 'notify', 'desks', 'WizardHandler', 'modal'];
-export function DeskConfigController($scope, $controller, notify, desks, WizardHandler, modal) {
+DeskConfigController.$inject = ['$scope', '$controller', 'notify', 'desks', 'modal', '$rootScope'];
+export function DeskConfigController($scope, $controller, notify, desks, modal, $rootScope) {
     // expecting $scope.desks to be defined
+
+    const stepsToWatch = ['general', 'people'];
 
     $scope.modalActive = false;
     $scope.numberOfUsers = 3;
@@ -12,13 +14,18 @@ export function DeskConfigController($scope, $controller, notify, desks, WizardH
     };
     $scope.desk = {
         edit: null,
+        orig: null,
     };
 
     $scope.openDesk = function(step, desk) {
-        $scope.modalActive = true;
-        $scope.step.current = step;
-        $scope.desk.edit = desk || {};
-        $scope.desk.edit.desk_metadata = $scope.desk.edit.desk_metadata || {};
+        desks.fetchDesks().then((_desks) => {
+            const updatedDesk = _desks._items.find((_desk) => desk != null && _desk._id === desk._id) || {};
+
+            $scope.modalActive = true;
+            $scope.step.current = step;
+            $scope.desk.edit = _.cloneDeep(updatedDesk) || {};
+            $scope.desk.orig = _.cloneDeep(updatedDesk) || {};
+        });
     };
 
     $scope.agg = $controller('AggregateCtrl', {$scope: $scope});
@@ -28,9 +35,14 @@ export function DeskConfigController($scope, $controller, notify, desks, WizardH
     };
 
     $scope.cancel = function() {
-        $scope.modalActive = false;
-        $scope.step.current = null;
-        $scope.desk.edit = null;
+        const diff = calculateDiff();
+        const newDesk = !$scope.desk.edit._id;
+
+        if (!newDesk && stepsToWatch.includes($scope.step.current) && Object.keys(diff).length > 0) {
+            $scope.confirmSave(diff, $scope.step.current, true);
+        } else {
+            closeModel();
+        }
     };
 
     $scope.remove = function(desk) {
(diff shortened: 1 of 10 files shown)


scripts/apps/desks/controllers/DeskConfigController.ts
```diff
@@ -1,24 +1,31 @@
 import _ from 'lodash';
 import {gettext} from 'core/utils';
 
-DeskConfigController.$inject = ['$scope', '$controller', 'notify', 'desks', 'WizardHandler', 'modal'];
-export function DeskConfigController($scope, $controller, notify, desks, WizardHandler, modal) {
+DeskConfigController.$inject = ['$scope', '$controller', 'notify', 'desks', 'modal', '$rootScope'];
+export function DeskConfigController($scope, $controller, notify, desks, modal, $rootScope) {
     // expecting $scope.desks to be defined
 
+    const stepsToWatch = ['general', 'people'];
```
tomaskikutis: Can you check if data changed, no matter which tab is opened?

scripts/apps/desks/controllers/DeskConfigController.ts
```diff
@@ -1,24 +1,31 @@
...
         edit: null,
+        orig: null,
     };
 
     $scope.openDesk = function(step, desk) {
-        $scope.modalActive = true;
-        $scope.step.current = step;
-        $scope.desk.edit = desk || {};
-        $scope.desk.edit.desk_metadata = $scope.desk.edit.desk_metadata || {};
+        desks.fetchDesks().then((_desks) => {
```
tomaskikutis: Why do you need to fetch it again? If you really need it, it's better to fetch only the one you need, not all of them. You can do this easily by using `dataApi.findOne<IDesk>(desk._id)`.

You shouldn't try to set a default value like `const updatedDesk = ... || {};`. If I understand correctly, the desk must be there. If it's not for some reason, log the error `logger.error` and exit. You may display a message to the user, but do not continue with the program when you know you have corrupted data.

scripts/apps/desks/controllers/DeskConfigController.ts
```diff
@@ -28,9 +35,14 @@ export function DeskConfigController($scope, $controller, notify, desks, WizardH
...
-        $scope.modalActive = false;
-        $scope.step.current = null;
-        $scope.desk.edit = null;
+        const diff = calculateDiff();
+        const newDesk = !$scope.desk.edit._id;
+
+        if (!newDesk && stepsToWatch.includes($scope.step.current) && Object.keys(diff).length > 0) {
+            $scope.confirmSave(diff, $scope.step.current, true);
+        } else {
+            closeModel();
```
tomaskikutis: It's modal, not model :)

scripts/apps/desks/controllers/DeskConfigController.ts
```diff
@@ -61,4 +73,70 @@ export function DeskConfigController($scope, $controller, notify, desks, WizardH
...
         return desks.deskMembers[desk._id];
     };
+
+    const closeModel = function() {
+        $scope.modalActive = false;
+        $scope.step.current = null;
+        $scope.desk.edit = null;
+    };
+
+    const calculateDiff = function() {
```
tomaskikutis: Do not use field names when generating a diff. There's a library I use to generate diffs, you can check the example here https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/helpers/CrudManager.tsx#L99

scripts/apps/desks/controllers/DeskConfigController.ts
```diff
@@ -61,4 +73,70 @@ export function DeskConfigController($scope, $controller, notify, desks, WizardH
...
+        modal.confirm(
+            gettext('You have unsaved changes in {{tab}} tab. Do you want to save them now?', {tab: step}),
+            gettext('Save changes?'),
+            gettext('Save'),
+            gettext('Ignore'))
+            .then(() => {
+                desks.save($scope.desk.orig, diff).then((res) => {
+                    _.merge($scope.desk.edit, res);
+                    _.merge($scope.desk.orig, res);
+                    if (diff.members) {
```
tomaskikutis: Is it possible to avoid using the field names when updating the object? How about iterating all members and updating them?

scripts/apps/desks/directives/DeskeditBasic.ts
```diff
@@ -105,6 +89,14 @@ export function DeskeditBasic(desks, WizardHandler, metadata, config,
                 return deskType === 'authoring' && config.features.noPublishOnAuthoringDesk;
             };
 
+            scope.$watch('desk.edit', (newVal, oldVal) => {
+                if ((newVal.content_expiry || oldVal.content_expiry != null) && !_.isEqual(newVal, scope.desk.orig)) {
```
tomaskikutis: Can you watch entire objects and not specify the exact fields?

scripts/apps/desks/controllers/DeskConfigController.ts
GyanP: Just to make sure there is no `_etag` mismatch error when we edit the desk again without refreshing and will change it to find only one desk.

scripts/apps/desks/controllers/DeskConfigController.ts
GyanP: Yyou mean to update the members every time even if they were not changed?

scripts/apps/desks/directives/DeskeditBasic.ts
GyanP: It' about `desk.edit` or `content_expiry`?

scripts/apps/desks/controllers/DeskConfigController.ts
tomaskikutis: When I look at this again, I think it would probably be best to avoid dealing with members explicitly at all. Just apply the entire diff and that's it. If members haven't changed, they shouldn't be in a diff.

scripts/apps/desks/directives/DeskeditBasic.ts
tomaskikutis: That's my bad, I think `desk.edit` is not a field, but a copy of the desk entity. I think saving should be enabled when anything in the object changes, so it doesn't make sense to check one or another field. What about iterating fields dynamically and checking all one by one, or generating a diff with a library I mentioned in another comment and enabling saving if the generated diff is not empty?

scripts/apps/desks/controllers/DeskConfigController.ts
GyanP: We can, but making changes in the other two tabs do not affect the desk. That's why avoided watching all the tabs.

scripts/apps/desks/directives/DeskeditBasic.ts
GyanP: Yup, I am looking onto that library. The issue with the `content_expiry` is that if it is set to `null` in the DB, on editing the desk it set back to `0`, this field always remains in the diff.

scripts/apps/desks/controllers/DeskConfigController.ts
tomaskikutis: Do that then. I'd rather have less code which is easier to maintain.

scripts/apps/desks/directives/DeskeditBasic.ts
tomaskikutis: I'm not sure I understand. If it's `null` in the database and when you edit the desk it becomes `0` it seems to me that it's correct that it's included in the diff. No?

GyanP: Yes. And because of this even if we don't change anything, on changing the tabs confirmation modal keeps popping up, as content_expiry always there included in the diff.

tomaskikutis: You can create a defaults object and apply it immutably on top of the database state before calculating the diff.

```
const database = {
    a: 1,
    b: 2,
    c: null,
};

const defaults = {
    c: 7,
}

const uiState = {
    a: 1,
    b: 2,
    c: 7,
}

diff = diffFunction(applyDefaults(database), uiState) // should produce an empty diff
```

scripts/apps/desks/controllers/DeskConfigController.ts
```diff
@@ -12,13 +14,28 @@ export function DeskConfigController($scope, $controller, notify, desks, WizardH
...
+        orig: null,
     };
 
     $scope.openDesk = function(step, desk) {
+        if (!desk) {
+            initializeDesk({}, step);
+            return;
+        }
+        desks.fetchDeskById(desk._id).then((_desk) => {
+            if (!_desk) {
```
tomaskikutis: Move logging to `fetchDeskById` method and reject the promise if it's not found.

PR #3002, opened 2019-06-25
Fix photo upload form

petrjasek:
SDANSA-282


scripts/apps/workspace/content/services/ContentService.ts
```diff
@@ -395,12 +395,14 @@ export function ContentService(api, superdesk, templates, desks, packages, archi
...
                 scope.editor = this.editor(profile, item.type);
                 scope.fields = this.fields(profile);
+                return profile;
             });
         } else {
-            scope.schema = this.schema(null, get(item, 'type', 'text'));
-            scope.editor = this.editor(null, get(item, 'type', 'text'));
-            scope.fields = null;
-            return $q.when();
+            return this.getCustomFields().then(() => {
```
tomaskikutis: Looks like you aren't using the result of `getCustomFields`. How is this supposed to work?

scripts/apps/authoring/media/MediaMetadataEditorDirective.ts
```diff
@@ -76,6 +119,8 @@ export default function MediaMetadataEditorDirective(metadata, deployConfig, fea
              * @return {Boolean}
              */
             scope.isDisabled = (field) => scope.disabled || (scope.associated && field.external);
+
+            scope.isExtra = (field): boolean => field.cv && field.cv.field_type === 'text';
```
tomaskikutis: Isn't a field considered to be "extra" if it can be dynamically configured via UI? I thought field type shouldn't matter. Or should it?

petrjasek: it does, because if it's controlled vocabulary it is stored as subject

scripts/apps/workspace/content/services/ContentService.ts
petrjasek: it populates `this._fields` which is used by `this.fields` call later

scripts/apps/authoring/media/MediaMetadataEditorDirective.ts
tomaskikutis: What I'm asking is why do you need `field.cv.field_type === 'text'`? As I understand the only difference between extra fields is that they can be configured dynamically and are stored in `IArticle.extra` namespace. Can't there be other types of extra fields than text?

petrjasek: there is also date and urls, have to fix the check

tomaskikutis: Is the check needed in the first place if we check against available types?

petrjasek: I think it might be different for type `media` and `related_content`, but will double check

scripts/apps/authoring/authoring/components/line-count.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import {get} from 'lodash';
+import {stripHtmlTags, gettextPlural} from 'core/utils';
+
+interface IProps {
+    item: string;
+    config: {
```
tomaskikutis: Import config(from `scripts/appConfig.ts`) instead of connecting a service. Connecting services is a crutch we use to get angular services, but the longer I think about it, the more I'm convinced that angular services are better of being standalone es6 modules. We would avoid mountains of boilerplate and wouldn't have to deal with async stuff when it's actually synchronous. What do you think?

scripts/apps/authoring/authoring/components/line-count.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import {get} from 'lodash';
+import {stripHtmlTags, gettextPlural} from 'core/utils';
+
+interface IProps {
+    item: string;
```
tomaskikutis: I'd call it `html`. Everything is an item :D

scripts/apps/authoring/authoring/index.ts
```diff
@@ -99,6 +100,16 @@ angular.module('superdesk.apps.authoring', [
     .directive('sdRemoveTags', directive.RemoveTagsDirective)
     .directive('tansaScopeSync', directive.TansaScopeSyncDirective)
     .directive('sdItemActionByIntent', directive.ItemActionsByIntentDirective)
+
+    .component('sdLineCount',
+        reactToAngular1(
+            LineCount,
+            ['item', 'html'],
```
tomaskikutis: `html` isn't used.

scripts/apps/authoring/authoring/components/line-count.tsx
petrjasek: we merge that config from core with superdesk repo config, so I can't just import it I think

tomaskikutis: You can. It's merged at build, not at runtime https://github.com/superdesk/superdesk-client-core/blob/master/webpack.config.js#L62

PR #3004, opened 2019-06-25
Port tests to TypeScript

tomaskikutis:


spec/helpers/metadata.ts
```diff
@@ -22,4 +23,5 @@ class MetadataHelper {
     }
 }
 
-module.exports = new MetadataHelper();
+export const metadata = new MetadataHelper();
+export default metadata;
```
pablopunk: Why do we need both? `import metadata, {metadata}`?

.gitignore
```diff
@@ -11,6 +11,7 @@ test-server/__pycache__
 e2e-test-results
 templates-cache.generated.js
 scripts/core/extension-imports.generated.ts
+spec/**/*.js
```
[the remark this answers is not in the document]
tomaskikutis: yep, these are compiled tests.

spec/helpers/metadata.ts
tomaskikutis: I'm not sure if we no longer need that. At first, my plan was to only port a subset of tests and I left both exports for compatibility with `commonjs` imports. It turned out faster and I ported all tests, so I'll remove these double exports.

PR #3005, opened 2019-06-26
Enable custom video and audio as well as related items to show up in the preview. (SDESK-4012)

GyanP:


scripts/apps/archive/directives/MediaPreview.ts
```diff
@@ -90,6 +91,27 @@ export function MediaPreview(api, $rootScope, desks, superdesk, content, storage
                 return _.map(scope.item.company_codes, 'qcode').join(', ');
             };
 
+            // Check if there is any association for fieldId
```
tomaskikutis: If the comment only repeats what is in the function signature, it doesn't provide any additional information and isn't worth having.

scripts/apps/archive/directives/MediaPreview.ts
```diff
@@ -90,6 +91,27 @@ export function MediaPreview(api, $rootScope, desks, superdesk, content, storage
...
+
+            // Return all associations for fieldId
+            scope.getAssociatedItems = function(associations, fieldId) {
+                var result = _.filter(associations, (association, key) => key.indexOf(fieldId) !== -1);
+
+                return result;
+            };
+
+            // Check if the rendition is image or not
+            scope.isImage = (rendition) => isImage(rendition);
```
tomaskikutis: Prefer code over comments. You could write it like this and it would be clear enough without any comments.

```
// utils file
export const renditions = {
    isImage: ...,
    isVideo: ...,
    isAudio: ...,
};
// this file
scope.renditions = renditions;
```

scripts/apps/authoring/authoring/controllers/AssociationController.ts
```diff
@@ -28,49 +28,14 @@ export function AssociationController(config, content, superdesk,
...
     };
 
-    /**
-     * @ngdoc method
-     * @name AssociationController#isImage
-     * @public
-     * @description Check if the rendition is image or not.
-     * @param {Object} rendition Rendition of the item.
-     */
-    this.isImage = function(rendition) {
```
tomaskikutis: Can you keep it in this file? Utils are meant for generic code, not related to any features. You can export it from this file and import it where you need it.

scripts/apps/archive/directives/MediaPreview.ts
```diff
@@ -90,6 +92,28 @@ export function MediaPreview(api, $rootScope, desks, superdesk, content, storage
                 return _.map(scope.item.company_codes, 'qcode').join(', ');
             };
 
+            /**
+             * @ngdoc method
+             * @name associationExists
+             * @private
```
tomaskikutis: > If the comment only repeats what is in the function signature, it doesn't provide any additional information and isn't worth having.

I meant to suggest to you to simply remove the comment. If the function signature is simple and clear, as it is in this case, no comments are needed. Can you remove all such comments in this PR?

GyanP: Remove from all those files that are included in this PR or from other files too?

tomaskikutis: Only those in this PR.

scripts/apps/authoring/authoring/controllers/AssociationController.ts
```diff
@@ -18,60 +18,12 @@ export function AssociationController(config, content, superdesk,
     mediaIdGenerator, authoring, renditions, notify) {
     const self = this;
 
-    /**
-     * @ngdoc method
-     * @name AssociationController#isMediaEditable
-     * @public
-     * @description Check if featured media can be edited or not. i.e. metadata/crops can be changed or not.
```
tomaskikutis: Don't remove the comment if it contains details which can't be inferred from the function signature. For example, this one mentions crops, which isn't mentioned in the signature nor in the function itself. Can you keep the comments that you didn't yourself add?

PR #3010, opened 2019-07-05
fix(metadata): Fixed issue with list scroll on metadata widget

pavlovicnemanja:


petrjasek:
seems like removing `vs-scroll-parent` from [template](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/metadata/views/metadata-terms.html#L66) fixes it too, that would be better imo

PR #3011, opened 2019-07-08
Fixed: 'Edit image' action done from article should not modify the original image file.(SDESK-4004)

GyanP:
Requires https://github.com/superdesk/superdesk-core/pull/1620

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
@@ -422,7 +422,6 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, deploy
 
         }}).then((result) => {
             $scope.data.item.renditions = result.renditions;
-            $scope.data.item._etag = result._etag;
             const editableMetadata = extractEditableMetadata($scope.data.metadata);
 
             $scope.data.metadata = Object.assign($scope.data.item, editableMetadata);


scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -422,7 +422,6 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, deploy
 
         }}).then((result) => {
             $scope.data.item.renditions = result.renditions;
-            $scope.data.item._etag = result._etag;
```
tomaskikutis: Why are you removing this?

GyanP: We don't save the edited media from the backend anymore, so `_etag` should remain the same.

tomaskikutis: If it's the same, there's no problem with updating it. Deciding on which resource to update is a server-specific implementation detail which client must not know about.

Can you close the PR?

GyanP: That would work fine for the images being edited from the story like feature media or from the article body. But it will throw the `_etag` mismatch error if we edit the article of type `picture` directly. Earlier it was done for the article for type `picture` only.

GyanP: https://drive.google.com/file/d/1p5is3U2DWDUpVuX9ODHmP2moOQ0fnH_t/view

tomaskikutis: > But it will throw the _etag mismatch error if we edit the article of type picture directly.
Why?

> Earlier it was done for the article of type picture only.
What was done earlier?

GyanP: Previously, when we were making a post request to `media_editor` API, we were updating the image item in the `archive` as well in the backend in the same request. So, the _etag of the image item gets changed. 
And when we edit the item of type picture directly we were updating it twice in the archive, once from the request to `media_editor` API, and once after performing the save from the authoring.
For this second update, we need to update that _etag in the client, so that there is no `precondition failed` error because of _etag mismatch.  
And this update to archive from the backend was the reason behind `original image` getting changed on any changes made to the image edited from inside of the article like feature-media.
But now we don't do that update to archive in the backend on request to `media_editor` API, that means _etag remains same for the image item. So, we do not need to change the _etag in the client. And if we change as you are saying, then we will get `precondition failed` error due to _etag mismatch.

PR #3013, opened 2019-07-09
Avoid extending inline styles when pressing space at end of block

pablopunk:
SDESK-4335

This includes all styles (annotations, comments, italic...)

pressing space

### DIFF
--- a/scripts/core/editor3/helpers/handleBeforeInputHighlights.ts
+++ b/scripts/core/editor3/helpers/handleBeforeInputHighlights.ts
@@ -1,4 +1,10 @@
-import {EditorState, Modifier, CharacterMetadata, DraftHandleValue} from 'draft-js';
+import {
+    EditorState,
+    Modifier,
+    CharacterMetadata,
+    SelectionState,
+    DraftHandleValue
+} from 'draft-js';
 import {List, OrderedSet} from 'immutable';
 import {uniq} from 'lodash';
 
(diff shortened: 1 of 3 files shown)


tomaskikutis:
I couldn't reproduce it on master, here's a [video](https://drive.google.com/file/d/1S0vKay-l9HWagqrEunuSNV4OKpTL6TS4/view?usp=sharing). I've put WIP on the PR for now. If you find that there is an issue, update the ticket with instructions on how to reproduce and ping me then :)

pablopunk:
I found it @tomaskikutis (sorry I was sick all day 😷). It only happens when you have a line break at the end. It actually says that on the task but I thought it was irrelevant, turns out it's not.

test

tomaskikutis:
Do you know why does it matter if you have a line-break at the end?

pablopunk:
@tomaskikutis it seems [here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/helpers/handleBeforeInputHighlights.ts#L23-L31) `resizeRightFailed` is false when there's a linebreak (which makes sense) so only one character is returned. If there's no linebreak, 2 are returned.

tomaskikutis:
Do you think the current solution is the best? Wouldn't it be better to either improve `getRelevantCharactersForCollapsedSelection` function or change how its result is treated?

pablopunk:
@tomaskikutis it's definitely better to improve that function, so I could fix this without any extra checks. I was just wondering how would it affect other features

tomaskikutis:
The function is only used in `handleBeforeInputHighlights`, which has a description of "prevents inheriting of highlight styles", so that's probably the only purpose and you shouldn't break anything by fixing it.

pablopunk:
@tomaskikutis I could break for example the ability to extend the highlight while you're typing in the middle of it, I'll take care of that tho

scripts/core/editor3/helpers/handleBeforeInputHighlights.ts
```diff
@@ -28,7 +36,7 @@ function getRelevantCharactersForCollapsedSelection(editorState, collapsedSelect
 
     characters = characters.concat(getDraftCharacterListForSelection(editorState, resizedSelection));
 
-    if (resizeRightFailed === true) {
+    if (resizeRightFailed === true || selectionAtTheEndOfTheBlock) {
```
tomaskikutis: Could you briefly explain what was the main problem? Why would it expand the annotation? Could you also rename `getRelevantCharactersForCollapsedSelection` so it's clear what does it do? (if you understand it :))

pablopunk: Sure, basically `getRelevantCharactersForCollapsedSelection` returns the characters before and after the cursor so we can check the inline styles of those.

If it fails resizing the selection to the right, it will return a character with an empty style (which prevents extending inline styles while writing at the end). The problem was that resizing to the right was not failing if there was another block after it, even though the cursor was at the end of the line. So I added that check to extend that behaviour.

I'm thinking now that I might have to check if the user is pressing space, because if they're not they might want to extend the highlight. I'll check with QA.

pablopunk: Ok, the "right" behaviour should be the one we already have with one line, so this PR should be ready

tomaskikutis: >  returns the characters before and after the cursor so we can check the inline styles of those.

> If it fails resizing the selection to the right, it will return a character with an empty style (which prevents extending inline styles while writing at the end).

Can you add this to the code? :)

> The problem was that resizing to the right was not failing if there was another block after it

How about resizing the selection by 2 chars to the right if it's at the end of the block. It seems like the same thing was done with resizeLeft, it would be good to keep the implementations the same if they do the same thing.

> `const resizeLeft = selectionAtTheStartOfTheLine ? 2 : 1;` // SDESK-2861

> I'm thinking now that I might have to check if the user is pressing space, because if they're not they might want to extend the highlight.

What if they want to extend it by adding another word or want to add punctuation which shoulnd't extend the annotation? I think it's best not to allow it and keep it consistent, that if you want to add anything to the annotation, you move one char before the end, write what you want and remove the last char.

If you were to check for space

tomaskikutis: > behaviour should be the one we already have with one line

What behavior are you referring to?

pablopunk: The one we have on master already when we only have one block (the one you show in [this video](https://drive.google.com/file/d/1S0vKay-l9HWagqrEunuSNV4OKpTL6TS4/view))

scripts/core/editor3/helpers/handleBeforeInputHighlights.ts
```diff
@@ -15,28 +21,40 @@ function isSelectionAtEndOfBlock(editorState: EditorState): boolean {
 }
 
 /**
- * @ngdoc method
- * @name addCommentsForServer
- * @param {EditorState} editorState
- * @param {SelectionState} collapsedSelection
- * @return {List<CharacterMetadata>}
+ * Returns the characters before and after the cursor.
```
tomaskikutis: Your comment on Github was clearer. This one doesn't explain why empty chars are pushed to the list.

PR #3014, opened 2019-07-09
Add edit crops screen after image upload to body field. (SDESK-4384)

GyanP:


scripts/core/editor3/actions/toolbar.tsx
```diff
@@ -66,13 +67,36 @@ export function removeFormat() {
...
+                if (media_list && Array.isArray(media_list) && media_list.length > 0) {
+                    const _media = media_list.shift();
+                    const _isImage = checkRenditions.isImage(_media.renditions.original);
+                    const options = {
+                        isNew: true,
+                        editable: true,
+                        defaultTab: _isImage ? 'crop' : 'view',
+                    };
+
+                    setTimeout(() => {
```
tomaskikutis: why do you need `setTimeout`?

scripts/core/editor3/actions/toolbar.tsx
```diff
@@ -66,13 +67,36 @@ export function removeFormat() {
...
+
+    const editedMedia = [];
 
     return (dispatch) => {
-        superdesk.intent('upload', 'media', files).then((media) => {
-            dispatch({
-                type: 'TOOLBAR_INSERT_MEDIA',
-                payload: {files: media, targetBlockKey: targetBlockKey},
-            });
+        superdesk.intent('upload', 'media', files).then((media_list) => {
```
tomaskikutis: It's better to not modify function arguments / promise results, especially when it's easy to avoid like here. The function that returns promises might be using the same references for caching or other purposes and you may break something by modifying it directly. Create a `const media_list_copy = [...media_list];` and remove items from that instead.

scripts/core/editor3/actions/toolbar.tsx
GyanP: I'm using this to let the current synchronous flow get over so that already opened modal gets closed and the new modal can be opened.

tomaskikutis: :+1: add this as a comment in this code

PR #3015, opened 2019-07-09
compile e2e tests post install. grunt task doesn't work in parent repositories

tomaskikutis:


MarkLark86:
@tomaskikutis @petrjasek I'm merging this PR as new PRs are all failing in other repos (event my PR in superdesk-client-core for branch 1.29).

PR #3018, opened 2019-07-11
List content marked for me

tomaskikutis:
SDBELGA-123

This PR only implements listing items marked for the current user but doesn't add filters in monitoring view.


scripts/core/helpers/CrudManager.tsx
```diff
@@ -12,9 +12,117 @@ import {
...
+
+    return ng.getServices(['config', 'session', 'api'])
+        .then((res: any) => {
+            const [config, session] = res;
+
+            const source = {
+                query: {
+                    filtered: {
+                        filter: {
+                            and: [
```
petrjasek: and and or queries are deprecated, pls use [bool query](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-bool-query.html) instead

scripts/core/ui/components/dropdown-tree.tsx
```diff
@@ -0,0 +1,108 @@
...
+                            <div style={{
+                                background: '#F8F8F8',
+                                maxHeight: 400,
+                                overflow: 'auto',
+                                boxShadow: '0 2px 10px rgba(0, 0, 0, 0.4), 0 3px 1px -2px rgba(0, 0, 0, 0.1)',
+                                ...(this.props.wrapperStyles || {}),
+                            }}>
+                                {
+                                    groups.map((group, i) => (
+                                        <div key={i}>{this.renderGroupRecursive(group, 0, 0)}</div>
```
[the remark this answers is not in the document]
tomaskikutis: I think using the iterator should be fine as long as the groups are immutable.

.eslintrc.js
```diff
@@ -14,6 +14,8 @@ module.exports = Object.assign({}, sharedConfigs, {
                 'arrow-body-style': 0,
                 'newline-per-chained-call': 0,
                 'quote-props': 0,
+                'arrow-body-style': 0,
+                'max-len': 0, // handled by tslint
```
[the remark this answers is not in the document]
tomaskikutis: > Is there something we want from eslint that tslint doesn't have?

I don't know, we'd need to go over the rules and check it. I don't expect significant things to be missing. I'll look into it.

scripts/core/helpers/CrudManager.tsx
tomaskikutis: It [doesn't seem to work](https://user-images.githubusercontent.com/2076953/61132181-24844a80-a4bb-11e9-8849-1f1be9e37e10.png). Can you post an example?

petrjasek: you should put
```
const source = {
    query: {
        bool: {
            must: [...], // what you have in and
            must_not: [...],
        }
    }
};
```
    ...

.eslintrc.js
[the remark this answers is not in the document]
tomaskikutis: I've checked the situation with linters. ESLint has significantly more rules/plugins as it's been out for longer. The company behind tslint is planning to deprecate tslint and do a better integration for typescript and eslint instead. TypeScript team seems to be on the same track. https://github.com/typescript-eslint/typescript-eslint#what-about-tslint

I've tried eslint with a typescript plugin, but it wasn't working good enough yet. For example, the unused variables rule was triggered for interfaces.

I think the best we can do now is to use eslint where it works and use tslint for typescript specific things that don't work with eslint.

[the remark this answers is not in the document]
pablopunk: Okay let's keep both for now then and keep an eye on eslint

scripts/core/helpers/CrudManager.tsx
tomaskikutis: The example in the elastic website shows `must` to be an object, not an array. I've [tried](https://user-images.githubusercontent.com/2076953/61212833-a2d32d80-a703-11e9-9f2c-00a4e898c55d.png) an array, but no luck with it either. Can you share a working query? Say the one that would return articles with `draft` state only.

MarkLark86: Try this one:
```
query: {
    bool: {
        must: Object.keys(filterValues).map((key) => ({term: {[key]: filterValues[key]}})),
        must_not: [
            {term: {state: 'spiked'}},
            {term: {package_type: 'takes'}},
        ],
        should: [{
            bool: {
                must: [
                    {term: {state: 'draft'}},
                    {term: {original_creator: session.identity._id}
                ]
            }
        }, {
            bool: {
                must_not: [
                    {term: {state: 'draft'}
                ]
            }
        }],
        minimum_should_match : 1,
    },
};
```

Using `should` with `minimum_should_match: 1` means that at least 1 query inside the `should` field must pass.

tomaskikutis: @MarkLark86  There were a few missing braces, but after fixing them, it  still didn't work. Can you post a simple working CURL example, say for getting all items with a state `draft`?

Here's the one I've tried:

```
tomas@p50:~$ curl 'https://sd-master.test.superdesk.org/api/search?aggregations=0&es_highlight=0&source=\{%22query%22:\{%22bool%22:\{%22must%22:\[\{%22term%22:\{%22state%22:%22draft%22\}\}\]\}\},%22sort%22:\[\{%22_updated%22:%22desc%22\}\],%22from%22:0\}' -H 'Authorization: Basic MDIzNWYzYzAtYzZhMS00OWQyLThlYmUtNDE1ODJiZDEwYzZkOg==' -H 'Content-Type: application/json'
<html>
  <head>
    <title>Internal Server Error</title>
  </head>
  <body>
    <h1><p>Internal Server Error</p></h1>
    
  </body>
</html>
tomas@p50:~$ 
```

MarkLark86: Sorry @tomaskikutis, I used notepad++ to construct that query.

The query is fine, the problem is the server is responding with an internal server error. After investigating in my local instance, the problem appears to be with the search endpoint and the use of eve_elastic at line https://github.com/petrjasek/eve-elastic/blob/master/eve_elastic/elastic.py#L162.

Eve elastic is expecting a filtered/filter query. So you will have to change your query. Try this instead:

```
const source = {
    query: {
        filtered: {
            filter: {
                bool: {
                    must: [
                        {term: {state: 'draft'}},
                    ],
                },
            },
        },
    },
};
```

tomaskikutis: Thanks for checking it. Getting drafts using the query above works now. What isn't working is `minimum_should_match`. If I add it, I get a server error. I assume it's related to the problem with search endpoint you linked.

scripts/apps/search/components/fields/slugline.tsx
```diff
@@ -0,0 +1,23 @@
...
+                data-test-id="field--slugline"
+            />
+        );
+    } else {
+        return null;
+    }
+}
+
+slugline['propTypes'] = {
+    item: PropTypes.any,
```
pablopunk: should this be an `PropTypes.object`?

scripts/core/helpers/CrudManager.tsx
MarkLark86: @tomaskikutis, sorry for the misdirection. `minimum_should_match` is an attribute on the bool query as this value is only applied to the parent bool query it is a member of (see https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-bool-query.html).

So it would be applied like so:
```
query: {
    filtered: {
        filter: {
            bool: {
                must: [....],
                must_not: [....],
                should: [....],
                minimum_should_match: 1,
            },
        },
    },
}
```

tomaskikutis: > minimum_should_match is an attribute on the bool query

Yep, that's where I've put it and I still get error 500 from the server.

```
query: {
    filtered: {
        filter: {
            bool: {
                must: Object.keys(filterValues).map((key) => ({term: {[key]: filterValues[key]}})),
                must_not: [
                    {term: {state: 'spiked'}},
                    {term: {package_type: 'takes'}},
                ],
                should: [
                    {
                        bool: {
                            must: [
                                {term: {state: 'draft'}},
                                {term: {original_creator: session.identity._id}},
                            ],
                        },
                    },
                    {
                        bool: {
                            must_not: [
                                {term: {state: 'draft'}},
                            ],
                        },
                    },
                ],
                // minimum_should_match : 1,
            },
        },
    },
},
```

scripts/apps/search/components/fields/slugline.tsx
tomaskikutis: It should be `IArticle`. I've added interfaces to all field components.

scripts/core/helpers/CrudManager.tsx
tomaskikutis: @petrjasek can you look into this? You can check out the code and uncomment `// minimum_should_match : 1,` to reproduce the error.

PR #3019, opened 2019-07-11
[SDESK-2300] fix(publish): Fixed issue with no-messages on multi publish button

pavlovicnemanja:
[SDESK-2300] Quick publish does not work and does not throw warning of missing fields

### DIFF
--- a/scripts/apps/search/controllers/MultiActionBarController.tsx
+++ b/scripts/apps/search/controllers/MultiActionBarController.tsx
@@ -204,6 +204,8 @@ export function MultiActionBarController(
                     notify.success(gettext('All items were published successfully.'));
                     multi.reset();
                 }
+            }, (err) => {
+                notify.error(gettext('Some items could not be published.'));
             });
     };
 }
--- a/scripts/apps/search/controllers/MultiActionBarController.tsx
+++ b/scripts/apps/search/controllers/MultiActionBarController.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable indent */
 
+import React from 'react';
 import _, {flatMap} from 'lodash';
 import {IArticle} from 'superdesk-api';
 import {gettext} from 'core/utils';
@@ -7,6 +8,11 @@ import {gettext} from 'core/utils';
 import {IExtensionActivationResult} from 'superdesk-api';
 import {extensions} from 'core/extension-imports.generated';
 import {showSpikeDialog} from 'apps/archive/show-spike-dialog';
+
+import {Modal} from 'core/ui/components/Modal/Modal';
+import {ModalHeader} from 'core/ui/components/Modal/ModalHeader'
+import {ModalBody} from 'core/ui/components/Modal/ModalBody';
+import {ModalFooter} from 'core/ui/components/Modal/ModalFooter';
 
 /**
  * @ngdoc controller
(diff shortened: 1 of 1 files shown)


superdesk.config.js
```diff
@@ -16,6 +16,11 @@ module.exports = function(grunt) {
                 subscriptions: true,
             },
             validatePointOfInterestForImages: true,
+            customAuthoringTopbar: {
```
petrjasek: why changing this?

pavlovicnemanja: Sorry, my bad! Left over on testing some other issue. Removing...

scripts/apps/search/controllers/MultiActionBarController.tsx
```diff
@@ -204,6 +204,8 @@ export function MultiActionBarController(
                     notify.success(gettext('All items were published successfully.'));
                     multi.reset();
                 }
+            }, (err) => {
+                notify.error(gettext('Some items could not be published.'));
```
tomaskikutis: What's the reason? Are there any user readable errors in the server response we could show?

pavlovicnemanja: This is the case when multi articles are selected and tried to be published. So if one of them didn't get through validation it fails. We discussed with @fritzSF and because we will need to show multiple lot of informations in notification (article name, fields not added) we agreed to go just with this message (as it was). This functionality is not used frequently.

tomaskikutis: How is the user supposed to fix the validation errors then if he doesn't know what's wrong? What about not using the notification, but displaying a scrollable modal with all the validation errors instead?

fritzSF: @tomaskikutis I wouldn't say it is worth doing something new for this specific action. To be honest I'm not even sure it should be possible to publish multiple items – exactly for this specific reason. Anyway, IMO this should be fixed to the state as it was and improved with a new ticket, which can be created right away.

tomaskikutis: It's not doing anything new, it's fixing an existing feature. I'm fine with removing the feature if we don't need it, but we shouldn't push half-broken stuff. It takes an hour of work to show the exact errors for each article. I see no reason for creating another ticket for the same thing.

fritzSF: I don't want to argue, but It is new as it didn't work that way before. How it worked might be very basic but not half-broken. Are you absolutely sure it would take an hour to do this? I somehow doubt it would take so little.

tomaskikutis: I'm not sure about how it used to work, but the `this.publish` function, Nemanja is changing has been in the codebase for at least 3 years.

> Are you absolutely sure it would take an hour to do this?

It depends how nice you want it to look, but a simple version using a modal and a simple list component from ui framework can be done in an hour.

pavlovicnemanja: This is old functionality that stopped working once we moved some extra validations on backend side. I just did it as it was before. We can create notifications that will show what is wrong in which article, but notification will be to big because it should contain article slugline or headline and error message. Second thing you suggested to create modal with list of actions, I think will take more than hour (probably day) because we should also create stylings for it. I think this functionality is not used that often so we can go with old implementation. This is not new feature, this is bug fix.

tomaskikutis: err

I've added the modal. It's not the most beautiful one, but does the job :)

pavlovicnemanja: It's ugly. I will do stylings here. It's not problem about modal, it's problem about UI/UX. And please don't do this in feature, It's my task, I will handle it. This was discussion if it is wort doing or not.

tomaskikutis: The priority goes to the feature working in the first place and only then to it looking nice. Having validation issues is common, and not displaying which items weren't published is a major issue. If it's used rarely enough to justify it not working properly, unstyled error messages should be an even lesser problem.

PR #3024, opened 2019-07-15
add unpublish action for published content

petrjasek:
it's like kill/takedown but without emails and does
not change the item (only its state).

you can select related published items while confirming
the action to unpublish them all.

SDFID-555

works best with https://github.com/superdesk/superdesk-core/pull/1629


scripts/apps/archive/constants.ts
```diff
@@ -0,0 +1,24 @@
+
+export const PUBLISHED = 'published';
+export const SCHEDULED = 'scheduled';
+export const CORRECTED = 'corrected';
+
+export const KILLED = 'killed';
+export const RECALLED = 'recalled';
+export const UNPUBLISHED = 'unpublished';
+
+export const PUBLISH_STATES = [
```
tomaskikutis: Why `PUBLISH_STATES` include `KILLED_STATES`? Wouldn't `ALL_STATES` be a better name in such case?

scripts/apps/archive/services/ArchiveService.ts
```diff
@@ -125,9 +126,7 @@ export function ArchiveService(desks, session, api, $q, search, $location, confi
      * @param {Object} item
      * @return boolean if the state of the item is in one of the published states, false otherwise.
      */
-    this.isPublished = function(item) {
-        return _.includes(['published', 'killed', 'scheduled', 'corrected', 'recalled'], item.state);
-    };
+    this.isPublished = isPublished;
```
tomaskikutis: Can you remove it from here and `AuthoringService` and import it directly where needed? It's hard to track usages when the same thing is aliased in multiple places.

scripts/apps/authoring/authoring/services/AuthoringService.ts
```diff
@@ -560,9 +558,15 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
             // in production
 
             action.duplicate = userPrivileges.duplicate &&
-                !_.includes(['spiked', 'killed', 'recalled'], currentItem.state);
-
-            action.add_to_current = !_.includes(['spiked', 'scheduled', 'killed', 'recalled'], currentItem.state);
+                !_.includes(['spiked', 'killed', 'recalled', 'unpublished'], currentItem.state);
+
+            action.add_to_current = !_.includes([
+                'spiked',
```
tomaskikutis: Can you use constants for these as well?

scripts/apps/authoring/authoring/services/AuthoringWorkspaceService.ts
```diff
@@ -158,6 +159,59 @@ export function AuthoringWorkspaceService($location, superdeskFlags, authoring,
...
+                    $scope.relatedItems = [];
+
+                    familyService.fetchItems(item.archive_item.family_id, item)
+                        .then((items) => {
+                            const published = {};
+
+                            $scope.relatedItems = items._items
+                                .filter((relatedItem) => {
+                                    if (!published[relatedItem.guid]) {
+                                        published[relatedItem.guid] = true; // only show each item once
```
tomaskikutis: Why does it return each item more than once?

petrjasek: if you publish something, then unpublish, and publish again you can get same item there multiple times

scripts/apps/archive/constants.ts
petrjasek: there are other states before items is published, those publish states are all states when something was sent to subscribers

scripts/apps/authoring/authoring/services/AuthoringWorkspaceService.ts
tomaskikutis: Which one of those multiple items do you want, last updated one?

> if you publish something, then unpublish, and publish again you can get same item there multiple times

This seems buggy. Why does it behave this way? Shouldn't publishing be only a matter of setting the state to `published` and unpublishing setting it back to `in-progress`?

scripts/apps/archive/constants.ts
tomaskikutis: Can you then add the remaining states or a comment with a link to code or docs with all possible states listed?

scripts/apps/authoring/authoring/index.ts
```diff
@@ -340,7 +348,38 @@ angular.module('superdesk.apps.authoring', [
...
+                privileges: {unpublish: 1},
+            })
+            .activity('edit.unpublished', {
+                label: gettext('Edit'),
+                priority: 100,
+                icon: 'edit-line',
+                group: 'corrections',
+                controller: ['data', 'authoringWorkspace', 'api',
+                    (data, authoringWorkspace, api) => {
+                        api.update('archive', data.item.archive_item, {state: 'in_progress'})
```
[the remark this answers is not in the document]
petrjasek: `unpublished`, it's checked in `additionalCondition:`

scripts/apps/authoring/authoring/services/AuthoringWorkspaceService.ts
petrjasek: sort of, but for the record we store in published everything that was sent out - even when something is unpublished/killed there is a copy with different metadata sent to subscribers

scripts/apps/authoring/authoring/index.ts
```diff
@@ -340,7 +343,68 @@ angular.module('superdesk.apps.authoring', [
...
+                label: gettext('Unpublish'),
+                priority: 50,
+                icon: 'kill',
+                group: 'corrections',
+                controller: ['data', 'authoring', 'familyService',
+                    (data, authoring, familyService) => {
+                        const item = data.item;
+                        let relatedItems = [];
+
+                        familyService.fetchItems(item.archive_item.family_id, item)
```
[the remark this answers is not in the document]
petrjasek: I would rather keep that unpublish endpoint consistent with other publishing actions for now,
it would make sense to have `/unpublish` with list of ids, it's not really important if an item is related or not, and the same would work for publishing from lists etc. but that's another task to refactor it all.
will check that related items endpoint, moving that logic with family id to server would be good for sure

petrjasek:
using new api for related items https://github.com/superdesk/superdesk-core/pull/1636

scripts/apps/archive/services/FamilyService.ts
```diff
@@ -102,6 +103,16 @@ export function FamilyService(api, desks) {
         return query(filter, 'versioncreated', 'asc');
     };
 
+    /**
+     * Fetch related items
```
tomaskikutis: The comment is made of 3 words, 2 of which are the in the function name. If that extra word is relevant, add it to the function name instead of having it as a comment duplicating the function name.

scripts/apps/archive/services/FamilyService.ts
```diff
@@ -102,6 +103,16 @@ export function FamilyService(api, desks) {
...
+     */
+    this.fetchRelatedByState = (item: IArticle, state: Array<string>, exclude: boolean=false) =>
```
tomaskikutis: add a comment on what exclude means.

PR #3027, opened 2019-07-17
Spellchecker should respond to language change immediately. (SDESK-4405)

GyanP:


scripts/core/editor3/components/spellchecker/SpellcheckerDecorator.tsx
```diff
@@ -20,45 +20,51 @@ export function getSpellcheckWarningsByBlock(
         return Promise.resolve({});
     }
 
-    const rangesByBlock: Array<{blockKey: string, startOffset: number, endOffset: number}> = [];
+    return ng.getService('spellcheck').then((spellcheck) => {
```
tomaskikutis: `getSpellcheckWarningsByBlock` too should only be called when spellchecker is already enabled.

It's checked [here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/components/Editor3Component.tsx#L148) and the property is coming from [here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/store/index.ts#L141). To disable the spellchecker you need to update `store.spellchecker.enabled`.

GyanP: If you look [here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/editor3/store/index.ts#L98), `spellcheckerDisabledInConfig` depends on the configuration settings. That means we can not enable and disable spellchecker from the UI. We can only decide if we want to run spellchecker or not using `Run automatically` toggle button in the UI. So, why to update the `store.spellchecker.enabled`?

tomaskikutis: Yeah, the code is a bit confusing. `spellcheckerDisabledInConfig` and `Run automatically` refer to spellchecker in general. It can be either for tansa or our internal one. `store.spellchecker.enabled` is only controlling our internal one, which doesn't have the concept of "run automatically" - it's either on or off.

Why do you need to disable the spellchecker in the first place?

GyanP: I don't need to disable the spellchecker. I thought setting `store.spellchecker.enabled` to `false` would mean disabling the spellchecker as you mentioned in the second comment.
So, the conclusion is if we turn on `Run automatically` then `store.spellchecker.enabled` should be set to true and vice versa, right?

tomaskikutis: > I don't need to disable the spellchecker.

What is this doing then?
```
if (spellcheck.isAutoSpellchecker !== true) {
    return Promise.resolve({});
}
```

> So, the conclusion is if we turn on Run automatically then store.spellchecker.enabled should be set to true and vice versa, right?

yes

GyanP: > What is this doing then?
 If `run automatically` is turned off `spellcheck.isAutoSpellchecker` is set to false and so, we don't need to call the `check` method for finding the errors.

tomaskikutis: You should set `store.spellchecker.enabled` to false for this use case. Let me know if you have any more doubts about anything.

scripts/core/editor3/actions/spellchecker.tsx
```diff
@@ -36,9 +37,10 @@ export function reloadSpellcheckerWarnings() {
     };
 }
 
-export function setSpellcheckerLanguage() {
+export function setSpellcheckerLanguage(language: string) {
```
tomaskikutis: Why take language as a parameter? Is there a case when some other language than currently selected is used?

scripts/core/editor3/service.ts
```diff
@@ -149,7 +149,7 @@ export class EditorService {
         }
 
         if (typeof spellcheck !== 'undefined') {
-            spellcheckerStores.map((s) => s.dispatch(action.setSpellcheckerLanguage()));
+            spellcheckerStores.map((s) => s.dispatch(action.setSpellcheckerLanguage(language)));
```
tomaskikutis: Change `map` to `forEach`. There's no point(except to confuse the developer looking at the code) in using `map` when you don't use its result.

scripts/core/editor3/actions/spellchecker.tsx
GyanP: Sometimes when we frequently change the language of an item, `state.spellchecking` holds the previous language so to make sure it holds the current language only passing it as a parameter.

tomaskikutis: Have you tried to debug what is happening? We should try to fix the incorrect state rather than working around it. If you can't figure it out, describe me how to reproduce it and I'll take a look.

GyanP: I did, and both the `state.item.language` and `state.spellchecking.language` not getting updated immediately. 
How to reproduce:
Remove language as a parameter from here `export function setSpellcheckerLanguage(language: string)`
It happens only when you type something in the body or headline, change the language and then undo the changes using `ctrl+z` and then again change the language, spellchecker would run with the previous language, not with the current one.

tomaskikutis: I've been checking out why it behaves that way and found that not only editor3 store changes, but also the article item itself which shouldn't be related to editor3. Can you debug what is causing the change? here's a video showing `item.language` change: https://drive.google.com/open?id=1qAmcvm3TcoJUQTFTgx3E4v9IKpfNX_WZ

GyanP: Do you mean undoing changes shouldn't impact header fields?

tomaskikutis: No. If you are editing a slugline for example and hit ctrl+z, it's all fine that it undoes the change. The issue is that I was focused on editor3 field in the content section when hitting ctrl+z. If I undo while being focused on one field, it shouldn't undo a different field. Did you find the code that does it?

GyanP: I checked it, and what I found is that we are watching and storing the whole `item` in the history stack, whenever we make a change to any field of the item a new version of that item is pushed into the history stack. So, when we click `ctrl+z`, the previous version of the item from the history stack is applied to the `scope.item`. If we undo once or twice our focus remains on the current field that we are editing depending on how much we have changed on that field. But when we keep on pressing `ctrl+z` continuously it will apply the last version of item from the history stack i.e. version of an item when we last clicked on `save` button.
That means undo operation has nothing to do with the field being focused or not, it depends how many times we click on `ctrl+z` and how many versions are available in the history stack.

tomaskikutis: Thanks for checking it. That's interesting. I never knew the feature existed. I've created SDESK-4569 to fix undo on the field causing global history to undo.

Anyway, spellchecker should work when global undo is used when no fields are focused. It seems to me that the underlying reason you are working around by adding a language parameter is that when you undo a global history, first the spellchecker actions are dispatched and only then the item is changed by the code that manages global history stack. That's probably why you have old data in state. Can you look into this and try to ensure that spellchecker actions are dispatched **after** the item history updated?

GyanP: I will look into it. And btw we are using this [angular history](https://github.com/decipherinc/angular-history) for managing this undo and redo.

GyanP: I looked at it but couldn't find any way to solve this issue. While investigating I found that history service broadcasts [this](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts#L113) message and then we apply the changes. 

Also what to do if external spellcheckers are not found in the backend? We should show some message to the user or simply return an empty array or maybe both?

tomaskikutis: > when you undo a global history, first the spellchecker actions are dispatched and only then the item is changed by the code that manages global history stack

Have you checked if this assumption is correct?

> Also what to do if external spellcheckers are not found in the backend? We should show some message to the user or simply return an empty array or maybe both?

If spellchecker is not found, display an error message and disable spellchecker for that article, without changing user preferences. If that's hard to do, return an empty array.

GyanP: > Have you checked if this assumption is correct?

Spellchecker actions are dispatched only after item gets changed as we are watching the `item.language` and if languages changes then only we run the spellchecker. But the state and store are not updated.

tomaskikutis: Let's keep the language parameter then, but add a **clear** comment describing this issue.

GyanP: For now, I am returning an empty array, as we don't have any way to know if external spellchecker is available or not. If I try to show a message inside the check method it will keep popping up every time we type something. So, not sure but maybe we can create an endpoint to get external spellchecker status based on language.

PR #3028, opened 2019-07-17
Force selection when closing popup on editor3

pablopunk:
SDESK-3885

I put a comment on the code explaining this because it's not intuitive just by looking at it:

Whenever we hide a popup, the ContentState is not changed so it will
trigger these two DraftJS events:

* editOnFocus
* editOnSelect

The first one renders the right selection but the second one uses global `window.getSelection()` (which doesn't exist, as the editor lost focus) to check if the editorState selection matches that one. As it doesn't, it renders the selection in the first character of the first block. Using `forceSelection` won't trigger those events and the selection will be correct.

Before, clicking Submit and Cancel had different behaviours on editor rendering (cancel would trigger 2 renders, submit none). Now they are consistent.

> NOTICE

The task is only about links but this also affects other popups like comments and annotations. If you open one of those popups and then you click cancel, the error was happening before this fix.

### DIFF
--- a/scripts/core/editor3/components/Editor3Component.tsx
+++ b/scripts/core/editor3/components/Editor3Component.tsx
@@ -469,7 +469,7 @@ export class Editor3Component extends React.Component<IProps> {
                         blockRenderMap={blockRenderMap}
                         blockRendererFn={getBlockRenderer({svc: this.props.svc})}
                         customStyleMap={{...customStyleMap, ...this.props.highlightsManager.styleMap}}
-                        onChange={(editorStateNext) => {
+                        onChange={(editorStateNext: EditorState) => {
                             // in order to position the popup component we need to know the position of editor selection
                             // even when it's not focused, or another input is focused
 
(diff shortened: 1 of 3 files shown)


tomaskikutis:
> Whenever we hide a popup, the ContentState is not changed so it will
trigger these two DraftJS events:

Is that because of hiding a popup or because of editor losing focus? Have you bisected to check if it's a regression? I remember working on a similar thing.

pablopunk:
@tomaskikutis I have not tried that, do you remember what the issue was exactly or when did you work on it?

pablopunk:
@tomaskikutis maybe this one? https://github.com/superdesk/superdesk-client-core/pull/2365

tomaskikutis:
Yep, it's this one. Does it work when you checkout that version? Does it brake in later commits?

Your fix is limited to `setPopup` which is good when on the other hand, I had made changes to draft-js in #2365. You can check out what happened in later commits, but I wouldn't put much more effort into doing this in a "better" way, since there may not be one.

pablopunk:
Actually, it did work checking out your branch `fix-cursor-position-when-focusing-programatically` on draft-js, I think that one fixes this issue.

PR #3029, opened 2019-07-18
Fixes and improvements for crud manager and generic list

tomaskikutis:
REQUIRED BY: https://github.com/superdesk/superdesk-planning/pull/1230


scripts/core/helpers/CrudManager.tsx
```diff
@@ -161,24 +133,12 @@ export const dataApi: IDataApi = {
         const queryString = '?' + Object.keys(query).map((key) =>
```
petrjasek: imo would make sense to call `encodeURI` on that

scripts/core/helpers/CrudManager.tsx
```diff
@@ -161,24 +133,12 @@ export const dataApi: IDataApi = {
...
             `${key}=${isObject(query[key]) ? JSON.stringify(query[key]) : query[key]}`).join('&');
```
petrjasek: or rather `encodeURIComponent(query[key])`

.eslintrc.js
```diff
@@ -1,6 +1,35 @@
 const sharedConfigs = require('superdesk-code-style');
 
 module.exports = Object.assign({}, sharedConfigs, {
+    rules: Object.assign(sharedConfigs.rules, {
```
pablopunk: Should include/discuss all this rules in https://github.com/superdesk/superdesk-client-code-style?

scripts/core/helpers/network.tsx
```diff
@@ -0,0 +1,139 @@
...
+}
+
+function httpRequestBase(options: IHttpRequestOptions): Promise<Response> {
+    const {method, url, payload, headers} = options;
+
+    return fetch(url, {
+        method,
+        headers: headers || {},
+        mode: 'cors',
+        body: typeof payload === 'undefined' ? undefined : JSON.stringify(payload),
```
pablopunk: I think the ternary is unnecessary here, what's the difference between this and `body: JSON.stringify(payload)`?

tomaskikutis: You are right, there's none. I thought `JSON.stringify` returns an `'undefined'` string as it does with `null` which turns out isn't true.

.eslintrc.js
tomaskikutis: I want to finish ansa stuff before going on holiday next week, but you can do it without me.

PR #3031, opened 2019-07-23
Copy and paste only metadata fields that are shown

pablopunk:
SDESK-4502

Centralize fields building on a service and only use those fields to copy&paste the metadata.

Before, all the object was being copied which resulted in the image being overwritten.


pablopunk:
I'm closing this PR in favor of https://github.com/superdesk/superdesk-client-core/pull/3035

Although there's still an issue with copy&pasting metadata, I'll create a ticket

PR #3032, opened 2019-07-23
fix search shortcuts

petrjasek:
- when using specific superdesk collections they were missing
- shortcuts with urgency/priority sent wrong query

SDANSA-307 SDANSA-308

### DIFF
--- a/scripts/apps/search/controllers/SearchMenuController.ts
+++ b/scripts/apps/search/controllers/SearchMenuController.ts
@@ -1,5 +1,6 @@
-import {get, isEqual} from 'lodash';
-import {ISavedSearch} from '../SavedSearch';
+import {get, isEqual, cloneDeep} from 'lodash';
+import {ISavedSearch, mapFiltersServerToClient} from '../SavedSearch';
+import {mapPredefinedDateFiltersServerToClient} from '../directives/DateFilters';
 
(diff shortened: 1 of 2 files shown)


scripts/apps/search/controllers/SearchMenuController.ts
```diff
@@ -71,17 +74,18 @@ export default function SearchMenuController(
...
-                    filter: search.filter,
-                }));
+                .map(mapFiltersServerToClient);
 
             // bundle repo and its shortcuts
             this.providers.forEach((provider) => {
                 providers.push(provider);
                 providers = providers.concat($filter('sortByName')(
-                    shortcuts.filter((shortcut) => get(shortcut, 'filter.query.repo', '') === provider._id),
+                    shortcuts.filter((shortcut) => {
```
tomaskikutis: Can you explain what are you filtering here? It's not obvious from the code at all to me. What is `provider._id`? any random string or can also be a name of one of the repositories?

It would also read easier to me if you used `x == null`, `x != null` explicitly instead of `!x` and `if(x)` when your intent is to check if the value is null.

petrjasek: will write it in there rather than here

PR #3033, opened 2019-07-24
Add uppercase and lowercase buttons to editor3

tomaskikutis:
SDANSA-304

DEPENDS ON: SDESK-4531 https://github.com/superdesk/superdesk-core/pull/1638


scripts/core/editor3/reducers/toolbar.tsx
```diff
@@ -244,4 +248,59 @@ const setPopup = (state: IEditorStore, {type, data}) => {
...
+
+    const {selection, changeTo} = payload;
+    const contentState = state.editorState.getCurrentContent();
+
+    let ended = false;
+    let started = false;
+    const startKey = selection.getStartKey();
+    const endKey = selection.getEndKey();
+
+    const nextBlockMap = state.editorState.getCurrentContent().getBlockMap().map((block) => {
```
petrjasek: just wondering - why not using [Modifier.replaceText](https://draftjs.org/docs/api-reference-modifier#replacetext) ?

tomaskikutis: To maintain inline styles I'd need to call `Modifier.replaceText` for every character. On top of that, it requires a `Selection` as one of the arguments, which means I'd need to fiddle with the offsets to account for backward selection.

tomaskikutis: I could use start/end offsets for anchor/focus offset and always create a selection with `backwards: false`, but I'd still need to call it for every char. Replacing ranges seems simpler. Are there any trade-offs you know I'm missing by not using replace-text?

petrjasek: not really, seemed simpler, but I was assuming those inline styles could be an issue, so wanted to check

PR #3036, opened 2019-07-25
As a user I want to be notified when a story is marked for me. (SDBELGA-124)

GyanP:
Requires https://github.com/superdesk/superdesk-core/pull/1634

### DIFF
--- a/scripts/core/menu/notifications/notifications.ts
+++ b/scripts/core/menu/notifications/notifications.ts
@@ -262,6 +262,8 @@ angular.module('superdesk.core.menu.notifications', ['superdesk.core.services.as
                 templateUrl: asset.templateUrl('core/menu/notifications/views/notifications.html'),
                 link: function(scope, elem, attrs, ctrl) {
                     scope.flags = ctrl.flags;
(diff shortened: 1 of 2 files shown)


tomaskikutis:
All features related to "mark for user" should go to `scripts/extensions/markForUser`. There's no extension point for notifications yet, so it will need to be added.

tomaskikutis:
Here's an addition to client API I came up with. An extension would indicate notification IDs it wants to handle and action buttons it wants to provide. It wouldn't provide the buttons themselves, but only a label and `onClick` handler function so API surface is smaller. Feel free to add suggestions @petrjasek @pablopunk 

screenshot_3

pablopunk:
@tomaskikutis sounds good, I'm guessing it should also require a payload to include some data about the notification

tomaskikutis:
Yep, we could add an argument to the click handler so it's `onClick(notification)`. The notification text will come from the server, so we don't need to handle that here. Unless we discover that we need it when implementing then we could adjust the interface.

pablopunk:
is this gonna be an extension?

PR #3037, opened 2019-07-26
Revert "upgrade react to 16.8"

MarkLark86:
Reverts superdesk/superdesk-client-core#3034

@tomaskikutis 
We need to revert this PR as it breaks Planning. 54 of the unit tests are failing and we cannot merge PRs with failing tests.


MarkLark86:
@tomaskikutis 
Nevermind, the culprit was enzyme's react adaptor needed upgrading too.

PR #3039, opened 2019-07-26
improve list performance

petrjasek:
- use pure components to avoid re-renders on mouse moves
- updated components which were highlighted via profiling

SDFID-562


tomaskikutis:
I've made significant changes to interfaces of field components in https://github.com/superdesk/superdesk-client-core/pull/3018. Can you wait for it to get merged to master and then update this PR?

petrjasek:
@tomaskikutis I might have to backport this to 1.30 and that would be problematic after refactoring

PR #3040, opened 2019-07-26
Save changes made to desk if user agrees by clicking save on confirmation prompt. (SDESK-4325)

GyanP:


scripts/apps/desks/controllers/DeskConfigController.ts
```diff
@@ -85,26 +87,46 @@ export function DeskConfigController($scope, $controller, notify, desks, modal)
...
-            gettext('Yes'),
-            gettext('No'));
+            gettext('Save'),
+            gettext('Ignore'))
+            .then(() => {
+                return desks.save($scope.desk.orig, diff).then((res) => {
+                    _.merge($scope.desk.edit, res);
+                    _.merge($scope.desk.orig, res);
+                    if (diff.members) {
+                        // if desk members were changed update them inside desks.deskMembers as well
```
tomaskikutis: You shouldn't change data owned by a service, outside of the service. If you need to update members, move it to the service.

PR #3042, opened 2019-07-29
Show message for style suggestions in the spellchecker

pablopunk:
SDBELGA-113

french spellchecker

### DIFF
--- a/scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.tsx
+++ b/scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.tsx
@@ -51,15 +51,10 @@ export class SpellcheckerContextMenuComponent extends React.Component<IProps> {
         const messageExists = Boolean(message);
 
         return (
-            <div
-                className={'dropdown open suggestions-dropdown'}
-                ref={(el) => (this.dropdownElement = el)}
-                style={{
-                    zIndex: 999,
-                    border: 'solid transparent',
-                    borderWidth: '6px 0',
-                }}
-                data-test-id='spellchecker-menu'
+            <div className={'dropdown open suggestions-dropdown'}
+                ref={(el) => this.dropdownElement = el}
+                style={{zIndex: 999, border: 'solid transparent', borderWidth: '6px 0'}}
+                data-test-id="spellchecker-menu"
             >
                 <ul className={'dropdown__menu'} style={{ position: 'static' }}>
                     <div className='form-label' style={{ margin: '0 16px' }}>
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.tsx
```diff
@@ -17,46 +21,71 @@ export class SpellcheckerContextMenuComponent extends React.Component<IProps> {
...
-                    <div className="form-label" style={{margin: '0 16px'}}>{gettext('Suggestions')}</div>
-                    {
-                        suggestions.length === 0
-                            ? <li><button>{gettext('SORRY, NO SUGGESTIONS.')}</button></li>
-                            : suggestions.map((suggestion, index) =>
+                <ul className={'dropdown__menu'} style={{ position: 'static' }}>
+                    <div className='form-label' style={{ margin: '0 16px' }}>
+                        {gettext('Suggestions')}
+                    </div>
+                    {messageExists && (
```
tomaskikutis: I see you've learned some angular :smile: 

You should never write multiple independent `if` statements for what is actually an `if/else`. When `if/else` is used, it's immediately clear that only one condition can run. When you use independent `if` statements, you remove that link and developer needs to analyze the conditions to understand that you really meant to write an `if/else`.

Use ternaries instead:

```
{
    messageExists
        ? (
            <div>a</div>
        )
        : (
            suggestions.length === 0
                ? (<div>a</div>)
                : (<div>b</div>)
        )
}
```

Nested ternaries might not look the nicest at the first glance either, but at least they represent the logic correctly.

scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.tsx
```diff
@@ -17,46 +21,71 @@ export class SpellcheckerContextMenuComponent extends React.Component<IProps> {
...
+                        <li>
+                            <button onMouseDown={() => this.onSuggestionClick(suggestions[0])}>{message}</button>
+                        </li>
+                    )}
+                    {!messageExists &&
```
tomaskikutis: Are you using boolean logic to simulate ternaries? :D why?

scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.tsx
```diff
@@ -91,4 +120,6 @@ export class SpellcheckerContextMenuComponent extends React.Component<IProps> {
     }
 }
 
-export const SpellcheckerContextMenu = connect()(SpellcheckerContextMenuComponent);
+export const SpellcheckerContextMenu = connect()(
```
tomaskikutis: Why change the formatting?

pablopunk: That was automatic, I'll revert it back

scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.tsx
pablopunk: At first it was not a ternary operator cause I didn't have the `!messageExist` part (it was showing the suggestions even if the message existed) so this was necessary. But yeah when I added the `!messageExists` part I didn't see the big picture

scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.tsx
pablopunk: See comment above

pablopunk:
@tomaskikutis I made it so the message is shown in the loop if it exists, cause the behaviour is the same, check the diff again, it's way less code

PR #3043, opened 2019-07-30
fix errors not highlighted in authoring for custom fields

petrjasek:
`isEmpty` returned true for `true`, so checking explicitly

SDESK-2390

### DIFF
--- a/scripts/core/ui/ui.ts
+++ b/scripts/core/ui/ui.ts
@@ -1048,7 +1048,7 @@ function validationDirective() {
             });
 
             scope.$watch(attrs.sdValidationError, (errors) => {
-                if (isEmpty(errors)) { // valid
+                if (errors == null || errors === '' || errors === false || (errors.length != null && errors.length === 0)) {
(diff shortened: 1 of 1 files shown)


scripts/core/ui/ui.ts
```diff
@@ -1048,7 +1048,7 @@ function validationDirective() {
             });
 
             scope.$watch(attrs.sdValidationError, (errors) => {
-                if (isEmpty(errors)) { // valid
+                if (errors == null || errors === '' || errors === false || (errors.length != null && errors.length === 0)) {
```
tomaskikutis: How about `Array.isArray(errors) && errors.length > 0`? Or isn't this what you are trying to check?

petrjasek:
sometimes the error is just string so that wouldn't work

tomaskikutis:
It would be helpful to add a type then `errors: string | Array<string>` (I'm not sure if the array contains strings, just added this as an example)

PR #3044, opened 2019-07-30
fix(poi):: Fixed issue with setting default poi

pavlovicnemanja:
[SDESK-4496] - Imported photos on Vijesti without point of interest

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
@@ -506,7 +506,7 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, deploy
     };
 
     // init poi if not set
-    if (!$scope.data.poi) {
+    if (!$scope.data.poi || !Object.keys($scope.data.poi).length) {
         $scope.data.poi = {x: 0.5, y: 0.5};
         if (!config.features.validatePointOfInterestForImages) {
             $scope.saveCrops(); // save it as defaults


scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -506,7 +506,7 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, deploy
     };
 
     // init poi if not set
-    if (!$scope.data.poi) {
+    if (!$scope.data.poi || !$scope.data.poi.length) {
```
tomaskikutis: It's not clear what is being checked. Can you write a condition without using `!`? Like `if ($scope.data.poi != null || $scope.data.poi.length > 0)`.

pavlovicnemanja: No, I can't. I need this to be executed if $scope.data.poi doesn't exist or if $scope.data.poi is empty, because in some cases it returns empty object.

pavlovicnemanja: In opposite way I need to wrap whole component.

tomaskikutis: > I need this to be executed if $scope.data.poi doesn't exist or if $scope.data.poi is empty

Can you use this? `if ($scope.data.poi == null || Object.keys($scope.data.poi).length === 0)`

pavlovicnemanja: Why to use that? This is more elegant way to write same thing with less characters. Also, every developer will understand what is going on here. If you want I can add comment above why this is set, but I think that there it is also.

I don't see importance and difference in two ways of writing something. This is my way.

tomaskikutis: When implicit checks like `if (!myVariable)` are used, there are multiple things that it might mean:
* a string is empty
* value is undefined
* value is null
* value is false

When another person reads such code, data types and the intention of the code is less clear. It's not about your or my way, but about code being easy to understand for others. Comments help, but why comment when you can write simple code in the first place?

By the way, is `$scope.data.poi` an array or an object? I see you are here accessing a property `length` which doesn't exist on objects, but in other places in the file, `$scope.data.poi.x` is accessed which wouldn't be present on an array.

pavlovicnemanja: It can't be confused, in logic way it's simple. If object doesn't exists or if object is empty execute this, if not, skip. Thats it. Simple.

This is stupid thing that we argue about. This is pretty simple part of code and means what it means. End. Thanks.

pavlovicnemanja: You are right about length, changing it to Object.keys($scope.data.poi).length

PR #3048, opened 2019-08-01
fix related item references including full item for published items

petrjasek:
SDESK-4574

### DIFF
--- a/scripts/apps/relations/directives/RelatedItemsDirective.ts
+++ b/scripts/apps/relations/directives/RelatedItemsDirective.ts
@@ -199,7 +199,7 @@ export function RelatedItemsDirective(authoringWorkspace, relationsService, noti
                         type: item.type, // used to display associated item types
                     };
                 } else {
-                    data[key] = item; // use full item for external items
+                    data[key] = item; // use full item for external items, like images from external search provider
                 }
 
                 scope.item.associations = angular.extend({}, scope.item.associations, data);
--- a/scripts/apps/relations/directives/RelatedItemsDirective.ts
+++ b/scripts/apps/relations/directives/RelatedItemsDirective.ts
@@ -1,6 +1,10 @@
 import {get} from 'lodash';
 import {getSuperdeskType} from 'core/utils';
 import {gettext} from 'core/utils';
+import {IArticle} from 'superdesk-api';
+
+const ARCHIVE_TYPES = ['archive', 'published'];
+const isInArchive = (item: IArticle) => item._type != null && ARCHIVE_TYPES.includes(item._type);
 
 /**
  * @ngdoc directive
@@ -189,10 +193,13 @@ export function RelatedItemsDirective(authoringWorkspace, relationsService, noti
                 const key = getNextKey(scope.item.associations || {}, scope.field._id);
                 let data = {};
 
-                if ('_type' in item && item._type !== 'archive') {
-                    data[key] = item;
+                if (isInArchive(item)) {
+                    data[key] = {
+                        _id: item._id,
+                        type: item.type, // used to display associated item types
+                    };
                 } else {
-                    data[key] = {_id: item._id};
+                    data[key] = item; // use full item for external items
                 }
 
                 scope.item.associations = angular.extend({}, scope.item.associations, data);
--- a/scripts/apps/relations/services/RelationsService.ts
+++ b/scripts/apps/relations/services/RelationsService.ts
@@ -2,7 +2,8 @@ import {pickBy, zipObject} from 'lodash';
 import {IArticle} from 'superdesk-api';
 import {isPublished} from 'apps/archive/utils';
 
-const RELATED_LINK_KEYS = 2;
+const RELATED_LINK_KEYS = 2; // links only have _id and type keys (and some old ones only _id)
+const isLink = (association) => association != null && Object.keys(association).length <= RELATED_LINK_KEYS;
 
 RelationsService.$inject = ['mediaIdGenerator', 'api', '$q'];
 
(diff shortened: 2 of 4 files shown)


scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -1,6 +1,10 @@
 import {get} from 'lodash';
 import {getSuperdeskType} from 'core/utils';
 import {gettext} from 'core/utils';
+import {IArticle} from 'superdesk-api';
+
+const ARCHIVE_TYPES = ['archive', 'published'];
```
tomaskikutis: add `'published'` to `IArticle['_type']`

scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -189,10 +193,13 @@ export function RelatedItemsDirective(authoringWorkspace, relationsService, noti
...
-                if ('_type' in item && item._type !== 'archive') {
-                    data[key] = item;
+                if (isInArchive(item)) {
+                    data[key] = {
+                        _id: item._id,
+                        type: item.type, // used to display associated item types
+                    };
                 } else {
-                    data[key] = {_id: item._id};
+                    data[key] = item; // use full item for external items
```
tomaskikutis: What's the use case for "external items". If you want to add an item as related, doesn't it make sense to fetch it and make it internal?

scripts/apps/relations/services/RelationsService.ts
```diff
@@ -2,6 +2,8 @@ import {pickBy, zipObject} from 'lodash';
 import {IArticle} from 'superdesk-api';
 import {isPublished} from 'apps/archive/utils';
 
+const RELATED_LINK_KEYS = 2;
```
tomaskikutis: What does `2` and comparisons to it mean?

scripts/apps/search/components/fields/AssociatedItemsList.tsx
```diff
@@ -14,32 +14,50 @@ interface IProps {
 }
 
 interface IState {
-    associations: Array<IArticle>;
+    types: Array<string>;
+}
+
+interface IRelated {
```
tomaskikutis:
```
interface IRelatedArticle {
    _id: IArticle['_id'];
    type: IArticle['type'];;
}
```

This would be clearer. I'm not sure if field relations are correct, so feel free to change them if they aren't.

scripts/apps/search/components/fields/AssociatedItemsList.tsx
```diff
@@ -14,32 +14,50 @@ interface IProps {
...
     getAssociations() {
         const {content} = this.props.svc;
+        const associations = this.props.item.associations || {};
+        const related = Object.values(associations)
+            .filter((_related) => _related != null);
+        const relatedTypes = related
+            .map((_related: IRelated) => _related.type)
+            .filter((type) => type != null);
+
+        if (relatedTypes.length === related.length) { // types for all
```
tomaskikutis: What is this checking? If all related items have types? Why does it matter?

scripts/apps/search/components/fields/AssociatedItemsList.tsx
```diff
@@ -14,32 +14,50 @@ interface IProps {
...
+            this.setState({types: relatedTypes});
+            return;
+        }
 
-        content.fetchAssociations(this.props.item).then((associations) => {
-            this.setState({associations: Object.values(associations)});
+        content.fetchAssociations(this.props.item).then((_associations: Array<IArticle>) => {
```
tomaskikutis: I assume this would fetch not only related items, but media galleries as well. Is that what you want?

petrjasek: yes

scripts/apps/search/components/fields/AssociatedItemsList.tsx
petrjasek: because that field displays how many items are there for each type. so it checks if all associations have types. it's for BC

scripts/apps/relations/services/RelationsService.ts
petrjasek: it checks if association is a link - so it only has _id and type set (or only _id for BC)

scripts/apps/relations/directives/RelatedItemsDirective.ts
petrjasek: that's items from external repos, like scanpix images or ansa photo archive

tomaskikutis: can you add this to the comment in the code?

scripts/apps/relations/services/RelationsService.ts
tomaskikutis: That's a sneaky one.

It's not a sustainable solution to identify it by the number of keys. Can you instead add a property `_internal: boolean` or similar and use that if you need to check it?

I don't mind if you do this in a separate PR. If you decide to keep `RELATED_LINK_KEYS` for now, add a comment describing the meaning behind the magic number :)

tomaskikutis: what is BC by the way?

petrjasek: backward compatibility

tomaskikutis: Backward compatibility is technical debt :) One of the better ones out there, but it's still better to avoid it if possible. Which is possible in this case by writing an upgrade script to add `_internal` property. As I said in a previous comment, it's fine to do this in a separate PR to be able to release the fix faster.

petrjasek: yep will fix that one separately

scripts/apps/search/components/fields/AssociatedItemsList.tsx
tomaskikutis: What kind of backward compatibility are you referring to? The one where associations might have an `_id`, but not `type` in which case you fetch it from a server separately? Can you add a comment to the code about this as well?

It's not easy to understand this as is, after a few months I'll be clueless why the code is the way it is.

tomaskikutis: When you write an upgrade script related to `_internal` property in another PR, It'd be useful if you could write an upgrade script for this as well and then remove the condition you added for compatibility.

petrjasek: yep makes sense

PR #3049, opened 2019-08-02
add desk sent output group to monitoring

petrjasek:
SDANSA-312


tomaskikutis:
Where is the "left toolbar button"? What does "sent from current desk" mean? Does it only refer to items sent to other desks or include publishing or other actions?

If this a new feature, can you add a screenshot/video/gif?

petrjasek:
I've updated the task description, at the end it's not a new button but rather a new stage type in monitoring settings for each desk called Output/Sent:
desk

PR #3051, opened 2019-08-06
Allow mark for user from dropdown menu while editing

pablopunk:
SDBELGA-133

Extends extension API to include a `groupId` on `IArticleAction` so we can include extension actions in existing groups of any menu, without refactoring current options.

I tried to use the existing `getActions` method (and it works) but there are some issues with that:

* The label is `Mark for User` while in this menu it should only be `User`
* If you mark the item for a user, close the editor, and open it again, you will have two options (`unmark` or `mark for another`). This makes sense but it's unpredictable for the user, becuase they would only see that after reopening the item for editing. Keeping the button `User` all the time is a better UX.

That's the reason I implemented a new `getActionsExtra`, so we only show the one action we want right now.

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
@@ -79,8 +79,8 @@ export function AuthoringTopbarDirective(
                     activationResult.contributions &&
                     activationResult.contributions.entities &&
                     activationResult.contributions.entities.article &&
-                    activationResult.contributions.entities.article.getActions
-                        ? activationResult.contributions.entities.article.getActions
+                    activationResult.contributions.entities.article.getActionsExtra
+                        ? activationResult.contributions.entities.article.getActionsExtra
                         : [],
             );
 
--- a/scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
@@ -4,6 +4,7 @@ import {getMarkForUserModal} from '../../../../extensions/markForUser/src/get-ma
 import {extensions} from 'core/extension-imports.generated';
 import {showModal} from 'core/services/modalService';
 import {IArticleAction} from 'superdesk-api';
+import {get, flatMap} from 'lodash';
 
 /**
  * @ngdoc directive
@@ -72,24 +73,23 @@ export function AuthoringTopbarDirective(
                 actionToTrigger.onTrigger();
             };
 
-            Object.keys(extensions).forEach((extension) => {
-                const {activationResult} = extensions[extension];
+            Promise.all(
+                Object.values(extensions)
+                    .map(({activationResult}) => {
+                        const getExtraActions = get(
+                            activationResult,
+                            'contributions.entities.article.getExtraActions',
+                            null,
+                        );
 
-                if (
-                    activationResult &&
-                    activationResult.contributions &&
-                    activationResult.contributions.entities &&
-                    activationResult.contributions.entities.article &&
-                    activationResult.contributions.entities.article.getExtraActions
-                ) {
-                    activationResult.contributions.entities.article.getExtraActions(scope.item)
-                        .then((actions) => {
-                            scope.extraActionsFromExtensions = [
-                                ...scope.extraActionsFromExtensions,
-                                ...actions,
-                            ];
-                        });
-                }
+                        if (getExtraActions) {
+                            return getExtraActions(scope.item);
+                        }
+                    })
+                    .filter(Boolean),
+            ).then((actions) => {
+                console.log(flatMap(actions));
+                scope.extraActionsFromExtensions = flatMap(actions);
             });
         },
     };
--- a/scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
@@ -73,19 +73,20 @@ export function AuthoringTopbarDirective(
                 actionToTrigger.onTrigger();
             };
 
-            const getExtraActionsFromExtensions = flatMap(
+            const getActionsExtraFromExtensions = flatMap(
                 Object.values(extensions).map((ext) => ext.activationResult),
                 (activationResult) =>
                     activationResult.contributions &&
                     activationResult.contributions.entities &&
                     activationResult.contributions.entities.article &&
-                    activationResult.contributions.entities.article.getExtraActions
-                        ? activationResult.contributions.entities.article.getExtraActions
+                    activationResult.contributions.entities.article.getActionsExtra
+                        ? activationResult.contributions.entities.article.getActionsExtra
                         : [],
             );
 
-            Promise.all(getExtraActionsFromExtensions.map((getPromise) => getPromise(scope.item)))
+            Promise.all(getActionsExtraFromExtensions.map((getPromise) => getPromise(scope.item)))
                 .then((actions) => {
+                    console.log(actions);
                     scope.extraActionsFromExtensions = flatMap(actions);
                 });
         },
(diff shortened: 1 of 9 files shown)


scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -56,6 +60,32 @@ export function AuthoringTopbarDirective(TranslationService, privileges, authori
...
+                if (
+                    activationResult &&
+                    activationResult.contributions &&
+                    activationResult.contributions.entities &&
+                    activationResult.contributions.entities.article &&
+                    activationResult.contributions.entities.article.getExtraActions
+                ) {
+                    activationResult.contributions.entities.article.getExtraActions(scope.item)
+                        .then((actions) => {
+                            scope.extraActionsFromExtensions = [
```
tomaskikutis: Use `flatMap` to get actions from all extensions at once as it's done in other places. Otherwise, you'll be triggering multiple re-renders when getting actions from each extension.

scripts/apps/authoring/views/authoring-topbar.html
```diff
@@ -249,8 +249,12 @@
                         {{ :: 'Desks' | translate }}
                       </button>
                       <ul class="dropdown__menu dropdown__menu--submenu-left" sd-mark-desks-dropdown></ul>
+
                     </div>
                   </li>
+                  <li ng-repeat="action in extraActionsFromExtensions">
```
tomaskikutis: It doesn't work when there is no group specified or another group is specified. You can leave it unimplemented, but you have to state it in the comments of the interface. It shouldn't be too hard to at least an ability to add ungrouped items.

scripts/core/superdesk-api.d.ts
```diff
@@ -35,12 +35,13 @@ declare module 'superdesk-api' {
 
     export interface IArticleAction {
         labelForGroup?: string;
+        groupId?: string; // action lists can specify which groups they wanna render via an id
```
tomaskikutis: You shouldn't add groupId here since it won't work in other places `IArticleAction` is used. `labelForGroup` is used for the same purpose in monitoring item action. I added it expecting that a group label might actually be rendered which it isn't as of now. I suggest either removing `labelForGroup` or using another interface.

Since `priority` and `icon` aren't implemented either, you should definitely not use this interface. You can extend the interface and omit some properties to make the interface match the implementation:

```
interface IArticleEditViewAction extends Omit<IArticleAction, 'priority' | 'icon'> {
    groupId?: string; // action lists can specify which groups they wanna render via an id
}
```

scripts/extensions/markForUser/src/managed-marked-user.tsx
```diff
@@ -1,7 +1,7 @@
 import {ISuperdesk, IArticle} from 'superdesk-api';
 import {getMarkForUserModal} from './get-mark-for-user-modal';
 
-export function manageMarkedUserForSingleArticle(superdesk: ISuperdesk, article: IArticle) {
+export function manageMarkedUserForSingleArticle(superdesk: ISuperdesk, article: IArticle, ignoreLock = false) {
```
tomaskikutis: Are you sure you need `ignoreLock`? It was already possible to change marked user from topbar in article edit view even when an article is opened for editing.

pablopunk: How? I can't do that

And without the `ignoreLock` I get:

Captura de pantalla 2019-08-07 a las 14 40 59

tomaskikutis: Never mind. I found a bug where no underscore prefixed properties weren't present on article object. That's why it worked for me without introducing `ignoreLock`. Keep it :)

scripts/core/superdesk-api.d.ts
```diff
@@ -62,6 +63,7 @@ declare module 'superdesk-api' {
                 article?: {
                     getActions?(article: IArticle): Promise<Array<IArticleAction>>;
                     getActionsBulk?(articles: Array<IArticle>): Promise<Array<IArticleActionBulk>>;
+                    getExtraActions?(article: IArticle): Promise<Array<IArticleAction>>;
```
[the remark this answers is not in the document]
pablopunk: But there's a difference between those two markForuser: [it's passing true](https://github.com/superdesk/superdesk-client-core/pull/3051/files#diff-ea9bad13dd1bfd970a7ad71b5b1af87dR10) to ignore that the item is locked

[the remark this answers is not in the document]
tomaskikutis: I'll be fixing a related bug and will take a look at this as well. Ideally, we should change `managed-marked-user.tsx` not to accept a different argument, but take locking into account when passing this argument `superdesk.entities.article.isLocked(article)`. If we made it `superdesk.entities.article.isLocked(article) && article._id !== currentArticleInEditMode` there wouldn't be a difference you mention. Of course, it would require to figure out how to check for the current article in edit mode, but that's an important API that we'll need to add anyway.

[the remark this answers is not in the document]
pablopunk: yep, that would be perfect

scripts/apps/authoring/views/authoring-topbar.html
pablopunk: How about showing them on top (no groups) https://github.com/superdesk/superdesk-client-core/pull/3051/files#diff-b8516f0f8a0f9b0d4d13ad9354c960d3R235-R241

scripts/core/superdesk-api.d.ts
pablopunk: @tomaskikutis I'm thinking about refactoring [this](https://github.com/superdesk/superdesk-client-core/blob/249dc467858084540efa5b36ead70cc87ad4eb60/scripts/apps/search/components/actions-menu/MenuItems.tsx#L216-L228) to use an ID instead of the label

scripts/core/superdesk-api.d.ts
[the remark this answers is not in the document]
tomaskikutis: Fixed in https://github.com/superdesk/superdesk-client-core/pull/3055

scripts/core/superdesk-api.d.ts
tomaskikutis: Sure, go ahead.

scripts/apps/authoring/views/authoring-topbar.html
tomaskikutis: I'd put it at the [bottom of currently ungrouped items](https://user-images.githubusercontent.com/2076953/62716832-19adce80-ba03-11e9-918b-31485d456314.png)

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -61,6 +66,31 @@ export function AuthoringTopbarDirective(
...
+            scope.extraActionsFromExtensions = [];
+
+            scope.triggerActionFromExtension = (actionToTrigger: IArticleAction) => {
+                actionToTrigger.onTrigger();
+            };
+
+            Promise.all(
+                Object.values(extensions)
+                    .map(({activationResult}) => {
+                        const getExtraActions = get(
```
tomaskikutis: You are being too creative here :smile: Just copy-paste it from here and replace the path `https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/search/components/actions-menu/MenuItems.tsx#L48`

- You don't need to map before using `flatMap`. You can return an empty array from `flatMap` callback and it will be ignored in the final result.
- Your current version uses `lodash.get` which means finding usages won't work.

scripts/apps/authoring/views/authoring-topbar.html
pablopunk: @tomaskikutis yeah that's where it is, I didn't explain it properly

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
pablopunk: THe code was doing kind of the same, I'll update it to keep it consistent and avoid `get()`

scripts/core/superdesk-api.d.ts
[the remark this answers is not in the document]
pablopunk: You meant removing `getExtraActions` and calling `getActions` directly or just include some of the actions from `getActions` inside `getExtraActions`?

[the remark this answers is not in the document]
tomaskikutis: The former - include everything from `getActions` and drop `getExtraActions` completely since it's the same thing.

petrjasek:
pls make it on develop branch

PR #3052, opened 2019-08-06
Add buttons for adding features and gallery photos

tomaskikutis:
SDANSA-270

REQUIRED FOR: https://github.com/superdesk/superdesk/pull/3108

demo: https://drive.google.com/open?id=1n-eNuO_P9svHR2rC0aYnBY1V4LgGFNEC


scripts/core/get-superdesk-api-implementation.tsx
```diff
@@ -119,11 +129,33 @@ export function getSuperdeskApiImplementation(
                     });
                 },
             },
+            contentProfile: {
+                get: (id) => {
+                    // Adding simple caching since the function will be called multiple times per second.
+
+                    // TODO: implement synchronous API(and a cache) for accessing
```
petrjasek: are you fixing that?

scripts/apps/authoring/views/article-edit.html
```diff
@@ -106,6 +107,7 @@
          data-allow-audio="true"
          data-item="item"
          data-rel="'featuremedia'"
+         data-field-name="'feature_media'"
```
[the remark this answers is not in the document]
tomaskikutis: I've added the following API for adding an image to a field in article edit view:

`addImage(field: string, image: IArticle): void;`

I didn't want hardcode any particular fields into the API itself, so API allows adding an image to any field given field type supports it. Support for additional field types can be added any time without touching the interface.

There can be many content profiles with different fields, so in order to decide whether to show an icon for adding an image, I need to check if that field is present in the content profile of the article in edit mode. If it is, an event is dispatched. The same event is dispatched no matter which field you want to add an image to. In order to only handle events for that particular field (image gallery or featured media), the field component has to listen to the event and check if it's addressed to it or another field. In order to do it, it has to know which field the component is being rendered for. For example `sd-item-association` is used for both `feature_image` and `feature_media`. For some reason, the field name wasn't previously passed. A property `rel` is being passed, but it contains a field name without underscores, so I can't use it for comparing with the value in the content profile. I've considered removing `rel` and using `field_name` everywhere, but there are too many usages including on the server.

scripts/core/get-superdesk-api-implementation.tsx
tomaskikutis: Not in this PR. It affects this task, but the problem isn't unique to it.

scripts/apps/search/components/Item.tsx
```diff
@@ -173,10 +180,16 @@ export class Item extends React.Component<IProps, IState> {
     }
 
     dbClick(event) {
-        // target can be an image or icon inside a button, so parents need to be checked
-        var parentButton = querySelectorParent(event.target, 'button');
+        const selector = 'button';
```
pablopunk: This function is a duplicate of `select(event)`

PR #3053, opened 2019-08-07
fix: Media gallery stops functioning properly when adding the last item before limit is reached. (SDESK-4596)

GyanP:
media-gallery

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
@@ -16,13 +16,12 @@ interface IScope extends ng.IScope {
     allowPicture: any;
     allowVideo: any;
     carouselItems: any;
-    crtIndex: any;
+    currentIndex: number;
     editable: any;
     field: any;
     item: any;
     items: any;
     maxUploads: any;
-    navCounter: number;
     rel: any;
     goTo(index: any): void;
     navNext(): void;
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -49,17 +50,18 @@ export function ItemCarouselDirective($timeout, notify) {
...
             scope.$watch('items', (items) => {
-                // Don't execute if there are no items or their length is same as before
-                if (items == null || items.length === previousItemsLength) {
+                let field = _.find(items, (item) => !item[item.fieldId]);
+
+                scope.rel = field ? field.fieldId : null;
+
+                // Don't execute if there are no items or their length is same as before and
+                // previousRel (previous next related-key of media item) is same as current rel(next key of media item)
+                if (items == null || items.length === previousItemsLength && previousRel === scope.rel) {
```
tomaskikutis: Why not use lodash `isEqual(items, scope.items)`?

GyanP: Actually, `items` and `scope.items` both are same, as we are watching `scope.iems` and `items` is the current value of `scope.items`.

tomaskikutis: Isn't `items` received in a callback from `watch` is an updated version which you could check against the current state in order to decide if you want to update the view?

GyanP: > items received in a callback from a watch is an updated version.

Yep, I do agree that `items` is an updated version, but it could be used to check against the previous state, not current state, as a watch is triggered only after the current state is changed.
watch-scope

tomaskikutis: When is `previousRel === scope.rel` check relevant? What is it checking?

GyanP: And the thing is we have to reload the carousel only if an item is removed or added, otherwise on every single change carousel will reload and the cursor will be lost. So, it's better to check the length than the whole content.

GyanP: > When is previousRel === scope.rel check relevant? What is it checking?

As I understood `scope.rel` holds the next key of the media that item, let's say we can add 3 items at max in a media gallery field, and we have already added two items with key `media--1, media--2`, then  `scope.rel` will be `scope.rel = media--3`. And when we try to add the third item `scope.rel` would become `null`. So, `previousRel === scope.rel` actually signifies that length is changed or not. So, I think there is no need to check the length of the items as well.

tomaskikutis: > And when we try to add the third item scope.rel would become null. So, previousRel === scope.rel actually signifies that length is changed or not. So, I think there is no need to check the length of the items as well.

These are side-effects. Check what you need directly. You mentioned that carousel needs to be reloaded when items are added or removed - check for that directly:

```
function getItemsCount(items) {
    return Object.keys(items)
        .filter((key) => items[key] != null)
        .length;
}

if (getItemsCount(previousItems) !== getItemsCount(currentItems)) {
    // reload carousel
}
```

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -111,29 +110,35 @@ export function ItemCarouselDirective(notify) {
                     waitForMediaToLoad(mediaItems).then(initCarousel);
                 });
             });
+
+            function getItemsCount(items: Array<any>): number {
+                return Object.values(items)
```
tomaskikutis: Why are you calling `Object.values` on an array?

`getItemsCount` function is pure and doesn't use any state from `ItemCarouselDirective`. Can you move it out of the directive and place it in the root file scope? It's easier to understand the code that way, because when you see a function in some scope, but it doesn't really use anything from that scope, you need to spend time to realize that. On the other hand, when you put it in the root scope, it's impossible for that function to use anything from another scope, so you don't even have to check.

GyanP: > place it in the root file scope?

Which file do you mean?

tomaskikutis: Same file `ItemCarouselDirective.ts`. Move the function to around line 34.

PR #3054, opened 2019-08-07
Fix: Drag and drop of media to body editor 3 from external folder is not working. (SDESK-4297)

GyanP:


scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -54,6 +54,19 @@ export function getValidMediaType(event) {
...
 
+/**
+ * Check if the files dragged from external folder are of allowed type or not
+ * @param mediaType holds type of data being dragged
+ * @param dataTransfer holds data being dragged
+ */
+export function isAllowedMediaType(mediaType: string, dataTransfer: DataTransfer): boolean {
+    if (mediaType === 'Files') {
+        return Object.values(dataTransfer.files).every(
+            (file: File) => /^image/.test(file.type) || /^audio/.test(file.type) || /^video/.test(file.type));
```
tomaskikutis: Don't use regex where exact comparison can be used against `VALID_MEDIA_TYPES`. Why is it a separate function instead of logic being added to `canDropMedia`?

GyanP: `canDropMedia` actually checks if the dragged item belongs to `VALID_MEDIA_TYPES`, which contains `text/html` or `text/uri-list` as well.  But [here](https://github.com/superdesk/superdesk-client-core/blob/66ed3a43f3cda1027fe16af0e240101d71a18e2b/scripts/core/editor3/components/BaseUnstyledComponent.tsx#L54) we have to allow the media only `audio, image, video or superdesk media`, so created a new method.

tomaskikutis: I don't get it. Your code in `isAllowedMediaType` and the code in `canDropMedia` serve the same purpose - determining if media can be dropped. Why do you want to separate it?

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -78,8 +64,17 @@ export function isAllowedMediaType(mediaType: string, dataTransfer: DataTransfer
...
     const {editorFormat, readOnly, singleLine} = editorConfig;
-    const isValidMedia = !!getValidMediaType(e.originalEvent);
     const supportsMedia = !readOnly && !singleLine && editorFormat.indexOf('media') !== -1;
+    const mediaType = getValidMediaType(e.originalEvent);
+    const dataTransfer = e.originalEvent.dataTransfer;
+    let isValidMedia = !!mediaType;
+
+    if (mediaType === 'Files' && dataTransfer.files) {
+        // checks if files dropped from external folder are valid or not
+        isValidMedia = Object.values(dataTransfer.files).every(
```
tomaskikutis: If `isValidMedia` is false at line 74, the entire function will always return false. Can you define a local variable and do an early return instead of overwriting the variable in another block scope?

```
if (mediaType === 'Files' && dataTransfer.files) {
    const isValidFileType = Object.values(dataTransfer.files).every(...);

    if (!isValidFileType) {
        return false;
    }
}

```

scripts/core/editor3/components/BaseUnstyledComponent.tsx
```diff
@@ -51,7 +51,8 @@ class BaseUnstyledComponent extends React.Component<any, any> {
         const blockKey = this.getDropBlockKey();
         const link = event.originalEvent.dataTransfer.getData('URL');
 
-        if (canDropMedia(event, this.props.editorProps) && mediaType.includes('application/superdesk')) {
+        if (canDropMedia(event, this.props.editorProps)
+            && (mediaType === 'Files' || mediaType.includes('application/superdesk'))) {
```
tomaskikutis: Move this part `&& (mediaType === 'Files' || mediaType.includes('application/superdesk'))` to the body of `canDropMedia` as well.

GyanP: It's there to differentiate if a file should go as media or embed.

PR #3058, opened 2019-08-13
[SDESK-4062] Fix scroll in Media Contacts

marwoodandrew:


scripts/apps/contacts/directives/ContactList.ts
```diff
@@ -148,9 +149,9 @@ export function ContactList(
                 return element.scrollTop + element.offsetHeight + 200 >= element.scrollHeight;
             }
 
-            elem.on('keydown', listComponent.handleKey);
+            scrollElem.on('keydown', listComponent.handleKey);
 
-            elem.on('scroll', handleScroll);
+            scrollElem.on('scroll', handleScroll);
```
petrjasek: I think you should also call `scrollElem.off();` on `$destroy` event

marwoodandrew: OK

marwoodandrew: Done

MarkLark86:
@petrjasek This bug fix is required in 1.29 & 1.30 as well (as you cannot load more items once reaching the bottom of the list).

PR #3059, opened 2019-08-13
As a user I want to see scheduled date and time for the items in the 'scheduled output' stage. (SDNTB-593)

GyanP:
scheduled_state

### DIFF
--- a/scripts/apps/search/components/fields/scheduledDateTime.tsx
+++ b/scripts/apps/search/components/fields/scheduledDateTime.tsx
@@ -2,21 +2,25 @@ import React from 'react';
 import {get} from 'lodash';
 import {IPropsItemListInfo} from '../ListItemInfo';
 
-export const scheduledDateTime: React.StatelessComponent<Pick<IPropsItemListInfo, 'item' | 'svc'>> = (props) => {
-    const datetime = props.svc.datetime;
+class ScheduledDateTime extends React.PureComponent<Pick<IPropsItemListInfo, 'item' | 'svc'>> {
+    render() {
+        const datetime = this.props.svc.datetime;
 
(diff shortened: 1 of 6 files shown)


scripts/apps/search/components/fields/scheduledDateTime.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import {get} from 'lodash';
+import {IPropsItemListInfo} from '../ListItemInfo';
+
+export const scheduledDateTime: React.StatelessComponent<Pick<IPropsItemListInfo, 'item' | 'svc'>> = (props) => {
```
petrjasek: would be good to make this a PureComponent so it won't rerender always

scripts/apps/search/components/fields/scheduledDateTime.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import {get} from 'lodash';
+import {IPropsItemListInfo} from '../ListItemInfo';
+
+class ScheduledDateTime extends React.PureComponent<Pick<IPropsItemListInfo, 'item' | 'svc'>> {
+    render() {
+        const datetime = this.props.svc.datetime;
+
+        if (this.props.item.state != null && this.props.item.state === 'scheduled') {
+            const scheduled = get(this.props.item, 'archive_item.schedule_settings.utc_publish_schedule');
```
pablopunk: is `get` necessary if you're not providing a default value nor checking null afterwards? I'd make it `const scheduled = this.props.item.archive_item.schedule_settings.utc_publish_schedule` and check for `null` in the previous `if`. I'm guessing it should never be null but in case it is, we shouldn't try to format it right?

Also `get` is not very good when we need typechecking :)

PR #3065, opened 2019-08-16
fix e2e tests

petrjasek:
use dropdown for multi actions when it's there


tomaskikutis:
How come these tests stopped working @petrjasek ?

petrjasek:
it didn't stop, but it was failing sometimes due to slow server response or animation maybe

PR #3067, opened 2019-08-20
[SDESK-4641] (fix): Item not being removed from workqueue on publish

MarkLark86:


MarkLark86:
@petrjasek This one will be required in 1.29 and 1.30 version branches

PR #3068, opened 2019-08-20
Fixed: Media carousel breaks once the order of the items is changed. (SDESK-4621)

GyanP:


scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -77,12 +86,13 @@ export function ItemCarouselDirective(notify) {
              * otherwise carousel height is messed up
              */
             scope.$watch('items', (items: Array<any>) => {
-                // Don't execute if there are no items or their length is same as before
-                if (items == null || previousItems && getItemsCount(items) === getItemsCount(previousItems)) {
+                // Don't execute if there are no items or their length is same as before and their order is unchanged
+                if (items == null || previousItems && getItemsCount(items) === getItemsCount(previousItems)
```
petrjasek: wouldn't [$watchCollection](https://docs.angularjs.org/api/ng/type/$rootScope.Scope#$watchCollection) do the same thing?

GyanP: Do you mean replacing $watch with $watchCollection?

petrjasek: exactly

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -76,13 +85,14 @@ export function ItemCarouselDirective(notify) {
              * Initialize carousel after all content is loaded
              * otherwise carousel height is messed up
              */
-            scope.$watch('items', (items: Array<any>) => {
-                // Don't execute if there are no items or their length is same as before
-                if (items == null || previousItems && getItemsCount(items) === getItemsCount(previousItems)) {
+            scope.$watchCollection('items', (items: Array<any>) => {
+                // Don't execute if there are no items or their length is same as before and their order is unchanged
+                if (items == null || previousItems && getItemsCount(items) === getItemsCount(previousItems)
+                    && !isOrderChanged(items, previousItems)) {
```
petrjasek: I think there should be no need for this now, it should trigger watch when order changes or there are items added/removed

GyanP: Actually, we have to prevent it from reloading the carousel on every change like if we are typing something in the description of media. If we don't do so on a single keypress watch gets triggered, carousel reloads and the cursor is lost and also blinking is observed.

petrjasek: that should be fine, it's shallow watching the items in collection, but should handle order changes and when there is an item added/removed

GyanP: You are right but [it's](https://drive.google.com/file/d/1l_R3-ag_wcybOIZjnCN6ISSkPJs5Jdoe/view) behaving quite strangely. Not sure but maybe because of [this](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts#L55) binding.

GyanP: the hash key is the reason why it gets triggered, I guess

carousel1

petrjasek: I think that's added by ng-repeat, you can use `track by $index` to avoid it here:
https://github.com/superdesk/superdesk-client-core/blob/46f6adca7761ab9f3360a02614f2619089aebf58/scripts/apps/authoring/views/item-carousel.html#L14
and here:
https://github.com/superdesk/superdesk-client-core/blob/46f6adca7761ab9f3360a02614f2619089aebf58/scripts/apps/authoring/views/item-carousel.html#L152

GyanP: Indeed it removed the `hashkey`, but the watch is still behaving strangely.

Screenshot from 2019-08-21 18-55-09

Here, if I change a field of media item i.e. description, then both previous, as well as a new value, is the same in the watch callback. But if I remove an item then previous value and new value in watch callback are different. Leaving aside swallow or deep watching the values in the first case must be different.

petrjasek: so it calls that watch listener if you edit description? where is that stored?

GyanP: yep
https://github.com/superdesk/superdesk-client-core/blob/46f6adca7761ab9f3360a02614f2619089aebf58/scripts/apps/authoring/views/item-carousel.html#L135

petrjasek: hm seems like problem is the autosave, it changes the `items` instance

PR #3074, opened 2019-08-22
Allow mark for user from dropdown menu while editing

pablopunk:
SDBELGA-133

[OLD PR](https://github.com/superdesk/superdesk-client-core/pull/3051)

Extends extension API to include a `groupId` on `IArticleAction` so we can include extension actions in existing groups of any menu, without refactoring current options.

I tried to use the existing `getActions` method (and it works) but there are some issues with that:

* The label is `Mark for User` while in this menu it should only be `User`
* If you mark the item for a user, close the editor, and open it again, you will have two options (`unmark` or `mark for another`). This makes sense but it's unpredictable for the user, becuase they would only see that after reopening the item for editing. Keeping the button `User` all the time is a better UX.

That's the reason I implemented a new `getActionsExtra`, so we only show the one action we want right now.

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
@@ -79,8 +79,8 @@ export function AuthoringTopbarDirective(
                     activationResult.contributions &&
                     activationResult.contributions.entities &&
                     activationResult.contributions.entities.article &&
-                    activationResult.contributions.entities.article.getActions
-                        ? activationResult.contributions.entities.article.getActions
+                    activationResult.contributions.entities.article.getActionsExtra
+                        ? activationResult.contributions.entities.article.getActionsExtra
                         : [],
             );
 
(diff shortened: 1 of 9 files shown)


scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -1,5 +1,10 @@
 import {AuthoringWorkspaceService} from '../services/AuthoringWorkspaceService';
 import {getSpellchecker} from 'core/editor3/components/spellchecker/default-spellcheckers';
+import {getMarkForUserModal} from '../../../../extensions/markForUser/src/get-mark-for-user-modal';
```
petrjasek: not used it seems

scripts/apps/authoring/views/authoring-topbar.html
```diff
@@ -232,6 +232,13 @@
                   </li>
                 </ul>
 
+                <ul ng-if="extraActionsFromExtensions">
+                    <!-- Extra actions from extensions without a group id -->
+                    <li ng-repeat="action in extraActionsFromExtensions">
```
petrjasek: I think you could add that ng-if here so there would be no empty `<li>` tags

PR #3075, opened 2019-08-22
Improvements to annotations library UI

pablopunk:
SDFID-558

* Instance language selected by default
* Removed language column
* Made names bold and description plain text

Kapture 2019-08-21 at 15 09 25

### DIFF
--- a/scripts/core/get-superdesk-api-implementation.tsx
+++ b/scripts/core/get-superdesk-api-implementation.tsx
@@ -193,7 +193,7 @@ export function getSuperdeskApiImplementation(
         },
         state: applicationState,
         instance: {
-            config: JSON.parse(JSON.stringify(config || {})), // deep clone of config object
+            config: JSON.parse(JSON.stringify(config || {})), // deep clone of config object
         },
         ui: {
             article: {
--- a/scripts/core/get-superdesk-api-implementation.tsx
+++ b/scripts/core/get-superdesk-api-implementation.tsx
@@ -199,7 +199,7 @@ export function getSuperdeskApiImplementation(
         },
         state: applicationState,
         instance: {
-            config: JSON.parse(JSON.stringify(config || {})), // deep clone of config object
+            config,
         },
         ui: {
             article: {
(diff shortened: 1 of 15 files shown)


scripts/core/get-superdesk-api-implementation.tsx
```diff
@@ -197,6 +198,9 @@ export function getSuperdeskApiImplementation(
             },
         },
         state: applicationState,
+        instance: {
+            config: JSON.parse(JSON.stringify(config || {})), // deep clone of config object
```
petrjasek: imo it would make sense to have it in some utils `utils.cloneDeep` or use lodash for it

scripts/core/helpers/UserHtmlSingleLine.tsx
```diff
@@ -1,8 +1,26 @@
 import React from 'react';
 
+interface IProps {
+    html: string;
+    showAsPlainText?: boolean; // if true, html is shown as plain text (no format)
+}
+
+function stripHtmlTags(html) {
```
petrjasek: we have this on `core.utils`

scripts/core/ui/components/ListPage/generic-list-page.tsx
```diff
@@ -175,35 +175,35 @@ export class GenericListPageComponent<T extends IBaseRestApiResponse>
...
+            const value = filterValues[key];
+
+            if (typeof value === 'string') {
+                let trimmedValue = value.trim();
+
+                if (trimmedValue.length > 0) {
+                    accumulator[key] = trimmedValue;
+                    return accumulator;
+                } else {
+                    return accumulator;
```
petrjasek: it could be like:
```
if (trimmedValue.length > 0) {
    accumulator[key] = trimmedValue;
}

return accumulator;
```

scripts/core/get-superdesk-api-implementation.tsx
pablopunk: I'm actually gonna pass the object directly because there's no advantage of cloning here and also the method I used is not reliable because JSON doesn't support values like Date, functions, etc

scripts/core/ui/components/ListPage/generic-list-page.tsx
pablopunk: I didn't write this function so I didn't really look into the implementation, but yeah you're right that's quite redundant

PR #3079, opened 2019-08-26
Use develop for comparing lint rules

pablopunk:


petrjasek:
seems like it doesn't get develop branch on travis, I guess running `git fetch develop` would be good, could be part of `.travis.yml`

.travis.yml
```diff
@@ -14,6 +14,7 @@ cache:
         - $HOME/.cache/yarn
 
 script:
+    - git fetch origin
```
petrjasek: maybe `git branch develop origin/develop` will do the trick

petrjasek: but you have to probably do that fetch first

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -212,7 +212,7 @@ export class Editor3Component extends React.Component<IProps> {
         return !canDropMedia(e, this.props);
     }
 
-    handleDropOnEditor(selection: SelectionState, dataTransfer: Object, isInternal: DraftDragType): DraftHandleValue {
+    handleDropOnEditor(selection: SelectionState, dataTransfer: any, isInternal: DraftDragType): DraftHandleValue {
```
tomaskikutis: `dataTransfer` argument is not used at all. Can you remove it?

pablopunk: @tomaskikutis the argument is coming from https://draftjs.org/docs/api-reference-editor.html#handledrop

PR #3083, opened 2019-08-29
handle "ignore" with external dictionaries

jerome-poisson:
`ignore` key can now be used with backend to ignore words with external
dictionaries. This patch does the client side.

SDBELGA-165

### DIFF
--- a/scripts/core/editor3/components/spellchecker/default-spellcheckers.tsx
+++ b/scripts/core/editor3/components/spellchecker/default-spellcheckers.tsx
@@ -12,7 +12,7 @@ const actions: {[key: string]: ISpellcheckerAction} = {
     ignoreWord: {
         label: gettext('Ignore word'),
         perform: (warning: ISpellcheckWarning) => ng.getService('spellcheck').then((spellcheck) => {
-            return spellcheck.addWord(warning.text, false);
+            return spellcheck.addWord(warning.text, true);
         }),
     },
 };
@@ -60,6 +60,7 @@ export function getSpellchecker(language: string): ISpellchecker {
         fr: 'grammalecte',
         nl: 'leuven_dutch',
     })[language];
+    const ignore = spellcheck.getIgnoredWords();
 
     if (spellcheckerName == null && spellcheck.isActiveDictionary === false) {
         return null;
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/components/spellchecker/default-spellcheckers.tsx
```diff
@@ -68,7 +69,10 @@ export function getSpellchecker(language: string): ISpellchecker {
         return {
             check: (str: string) => httpRequestJsonLocal<{errors: Array<ISpellcheckWarning>}>({
                 method: 'POST',
-                payload: {spellchecker: spellcheckerName, text: str},
+                payload: {
+                    spellchecker: spellcheckerName,
+                    text: str,
+                    ignore: ignore},
```
tomaskikutis: Aren't ignored words stored in the database? Why send it over the wire every time spellcheck runs instead of accessing it from the back-end directly?

jerome-poisson: Hi Tomas,

no they are not stored in database, that the case only with the `add to dictionary` feature. Ignored words are per session only and should disappear with a single refresh (see the ticket for details).

scripts/core/spellcheck/spellcheck.ts
```diff
@@ -389,6 +389,16 @@ function SpellcheckService($q, api, dictionaries, $rootScope, $location, _, pref
         getItemIgnored()[word] = 1;
     }
 
+    /**
+     * Return list of ignored words
+     *
+     * @return {Array} list of words
+     */
+    this.getIgnoredWords = getIgnoredWords;
```
tomaskikutis: Why not inline the function?

```
this.getIgnoredWords = getIgnoredWords(): Array<string> {
    return Object.keys(getItemIgnored());
}
```

scripts/core/editor3/components/spellchecker/default-spellcheckers.tsx
tomaskikutis: Got it :+1:

scripts/core/spellcheck/spellcheck.ts
jerome-poisson: I just did like the other methods for consistency (e.g. `ignoreWord` above), as I'm usually not working with client I'm not sure of coding style :). I'll change that :+1:

PR #3086, opened 2019-08-30
Publisher side menu icon

darconny:
- active state icon fix in Output control page


darconny:
What is better, to complicate things or solve it in a simple way? This is a one line fix, instead you have to make extension point for the minor issue like this one (active state of the side menu button). No problem, I will create a ticket and assign it to you.

tomaskikutis:
I'm all for simplicity, but it is not only lines of code that matter. The separation between systems/concerns is very important for maintainability. As time passes, "quick fixes" accumulate and coupling makes subsequent changes harder to do.

PR #3087, opened 2019-09-02
[SDESK-4643]

fritzSF:


styles/sass/archive-preview.scss
```diff
@@ -1323,3 +1323,16 @@ figure.preview-overlay-on {
         }
     }
 }
+
+.content-item-preview, .preview-content {
+    .preview-label {
```
tomaskikutis: Why are you namespacing it to `.content-item-preview, .preview-content`? It's hard to work with classes when they are styled differently depending on where in the DOM tree they are inserted.

fritzSF: In this case, it is needed as the CSS is heavily nested.

PR #3088, opened 2019-09-03
Prevent drag and drop of picture from preview pane. (SDESK-4665)

GyanP:


tomaskikutis:
Can you instead check where the image is coming from on `drop` event? It would require fewer changes in the code that way. For example `ItemRendition.ts` file wouldn't need to be changed.

GyanP:
> Can you instead check where the image is coming from on `drop` event?

Do you mean check if an image URL belongs to superdesk or current domain? And based on that reject it on the drop event. If yes then, we will have to handle it separately for editor2 and editor3.
So, IMO prevent the dragging is a much better option.

tomaskikutis:
I meant you to check the dropped DOM element, go up the DOM tree and then check if any of the parents is preview pane and cancel the operation if it is. I see that drop event probably doesn't contain the DOM element so this wouldn't work. Let's keep the current implementation then.

PR #3091, opened 2019-09-04
add error/success notifications for unpublish actions

petrjasek:
Unpublish action can fail when the item is in a package.

SDFID-573

### DIFF
--- a/scripts/apps/authoring/authoring/index.ts
+++ b/scripts/apps/authoring/authoring/index.ts
@@ -1,4 +1,5 @@
 import 'angular-history/history';
+import {get} from 'lodash';
 
 import * as svc from './services';
 import * as directive from './directives';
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/index.ts
```diff
@@ -359,20 +360,35 @@ angular.module('superdesk.apps.authoring', [
...
+                            notify.success(gettext('Item was unpublished successfully.'));
+                        };
+
                         familyService.fetchRelatedByState(item.archive_item, [ITEM_STATE.PUBLISHED])
                             .then((items) => {
                                 relatedItems = items;
 
                                 const unpublish = (selected) => {
-                                    authoring.publish(item.archive_item, {}, 'unpublish');
+                                    authoring.publish(item.archive_item, {}, 'unpublish')
```
tomaskikutis: What about handling the error in the authoring service itself?

scripts/apps/authoring/authoring/index.ts
```diff
@@ -359,20 +360,35 @@ angular.module('superdesk.apps.authoring', [
...
-                controller: ['data', 'authoring', 'familyService',
-                    (data, authoring, familyService) => {
+                controller: ['data', 'authoring', 'familyService', 'notify',
+                    (data, authoring, familyService, notify) => {
                         const item = data.item;
                         let relatedItems = [];
 
+                        const handleError = (reason) => {
+                            if (reason != null && reason.status === 400 && get(reason, 'data._issues')) {
+                                notify.error(gettext(`This item is in a package.
```
tomaskikutis: Use a message received from the server for displaying a notification. Just because it's error 400 and has some issues in the payload doesn't mean it's a package related issue. It might be the case at this point in time, but it's naive to think we won't change publishing logic in the future.

petrjasek: you know what - you might be right :wink:

PR #3092, opened 2019-09-05
fix(editor3) dragging text on chrome

pablopunk:
SDESK-4575

htmlComesFromDraftjsEditor was not working when manually selecting
one/more words on the same line on chrome, as it removed all html
references of draft-js

### DIFF
--- a/scripts/core/editor3/components/BaseUnstyledComponent.tsx
+++ b/scripts/core/editor3/components/BaseUnstyledComponent.tsx
@@ -5,6 +5,7 @@ import {getValidMediaType, canDropMedia} from './Editor3Component';
 import {moveBlock, dragDrop, embed} from '../actions/editor3';
 import {getEmbedObject} from './embeds/EmbedInput';
 import {htmlComesFromDraftjsEditor} from 'core/editor3/helpers/htmlComesFromDraftjsEditor';
+import {htmlIsPlainTextDragged} from 'core/editor3/helpers/htmlIsPlainTextDragged';
 
 const EDITOR_BLOCK_TYPE = 'superdesk/editor3-block';
 
@@ -14,17 +15,6 @@ export function isEditorBlockEvent(event) {
 
 export function getEditorBlock(event) {
     return event.originalEvent.dataTransfer.getData(EDITOR_BLOCK_TYPE);
-}
-
-// Dragging a single line in chrome removes DraftJS's data
-// and only sends a <span> containing the dragged text next
-// to a <meta> tag
-function htmlIsPlainTextDragged(html: string): boolean {
-    const parser = new DOMParser().parseFromString(html, 'text/html');
-
-    return parser.querySelectorAll('span').length === 1 &&
-        parser.querySelectorAll('meta').length === 1 &&
-        parser.querySelectorAll(':not(span):not(meta):not(html):not(head):not(body)').length === 0;
 }
 
 class BaseUnstyledComponent extends React.Component<any, any> {
--- a/scripts/core/editor3/components/BaseUnstyledComponent.tsx
+++ b/scripts/core/editor3/components/BaseUnstyledComponent.tsx
@@ -16,6 +16,17 @@ export function getEditorBlock(event) {
     return event.originalEvent.dataTransfer.getData(EDITOR_BLOCK_TYPE);
 }
 
+// Dragging a single line in chrome removes DraftJS's data
+// and only sends a <span> containing the dragged text next
+// to a <meta> tag
+function htmlIsPlainTextDragged(html: string): boolean {
+    const parser = new DOMParser().parseFromString(html, 'text/html');
+
+    return parser.querySelectorAll('span').length === 1 &&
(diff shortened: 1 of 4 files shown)


tomaskikutis:
Does drag & dropping text fragments work with comments/suggestions/formatting? If it doesn't it might be better to disable it entirely.

scripts/core/editor3/helpers/htmlComesFromDraftjsEditor.ts
```diff
@@ -1,2 +1,18 @@
...
+    const parser = new DOMParser().parseFromString(html, 'text/html');
+    const comesFromeditor = parser.querySelector('[data-offset-key]') != null;
+
+    if (comesFromeditor) {
+        return true;
+    }
+
+    // Dragging a single line in chrome removes DraftJS's data
+    // and only sends a <span> containing the dragged text next
+    // to a <meta> tag
```
tomaskikutis: How do you know then if it's a meta+span from superdesk or some text from wikipedia for example? Wouldn't dragging from anywhere produce meta+span ?

pablopunk: well I assume if you only drag text, then the embed shouldn't be created, it should just drop that text there, which is what is doing right now

tomaskikutis: It sounds like you can't determine where the text is coming from and make the assumptions based on the content. I call this lying in the code :)

You can make assumptions, but you need naming to reflect that. If it was called `htmlMaybeComesFromDraftjsEditor` it would be more appropriate, because it indicates to the developer that the function is imprecise.

The problem with the current code is that someone may use the function in the future and it will produce false positives. It's best to avoid false positives / false negatives entirely. For example, in this case, if there is no way to determine if the text is coming from draft-js or not, you could rename the function(as it was before your change) to `htmlContainsDraftJsBlock` and it would then be accurate to its name and wouldn't produce false positives/negatives.

Somewhere in the embed/pasting code, you could then add some code doing the guesswork based on content which would produce the same end result without making a function behave differently than its name implies.

pablopunk: I can get the old function back and make a new one to "guess" it's just plain text and check both when dropping to determine if the `embed` will be created 👍

scripts/core/editor3/components/BaseUnstyledComponent.tsx
```diff
@@ -16,6 +16,17 @@ export function getEditorBlock(event) {
...
 
+// Dragging a single line in chrome removes DraftJS's data
+// and only sends a <span> containing the dragged text next
+// to a <meta> tag
+function htmlIsPlainTextDragged(html: string): boolean {
+    const parser = new DOMParser().parseFromString(html, 'text/html');
+
+    return parser.querySelectorAll('span').length === 1 &&
+        parser.querySelectorAll('meta').length === 1 &&
+        parser.querySelectorAll(':not(span):not(meta):not(html):not(head):not(body)').length === 0;
```
tomaskikutis: What if it contains a link or a div. Wouldn't it be better to check by locating the root span and comparing `innerHTML` with `innerText`?

pablopunk: but a link or a div probably indicates that an `embed` should be created shouldn't it?

tomaskikutis: Probably. But if you have `<span><a href="...">...</a><span>` your function will determine that it's plain text, won't it? That's why I suggest comparing `innerHTML` / `innerText` instead.

pablopunk: It won't determine that's plain text because of the last check. I'll add a test to confirm that. We can keep adding cases in case we ever run into this issue again.

tomaskikutis: I now see it won't. It would think that `<span><body>text</body></span>` is plain text though. I know it's invalid HTML, but why wouldn't you use `innerHTML`/`innerText` comparison?

The current condition looks like checking if a button is green, by adding a condition that it's not blue, not red, not yellow etc.

pablopunk: it's more like checking it's green and it's not any other color than green, which sounds right to me.

What would you do with the innerHTML? If you create a DOMParser it will still add a html/body/head tag

tomaskikutis: If I understand correctly, you want to check that HTML contains a single span which itself only contains plain text. This looks more clear and straight-forward to me:

```
var doc = new DOMParser().parseFromString('...', 'text/html');
doc.body.childElementCount === 1 && doc.body.childNodes[0].tagName === 'SPAN' && doc.body.childNodes[0].innerHTML === doc.body.childNodes[0].innerText
```

PR #3093, opened 2019-09-06
fix: explanation of the mistake is mixed with suggestions with external spellchecker. (SDBELGA-171)

GyanP:


scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.tsx
```diff
@@ -54,6 +54,12 @@ export class SpellcheckerContextMenuComponent extends React.Component<IProps> {
                 data-test-id="spellchecker-menu"
             >
                 <ul className={'dropdown__menu'} style={{position: 'static'}}>
+                    {messageExists &&
+                        <Fragment>
+                            <div style={{margin: '0 16px'}}>{message}</div>
```
tomaskikutis: `<div>` is not valid as a child of `<ul>`. Use `<li>`.

scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.tsx
```diff
@@ -66,7 +72,7 @@ export class SpellcheckerContextMenuComponent extends React.Component<IProps> {
                                         }
                                         data-test-id="spellchecker-menu--suggestion"
                                     >
-                                        {messageExists ? message : suggestion.text}
+                                        {suggestion.text}
```
tomaskikutis: Did you check if it works well with `grammalecte`? I think @pablopunk added the ternary to handle a case when `suggestion.text` is whitespace. The message is then was used so the button can be clicked. Can you test if your change works with white-space suggestions?

GyanP: yep, it doesn't look good when the suggestion is whitespace.

space-suggestion

scripts/core/editor3/components/spellchecker/SpellcheckerContextMenu.tsx
```diff
@@ -43,9 +43,11 @@ export class SpellcheckerContextMenuComponent extends React.Component<IProps> {
         const {suggestions, message} = this.props.warning;
         const {spellchecker} = this.props;
 
-        // If the message exists, use it as the button text
-        // instead of the suggestion
+        // If the message exists, and suggestion is whitespace suggestion
+        // use message as the button text instead of the suggestion
         const messageExists = Boolean(message);
+        const isWhitespaceSuggestion = suggestions.length === 1
+            && (suggestions[0].text === ' ' || suggestions[0].text === '');
```
tomaskikutis: What if there are more than one suggestions and only one of them is whitespace? What if whitespace is multiple characters instead of one? While the second one is unlikely, it's not hard to handle so why not? Use `trim().length > 0` to check for whitespace.

PR #3095, opened 2019-09-06
Marked for user  bugfix marking from topbar

tomaskikutis:
SDBELGA-123


end-to-end-testing-helpers/tsconfig.json
```diff
@@ -0,0 +1,30 @@
+{
+    "compilerOptions": {
```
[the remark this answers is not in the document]
tomaskikutis: I'll try doing this.

tomaskikutis:
I've looked at lerna webpacge, but it's not clear to me what the tool is for. What do you expect it would help with?

petrjasek:
it should be possible to eg run lint check on all extensions/packages via single command, run tests, install dependencies

tomaskikutis:
In our current setup, all dependencies for extensions are already being installed in `postinstall` script in the root package. Lint task for the main package checks the extensions too. There aren't any unit tests in extensions yet, but we could include them in the same run from the root package as we do with linting.

I'm not convinced it's worth adding lerna. Is there anything I'm missing?

petrjasek:
yep might not be worth it, imo it's targeting monorepos where you publish packages to npm

scripts/extensions/markForUser/spec/mark_for_user_spec.ts
```diff
@@ -0,0 +1,44 @@
...
+        // mark item for current user
+        const articleItem = els(['article-item']).get(0);
+
+        hover(articleItem);
+
+        el(['context-menu-button'], null, articleItem).click();
+        el(['context-menu']).element(by.buttonText('Mark for user')).click();
+        el(['mark-for-user-modal', 'modal-body', 'select-user-dropdown', 'dropdown-button']).click();
+        els(['mark-for-user-modal', 'modal-body', 'select-user-dropdown', 'option']).get(1).click();
+        el(['mark-for-user-modal', 'modal-footer', 'confirm']).click();
```
petrjasek: I think we should use those helpers/page objects, this is too low level, hard to read

petrjasek: I mean this whole `it` function not just this line

tomaskikutis: I don't like helper modules in e2e tests. It's not that easy to see which elements actions are being performed on when it's distributed in multiple functions which can call each other. It's not been a pleasure to debug either.

Helper modules would help for complex interactions with reusable elements like selecting a value from a multi-level dropdown or to adding values to tag input. I'd rather keep it simple since none of the code here is used anywhere else.

petrjasek: it would make the spec readable without knowing the markup, being reusable is a bonus.

tomaskikutis:
```
// mark item for current user
// open marked item in read-only mode from "marked for me" dropdown
// change marked user
// check if marked user change is persisted after reload
```

If I split it into funcions it would probably read similarly as comments above which are already there. The only difference would be that the code would be omitted which I don't see as giving more readability. If you aren't interested in the code you can skip reading it now as well and only look at the comments.

It's test code anyway, no one is going to read it if it works, and when it doesn't and needs debugging, it will be helpful to have everything in the same place.

petrjasek: if you debug it the function names make a difference, you will see those in tracelog but not the comments

tomaskikutis: If line numbers aren't visible in the stack trace, wrapping stuff in functions is not how it's solved :slightly_smiling_face:

I don't remember it being the issue. I see the line numbers when tests fail. The only improvement regarding stack traces that'd be helpful to me is source maps, so line numbers are displayed relative to typescript files, not generated `.js` ones. Even though it's not that hard to open the generated file and map the line manually.

MarkLark86: The style guide from Protractor (https://www.protractortest.org/#/style-guide) recommends storing page objects to variables and using those. Reasons they include are:

~~~
Page Objects help you write cleaner tests by encapsulating information about
the elements on your application page. A page object can be reused across
multiple tests, and if the template of your application changes, you only need to
update the page object.

Use Page Objects to interact with page under test
Why?
- Encapsulate information about the elements on the page under test
- They can be reused across multiple tests
- Decouple the test logic from implementation details
~~~

tomaskikutis: Information about the elements is already decoupled from tests here. Tests only reference to `data-test-id` attributes which won't change unless a feature itself changes. Changes in implementation won't impact the tests.

Suggestions from protractor page are useful in terms of how tests are generally written using protractor which is by highly coupling tests to the implementation details by referring to bindings, models, and repeaters in tests. It's bad practice to do so in the first place, so advice on how to deal with the consequences of this isn't that relevant here where that coupling isn't present.

I'd consider moving code into functions if either that code was required to be used somewhere else or had high exposure to many scope variables which would make it harder to track their usages, but neither is the case. Splitting the code further would add another layer to go through when reading or debugging the code without any benefits I can see.

MarkLark86: The following are common to all of superdesk:

- `el(['workspace-navigation', 'Monitoring']).click();`: (place in`spec/helpers/workspace.ts`)
- `const articleItem = els(['article-item']).get(0);`: (place in `spec/helpers/search.ts`)
- `el(['context-menu']).element(by.buttonText('Mark for user')).click();` (place in `spec/helpers/search.ts`)
- el(['authoring-topbar', 'marked-for-user', 'user-avatar']).getAttribute('title') (place in `spec/helpers/authoring.ts`)

As you can see there are page objects in your tests that are not specific to markForUser. You cannot assume the user will know the `data-test-id` of a specific UI element (event the developer who added this id will forget it 6 months down the track). And if the client-core get's refactored, these specifics might change, but using helpers will help with this.

The idea behind using page helpers is to reduce the required refactoring when changes are made by re-using the same helper functions.

It also aids in reading the specs, i.e. `workspace.sideNav('Monitoring').click()`

tomaskikutis: It's important to understand that `data-test-id`s are coupled not to elements, but to user interaction. They are placed on elements, but DOM structure can change and IDs can be put on other elements without affecting the tests until user interaction is the same. This method is already quite robust and does not break easily when the implementation changes.

It would make sense to abstract some of it if there was a significantly long chain of steps to complete an action, like selecting a value a multi-level dropdown or to adding values to tag input. The code you quoted here are one-liners which simply query a component by `data-test-id`. I see an attempt to abstract it furher similar to trying to abstract an "if" statement because it's repeated it many places.

> You cannot assume the user will know the data-test-id of a specific UI element (event the developer who added this id will forget it 6 months down the track)

Developers don't need to remember IDs. It's [easy to query them in the DOM](https://user-images.githubusercontent.com/2076953/65759886-408fa380-e11c-11e9-9f8c-4c40b353ce22.png) before writing tests.

> It also aids in reading the specs, i.e. workspace.sideNav('Monitoring').click()

Clear naming of IDs accomplishes the same:
```el(['workspace', 'sideNav', 'Monitoring']).click()```

petrjasek: I think this it's about the level of abstraction and it's kind of subjective, like what would be this?
>  significantly long chain of steps to complete an action

imo it's very close to 2 :wink:

tomaskikutis: If it's specific to a feature like "mark for user" and is not being reused, there's no reason to abstract anything even if it's 10 steps. I was referring to reusable things like "selecting a value from a multi-level dropdown or adding values to tag input".

I've moved a few reusable steps to e2e helpers. Everything else is specific to the feature.

PR #3096, opened 2019-09-06
Fix duplicate actions in item menu

pablopunk:
SDESK-4677

Use actions from the proper group instead of the stack group it belongs


petrjasek:
any idea if this is only broken on develop branch or also on master?

pablopunk:
@petrjasek I don't think so, the bug was introduced on this PR https://github.com/superdesk/superdesk-client-core/pull/3074

PR #3098, opened 2019-09-06
add embedded images without changes to crop

petrjasek:
SDESK-4679

### DIFF
--- a/scripts/core/editor3/actions/toolbar.tsx
+++ b/scripts/core/editor3/actions/toolbar.tsx
@@ -91,7 +91,7 @@ export function insertMedia(files?, targetBlockKey = null) {
                             editedMedia.push(cropped);
                             editNext();
                         }, (reason) => {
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/actions/toolbar.tsx
```diff
@@ -90,6 +90,12 @@ export function insertMedia(files?, targetBlockKey = null) {
                         renditions.crop(_media, options).then((cropped) => {
                             editedMedia.push(cropped);
                             editNext();
+                        }, (reason) => {
+                            if (reason != null && reason.done != null && reason.done) {
```
tomaskikutis: Use explicit check `if (reason != null && reason.done != null && reason.done === true)`. I can see it's expected to be a boolean now since it's in the same diff, but for future readers, it won't be obvious if `reason.done` is supposed to be a string or a boolean or something else.

PR #3099, opened 2019-09-06
Ensure that menu items are displayed as highlighted for child paths.

tomaskikutis:
SDESK-4674

### DIFF
--- a/scripts/core/menu/menu.ts
+++ b/scripts/core/menu/menu.ts
@@ -1,6 +1,5 @@
 import {reactToAngular1} from 'superdesk-ui-framework';
 import {GlobalMenuHorizontal} from './GlobalMenuHorizontal';
-import {trimStartExact} from 'core/helpers/utils';
 
 SuperdeskFlagsService.$inject = ['config'];
 function SuperdeskFlagsService(config) {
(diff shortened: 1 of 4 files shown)


scripts/core/menu/menu.ts
```diff
@@ -91,6 +104,71 @@ angular.module('superdesk.core.menu', [
...
+                            scope.items.push(item);
+                        });
+                    // menu items and groups - end
+
+                    /*
+                        Marking item as active even if current path doesn't match exactly to item href path
+                        For example, if current path is /a/b/c and there is only menu item for /a/b,
+                        /a/b should be set active.
+                    */
+                    function getActiveMenuItemPath(currentPath: string) {
```
petrjasek: imo you're looking for longest matching prefix in strings, so this could be something like:
```
scope.items
    .filter((item) => typeof item.href === 'string' && item.href.length > 0 && currentPath.startsWith(item.href))
    .map((item) => item.href)
    .reduce((longest, current) => longest == null || longest.length < current.length ? current : longest);
```
which imo makes it more obvious and wound't need all those comments.

tomaskikutis: :+1: Great suggestion. I only changed it to compare the number of occurrences of `/` instead of character count, since `a/b/c/d/e/f` is deeper even if `a/b/c/ddddddddddd` is longer by the number of characters.

petrjasek: but they wouldn't match both as a prefix if they are different, would they?

petrjasek: so at the end one must be prefix of another if they are both a prefix of current url, and only thing that matters is which one is longer imo

tomaskikutis: You're right. I've removed the splitting part.

PR #3102, opened 2019-09-10
save new annotations to annotations library

petrjasek:
SDFID-560


scripts/core/superdesk-api.d.ts
```diff
@@ -642,6 +642,7 @@ declare module 'superdesk-api' {
         label: string;
         selectedByDefault(annotationText: string, mode: 'create' | 'edit'): Promise<boolean>;
         component: React.ComponentType<IPropsAnnotationInputComponent>;
+        onCreate(language: string, annotationText: string, definitionHtml: string): void;
```
tomaskikutis: Can you rename it to `onAnnotationCreate`? By only looking to the interface one might as well think `onCreate` refers to tab creation instead of annotation creation.

PR #3110, opened 2019-09-13
add translation info to content lists

petrjasek:
SDBELGA-138

### DIFF
--- a/scripts/apps/search/components/fields/translations-list.tsx
+++ b/scripts/apps/search/components/fields/translations-list.tsx
@@ -4,8 +4,6 @@ import {dataApi} from 'core/helpers/CrudManager';
 
 interface IProps {
     ids: Array<IArticle['_id']>;
-    label: string;
-    onClose: () => void;
     onClick: (item: IArticle) => void;
 }
 
@@ -29,20 +27,14 @@ export class TranslationsList extends React.PureComponent<IProps, IState> {
 
     render() {
         return (
-            <ul className="highlights-list-menu open">
-                <li>
-                    <div className="dropdown__menu-label">{this.props.label}</div>
-                    <button className="dropdown__menu-close" onClick={() => this.props.onClose()}>
-                        <i className="icon-close-small icon--white" />
-                    </button>
-                </li>
+            <ul className="simple-list" style={{padding: 0, flexGrow: 1}}>
                 {this.state.items.length === 0 && (
                     <li style={{minHeight: this.props.ids.length * 1.5 + 'em'}}>
                         <div className="sd-loader sd-loader--dark-ui" />
                     </li>
                 )}
                 {this.state.items.map((item) => (
-                    <li key={item._id}>
+                    <li key={item._id} className="simple-list__item">
                         <b className="label label--hollow">{item.language}</b>
                         &nbsp;
                         <a className="sd-overflow-ellipsis"
--- a/scripts/apps/search/components/fields/translations-list.tsx
+++ b/scripts/apps/search/components/fields/translations-list.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import {IArticle} from 'superdesk-api';
+import {dataApi} from 'core/helpers/CrudManager';
 
 interface IProps {
-    svc: any;
     ids: Array<IArticle['_id']>;
     label: string;
     onClose: () => void;
@@ -20,12 +20,11 @@ export class TranslationsList extends React.PureComponent<IProps, IState> {
     }
 
     componentDidMount() {
-        const {api} = this.props.svc;
-
-        Promise.all(this.props.ids.map((id) => api.find('archive', id)))
-            .then((items) => {
-                this.setState({items});
-            });
+        Promise.all(this.props.ids.map((id) =>
+            dataApi.findOne<IArticle>('archive', id),
+        )).then((items) => {
+            this.setState({items});
+        });
     }
 
     render() {
--- a/scripts/apps/search/components/fields/translations-list.tsx
+++ b/scripts/apps/search/components/fields/translations-list.tsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import {IArticle} from 'superdesk-api';
+
+interface IProps {
+    svc: any;
+    ids: Array<IArticle['_id']>;
+    label: string;
+    onClose: () => void;
+    onClick: (item: IArticle) => void;
+}
+
+interface IState {
+    items: Array<IArticle>;
+}
+
+export class TranslationsList extends React.PureComponent<IProps, IState> {
+    constructor(props) {
+        super(props);
+
+        this.state = {items: []};
+    }
+
+    componentDidMount() {
+        const {content} = this.props.svc;
+
+        content.fetchIds(this.props.ids).then((items) => {
+            this.setState({items});
+        });
+    }
+
+    render() {
+        return (
+            <ul className="highlights-list-menu open">
+                <li>
+                    <div className="dropdown__menu-label">{this.props.label}</div>
+                    <button className="dropdown__menu-close" onClick={() => this.props.onClose()}>
+                        <i className="icon-close-small icon--white" />
+                    </button>
+                </li>
+                {this.state.items.map((item) => (
+                    <li key={item._id}>
+                        <b className="label label--hollow">{item.language}</b>
+                        &nbsp;
+                        <a className="sd-overflow-ellipsis"
+                            onClick={() => this.props.onClick(item)}>
+                            {item.headline || item.slugline}
+                        </a>
+                    </li>
+                ))}
+            </ul>
+        );
+    }
+}
--- a/scripts/apps/search/components/fields/translations-list.tsx
+++ b/scripts/apps/search/components/fields/translations-list.tsx
@@ -16,16 +16,16 @@ interface IState {
 export class TranslationsList extends React.PureComponent<IProps, IState> {
     constructor(props) {
         super(props);
-
         this.state = {items: []};
     }
 
     componentDidMount() {
-        const {content} = this.props.svc;
+        const {api} = this.props.svc;
 
-        content.fetchIds(this.props.ids).then((items) => {
-            this.setState({items});
-        });
+        Promise.all(this.props.ids.map((id) => api.find('archive', id)))
+            .then((items) => {
+                this.setState({items});
+            });
     }
 
     render() {
@@ -37,6 +37,11 @@ export class TranslationsList extends React.PureComponent<IProps, IState> {
                         <i className="icon-close-small icon--white" />
                     </button>
                 </li>
+                {this.state.items.length === 0 && (
+                    <li style={{minHeight: this.props.ids.length * 1.5 + 'em'}}>
+                        <div className="sd-loader sd-loader--dark-ui" />
+                    </li>
+                )}
                 {this.state.items.map((item) => (
                     <li key={item._id}>
                         <b className="label label--hollow">{item.language}</b>
--- a/scripts/apps/search/components/fields/translations.tsx
+++ b/scripts/apps/search/components/fields/translations.tsx
@@ -16,8 +16,8 @@ export class Translations extends React.PureComponent<IProps> {
         if (this.props.item.translated_from != null) {
             output.push(
                 <button key="translated" className="label label--hollow"
-                    onClick={($event) => {
-                        this.renderOriginalArticle($event.target);
+                    onClick={(event) => {
+                        this.renderOriginalArticle(event.target);
                     }}>
                     {gettext('translation')}
                 </button>,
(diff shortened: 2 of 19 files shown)


scripts/apps/search/components/fields/translations.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    renderTranslations() {
+        this.renderPopup(
+            gettext('Translations'),
+            this.props.item.translations,
+            this.translationsRef,
+        );
+    }
+
+    renderPopup(label: string, ids: Array<IArticle['_id']>, ref: React.ReactNode) {
```
tomaskikutis: [popup](https://user-images.githubusercontent.com/2076953/65036222-3b2b9f80-d94b-11e9-9609-8c14d729bcbf.gif) should disappear when scrolling.

scripts/apps/search/components/fields/translations.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    render() {
+        const output = [];
+
+        if (this.props.item.translated_from != null) {
+            output.push(
+                <span key="translated" className="label label--hollow"
+                    onClick={($event) => {
+                        $event.stopPropagation();
+                        this.renderOriginalArticle();
```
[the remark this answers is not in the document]
petrjasek: I'm not sure we can avoid that state in components using the popup, otherwise based on what would you actually render that popup?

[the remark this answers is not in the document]
tomaskikutis: Yeah, I forgot about that. We can do the same as with closing from within the inside of the popup:
```
reference={(open, close, toggle) =>
    (
        <button className="label label--hollow" onClick={toggle}>
            {gettext('translation')}
        </button>
    )
}
```

scripts/apps/search/components/fields/translations-list.tsx
```diff
@@ -0,0 +1,53 @@
...
+        const {content} = this.props.svc;
+
+        content.fetchIds(this.props.ids).then((items) => {
+            this.setState({items});
+        });
+    }
+
+    render() {
+        return (
+            <ul className="highlights-list-menu open">
```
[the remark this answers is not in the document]
petrjasek: that doesn't fit the design really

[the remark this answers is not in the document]
tomaskikutis: If the height were different and transparent background was supported it should probably fit. But yeah, you can leave it as-is for now and we should be able to improve it as more components land in the UI framework.

scripts/apps/workspace/content/services/ContentService.ts
```diff
@@ -356,6 +356,10 @@ export function ContentService(api, superdesk, templates, desks, packages, archi
         })).then((values) => zipObject(keys, values));
     };
 
+    this.fetchIds = (ids: Array<IArticle['_id']>) => {
```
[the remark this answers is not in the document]
petrjasek: you can get published from archive by id, they just won't be listed when you search archive

scripts/apps/search/components/fields/translations-list.tsx
```diff
@@ -16,16 +16,16 @@ interface IState {
 export class TranslationsList extends React.PureComponent<IProps, IState> {
     constructor(props) {
         super(props);
-
         this.state = {items: []};
     }
 
     componentDidMount() {
-        const {content} = this.props.svc;
+        const {api} = this.props.svc;
```
tomaskikutis: use `dataApi` instead. It supports types and can be imported as es6 module

scripts/apps/search/components/fields/translations.tsx
```diff
@@ -10,53 +10,49 @@ interface IProps {
...
-    translationsRef: React.ReactNode;
-
     render() {
         const output = [];
 
         if (this.props.item.translated_from != null) {
             output.push(
-                <span key="translated" className="label label--hollow"
+                <button key="translated" className="label label--hollow"
                     onClick={($event) => {
```
tomaskikutis: Why are you using `$` sign in the name? That implies that it'd be a jQuery even, but it isn't.

scripts/apps/search/components/fields/translations-list.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import {IArticle} from 'superdesk-api';
+
+interface IProps {
+    svc: any;
```
tomaskikutis: Can you remove `svc` from props? It's never clear what it contains and it's a hassle to drag it around the component tree. Use `getService` instead to get the services you need at any level of the component tree.

scripts/apps/search/components/item-list-popup.tsx
```diff
@@ -0,0 +1,60 @@
...
+        super(props);
+        this.close = this.close.bind(this);
+    }
+
+    componentDidMount() {
+        positionPopup(this.props.target);
+    }
+
+    close(event: SyntheticEvent) {
+        event.stopPropagation();
```
tomaskikutis: Why do you need this?

scripts/apps/search/components/item-list-popup.tsx
```diff
@@ -0,0 +1,60 @@
...
+                <li>
+                    <div className="dropdown__menu-label">{this.props.label}</div>
+                    <button className="dropdown__menu-close" onClick={this.close}>
+                        <i className="icon-close-small icon--white" />
+                    </button>
+                </li>
+                <li>
+                    {this.props.children}
+                </li>
+                {createPortal(
```
tomaskikutis: Why do you need nested portals?

scripts/apps/search/components/fields/translations.tsx
```diff
@@ -38,41 +52,47 @@ export class Translations extends React.PureComponent<IProps> {
...
         return output.length > 0 ? output : null;
     }
 
     renderOriginalArticle(elem: EventTarget) {
-        this.renderPopup(
-            gettext('Original Article'),
-            [this.props.item.translated_from],
-            elem,
-        );
+        this.setState({
```
tomaskikutis: Looks like you are going to end up state being out-of-sync from UI. When you click the close button, the DOM will be imperatively modified while this state will still contain the data of now closed popup.

If you were to go with the popup interface I suggested, that wouldn't happen. Why did you decide against it? Do you think it was a bad approach or are you trying to save time?

petrjasek: the dom will be only modified after the state changes

scripts/apps/search/components/item-list-popup.tsx
petrjasek: that backdrop must be rendered outside react placeholder, it's positioned absolutely so it wouldn't be possible to cover the whole screen if it's inside it

scripts/apps/search/components/item-list-popup.tsx
petrjasek: to avoid the click propagate to list item and open preview

tomaskikutis: The popup is outside of the list component it should not send it any events anyhow. It's probably because you are using a portal. Can you comment that so it's clear for future readers?

scripts/apps/search/components/item-list-popup.tsx
tomaskikutis: What are you using a backdrop for here? It's a popup, not a modal, so it shouldn't make the entire screen darker when it opens. Or is it for something different?

scripts/apps/search/components/item-list-popup.tsx
```diff
@@ -0,0 +1,60 @@
...
+                    <div style={BACKDROP_STYLE} onClick={this.close} onScroll={this.close}>
```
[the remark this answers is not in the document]
petrjasek: it shouldn't do much, it scrolls that invisible div which shouldn't be visible

scripts/apps/search/components/item-list-popup.tsx
```diff
@@ -0,0 +1,60 @@
+import React, {ElementType, CSSProperties, SyntheticEvent} from 'react';
+import {createPortal} from 'react-dom';
+import {positionPopup} from '../helpers';
+
+interface IProps {
+    label: string;
+    close: () => void;
+    target: ElementType | EventTarget;
```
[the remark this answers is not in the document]
petrjasek: I use to avoid storing refs in translations component

scripts/apps/search/components/item-list-popup.tsx
petrjasek: to be able to catch scroll or click outside of it

scripts/apps/search/components/item-list-popup.tsx
petrjasek: yep that's how portals work

scripts/apps/search/components/item-list-popup.tsx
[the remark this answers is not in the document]
tomaskikutis: I'm suggesting to change the type from `target: ElementType | EventTarget;` to `target: HTMLElement;`. It shouldn't have to do anything with refs, you would still store it in the state as you do now.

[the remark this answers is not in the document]
petrjasek: wouldn't that re-render just to store those refs?

scripts/apps/search/components/item-list-popup.tsx
tomaskikutis: That's why we shouldn't be using a portal here :) Components like modals/dropdowns/popups should be isolated and not forward any events to where they have been initiated from. We can sprinkle `preventDefault`s and `stopPropagation` calls, but that's a sub-optimal solution.

scripts/apps/search/components/fields/translations.tsx
tomaskikutis: Right, I've mistaken how the data flows. Can you address the second question? :slightly_smiling_face:

petrjasek: I didn't see the need to complicate it, so instead of using a function it just renders the children like any other component.

scripts/apps/search/components/item-list-popup.tsx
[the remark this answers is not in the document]
tomaskikutis: If you used refs, it would re-render. It doesn't matter if it re-renders since it's a fraction of a second anyway. That said, I don't suggest to use refs.

```
popup: {
    label: gettext('Translations'),
    ids: this.props.item.translations,
    target: elem, // <-- `HTMLElement`
},
```

I've tracked the usages and the `elem` you use in the current `setState` call is already `HTMLElement` it's just that TypeScript can't guarantee it, in which case it's fine to use type casting and do ` target: elem as HTMLElement`.

scripts/apps/search/components/item-list-popup.tsx
petrjasek: I'm not sure what's better actually, there must be some kind of connection because if that parent component (item) is not rendered it shouldn't be visible anymore.

tomaskikutis: I agree, we need a connection for removing the popup when the parent is umounted, but we can do it through a wrapper component :
```
class Wrapper extends React.PureComponent {
    componentDidMount() {
        // render to DOM
    }
    componentDidUnmount() {
        // clean up DOM
    }
    render() {
        return null;
    }
}
```

scripts/apps/search/components/fields/translations.tsx
tomaskikutis: I hear you, but your version isn't without "complications" either. You now have to store ids,label and dom node in the state as well as have `renderOriginalArticle`, `renderTranslations` and `closePopup` methods in the component, which you wouldn't have needed otherwise.

scripts/apps/search/components/item-list-popup.tsx
petrjasek: yep, or we use portal

tomaskikutis: What I like about the wrapper approach is that it gives confidence that bugs similar to this one are less likely.

Here's one present right now: Let's say I have a popup open and want to highlight a word to be copied. I use the browser feature to double click the word in order to highlight it and article edit view opens :beetle:

petrjasek: I'm not sure it's a bug or a feature. This wrapper is a workaround to avoid this click propagate and imo that's less explicit than adding click handler inside the component which would stop the propagation.

tomaskikutis: It sounds like you want it to be a feature so you don't have to change the code :smile:

The current code is fine for this use case. There is a small bug with the double click, but it's fine. I expect to replace this with a generic popup component as soon as we get it in the UI framework.

Being explicit is fine, but I don't see this being about explicit/implicit code. It's more about building an abstraction and isolating it. When React calls `document.createElement` you don't see it being too implicit and that you'd rather call it yourself. You can also look at a native select dropdown. If you double click an option it won't send the event to the parent where that select is rendered in. I see no uses cases where it would make sense.

tomaskikutis:
> use fragment instead of an array

Did you bump into a bug with this? Or did you change it so it's easier to read?

PR #3111, opened 2019-09-16
fix(menus) Wait for privileges to load before populating list

MarkLark86:


MarkLark86:
@petrjasek @tomaskikutis Hopefully this is the last change to fix the menu list populating. This bug was not easily reproducible in local dev builds, only in live builds like sdp-master.

PR #3112, opened 2019-09-16
fix: Prevent drag and drop of media items to the read-only fields. (SDESK-4632-1)

GyanP:


scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
```diff
@@ -99,6 +99,12 @@ export function ArticleEditDirective(
                     superdesk.intent('preview', 'item', item);
                 };
 
+                elem.on('drop dragdrop', (event) => {
```
petrjasek: you should also de-register this listener on scope `$destroy` event

tomaskikutis: Which element does `elem` refer to?

GyanP: @tomaskikutis Sorry, I totally missed this. It's there to reject all the items dropped when the `editable` is set to `false`. What do you mean by which element?

tomaskikutis: You are adding an event listener on an element/DOM node. I was asking which element is that. Article edit view is made up of multiple fields - headline, body, media gallery etc. I see you are most likely adding the event listener on a root element wrapping the entire article edit view, which is fine. Though you should probably add the code removing the event listener when the view is unmounted.

PR #3117, opened 2019-09-18
fix: drag and drop of media from external folder to feature-media and custom-media fields. (SDESK-4655)

GyanP:


scripts/apps/authoring/authoring/controllers/AssociationController.ts
```diff
@@ -149,8 +149,12 @@ export function AssociationController(config, content, superdesk,
             return renditions.crop(item, cropOptions)
                 .then((rendition) => {
                     self.updateItemAssociation(scope, rendition, options.customRel, callback);
-                })
-                .finally(() => {
+                }, (response) => {
```
pablopunk: maybe this could be a `catch` instead? It looks more readable

scripts/apps/authoring/authoring/directives/ItemAssociationDirective.ts
```diff
@@ -76,12 +76,14 @@ export function ItemAssociationDirective(renditions) {
...
                 elem.on('drop dragdrop', (event) => {
+                    const type = getSuperdeskType(event);
+
                     // drop event should only fire if dragover event is prevented
                     // however, `ng-file-upload` library calls preventDefault on this event
                     // which it shouldn't do, since the element is not controlled by the library.
                     // Because of this, drop is triggered when it shouldn't have been on firefox, but not on chrome.
                     // The same media types check is added from `dragover` method to work around the situation.
-                    if (MEDIA_TYPES.includes(getSuperdeskType(event))) {
+                    if (MEDIA_TYPES.includes(type) || type === 'Files') {
```
pablopunk: what about adding `'Files'` to `MEDIA_TYPES`?

GyanP: yep, that makes sense.

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -79,6 +79,8 @@ export function ItemCarouselDirective(notify) {
                 .filter((key) => allowed[key] === true)
                 .map((key) => 'application/superdesk.item.' + key);
 
+            ALLOWED_TYPES.push('Files');
```
tomaskikutis: A file is not a media type.

```
allowed: {picture: boolean, video: boolean, audio: boolean, file: boolean}
```
Having `file` inside allowed doesn't make sense. Are images, videos or audio not files? They are. Why then `files` is there? Does it support PDFs or excel files? No, it doesn't.

It's confusing for other developers and will even be confusing for you after 6 months when you forget why was it done this way.

Separate the concerns in the code. Make sure that uploading files from the file system is done in one place and when you need to add an image to the gallery, you don't need to care about where it's coming from - a file system, superdesk images. See [this diagram](https://drive.google.com/file/d/1KDzE171U79Yo7sJV75YApey-qrYcX3Ly/view?usp=sharing) I made.

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -152,7 +154,7 @@ export function ItemCarouselDirective(notify) {
                 carousel.trigger('to.owl.carousel', [index]);
             };
 
-            function canAddImage(image: IArticle): boolean {
+            function canAddImage(image: IArticle, files?: FileList): boolean {
```
tomaskikutis: Try refactoring this so it's more similar to the diagram I referenced in another comment. Make this function `canAddImages(images: Array<IArticle>): boolean`. Make it possible to accept multiple images and remove reference to files.

scripts/apps/authoring/authoring/directives/ItemAssociationDirective.ts
```diff
@@ -149,3 +141,44 @@ export function ItemAssociationDirective(renditions) {
         },
     };
 }
+
+export function isAllowedMediaType(scope, event) {
+    const {allowAudio, allowPicture, allowVideo} = scope;
+    const VALID_MEDIA_TYPES = ['Files'];
```
tomaskikutis: What is the significance of a string `'Files'`? Is it an arbitrary string or something the browser adds into the data transfer object? If it's not a browser thing, extract it to constants across the codebase.

GyanP: It's being added by the browser itself to differentiate between the type of data dropped eg: text/plain, text/uri-list, Files, etc.

scripts/apps/authoring/authoring/services/RenditionsService.ts
```diff
@@ -156,6 +156,11 @@ export function RenditionsService(metadata, $q, api, superdesk, _, notify) {
                         .catch(() => {
                             notify.error(gettext('Failed to generate picture crops.'));
                         });
+                }).catch((response) => {
+                    // if new crops not generated continue with default one
```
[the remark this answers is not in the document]
GyanP: Well. the situation is that whenever a user uploads a new media from an external folder, we ask the user to play with the crops after image upload by opening edit crops modal. Earlier we used to prompt the user to click on `save` or `cancel` and then on `Done` button to close the modal. But later it was decided not to prompt the user to click `save` or `cancel`, unless the user changes the crops, and the modal can be closed by clicking `Done` button directly.
In that case, we are sending [this](https://github.com/petrjasek/superdesk-client-core/blob/58d5d41d83a4e1e345d6eefe2182064e1172daa7/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts#L250) response. We have to proceed with default behavior whether the crops changed or not. That's why I decided to handle it inside `RenditionService.ts`, as it takes care of the crops.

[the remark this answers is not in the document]
tomaskikutis: Thanks for explaining. Can you add that link from your GitHub comment to the code comment as well?

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -148,15 +149,15 @@ export function ItemCarouselDirective(notify) {
...
+            function canAddMediaItems(internalIds: Array<IArticle['_id']>, externalItemsCount: number = 0): boolean {
                 const mediaItemsForCurrentField = Object.keys(scope.item.associations || {})
                     .filter((key) => key.startsWith(scope.field._id) && scope.item.associations[key] != null)
                     .map((key) => scope.item.associations[key]);
 
                 const currentUploads = mediaItemsForCurrentField.length;
 
                 const itemAlreadyAddedAsMediaGallery = mediaItemsForCurrentField.some(
-                    (mediaItem) => mediaItems.find((_item) => _item != null && _item._id === mediaItem._id),
+                    (mediaItem) => internalIds.find((_id) => _id === mediaItem._id),
```
tomaskikutis: Use `internalIds.includes(mediaItem._id)`

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -204,11 +205,12 @@ export function ItemCarouselDirective(notify) {
 
                     if (isAllowedMediaType(scope, event)) {
                         const itemStr = event.originalEvent.dataTransfer.getData(type);
-                        const mediaItems = typeof itemStr === 'string' && itemStr.length > 0
-                            ? [angular.fromJson(itemStr)]
-                            : Object.values(event.originalEvent.dataTransfer.files);
+                        const internalIds = typeof itemStr === 'string' && itemStr.length > 0
+                            ? [angular.fromJson(itemStr)._id]
```
tomaskikutis: Why not use `JSON.parse` instead of `angular.fromJson`?

scripts/apps/authoring/authoring/directives/ItemAssociationDirective.ts
```diff
@@ -66,6 +66,15 @@ export function ItemAssociationDirective(renditions, notify) {
                     // which it shouldn't do, since the element is not controlled by the library.
                     // Because of this, drop is triggered when it shouldn't have been on firefox, but not on chrome.
                     // The same media types check is added from `dragover` method to work around the situation.
+                    const externalItemsCount = Object.values(event.originalEvent.dataTransfer.files || []).length;
+
+                    if (externalItemsCount > 1) {
```
tomaskikutis: The comment block above refers to the `if` statement that was directly below it. You have now added code between the code and its comment which makes the comment look like it's about your newly added code, which is not the case. Move the comment block down so its directly above this line `if (isAllowedMediaType(scope, event)) {`.

Also, add a comment on what your code is doing and what use case does it handle. Why wouldn't you want to allow adding multiple files?

GyanP: I feature-media field we support only one media, for more than one media fields we have custom-media fields.

PR #3118, opened 2019-09-19
Introduce configuration for the main left hand side toolbar

tomaskikutis:
SDESK-4668

### DIFF
--- a/scripts/core/index.ts
+++ b/scripts/core/index.ts
@@ -38,6 +38,7 @@ import {reactToAngular1} from 'superdesk-ui-framework';
 import {ExtensionPage} from './extension-page';
 import {registerExtensions} from './register-extensions';
 import {AuthoringWorkspaceService} from 'apps/authoring/authoring/services/AuthoringWorkspaceService';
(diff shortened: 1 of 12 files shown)


scripts/core/index.ts
```diff
@@ -86,10 +87,29 @@ let _superdesk;
 
 core.component('sdExtensionPage', reactToAngular1(ExtensionPage, [], [], styles));
 core.config(['$routeProvider', 'superdeskProvider', ($routeProvider, superdesk) => {
+    // set initial default route to personal
+    // when user is logged in, it will be overwritten by a default route
+    // from configs if user has permissions to that route
     $routeProvider.when('/', {
-        redirectTo: appConfig.defaultRoute,
+        redirectTo: '/workspace/personal',
```
pablopunk: shouldn't this be appConfig.defaultRoute so you don't have to redirect again below?

tomaskikutis: That's what I tried first. I can't do this synchronously since I first need to check permissions for `defaultRoute` which is async. When it's not registered synchronously the login page breaks.

PR #3120, opened 2019-09-24
Internal attachments

pablopunk:
SDFID-574

Server part: https://github.com/superdesk/superdesk-core/pull/1676

### DIFF
--- a/scripts/core/editor3/components/links/AttachmentList.tsx
+++ b/scripts/core/editor3/components/links/AttachmentList.tsx
@@ -71,7 +71,7 @@ export class AttachmentList extends React.Component<IProps, IState> {
 
     render() {
         if (!this.state.attachments) {
-            return <div />; // wait for attachments
+            return null; // wait for attachments
         }
 
         const publicAttachments = this.state.attachments.filter(({internal}) => !internal);
--- a/scripts/core/editor3/components/links/AttachmentList.tsx
+++ b/scripts/core/editor3/components/links/AttachmentList.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-multi-comp */
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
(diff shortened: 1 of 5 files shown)


scripts/core/editor3/components/links/AttachmentList.tsx
```diff
@@ -22,38 +24,47 @@ export class AttachmentList extends React.Component<any, any> {
             });
     }
 
-    render() {
-        if (!this.state.attachments) {
-            return <div />; // wait for attachments
+    renderAttachment(attachment: IAttachment) {
```
tomaskikutis: Extract this into an `AttachmentItem` component. I've noticed that when people put JSX in a method other than `render`, most of the time the code better of being a separate component. Ignore that no components in a single file rule. That only matters if there's a significant number of state in the file scope.

scripts/core/editor3/components/links/AttachmentList.tsx
```diff
@@ -22,38 +24,47 @@ export class AttachmentList extends React.Component<any, any> {
...
                 </div>
-            );
-        });
+            </div>
+        );
+    }
+
+    render() {
+        if (!this.state.attachments) {
+            return <div />; // wait for attachments
```
tomaskikutis: Why return `<div />` instead of `null`?

scripts/core/editor3/components/links/AttachmentList.tsx
```diff
@@ -22,38 +24,47 @@ export class AttachmentList extends React.Component<any, any> {
...
+        }
+
+        const attachments = this.state.attachments
+            .map(this.renderAttachment)
+            .filter(Boolean);
```
tomaskikutis: Don't use implicit casting. When you have `if (x)` it's clearer when it is written as `if (x != null)` if that's what you mean.

I would avoid filtering on data rather than UI components here:

```
render() {
    if (this.state.attachments != null) {
        return null; // wait until loaded
    }

    const externalAttachements = this.state.attachments.filter((attachment) => attachment.internal !== true);

    if (externalAttachements.length > 0) {
        return (
            <div>
                {externalAttachements.map((attachment, i) => {
                    <AttachmentItem attachment={attachment} />
                })}
            </div>
        );

    } else {
        return (
            <p style={{padding: 20, margin: 0}}>
                {gettext('There are no attachments yet. Upload some first using Attachments widget.')}
            </p>
        );
    }
}
```

scripts/core/editor3/components/links/AttachmentList.tsx
```diff
@@ -5,6 +5,7 @@ import classNames from 'classnames';
 import ng from 'core/services/ng';
 import {FileiconFilter, FilesizeFilter} from 'core/ui/ui';
 import {gettext} from 'core/utils';
+import {IAttachment} from 'apps/authoring/attachments';
 
 export class AttachmentList extends React.Component<any, any> {
```
tomaskikutis: Add types for state and props :)

scripts/core/editor3/components/links/AttachmentList.tsx
pablopunk: That was like that already, I'll change it tho

scripts/core/editor3/components/links/AttachmentList.tsx
pablopunk: ha, I didn't read this comment but I've already done something similar on my latest commit

PR #3122, opened 2019-09-25
fix multi image edit for custom fields

petrjasek:
SDANSA-318


scripts/apps/search/MultiImageEdit.ts
```diff
@@ -150,10 +154,45 @@ export function MultiImageEditController(
...
             copyrightnotice: compare('copyrightnotice'),
-            extra: compare('extra'),
+            extra: compareExtra(),
             language: compare('language'),
             creditline: compare('creditline'),
         };
+
+    }
+
+    function compareExtra() {
```
tomaskikutis: Could you add a comment why this one needs special handling and what are you doing in every loop?

scripts/apps/search/MultiImageEdit.ts
```diff
@@ -55,19 +55,23 @@ export function MultiImageEditController(
...
+    $scope.selectImage = (image, update: boolean = true) => {
         if ($scope.images.length === 1) {
             $scope.images[0].selected = true;
         } else {
             image.selected = !image.selected;
         }
-        updateMetadata();
+
+        if (update) {
+            updateMetadata();
```
tomaskikutis: Can you add a comment on why would metadata be updated upon selecting an image?

petrjasek:
@tomaskikutis yep I agree that this whole image editing should be re-factored (or rather rewritten), I wasn't even sure it's possible to fix that issue, but somehow I did so I've made that PR..

scripts/apps/search/MultiImageEdit.ts
```diff
@@ -161,23 +162,34 @@ export function MultiImageEditController(
...
+            .filter((item) => item.extra != null && item.extra[field] != null && item.extra[field] !== '')
+            .map((item) => item.extra[field])
+            .forEach((value) => uniqueValues[value] = 1);
+        return Object.keys(uniqueValues);
+    }
+
+    /**
+     * Populate .extra metadata for editing.
+     *
+     * Works like compare() but for custom fields stored in .extra.
```
tomaskikutis: Could the same code be reused between `compareExtra` and `compare` with a different argument specifying which part of the object to check?

PR #3127, opened 2019-09-30
fix: annotations library breaks when creating new item with some active filters. (SDFID-575)

GyanP:


scripts/core/ui/components/ListPage/generic-list-page.tsx
```diff
@@ -548,11 +548,24 @@ export class GenericListPageComponent<T extends IBaseRestApiResponse>
...
+                                            this.props.notify.success(gettext('The item has been created.'));
+                                            this.closeNewItemForm();
+                                            // if there are any active filters and new item get filtered out
+                                            // the annotations library breaks as we don't have item to show preview, so
+                                            // remove all the active filters and apply language filter as item language
+                                            // so that new item does not get filtered out as we show preview of item
+                                            this.setState((prevState) => ({
+                                                ...prevState,
+                                                filterValues: {
+                                                    language: res.language,
```
tomaskikutis: You can't set language here. As the file name implies, this view is generic and mustn't know anything about specific field names. For example, the same view is used in "internal destinations" settings page which doesn't have a language field and applying such filter would be a bug.

There are two ways you can solve this properly:
* Only display a preview if the item is in the list(it can also end up in another page even if nothing is filtered).
* Refactor how the preview works so it can work well even if the item is not visible in the list.

PR #3128, opened 2019-09-30
Display marked for user notifications

tomaskikutis:
SDBELGA-123

This is branched from https://github.com/superdesk/superdesk-client-core/pull/3095. Only the last commit is relevant to the feature.


GyanP:
@tomaskikutis as per Karel this should be the format of notifications `Item <headline> has been marked for <user> by <user>`

tomaskikutis:
See my comment in `SDBELGA-124`. I want you to send the full message from the server. The client will display any message that is sent given it's in plain text.

PR #3131, opened 2019-10-02
Allow filtering users by username

tomaskikutis:
SDBELGA-176


scripts/core/ui/components/SelectUser.tsx
```diff
@@ -41,14 +41,21 @@ export class SelectUser extends React.Component<IPropsSelectUser, IState> {
...
                     this.props.onSelect(keyedUsers[value]);
                 }}
                 renderItem={(user) => (
                     <div style={{display: 'flex', alignItems: 'center'}}>
                         <UserAvatar displayName={user.display_name} pictureUrl={user.picture_url} />
-                        <div style={{paddingLeft: '0.5em'}}>{user.display_name}</div>
+                        <div style={{paddingLeft: '0.5em', textAlign: 'left'}}>
+                            <div><strong>{user.display_name}</strong></div>
+                            <div
+                                style={{fontSize: '11px', lineHeight: '11px', opacity: 0.5, marginTop: '-2px'}}
```
MarkLark86: Inline styles should be avoided at all cost. The preference should be:
- UI-Framework
- Stylesheet/SASS (only if the UI-Framework doesn't have what you require).

This is so @fritzSF and his team are able to better support and maintain the styles used in Superdesk. @fritzSF Are you able to provide some assistance here to make this section adhere to UI-Framework standards?

tomaskikutis: What's the benefit to copy the same content over to another file? Markup and styles are solving the same problem, I don't see a reason to separate them if nothing is gained from it. Inline styles don't work for pseudo-selectors for example, but I'm not using it :)

MarkLark86: Using custom styles makes it hard to support and maintain a unified look and feel across the entire product. This is a list of users, so I would use the UI-Framework's list item:

```
<div class="sd-list-item">
    <div class="sd-list-item__column sd-list-item__column--no-border">
        <figure class="avatar avatar--small">BM</figure>
    </div>
    <div class="sd-list-item__column">
        <div class="sd-list-item__row">
            <span>Pittaway, Mark</span>
        </div>
        <div class="sd-list-item__row">
            <span class="sd-list-item__text-label">@mpittaway</span>
        </div>
    </div>
</div>
```

There are modifiers available for the column/row/text in this format.

tomaskikutis: I agree with the unified look argument. The reason I didn't use it initially is that I didn't think it would look good enough, but it came quite alright.

petrjasek:
this doesn't do the filtering right? just wondering given the PR description

tomaskikutis:
It used to specify which fields are to be used for filtering by select2 component here https://github.com/superdesk/superdesk-client-core/pull/3131/files#diff-826f0d1d82996f6fa51ac07311b81b56R80

To fix the issue with not all users being visible due to pagination, I redone the filtering to work like autocomplete, so yeah, this PR is no longer doing the filtering.

PR #3135, opened 2019-10-07
Allow search by username when adding authors to the story. (SDBELGA-177)

GyanP:


scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -770,13 +770,16 @@ function MetaTermsDirective(metadata, $filter, $timeout, preferencesService, des
 
                         searchObj[scope.uniqueField] = t[scope.uniqueField];
                         if (searchUnique) {
-                            return (t.name.toLowerCase().indexOf(term.toLowerCase()) !== -1 ||
-                                    t[scope.uniqueField].toLowerCase().indexOf(term.toLowerCase()) !== -1) &&
-                                !_.find(scope.item[scope.field], searchObj);
+                            return (t.name.toLowerCase().indexOf(term.toLowerCase()) !== -1
```
tomaskikutis: Extra parentheses aren't needed here, remove them. Use `Array.includes` instead of `Array.indexOf`. Term `toLowerCase` is called a lot. Create a variable like `termLower` and use that instead.

Could you also explain the following lines?

```
|| t[scope.uniqueField].toLowerCase().indexOf(term.toLowerCase()) !== -1)
&& !_.find(scope.item[scope.field], searchObj);
```

scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -770,22 +770,27 @@ function MetaTermsDirective(metadata, $filter, $timeout, preferencesService, des
...
+                            // {qcode: "1", name: "Arbeidsliv", scheme: "subject_custom"}  is already added
+                            // and if user search for "Arbeidsliv" again he shouln't get any search results
                             && !_.find(scope.item[scope.field], searchObj);
                     }));
                     scope.activeList = true;
                 }
                 return scope.terms;
             };
 
+            const toLower = (term) => term.toLowerCase();
```
tomaskikutis: I wasn't suggesting to create a function, but a variable like `const termLower = term.toLowerCase()`. There isn't a point to create another function that does the same.

PR #3136, opened 2019-10-07
copy metadata in media editor based on picture schema

petrjasek:


tomaskikutis:
Why is it going directly to a release branch instead of to `develop` and only then to release?

petrjasek:
because I've started to prepare the release already for 1.32, so fixes go there and when the release is done it will be merged back to develop

petrjasek:
there is no ticket. when there are custom metadata fields on images copy&paste metadata ignores these, you can see that on sd-ansa fields like city/nation

PR #3142, opened 2019-10-09
[SDESK-4730] Function point to have planning prompt for further action when an news story is updated

nrvikas:


nrvikas:
Using FunctionPointsService now because Superdesk API is not available in planning as API is in 1.31.

MarkLark86:
@petrjasek This should only be required in Superdesk v1.30.

Once we update to Superdesk v1.31 we can start to use the Superdesk API and extensions to provide this functionality. Then we can remove the functionPointService entirely

petrjasek:
@MarkLark86 this should go to 1.30 branch and not develop then

petrjasek:
changing the base makes a mess here (there are too many other changes), can you close this PR and create a new one on 1.30 branch @nrvikas ?

nrvikas:
@petrjasek , absolutely, thanks. I will create a new one.

nrvikas:
Closing this PR to create a new one.

PR #3145, opened 2019-10-09
fix(build): Only compile extension e2e if protractor is installed

MarkLark86:


petrjasek:
@tomaskikutis why is every extension a package? seems like we just copy/paste the metadata without adding anything specific

tomaskikutis:
For one, it allows extensions to include their own dependencies so bundle size doesn't increase for instances where these extensions aren't being used. There aren't any custom dependencies yet, but some will probably come when video editing extension is done.

It also allows extensions to use different tools. We only care about the JavaScript bundle the extension produces. If someone wants to use a different testing library or write the code in CoffeeScript, they can do that without affecting the core. It's probably not relevant to our internal extensions, but it doesn't hurt to keep the `package.json`. It's also how independent builds are currently implemented. The core build script runs `compile` npm script for every extension.

petrjasek:
we can probably do the generated import still no matter if those are packages or not, so bundle size shouldn't be a factor here.
seems like there is no real benefit to maintain that structure, it allows us doing something we don't actually need. so why not dump it?

tomaskikutis:
> we can probably do the generated import still no matter if those are packages or not, so bundle size shouldn't be a factor here.

We can, but for one it's less clean since when you need a dependency for an extension, you would need to add it to the core.

More importantly, I'm thinking about improving the build in general which would include publishing extensions to npm. Our current build setup is quite messy due to every repo sharing the same build process from `client-core`. When we publish `client-core` to npm, we should publish an already bundled code and not the source. Other packages which depend on `client-core` should not be able to access the build process nor the source code like they can now.

I was thinking about how to gradually port planning to the extensions API and figured it would be easier if our extensions didn't need to be built in the core but would be imported and passed as an argument from parent repo(parent repo would import `startApp` from `client-core`). It would also allow easier development for extensions in separate repositories(like planning) by allowing to `npm link` the extensions which is not possible now due to our messy imports spanning across multiple repositories and all depending on client-core build process.

I've made a little [diagram](https://drive.google.com/file/d/1XnBlsziHeDmFf9lcXsXUAIhb5I71t69u/view?usp=sharing), see if it makes sense to you.

petrjasek:
yep that would make sense

PR #3146, opened 2019-10-10
re-implement hiding of nested items

petrjasek:
when it's activated filter out nested items via
query and fetch previous items from server when
needed.

SDFID-580 SDESK-4709

### DIFF
--- a/scripts/apps/search/components/Item.tsx
+++ b/scripts/apps/search/components/Item.tsx
@@ -55,7 +55,7 @@ interface IProps {
     svc: any;
     scope: any;
     swimlane: any;
-    item: IArticle;
+    item: IArticle | IPublishedArticle;
     profilesById: any;
     highlightsById: any;
     markedDesksById: any;
--- a/scripts/apps/search/components/Item.tsx
+++ b/scripts/apps/search/components/Item.tsx
@@ -12,7 +12,7 @@ import {ItemSwimlane} from './ItemSwimlane';
 import {ItemPhotoGrid} from './ItemPhotoGrid';
 import {ListItemTemplate} from './ItemListTemplate';
 import {ItemMgridTemplate} from './ItemMgridTemplate';
-import {IArticle, IDesk} from 'superdesk-api';
+import {IArticle, IDesk, IPublishedArticle} from 'superdesk-api';
 import {querySelectorParent} from 'core/helpers/dom/querySelectorParent';
 import {AuthoringWorkspaceService} from 'apps/authoring/authoring/services/AuthoringWorkspaceService';
 
@@ -70,7 +70,6 @@ interface IProps {
     narrow: any;
     hideActions: boolean;
     multiSelectDisabled: boolean;
-    nested: Array<IArticle>;
     isNested: boolean;
     actioning: boolean;
 }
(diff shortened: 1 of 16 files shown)


scripts/core/superdesk-api.d.ts
```diff
@@ -276,6 +278,15 @@ declare module 'superdesk-api' {
         };
     }
 
+    export interface IPublishedArticle extends IArticle {
```
tomaskikutis: As I've written in `SDESK-4687` it would be good to unify the articles so the structure is the same whether it's published or not. Do you know why `archive_item` and `item_id` is used? Wouldn't `status: 'published;` work while maintaining the same structure?

scripts/core/superdesk-api.d.ts
```diff
@@ -236,7 +237,8 @@ declare module 'superdesk-api' {
         broadcast: any;
         flags: any;
         source: string;
-        correction_sequence: any;
+        correction_sequence: number;
+        rewrite_sequence: number;
```
tomaskikutis: Are there any docs on what do these do? It doesn't have to be in this repo, a link would work fine too.

scripts/apps/search/components/Item.tsx
```diff
@@ -226,9 +234,35 @@ export class Item extends React.Component<IProps, IState> {
 
     toggleNested(event) {
         event.stopPropagation();
+
+        const showNested = !this.state.showNested;
+
+        if (showNested && !this.state.loading && !this.state.nested.length) {
+            this.setState({loading: true});
+            this.fetchNested(this.props.item as IPublishedArticle);
```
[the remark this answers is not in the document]
petrjasek: only published items have nested items

scripts/core/superdesk-api.d.ts
petrjasek: item_id is there because there are multiple versions of the same item in published collection, so there is another id for that. why there is archive_item I'm not sure exactly

scripts/apps/search/components/Item.tsx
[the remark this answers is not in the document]
tomaskikutis: I do not doubt that. What I'm saying is that at this line*, you are converting `this.props.item` to `IPublishedArticle`. By doing this you are assuring that it has the fields specific to `IPublishedArticle` while at line 54, where `IProps` interface is defined, you are saying that `this.props.item` is `IArticle` and doesn't have the additional fields that `IPublishedArticle` has, which is not true. `this.props.item` will sometimes have the fields and sometimes will not, so it needs to be typed `{item: IArticle | IPublishedArticle}`

\* `this.fetchNested(this.props.item as IPublishedArticle);`

scripts/core/superdesk-api.d.ts
tomaskikutis: Items that aren't published also have different versions you can revert back to. No matter the state of an article, it could probably use the same versioning management.

petrjasek: yep but there is only 1 version visible at all times, while with published we display them all together

tomaskikutis: I don't know how it's relevant, but it sounds like just another requirement to keep in mind rather than a limitation that would make it impossible to unify `published` and `archive` into a single entity. What do you think?

petrjasek: I'm sure it's possible

scripts/apps/search/components/Item.tsx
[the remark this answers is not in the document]
tomaskikutis: Remove `as IPublishedArticle`. Use a function to determine the type like [here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/core/ui/components/generic-form/interfaces/form.ts#L15)

PR #3149, opened 2019-10-11
fix(kill): Item schema not loading when killing from archived

MarkLark86:


MarkLark86:
This bug is affecting all versions >= 1.30

PR #3151, opened 2019-10-11
Allow custom avatars

tomaskikutis:
SDBELGA-149


scripts/index.ts
```diff
@@ -45,4 +61,10 @@ body.ready(() => {
     ].concat(appConfig.apps || []), {strictDi: true});
 
     window['superdeskIsReady'] = true;
-});
+}
+
+setTimeout(() => {
```
MarkLark86: What is the reason for replacing body.ready callback to a 500ms timeout? Using a timeout to control when to bootstrap the application is not ideal. What specific condition(s) are you waiting for?

tomaskikutis: I'm changing how superdesk client is being started. As you can see, `startApp` function is being exported. It's supposed to be imported and called from the repository which has `superdesk-core` as a dependency in `package.json`. For now, it allows providing a custom component to be used as an avatar. In the long term, I'm planning to use the same principle for passing the config(instead of `superdesk.config.js`), registering extensions and eventually decoupling `client-core` build process and only publishing a pre-built bundle(without source files) to npm.

I'm maintaining the current behavior for now so we don't have to immediately change how e2e tests are ran. It also works without adding `startApp` calls from main repositories, but that's only a coincidence, I'm planning to add the calls anyway.

The bundle is included in the bottom of the `body` tag, so there probably is no reason to wait for `body.ready` anyway. The only reason I can see is a likely mistake of a config file being included [after](https://user-images.githubusercontent.com/2076953/66734875-143a8d80-ee65-11e9-9904-0b1420c2bfc8.png) the main bundle which I fixed in this PR.

MarkLark86: I understand the reasons behind changing how the application is being started, and I'm all for that. My question is why are you using setTimeout? It is better to wait for a specific condition than to use a timeout (also resources are not guaranteed to arrive in order, i.e. it is possible that `app.bundle.js` could finish loading in the browser before `config.js`).

To make sure this app start code runs after the config has loaded, you will either have to use an event listener or the newer `defer` attribute on the `app.bundle.js` script element.

tomaskikutis: When I'm calling `startApplication` [from a parent repository](https://github.com/superdesk/superdesk-belga/pull/74/files#diff-61464acc9ecda49cb80e3c3e4d58ad4eR20), I'm using a timeout to avoid circular imports issue which can't be resolved without significant refactoring.

Since I already use a timeout there, I need another timeout to check whether `startApp` was called. It's not ideal, but shouldn't cause issues either.

By the way, the scripts **are** guaranteed to be executed in order. If a script hasn't downloaded yet, the browser will wait for it before executing the next script. There are exceptions if you lazy load or add an `async` attribute, but if the scripts are the last thing in the body and there are no special attributes, it will execute in order.

MarkLark86: > scripts are guaranteed to be executed in order

I apologise, I must have read the specifications incorrectly.

> I need another timeout to check whether startApp was called. It's not ideal

It seems like this `setTimeout` is a fallback in case the parent repository does not call `startApp`. If this is the case please add a comment here explaining this is a fallback only, and that the parent repository should be running this function instead.

scripts/index.ts
```diff
@@ -22,7 +24,21 @@ function initializeConfigDefaults(config) {
...
+) {
+    if (started === true) {
+        return;
+    }
+
+    started = true;
+
+    for (const key in customUiComponents) {
+        CC[key] = customUiComponents[key];
+    }
```
pablopunk: It's a bit safer to include a check here for `if (customUiComponents.hasOwnProperty(key)` before the assignment, so you only loop and assign the instance keys and not the unknown keys from the prototype chain.

PR #3152, opened 2019-10-13
[SDESK-4708] Create filter condition on on existence of featuremedia

marwoodandrew:


MarkLark86:
@petrjasek This one is required to be backported right through to 1.30

PR #3153, opened 2019-10-14
[SDCP-6] - Users should send only to member desks if no privilege

akintolga:
- users with `move` privilege can send content to any desk (no change)
- users with no `move` privilege can send still send content but only to member desks

### DIFF
--- a/scripts/apps/authoring/authoring/directives/SendItem.tsx
+++ b/scripts/apps/authoring/authoring/directives/SendItem.tsx
@@ -816,8 +816,12 @@ export function SendItem($q,
             }
 
             /**
-             * Set the last selected desk based on the user action.
-             * To be called after currentUserAction is set
+             * Sets the desk and stage list for user to move the story to.
+             * Also sets the last selected desk based on the user action.
+             * if user has the move privilege and if the story is not published she can send the story to any desk.
+             * if user doesn't have the move privilege or if the story is published
+             * then story can be sent only to member desks.
+             * This function to be called after currentUserAction is set
              */
             function setDesksAndStages() {
                 if (!scope.currentUserAction) {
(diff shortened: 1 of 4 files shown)


scripts/apps/authoring/authoring/directives/SendItem.tsx
```diff
@@ -824,8 +824,16 @@ export function SendItem($q,
                     return;
                 }
                 // set the desks for desk filter
-                if (scope.currentUserAction === ctrl.userActions.publish) {
+                if (scope.currentUserAction === ctrl.userActions.publish || !privileges.privileges.move) {
```
tomaskikutis: Can you explain this code? It looks strange that a particular piece of code gets executed if action is `publish` or the user doesn't have a `move` privilege. What does publishing have to do with moving? Shouldn't these be independent actions?

akintolga: Publish action is a compound action consisting of publish and optional send actions. When a story gets published users could choose to send the published story to the desks she is a member of. But they could send an unpublished story to any desk. Now we'd like to limit the desks to member desks for sending the (unpublished) stories if user doesn't have the move privilege.

tomaskikutis: The requirements make sense, but the implementation is confusing. Even if publishing is a compound action from a user perspective, can you separate it in the code?

akintolga: This function sets desks and stages for user to send the story to. There are two things that affect the list of desks and stages, one if the action is publish the other is if the user has the move privilege. I'm not sure if I need to separate it, but I can add some comments here. What do you think?

tomaskikutis: Yeah, a comment might do for now. Your first reply explains it pretty well so I'd just use that.

PR #3156, opened 2019-10-15
search users over the network

tomaskikutis:
SDBELGA-190


scripts/core/ui/components/SelectUser.tsx
```diff
@@ -9,29 +9,54 @@ import {UserAvatar} from 'apps/users/components/UserAvatar';
...
+        this.setState({loading: true, fetchedUsers: null});
+
+        dataApi.query<IUser>(
+            'users',
+            1,
+            {field: 'display_name', direction: 'ascending'},
+            (
+                searchString.length > 0
+                    ? {
+                        display_name: {
```
GyanP: I think it should be possible to search by username as well.

tomaskikutis: :+1:  Good catch, I forgot about that feature.

PR #3158, opened 2019-10-16
Add order to the media items when added to media gallery. (SDESK-4423)

GyanP:


tomaskikutis:
Can you describe what was the issue and how you fixed it?

GyanP:
As of now, we don't add the `order` field when the media items are added to the media-gallery field. It gets added only after reordering is done. So, when an article is published, in the output, the last added media item will show up as first due to timestamp.
In this PR, I made sure to add the `order` field as soon as the media is added to the media-gallery field and took care of reordering them whenever a media is removed from the gallery.

scripts/apps/authoring/authoring/controllers/AssociationController.ts
```diff
@@ -91,10 +91,14 @@ export function AssociationController(config, content, superdesk,
      */
     this.updateItemAssociation = function(scope, updated, customRel, callback = null, autosave = false) {
         let data = {};
-        // if the media is of type media-gallery, update same association-key not the next one
+        // when sending correction if the media is of type media-gallery, update same association-key not the next one
```
tomaskikutis: Is this specific to corrections and media galleries? Why?

GyanP: Sorry, my bad, it wasn't required at all, not sure why I did this.

tomaskikutis: Comments are never "required", but it's good to add them when you figure something out which isn't obvious otherwise and is useful for understanding the code. It's best to be precise with comments so that's why I asked if it's related to corrections :slightly_smiling_face:

GyanP: > What others there are, only related items?

feature-media, related items, and media-gallery

> do association keys work differently for them than other types of associations?

I think it works the same way as for other associations. The only issue is https://github.com/superdesk/superdesk-client-core/pull/2962#discussion_r285011218

PR #3160, opened 2019-10-17
[SDANSA-329] fix(carousel): Fixed issue with carousel on resize

pavlovicnemanja:


scripts/apps/workspace/directives/WorkspaceSidenavDirective.ts
```diff
@@ -34,6 +34,9 @@ export function WorkspaceSidenavDirective(superdeskFlags, Keys, config,
                 } else {
                     superdeskFlags.flags.hideMonitoring = false;
                 }
+
+                // Trigger resize event to update elements
+                $timeout(() => window.dispatchEvent(new Event('resize')), 500, false);
```
tomaskikutis: Implementation details specific to UI components must not leak outside. Handle it inside the carousel component instead.

pavlovicnemanja: This is not just for carousel, carousel has listener for resizing screen, as hideMonitoring is affecting size of editor or monitoring view, it will also behave better on splitter and responsive buttons in monitoring. Here I'm just triggering default resize event as it should be.

pavlovicnemanja: Carousel already have behaviour for this functionality, so creating something else on behave of this will be dirty fixing carousel plugin.

tomaskikutis: You can also use the [intersection observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to monitor the dimensions of a certain element like authoring panel or carousel. That way, the logic can stay in the respective components.

I've checked your current fix, and it doesn't seem to work when [resizing the pane](https://user-images.githubusercontent.com/2076953/67077131-e1f39d80-f18e-11e9-8748-6cb441842416.gif).

scripts/core/menu/menu.ts
```diff
@@ -1,5 +1,6 @@
 import {reactToAngular1} from 'superdesk-ui-framework';
 import {GlobalMenuHorizontal} from './GlobalMenuHorizontal';
+import { flags } from 'apps/search/components/fields/flags';
```
tomaskikutis: Is this used anywhere?

pavlovicnemanja: It's not, added by VS Code automatically... Removing...

scripts/apps/workspace/directives/WorkspaceSidenavDirective.ts
pavlovicnemanja: Resizing pane also fixed

PR #3163, opened 2019-10-18
implement sent items filter in personal space

petrjasek:
SDANSA-339

### DIFF
--- a/scripts/apps/monitoring/directives/MonitoringGroup.ts
+++ b/scripts/apps/monitoring/directives/MonitoringGroup.ts
@@ -231,7 +231,7 @@ export function MonitoringGroup(
                 }
 
                 let limited = !(monitoring.singleGroup || scope.group.type === 'highlights'
-                || scope.group.type === 'spike');
+                || scope.group.type === 'spike' || scope.group.type === 'personal');
 
                 if (!_.isNil(scope.forceLimited)) {
                     limited = JSON.parse(scope.forceLimited);
@@ -668,6 +668,8 @@ export function MonitoringGroup(
                             searchCriteria.source.size = PAGE_SIZE;
                         }
                     }
+                } else if (scope.group != null && scope.group.type === 'personal') {
+                    provider = 'production';
                 } else {
                     provider = 'archive';
                 }
--- a/scripts/apps/monitoring/directives/MonitoringGroup.ts
+++ b/scripts/apps/monitoring/directives/MonitoringGroup.ts
@@ -669,7 +669,7 @@ export function MonitoringGroup(
                         }
                     }
                 } else if (scope.group != null && scope.group.type === 'personal') {
-                    provider = 'production';
+                    provider = 'news';
                 } else {
                     provider = 'archive';
                 }
--- a/package.json
+++ b/package.json
@@ -140,8 +140,7 @@
     "react-test-renderer": "^16.2.0",
     "request": "^2.83.0",
     "superdesk-code-style": "^1.1.1",
-    "typescript-eslint-parser": "^18.0.0",
-    "webdriver-manager": "12.1.6"
+    "typescript-eslint-parser": "^18.0.0"
   },
   "scripts": {
     "setup-extensions": "node ./tasks/namespace-css-from-extensions.js && node ./tasks/install-extensions.js && node ./tasks/generate-extension-imports.js",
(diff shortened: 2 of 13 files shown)


petrjasek:
uses https://github.com/superdesk/superdesk-core/pull/1698

scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -668,6 +668,8 @@ export function MonitoringGroup(
                             searchCriteria.source.size = PAGE_SIZE;
                         }
                     }
+                } else if (scope.group != null && scope.group.type === 'personal') {
+                    provider = 'production';
```
tomaskikutis: Why doesn't archive work?

petrjasek: it doesn't have published items

petrjasek:
sent means what was sent to desk

scripts/apps/monitoring/directives/MonitoringGroup.ts
tomaskikutis: how about `published` endpoint? It should have published items.

tomaskikutis:
> sent means what was sent to desk

Does it matter where was it sent from?

petrjasek:
not really, it's in personal space, so it's for users to see items they've created which are not anymore in their private space but are on some desk or published

scripts/apps/monitoring/directives/MonitoringGroup.ts
petrjasek: it has only published, but not items in progress. so there is new api which has both

tomaskikutis:
> which are not anymore in their private space

Does it work if it never was in their private space? If I create an item in monitoring and send it to another desk, will it count as sent?

petrjasek:
yep it will have those too

scripts/apps/monitoring/directives/MonitoringGroup.ts
tomaskikutis: Is this new API a part of the effort to unify the endpoints for articles?(like SDESK-4687) Does it have personal items? Is it read-only or supports post/put/patch as well?

petrjasek: it is, for now it's read-only

tomaskikutis: Does the querying work identically to `archive` / `published`? Can we safely replace  `archive` / `published` queries to `production`?

I'd also pick a better name for it like `articles` or `publishable`(so name includes media).

petrjasek: it works like archive, I would do more testing before replacing it everywhere.
I think production is what we usually use when referring to this type of content, created by superdesk users and kind of live, so it's not ingest anymore and not archive yet

tomaskikutis: What I don't like in the name is that it's closer to an adjective than a noun. It would be unclear to name a collection `saved`, because it's not clear **what** is saved. So we call it `saved_searches`.

The same goes for `published` or `archived`. What is published or archived? Articles. Calling it `articles_archived`, `articles_published` or `articles_production` is much clearer.

petrjasek: yep saved/archived/published that's adjective, but production?

tomaskikutis: `production` describes the state, not an entity. For example, it's possible to have production planning items, production events, production desk stages, production workspaces. It's not how it's currently structured, but if you aren't familiar with the codebase there is no way to know what is it referring to when the name of an entity is not included in the name.

petrjasek: it's more about the process than state. adding articles to it is imo a bit misleading because there are pictures, packages, videos etc. `news_production` maybe, but what else would you produce in superdesk anyhow. for events that process is called planning. and as you said we don't use it anywhere else

tomaskikutis: > what else would you produce in superdesk anyhow

some kind of templates, media, analytics reports. It sounds silly for us since we know that we don't have that, but for someone not too familiar with the system it's hard to tell.

I'd say it's a problem that we haven't come up with a name for the main entity in Superdesk yet. It seems we agree that `archived` or `published` or `saved` are bad names because they are specific to one entity and that entity name is not a part of a collection/endpoint name.

`IArticle` in the client is misleading for the same reason as you pointed out - it can contain a package or a media item too. It would be good to come up with a better name for that entity and use it everywhere. I like `news` / `news item` you suggested. Since `production` is the default stage, I'd rename `production` collection/endpoint to be simply `news` and have others be different like `news_archived` or `news_ingest`.

petrjasek: :ok_hand: sounds like a plan

PR #3164, opened 2019-10-21
As a user I want to "save, close, and update" a story I am editing. (SDBELGA-197)

GyanP:


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -138,6 +138,7 @@ export function AuthoringDirective(
             $scope.fullPreviewUrl = '/#/preview/' + $scope.origItem._id;
             $scope.proofread = false;
             $scope.referrerUrl = referrer.getReferrerUrl();
+            $scope.gettext = gettext;
```
petrjasek: you are using both gettext and translate in template, what about using just gettext:
```
$scope.ACTION_SAVE_AND_CONTINUE_LABEL = gettext('Close and Continue');
$scope.ACTION_SAVE_AND_CONTINUE_LABEL_SHORT = gettext('C & C');
```
and use that in template

tomaskikutis: I'd simply pass `gettext` to the scope instead of adding a scope variable for each label. Having a variable for each label is more work without any benefits I can see.

There isn't much difference for 2 labels, but if we were apply this principle for all templates, it certainly doesn't make sense to have a variable per translation string, does it?

petrjasek: if gettext works in html that's fine too, we can just avoid using `translate` directive

tomaskikutis: What's the problem with a `translate` directive? We'll get rid of angular templates in the long run anyway, I don't think it's worth being too pedantic about the syntax when both work the same without any known bugs.

petrjasek: but why use gettext then and not `title="{{ 'Close and Continue' | translate }}" ` ?

tomaskikutis: Because string extraction doesn't work when translate **filters** are used.

String extraction doesn't work:
`<span>{{'Sample text' | translate}}</span>`

Everything works fine:
`<span translate>Sample text</span>`

petrjasek: I see, my bad

PR #3165, opened 2019-10-21
Fix moving media inside editor3

tomaskikutis:
SDESK-4726
SDESK-2420

Fixing a regression caused by #3081


tomaskikutis:
I haven't checked ` SDESK-4575`, but I've added a comment to the JIRA task and QA will re-test it.

PR #3169, opened 2019-10-24
Annotations library pre fill language

tomaskikutis:
SDFID-561


scripts/index.ts
```diff
@@ -13,22 +13,17 @@ if (appConfig.features.useTansaProofing) {
...
-        Object.assign(appConfig, serverConfig);
-        callback();
-    });
-    oReq.open('GET', appConfig.server.url + '/client_config');
-    oReq.send();
+function loadConfigs() {
+    return fetch(appConfig.server.url + '/client_config', {
+        method: 'GET',
+        mode: 'cors',
+    }).then((res) => res.json().then((json) => {
```
petrjasek: you should check response status: `(res) => res.ok ? res.json() : Promise.reject(res)`

tomaskikutis: the application won't start if there's an error here, but I can add error handling so it looks nice :smile:

PR #3174, opened 2019-10-27
Disallow editing embeds in read only mode

tomaskikutis:
SDESK-4441

### DIFF
--- a/superdesk.config.js
+++ b/superdesk.config.js
@@ -18,7 +18,6 @@ module.exports = function(grunt) {
             validatePointOfInterestForImages: true,
             hideRoutedDesks: false,
         },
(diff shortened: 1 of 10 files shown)


superdesk.config.js
```diff
@@ -18,6 +18,7 @@ module.exports = function(grunt) {
             validatePointOfInterestForImages: true,
             hideRoutedDesks: false,
         },
+        iframely: {key: '4358cefd09a66b1126c036'},
```
petrjasek: should the key be here? I assume it's free, but if we start deploying it by default it won't work eventually

tomaskikutis: Yep, it shouldn't. I've added it for testing and forgot to remove it.

PR #3178, opened 2019-10-29
fix(products): Fixed issue with removing products after reverting version

pavlovicnemanja:
[SDANSA-331]

### DIFF
--- a/scripts/apps/authoring/metadata/metadata.ts
+++ b/scripts/apps/authoring/metadata/metadata.ts
@@ -895,7 +895,7 @@ function MetaTermsDirective(metadata, $filter, $timeout, preferencesService, des
                     filteredArray = _.without(subjectCodesArray, term);
 
                 if (subjectCodesArray && filteredArray.length === subjectCodesArray.length) {
-                    _.remove(filteredArray, {name: term});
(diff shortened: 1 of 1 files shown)


tomaskikutis:
Can you describe what ended up being the issue?

pavlovicnemanja:
@tomaskikutis when items were reverted, product objects are not filtered well with` _.without`, so it will falling back to `_.remove` by string, so I added here first to check if `term` is object and has `name` property so it can by removed with `_.remove` functionality.

tomaskikutis:
Why was it failing using without? Can you add TypeScript types to the `MetaTermsDirective` so it's more clear? You can check the example on how to add types in `ItemCarouselDirective.ts`

pavlovicnemanja:
Fields were not a problem, they were the same, I think problem was caused by `__proto__`.

tomaskikutis:
What problems did `__proto__` cause?

tomaskikutis:
:) yeah, I understand that the item wasn't being removed, but why do you think it's related to `__proto__` and in what way did it influence the `without` function?

pavlovicnemanja:
That was only thing that was different in object. Because of that I thought that _.without function was not working. Can we move further discussion on slack channel? It's faster.

PR #3179, opened 2019-10-29
fix(multiedit): Fixed saving media on multiedit view

pavlovicnemanja:
[SDESK-2366]

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
@@ -1,9 +1,51 @@
 import _ from 'lodash';
+import {IArticle} from 'superdesk-api';
 import {FIELD_KEY_SEPARATOR} from 'core/editor3/helpers/fieldsMeta';
 import {getLabelNameResolver} from 'apps/workspace/helpers/getLabelForFieldId';
 import {MEDIA_TYPES} from 'apps/vocabularies/constants';
 import {isPublished} from 'apps/archive/utils';
 import {resetFieldMetadata} from 'core/editor3/helpers/fieldsMeta';
+
+interface IScope extends ng.IScope {
+    handleUrlsChange: any;
+    toggleDetails: any;
+    errorMessage: any;
+    contentType: number;
+    canListEditSignOff: any;
+    editSignOff: any;
+    mediaLoading: any;
+    validator: any;
+    features: any;
+    item: IArticle;
+    origItem: IArticle;
+    label: any;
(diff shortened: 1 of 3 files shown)


scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
```diff
@@ -324,9 +324,13 @@ export function ArticleEditDirective(
...
                             resetFieldMetadata(scope.item);
-                            scope.refresh();
 
-                            if (isPublished(scope.item)) {
+                            // On multiedit mode, there is no refresh function
+                            if (typeof scope.refresh === 'function') {
+                                scope.refresh();
+                            }
+
+                            if (isPublished(scope.item) || scope.articleEdit) {
```
tomaskikutis: What are you checking with ` scope.articleEdit`? I don't see such a variable in the directive.

scripts/core/editor3/helpers/fieldsMeta.ts
```diff
@@ -50,7 +50,9 @@ export function setFieldMetadata(item, fieldKey, contentKey, contentValue) {
 }
 
 export function resetFieldMetadata(item) {
-    Object.keys(item[META_FIELD_NAME]).forEach((key) => {
-        setFieldMetadata(item, key, fieldsMetaKeys.draftjsState, null);
-    });
+    if (item[META_FIELD_NAME]) {
```
tomaskikutis: When does it need this condition? Is it `null` sometimes? When?

scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
pavlovicnemanja: If scope.articleEdit is defined then it means it's called from Multiedit directive, which doesn't have item and originItem in scope, so autosave is not working.

scripts/core/editor3/helpers/fieldsMeta.ts
pavlovicnemanja: When this is triggered from multiedit it's failing:
https://github.com/superdesk/superdesk-client-core/pull/3179/files#diff-1718c3ee2990cd8f6be76101964abc0fR326

scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
tomaskikutis: Makes sense. Can you add TypeScript types for `scope` and add a comment in the interface that `articleEdit` means that the item is being edited in multi-edit mode?

PR #3180, opened 2019-10-30
add config for editor3 draft features

petrjasek:
SDANSA-341

### DIFF
--- a/scripts/apps/authoring/views/article-edit.html
+++ b/scripts/apps/authoring/views/article-edit.html
@@ -41,8 +41,9 @@
          data-read-only="!_editable"
          data-item="item"
          data-refresh-trigger="refreshTrigger"
-         data-tabindex="editor.headline.order">
-    </div>
+         data-tabindex="editor.headline.order"
+         data-strip-pasted-styles="editor.headline.cleanPastedHTML"
+    ></div>
 </div>
 
 <div class="field"
@@ -168,8 +169,9 @@
          data-read-only="!_editable"
          data-item="item"
          data-refresh-trigger="refreshTrigger"
-         data-tabindex="editor.abstract.order">
-    </div>
+         data-tabindex="editor.abstract.order"
+         data-strip-pasted-styles="editor.abstract.cleanPastedHTML"
+    ></div>
 </div>
 
 <div class="field"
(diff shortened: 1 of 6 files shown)


tomaskikutis:
I don't think that native spellchecker will work without breaking other stuff. If it's really needed, get QA to test applying native browser spellchecker suggestions in combination with one of the inline styles(like bold), links, comments, annotations, and all suggestion mode options.

scripts/apps/authoring/views/article-edit.html
```diff
@@ -41,8 +41,9 @@
          data-read-only="!_editable"
          data-item="item"
          data-refresh-trigger="refreshTrigger"
-         data-tabindex="editor.headline.order">
-    </div>
+         data-tabindex="editor.headline.order"
```
tomaskikutis: Why different names? `cleanPastedHTML` / `stripPastedStyles`

petrjasek: not sure, was using draftjs config name but I guess I can only use that on editor component

tomaskikutis:
Did you get the native spellchecker tested yet? It's very likely to not work with existing editor3 features, so I'd rather not merge it until tested. You can enable the config here so QA can test it out on a PR instance.

petrjasek:
I did test it and everything seems working, it maintains inline styles so everything works (annotations, comments, suggestions)

petrjasek:
but will assign it to qa for sure

tomaskikutis:
It's too good to be true ;D I'll check it out myself :)

tomaskikutis:
I've found the following issues:
* if suggestion mode is on and a spellchecker fix is applied, it isn't marked as a suggestion
* if there's a suggestion to replace [one word with another](https://user-images.githubusercontent.com/2076953/68132389-e5bd4900-ff1e-11e9-98e3-1becd672cc66.png), the spellchecker sees it as a single word and marks as a spelling error even though both words are correct.
* if there's a [suggestion in the middle of the word](https://user-images.githubusercontent.com/2076953/68132751-7f84f600-ff1f-11e9-9601-177254d53e02.png), like correcting a spelling mistake, it still displays a spellchecker error.

I'm wondering if the same issues are present when using our built-in spellchecker.

scripts/core/superdesk-api.d.ts
```diff
@@ -879,6 +879,10 @@ declare module 'superdesk-api' {
         };
         confirm_spike: boolean;
         language: string; // default client language
+        editor3: {
+            /** set if browser spellcheck is enabled */
+            spellCheck: boolean;
```
tomaskikutis: Can you rename it to `enableBrowserSpellchecker`? We have 2 spellcheckers so it'd be clearer which one is being referred to. I see you clarified it in the comment. Why not put it in the config name?

PR #3182, opened 2019-10-31
[SDESK-4790] - Shortcut buttons checking user privileges

akintolga:
- Also addressing an issue in `To Desk` shortcut that moves the story to the next stage

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
@@ -172,7 +172,7 @@ export function AuthoringDirective(
              */
             $scope.canPublishOnDesk = function() {
                 return !($scope.deskType === 'authoring' && config.features.noPublishOnAuthoringDesk) &&
-                    !!$scope.privileges.publish;
+                    privileges.userHasPrivileges({publish: 1});
             };
 
             getDeskStage();
@@ -863,7 +863,13 @@ export function AuthoringDirective(
             };
 
             $scope.sendToNextStage = function() {
-                var currentDeskId = desks.getCurrentDeskId() || (userDesks && userDesks[0]);
+                var currentDeskId = desks.getCurrentDeskId();
+
+                if (_.isNil(currentDeskId)) {
+                    notify.error(gettext('Failed to send to next stage.'));
+                    return;
+                }
+
                 var stageIndex, stageList = desks.deskStages[currentDeskId];
                 var selectedStage, selectedDesk = desks.deskLookup[currentDeskId];
 
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
@@ -171,7 +171,8 @@ export function AuthoringDirective(
              * @returns {Boolean}
              */
             $scope.canPublishOnDesk = function() {
-                return !($scope.deskType === 'authoring' && config.features.noPublishOnAuthoringDesk);
+                return !($scope.deskType === 'authoring' && config.features.noPublishOnAuthoringDesk) &&
+                    !!$scope.privileges.publish;
             };
 
             getDeskStage();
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -171,7 +171,8 @@ export function AuthoringDirective(
              * @returns {Boolean}
              */
             $scope.canPublishOnDesk = function() {
-                return !($scope.deskType === 'authoring' && config.features.noPublishOnAuthoringDesk);
+                return !($scope.deskType === 'authoring' && config.features.noPublishOnAuthoringDesk) &&
+                    !!$scope.privileges.publish;
```
tomaskikutis: Is `!!$scope.privileges.publish` meant to mean `$scope.privileges.publish != null`? Use the explicit version.

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -862,8 +863,9 @@ export function AuthoringDirective(
             };
 
             $scope.sendToNextStage = function() {
-                var stageIndex, stageList = desks.deskStages[desks.activeDeskId];
-                var selectedStage, selectedDesk = desks.deskLookup[desks.activeDeskId];
+                var currentDeskId = desks.getCurrentDeskId() || (userDesks && userDesks[0]);
```
tomaskikutis: What's the change here? Does something not work with using `activeDeskId`? Why is fallback to `userDesks[0]` used?

scripts/apps/authoring/views/authoring-topbar.html
```diff
@@ -58,28 +58,25 @@
...
                 type="submit"
-                ng-if="canPublishOnDesk()"
                 ng-click="saveAndContinue(publish, true)"
                 title="{{ :: 'Publish' | translate }}">
                 <span class="btn__text" translate>Publish</span>
                 <span class="btn__text--short" translate>P</span>
         </button>
         <button
-                ng-if="$root.config.features.customAuthoringTopbar.publishAndContinue === true"
+                ng-if="$root.config.features.customAuthoringTopbar.publishAndContinue === true && canPublishOnDesk() && canRewriteArticle()"
```
tomaskikutis: Why is `canRewriteArticle` checked for `Publish and continue` action, but not for `Publish`?

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
akintolga: I think `getCurrentDeskId` uses `activeDeskId` so it is better to use the `getCurrentDeskId` here. It also considers users desks as well, so I removed some of the redundant bits there. Also added an error message where there is no current desk id (esp if the user is working in his personal space).

scripts/apps/authoring/views/authoring-topbar.html
akintolga: `Publish and continue` publishes a story and creates an update right after, That's why it requires `canRewriteArticle` permission.

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -862,8 +863,15 @@ export function AuthoringDirective(
             };
 
             $scope.sendToNextStage = function() {
-                var stageIndex, stageList = desks.deskStages[desks.activeDeskId];
-                var selectedStage, selectedDesk = desks.deskLookup[desks.activeDeskId];
+                var currentDeskId = desks.getCurrentDeskId();
+
+                if (_.isNil(currentDeskId)) {
+                    notify.error(gettext('Failed to send to next stage.'));
```
tomaskikutis: When is `currentDeskId` null/undefined? Is it a valid use case or an error? If it's an error, add a `logger.warn` call so the error gets sent to sentry and we can find and fix it.

akintolga: currentDeskId could be null if the user is working on his personal space so I don't think it is an error

tomaskikutis: Can you add a reason to the error message so it's clear to the user why the action doesn't work?

PR #3184, opened 2019-11-04
Remove responsive iframes

pablopunk:
SDESK-4786

When the user inputs an embed code, it won't be modified. If the user inputs an url, iframely will take care of responsive layout.


tomaskikutis:
In order to make iframes responsive & maintain aspect ratio, [padding hack](https://davidwalsh.name/responsive-iframes) is used. The problem here is that the embed code you provided comes with a padding hack already applied which messes things up.

What your current changes do, is basically remove the padding hack which doesn't maintain [aspect ratio](https://user-images.githubusercontent.com/2076953/68116831-12ac3480-fefc-11e9-9d5c-a1a819a00863.gif).

I've talked to Karel and he said we're fine with embeds not being responsive if they are added via embed-code instead of a URL. Can you remove `makeIframesResponsive` entirely?

Embeds from iframely come with responsiveness hack already applied. Some 3rd parties like streamable provide the hack with their embed code too. If we ever need to support responsive embeds for a service that isn't handled by iframely and doesn't provide responsive embeds hack in its own embed code, we could build a custom iframely like URL -> embed code mapper.

tomaskikutis:
One more thing, can you add an `i` icon with a tooltip somewhere near the embed input field explaining that embed codes might not work with responsive design unless the provider of the embed code has taken care of that? I'd also add a suggestion to use URL embeds instead of embed codes.

scripts/core/editor3/components/embeds/EmbedInput.tsx
```diff
@@ -131,6 +132,14 @@ export class EmbedInputComponent extends React.Component<any, any> {
 
         return (
             <form onSubmit={this.onSubmit} className="embed-dialog" onKeyUp={this.onKeyUp}>
+                <OverlayTrigger overlay={
+                    <Tooltip id="create_new_embed_tooltip">
+                        <p translate>To get a responsive embed code, paste a URL.</p>
+                        <p translate>If you paste an embed code, it will be used &quot;as is&quot;.</p>
```
tomaskikutis: `translate` attribute only works in angular templates. Use `gettext` elsewhere. `<p>{gettext('To get a responsive embed code, paste a URL.')}</p>`

PR #3187, opened 2019-11-05
Fix: Typeahead (autocompletion) does not work well for multilingual CVs. [SDESK-4791]

GyanP:


scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -768,8 +768,18 @@ function MetaTermsDirective(metadata, $filter, $timeout, preferencesService, des
                     scope.terms = $filter('sortByName')(_.filter(filterSelected(searchList), (t) => {
                         var searchObj = {};
                         const termLower = term.toLowerCase();
+                        // if there are translations available and the item language is other than "en",
+                        // search the term from the translations
+                        const searchFromTranslations = t.translations != null && t.translations.name != null
+                            && t.translations.name[scope.item.language] != null && scope.item.language !== 'en';
```
petrjasek: I wonder why there is this last `&& scope.item.language !== 'en';` ? if there is a translation for item language lets use it, even if it's english

PR #3188, opened 2019-11-05
Allow extensions from external repositories

tomaskikutis:
SDBELGA-210

This changes how extensions are setup. Instead of being configured in `superdesk.config.js`, extensions are now imported from root repositories like [here](https://github.com/superdesk/superdesk-belga/pull/82) and are passed as an argument to `startApp` function. It makes it easier to develop extensions in external repositories like planning. Previously, generated imports file in client-core had to be adjusted every time when developing extensions from external repositories.

REQUIRED BY: https://github.com/superdesk/superdesk-belga/pull/82
REQUIRED BY: https://github.com/superdesk/superdesk-planning/pull/1383


petrjasek:
I wonder if it makes sense to keep that separate install/compile step for extensions, we import it anyhow while compiling the app, not sure what is the benefit..

tomaskikutis:
It's beneficial for keeping extensions self-contained. If your extension requires a dependency, you add it to the extension instead of the entire app. If you want to remove the extension, you delete the folder and can be sure that there's no related code in the core. The separate build process gives more control for extensions. For example, all extensions use stricter TypeScript configs than we can use in the core.

The long term plan is to separate the build processes between client core / parent repos / planning and only publish a pre-built bundle of client-core to npm instead of the entire source.

petrjasek:
I think it would make sense then to make those extensions dependencies of the core and handle compilation in their package.json file and just import them later, same thing like in [belga repo](https://github.com/superdesk/superdesk-belga/pull/82#discussion_r342970242)

tomaskikutis:
I don't see a way to make it work like that. npm modules are designed to work as pre-built bundles and when we try to hack it to build itself on install, we run into problems.

For example, `markForUser` module depends on `end-to-end-testing-helpers`. Both are local modules and are currently expected to build themselves on install. When one of them is put inside `node_modules` directory, the path to another module changes and it won't build.

Also, `devDependencies` won't get installed if I add it as a module.

I suggest to keep it as is for now and later build automation to bundle modules and push them to npm on every push to the remote. Do you have any other ideas?

petrjasek:
I don't understand what is the problem with npm. It should call each extensions install script if there is any on superdesk-core install. and it should allow imports like `import extension from 'markForUser';` without exposing extensions structure when using it.

tomaskikutis:
If you want to have a local package, you must specify a **relative** path in [package.json](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/extensions/markForUser/package.json#L12). The same relative path can't work for running `npm install` from `scripts/extensions/markForUser` and `node_modules/markForUser`

petrjasek:
are we talking about `end-to-end-testing-helpers` ? or is there any other path which is problematic?

tomaskikutis:
If you have one local module depending on another local module and expect both to build themselves, every path will be problematic.

You can check out my branch and try it out yourself.

PR #3197, opened 2019-11-07
draft did remove some styles, but not all

petrjasek:
SDANSA-341

### DIFF
--- a/scripts/core/editor3/components/handlePastedText.ts
+++ b/scripts/core/editor3/components/handlePastedText.ts
@@ -82,6 +82,10 @@ export function handlePastedText(text: string, _html: string): DraftHandleValue 
         return 'handled';
     }
 
+    if (this.props.cleanPastedHtml) { // force plain text, but keep line breaks
+        html = '<p>' + text.split('\n').join('</p><p>') + '</p>';
+    }
+
     if (suggestingMode) {
         if (!Suggestions.allowEditSuggestionOnLeft(editorState, author)
             && !Suggestions.allowEditSuggestionOnRight(editorState, author)) {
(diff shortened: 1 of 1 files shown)


scripts/core/editor3/components/handlePastedText.ts
```diff
@@ -82,6 +82,10 @@ export function handlePastedText(text: string, _html: string): DraftHandleValue
         return 'handled';
     }
 
+    if (this.props.cleanPastedHtml) { // force plain text, but keep line breaks
```
tomaskikutis: Shouldn't pasting multi-line text always work, not only when you have `cleanPastedHtml` enabled?

petrjasek: it works, but it keeps some html if it's enabled like headlines, lists, probably all supported blocks.

tomaskikutis: I've checked the PR locally and it doesn't work without enabling `cleanPastedHtml`

lines

petrjasek: yep that's the idea, you have to enable that in content profile

petrjasek: or you mean it should preserve line breaks when pasting text? that's probably how it works now on master too

tomaskikutis: > yep that's the idea, you have to enable that in content profile

To enable what? preserving lines when pasting? It should work out the box. It does on draftjs demo website. I know it doesn't work on master either, but if you are working on the feature, it would be good to fix it in general instead of addressing a single use-case.

petrjasek: sure, I thought you're talking about removing formatting

PR #3198, opened 2019-11-07
Keep required value of editor fields in sync with the schema fields for Content Profiles.

GyanP:
fixes SDNTB-615

When creating a new CP or adding a new field to it the `required` field is present in schema fields but missing from the editor fields, it gets inserted only if we toggle the required checkbox in the CP.

### DIFF
--- a/scripts/apps/workspace/content/controllers/ContentProfileFields.ts
+++ b/scripts/apps/workspace/content/controllers/ContentProfileFields.ts
@@ -117,6 +117,8 @@ export default function ContentProfileFields($scope, content, vocabularies, meta
 
                 this.sections[section].enabled.push(key);
                 this.model.editor[key].order = index + 1; // keep order in sync
+                // keep required value of editor fields in sync with the schema fields SDNTB-615
+                this.model.editor[key].required = this.model.schema[key].required;
             });
 
             Object.keys(this.model.editor)


scripts/apps/workspace/content/controllers/ContentProfileFields.ts
```diff
@@ -117,6 +117,8 @@ export default function ContentProfileFields($scope, content, vocabularies, meta
 
                 this.sections[section].enabled.push(key);
                 this.model.editor[key].order = index + 1; // keep order in sync
+                // keep required value of editor fields in sync with the schema fields SDNTB-615
+                this.model.editor[key].required = this.model.schema[key].required;
```
tomaskikutis: Why do we have `required` field in 2 places? Shouldn't it only be in the content profile? What's the distinction between `editor` and `schema`?

GyanP: A schema defines validation rules etc and is about the fields in DB. while the editor defines what is visible in the editor. Some editor fields are stored in different schema fields, ex: any custom cv is stored in a subject field. This is what Petr told me today morning.

In the authoring, we show the required indication using both [schema](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/authoring/views/authoring-header.html#L255) as well as [editor](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/authoring/views/authoring-header.html#L332) for different fields.

tomaskikutis: Thanks for explaining. As I understand it, `schema` is a list of fields that are supported for configuring which fields appear in authoring view, and that configuration is stored in `editor`.

PR #3202, opened 2019-11-08
Fix editor3 paste initial space

petrjasek:
hacking draft convertFromHTML

SDANSA-341

### DIFF
--- a/scripts/core/editor3/components/handlePastedText.ts
+++ b/scripts/core/editor3/components/handlePastedText.ts
@@ -59,9 +59,10 @@ function pasteContentFromOpenEditor(
     return 'not-handled';
 }
 
-export function prepareTextForPaste(text: string) {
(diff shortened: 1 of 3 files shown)


tomaskikutis:
Can you explain the problem you are solving here? What does "initial space" mean?

petrjasek:
spaces at the beginning of each line. there was always 1 less in the editor than in pasted text

scripts/core/editor3/components/handlePastedText.ts
```diff
@@ -83,8 +87,9 @@ export function handlePastedText(text: string, _html: string): DraftHandleValue
     }
 
     // preserve line breaks when pasting or forcing plain text
+    // \r are important for draft convertFromHTML to preserve initial spaces on each line
```
tomaskikutis: Can you move the comment above `prepareTextForPaste`?

PR #3205, opened 2019-11-11
Format options for custom text fields. [SDESK-4805]

GyanP:


scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
```diff
@@ -126,10 +126,9 @@ export function ContentProfileSchemaEditor(vocabularies) {
 
             scope.getEditor3FormattingOptions = (fieldName) => {
                 const isCustomPlainTextField = typeof scope.fields[fieldName] === 'object'
-                    && typeof scope.fields[fieldName].field_options === 'object'
-                    && scope.fields[fieldName].field_options.single === true;
```
tomaskikutis: What does `field_options.single` mean?

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
```diff
@@ -126,10 +126,9 @@ export function ContentProfileSchemaEditor(vocabularies) {
...
+                    && scope.fields[fieldName].field_type === 'text';
```
tomaskikutis: What other `field_type`s there are?

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
GyanP: Setting `field_options.single` to `true` means that the field is a single-line field. eg: slugline is a single-line field.

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
GyanP: `related_content`, `media`, `date_field` etc.

tomaskikutis: > related_content, media, date_field etc.

None of these can use editor3, does it make sense to check them?

GyanP: Not at all, we have to check it for `text` fields only.

PR #3206, opened 2019-11-11
fix invisible ingest routing schemas

petrjasek:
temporary fix for missing pagination

SDANSA-350


petrjasek:
I've set 200 that's maximum server allows

PR #3210, opened 2019-11-13
Datetime field

tomaskikutis:
SDBELGA-217

Implemented date time field & made it possible to add custom fields from extensions to article header.

DEPENDS ON: https://github.com/superdesk/superdesk-ui-framework/pull/395

captura_2


scripts/extensions/datetimeField/src/extension.tsx
```diff
@@ -0,0 +1,55 @@
...
+} from 'superdesk-api';
+import {getDateTimeField} from './getDateTimeField';
+import {getConfigComponent} from './getConfigComponent';
+
+class DateTimePreview extends React.PureComponent<IPreviewComponentProps> {
+    render() {
+        if (this.props.value == null) {
+            return null;
+        } else {
+            return <div>{new Date(this.props.value).toLocaleString()}</div>;
```
petrjasek: imo we should provide helpers for datetime formatting via ISuperdesk, we have configs for that so we should keep it consistent

PR #3211, opened 2019-11-14
Fix search and edit on annotations library

pablopunk:
SDESK-4802

There were several issues when searching & editing items inside the annotations library, caused by filters only being formatted the right way the first time the page load, not after every refresh while editing.

### DIFF
--- a/scripts/core/ui/components/ListPage/generic-list-page.tsx
+++ b/scripts/core/ui/components/ListPage/generic-list-page.tsx
@@ -33,6 +33,7 @@ import {
     IGenericListPageComponent,
     ICrudManagerFilters,
     ICrudManager,
+    IFormGroup,
 } from 'superdesk-api';
 import {gettext} from 'core/utils';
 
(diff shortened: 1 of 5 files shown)


scripts/core/ui/components/ListPage/generic-list-page.tsx
```diff
@@ -613,12 +601,33 @@ export class GenericListPageComponent<T extends IBaseRestApiResponse>
...
-    connectServices<IPropsGenericForm<T>>(
-        connectCrudManager<IPropsGenericForm<T>, IPropsConnected<T>, T>(
-            GenericListPageComponent,
-            'items',
-            resource,
-        )
-        , ['modal', '$rootScope', 'notify'],
-    );
+export const getGenericListPageComponent =
+    <T extends IBaseRestApiResponse>(resource: string, formConfig?: IFormGroup) =>
```
tomaskikutis: Can you make `formConfig` non-optional and pass it from all the places that call the function?

scripts/core/ui/components/ListPage/generic-list-page.tsx
```diff
@@ -613,12 +601,33 @@ export class GenericListPageComponent<T extends IBaseRestApiResponse>
...
+        connectServices<IPropsGenericForm<T>>(
+            connectCrudManager<IPropsGenericForm<T>, IPropsConnected<T>, T>(
+                GenericListPageComponent,
```
tomaskikutis: It would be better if the component returned from `getGenericListPageComponent` wouldn't have a `formConfig` prop and would use the one passed to function as a parameter, but it's complex to do, so don't bother with it. I just wanted to mention a possible improvement.

scripts/core/ui/components/ListPage/generic-list-page.tsx
pablopunk: @tomaskikutis what about [here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/extensions/annotationsLibrary/src/AnnotationInputWithKnowledgeBase.tsx#L49), as you can see it's not using any formConfig but it's [generating the filters](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/extensions/annotationsLibrary/src/AnnotationInputWithKnowledgeBase.tsx#L27) anyway

tomaskikutis: Do you see an issue with it? `formatConfig` is needed for retrieving the field type and passing it along with a value to `generateFilterForServer`. Here it's done directly.

pablopunk: it's not an issue if `formConfig` is optional in `connectCrudManager`

pablopunk: ok nevermind I thought you were saying formConfig should be non-optional in crudmanager, not genericListPageComponent

PR #3212, opened 2019-11-15
[SDESK-4766] Allow a Contact to have a 'contact_type'

MarkLark86:
Depends on https://github.com/superdesk/superdesk-core/pull/1727


MarkLark86:
@petrjasek This one is required to be backported right through to 1.30

scripts/apps/contacts/Contacts.d.ts
```diff
@@ -0,0 +1,75 @@
+
```
tomaskikutis: Use \*.ts/\*.tsx files instead of *d.ts when you can. *d.ts ones don't get type-checked properly. It might be an issue with our setup, but it's easier to change the extension than investigate it.

scripts/apps/contacts/components/Form/ProfileDetail.tsx
```diff
@@ -1,13 +1,50 @@
...
+    mobileUsages: any;
+    displayOtherStateField: any;
+    requiredField: any;
+    touched: any;
+    organisations: any;
+    orgValue: any;
+    contactTypes: any;
+}
+
+export class ProfileDetail extends React.Component<IProps, IState> {
```
tomaskikutis: extend from `React.PureComponent` when possible.

scripts/apps/contacts/components/Form/ProfileDetail.tsx
```diff
@@ -129,17 +171,43 @@ export class ProfileDetail extends React.Component<any, any> {
                     </div>
                 }
 
-                {!readOnly &&
-                    <Row>
+                <Row flex={true}>
+                    {get(this.state, 'contactTypes.length', 0) > 0 && (
```
tomaskikutis: Use optional chaining and nullish coalescing instead of `lodash.get` to preserve type safety: `this.state?.contactTypes?.length ?? 0 > 0`

scripts/apps/contacts/helpers.ts
```diff
@@ -75,20 +76,30 @@ export function validateRequiredField(contact) {
...
     });
 }
 
+export function validateAssignableType(contact: IContact): boolean {
+    if (!_.get(contact, 'contact_type.assignable')) {
+        return true;
+    }
+
+    return _.get(contact, 'contact_email.length', 0) > 0 &&
+        !_.some(contact.contact_email, (v) => _.isEmpty(v));
```
tomaskikutis: Why not use native [Array.prototype.some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)?

Is `isEmpty` being used simply as `x != null`? If so, use the native construct instead of a library function.

scripts/apps/contacts/services/ContactsService.ts
```diff
@@ -140,9 +164,37 @@ export class ContactsService {
...
+    }
+
+    convertForClient(response: IContact) {
+        if (!response) {
+            return response;
+        }
+
+        const contact: IContact = replaceUrls(response);
+
+        if (typeof contact.contact_type === 'string' && get(this.metadata, 'values.contact_type.length', 0) > 0) {
```
tomaskikutis: What's the purpose of the conversion? Why doesn't `contact_type` from the server work?

scripts/apps/contacts/components/Form/ProfileDetail.tsx
MarkLark86: This PR has to be merged through to v1.30 of Superdesk which uses Typescript 3.0, where optional chaining and nullish coalescing is not available.

scripts/apps/contacts/services/ContactsService.ts
MarkLark86: The server stores `contact_type` as a string, where as we want to use the object version of the value in the client `IContactType` (which includes `qcode`, `name` and `assignable` fields)

tomaskikutis: Wouldn't it be cleaner to keep the entity exactly as it is on the server and load additional data outside of that data structure?

```
export interface IContact {
    _id: string;
    contact_type?: string;
}

const contacts: Array<IContact> = getContacts();

contacts.forEach((contact) => {
    const contactTypeObj = metadata[contact.contact_type];

    console.log(contactTypeObj.name);
});
```

We do a similar thing with users frequently. If we need to get a display name of a user, we don't patch the object which has the user as an ID string. We load full user objects separately and can then retrieve these objects by IDs.

PR #3213, opened 2019-11-15
implement video preview component with poster support

petrjasek:
SDCP-67

### DIFF
--- a/scripts/core/superdesk-api.d.ts
+++ b/scripts/core/superdesk-api.d.ts
@@ -279,6 +279,16 @@ declare module 'superdesk-api' {
             archive?: boolean;
             externalsource: boolean;
         };
+
+        // media renditions
+        renditions: {
+            [key: string]: {
+                href: string;
+                width?: number;
+                height?: number;
+                mimetype: string;
+            };
+        };
     }
 
     export interface IPublishedArticle extends IArticle {
--- a/scripts/core/superdesk-api.d.ts
+++ b/scripts/core/superdesk-api.d.ts
@@ -280,13 +280,20 @@ declare module 'superdesk-api' {
             externalsource: boolean;
         };
 
-        // media renditions
+        /** Wrapper for different renditions of non-textual content of the news object */
         renditions: {
             [key: string]: {
+                /** The URL for accessing the rendition as a resource */
                 href: string;
-                width?: number;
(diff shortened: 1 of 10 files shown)


scripts/core/superdesk-api.d.ts
```diff
@@ -279,6 +279,16 @@ declare module 'superdesk-api' {
             archive?: boolean;
             externalsource: boolean;
         };
+
+        // media renditions
+        renditions: {
+            [key: string]: {
+                href: string;
+                width?: number;
```
tomaskikutis: Shouldn't width/height be mandatory for a rendition?

petrjasek: I don't get such info from some external apis

tomaskikutis: I thought renditions are our internal thing. If you have an image/video and want to have it in different sizes, you define a point of interest, original media gets cropped and renditions generated. Isn't it how it works?

petrjasek: yep but we use the same structure for external media archive search where that info could be missing. also there are renditions for audio files which won't have any width/height

tomaskikutis: Can you add your explanation as a comment in the code?

What's your opinion on this data structure in the context of how we are using it? Do you think it's good as-is and we don't need to change anything or that it's no good and we should be using X instead?

petrjasek: it's actually part of the standard https://www.iptc.org/std/ninjs/ninjs-schema_1.1.json and similar thing is in NewsMLG2 so we need to map to it in the output and I think it works fine internally too.

tomaskikutis: ninjs/newsml standards only matter at the application boundaries when we need to import/export content. We need to have certain information available, but we can use whatever data structures we want internally.

Can you make comments cleaner? Avoid dummy comments like `the width of the display area measured in pixels`. All interface fields are clear and don't need any comments. What isn't obvious is how the renditions field is used. Like what's the format of the key? Mention the difference between internal renditions feature that generates different sizes for source images, that it only works for images and not other media types(the internal renditions). Mention that when importing from outside of Superdesk, media gets inserted into this field.

PR #3215, opened 2019-11-16
allow extensions to provide widgets for search panel

petrjasek:
SDCP-67

### DIFF
--- a/scripts/apps/search/components/search-panel-widgets.tsx
+++ b/scripts/apps/search/components/search-panel-widgets.tsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import {flatMap} from 'lodash';
+import {extensions} from 'appConfig';
+
+interface IProps {
+    provider: any;
+    params: any;
+    setParams: (updates: any) => void;
+}
+
+export class SearchPanelWidgets extends React.PureComponent<IProps> {
+    render() {
+        const widgets = flatMap(
+            Object.values(extensions)
+                .map((extension) => extension.activationResult.contributions.searchPanelWidgets)
+                .filter((_widgets) => _widgets != null),
+        );
+
+        if (widgets.length === 0) {
+            return null;
+        }
+
+        return (
+            <React.Fragment>
+                {widgets.map((Widget, index) =>
+                    <Widget key={index}
+                        provider={this.props.provider}
+                        params={this.props.params}
+                        setParams={(updates) => this.props.setParams(updates)}
+                    />,
+                )}
+            </React.Fragment>
+        );
+    }
+}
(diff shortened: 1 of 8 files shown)


tomaskikutis:
Can you explain how it works? Do interface `params` act simply as additional filters? I see that a component from extensions is getting rendered into the "advanced search" section of the search, but other than that I don't understand how an actual extension might work. It's not that there's something wrong with your code. I'm simply not familiar with search providers.

Can you go over how do custom media providers work in general? I assume it's not only a jpeg/mp4 files that we are getting from these providers, but also metadata. Do we have a transform function somewhere on the back-end which converts outside data structures to a Superdesk format?

petrjasek:
yes the extensions provide additional filters based on external api capabilities, eg portrait/landspace pictures filter. params are sent to server together with query.
on the server there is a proxy for external api which sends request and then converts the results to superdesk like format, as you said.
usage: https://github.com/superdesk/superdesk-cp/pull/2

scripts/core/superdesk-api.d.ts
```diff
@@ -77,9 +77,15 @@ declare module 'superdesk-api' {
                 };
             };
             iptcMapping?(data: IPTCMetadata, item: Partial<IArticle>): Promise<Partial<IArticle>>;
+            searchPanelWidgets?: Array<React.ComponentType<ISeachPanelWidgetProps>>;
         }
     }
 
+    export type ISearchPanelWidgetProps = {
+        provider: string;
+        params: any;
```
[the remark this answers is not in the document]
petrjasek: I don't see any common params. from/to is orangelogic specific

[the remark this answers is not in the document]
tomaskikutis: I see. Keep it generic then: `params: T;` and define from/to in the extension.

scripts/apps/search/components/search-panel-widgets.tsx
```diff
@@ -12,7 +12,7 @@ export class SearchPanelWidgets extends React.PureComponent<IProps> {
     render() {
         const widgets = flatMap(
             Object.values(extensions)
-                .map((extension) => extension.activationResult?.contributions?.searchPanelWidgets ?? [])
+                .map((extension) => extension.activationResult?.contributions?.searchPanelWidgets || [])
```
tomaskikutis: How is this a lint fix? Out of curiosity, I've checked it out the commit before this one and `npm run lint` passes without any errors.

petrjasek: npm run lint did pass, but not `grunt eslint:app` which we use on travis

tomaskikutis: Right. I now remember that I had the same issue myself and solved it by removing `eslint:app` task. Previously, 2 lint tasks were always being executed - `eslint:app` and `npm run lint`.

PR #3216, opened 2019-11-18
SDBELGA-152 keycloak

idsvn:
Hi @tomaskikutis , this is task for SSO(keycloak) of Belga.
So, IMO, We move it into extension of Belga, So we need a api Auth to use in Belga.
Please support us or you have any idea for it.


package.json
```diff
@@ -84,6 +84,7 @@
     "json-loader": "~0.5.4",
     "json-merge-patch": "0.2.3",
     "json5": "0.5.1",
+    "keycloak-js": "^6.0.1",
```
tomaskikutis: Lock to the exact version: `"keycloak-js": "6.0.1",`

scripts/core/auth/keycloak-directive.ts
```diff
@@ -0,0 +1,44 @@
...
+angular.module('superdesk.core.auth.keycloak', []).directive('sdKeycloakModal', [
+    'session',
+    'auth',
+    'deployConfig',
+    function(session, auth, deployConfig) {
+        return {
+            link: function(scope, $element) {
+                const init = async () => {
+                    const methods = await deployConfig.all({ oidc: 'oidc_auth' })
+                    const isOIDCEnabled = await methods.oidc
```
tomaskikutis: Use promises instead of `async/await`

scripts/core/auth/keycloak-directive.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/*
+ * Keycloak modal is watching session identity, redirect to login screen
+ * and call logout when session is cleared
+ */
+angular.module('superdesk.core.auth.keycloak', []).directive('sdKeycloakModal', [
+    'session',
+    'auth',
+    'deployConfig',
+    function(session, auth, deployConfig) {
```
tomaskikutis: replace usages of `deployConfig` with `appConfig` from `scripts/appConfig.ts`

tomaskikutis: Don't forget to remove `deployConfig` from here

scripts/core/auth/keycloak-directive.ts
```diff
@@ -0,0 +1,36 @@
...
+                    scope.keycloak.init({
+                        onLoad: 'login-required',
+                    }).success(() => {
+                        auth.loginOIDC(scope.keycloak.token);
+                    });
+                };
+
+                init();
+
+                scope.$watch(() => session.identity, () => {
```
petrjasek: I guess you could set this watch only on auth success so it won't run if there is no keycloak used

scripts/core/auth/keycloak-directive.ts
```diff
@@ -0,0 +1,36 @@
+import Keycloak from 'keycloak-js';
+import {appConfig} from 'appConfig';
+/*
+ * Keycloak modal is watching session identity, redirect to login screen
+ * and call logout when session is cleared
+ */
+angular.module('superdesk.core.auth.keycloak', []).directive('sdKeycloakModal', [
```
petrjasek: actually why is this directive? seems like you don't need any element, do you?

scripts/core/superdesk-api.d.ts
```diff
@@ -903,6 +903,8 @@ declare module 'superdesk-api' {
         saml_auth: any;
         google_auth: any;
         saml_label: any;
+        oidc_auth: any;
+        keycloak_config: any;
```
petrjasek: not sure why we need both settings? wouldn't just `keycload_config` do?

scripts/core/auth/keycloak-service.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Keycloak modal is watching session identity, redirect to login screen
+ * and call logout when session is cleared
+ */
+angular.module('superdesk.core.auth.keycloak', []).service('keycloak', [
+    'session',
+    'auth',
+    '$rootScope',
+    function(session, auth, $rootScope) {
+        this.keycloakAuth = () => {
+            $rootScope.keycloak = Keycloak(appConfig.keycloak_config);
```
petrjasek: not sure why to use `$rootScope` here, you could just use local variable:
```js
function (session, auth, $rootScope) {
    let keycloak;
  
    this.keycloakAuth = () => {
        keycloak = ...;
    };
}
```

scripts/core/superdesk-api.d.ts
idsvn: It's for consistency, because other auth methods have variable format `method`_auth with boolean type to check if it's enabled so we only do the same for OIDC. Should we move `oidc_auth` to `keycloak_config`?

petrjasek: makes sense to keep it consistent :+1:

idsvn:
We have finished moving Keycloak from belga to core, sorry for late update, there are some issues on the clients that we have not addressed before, could you give us your insight on these:
- We still show email/password login screen before redirect to Keycloak (both login and logout) so sometime if it's happened too fast, it would cause [flashing screen](https://user-images.githubusercontent.com/45254285/71951266-15354e00-320d-11ea-8794-ff451f5142a7.gif) 
- If you log in via `/#/workspace` (or click workspace then log out), the [hamburger button will no longer work](https://user-images.githubusercontent.com/45254285/71951458-b2908200-320d-11ea-8a96-6c7ae3029da5.gif), we wonder if it's because Keycloak append param to browser url 
e.g. `http://localhost:9000/#/workspace&state=d0e4d347-ad30-4439-9bd1-ee5ca2b4e7c1&session_state=6a20ade0-f7b8-46a6-850e-334d04dab7e8&code=d1217351-9b20-41b7-9580-ae741713e361.6a20ade0-f7b8-46a6-850e-334d04dab7e8.987248c3-011e-4231-ae95-b49a9981127c`

idsvn:
Hi @tomaskikutis @petrjasek We fixed issue redirect when re-login after logged out and upgraded Keycloak to `10.0.2` to address `Samesite` cookie change in Chrome (see [Keycloak changelog](https://www.keycloak.org/docs/latest/release_notes/#keycloak-8-0-2) and [Chromium update](https://www.chromium.org/updates/same-site)).
Latest version is `11.0.0` but we encounter [this bug](https://stackoverflow.com/questions/63111877/keycloak-with-angular-and-spring-error-get-http-localhost8180-auth-realms-st) so have to downgrade it to `10.0.2`
For details how to setup Keycloak, please check our comment in the ticket.

ride90:
Hi @idsvn, can you make all test pass for this pr?

ride90:
@idsvn do I need to switch/configure something on a client side or only server settings?
Currently  I have a keycloak server running on my local, realm, users and so on is configured, but it does not work. 
I see client sends requests to `api/auth_db`
Here is my server config:
```
OIDC_ENABLED = True
OIDC_ISSUER='http://localhost:8080/auth/realms/SUPERDESK'
OIDC_SERVER_CLIENT='superdesk-server'
OIDC_SERVER_CLIENT_SECRET='b7bb2c61-2e3a-47d0-b0c3-5cede34f2d46'
OIDC_CLIENT_REDIRECT_URI='http://localhost:9000'
OIDC_WEB_CLIENT='web-client'
```

idsvn:
> @idsvn do I need to switch/configure something on a client side or only server settings?

It needs only server settings, I think you are missing `SECRET_KEY`, it requires both `OIDC_ENABLED` and `SECRET_KEY` ([link](https://github.com/superdesk/superdesk-core/blob/ce933bb4d345a3a29c704cab9b3837a367a13548/apps/auth/oidc/__init__.py#L18)), I also updated instruction on the ticket to include `SECRET_KEY`

Maybe it's better to log warning when `OIDC_ENABLED` but `SECRET_KEY` is missing?

ride90:
Hi @idsvn , thanks for updating a docs.

> Maybe it's better to log warning when OIDC_ENABLED is set to True but SECRET_KEY is missing?

Yes, makes sense

ride90 (approved):

tomaskikutis (approved):

PR #3217, opened 2019-11-18
Prevent user from adding special characters in slugs. [SDCP-83]

GyanP:
invalid-slug

### DIFF
--- a/scripts/apps/authoring/authoring/ValidateCharacters.tsx
+++ b/scripts/apps/authoring/authoring/ValidateCharacters.tsx
@@ -16,7 +16,7 @@ export class ValidateCharacters extends React.PureComponent<IProps> {
             return null;
         }
 
-        if (appConfig != null && appConfig.disallowed_characters != null) {
+        if (appConfig?.disallowed_characters != null) {
             const invalidCharString = appConfig.disallowed_characters.filter((char) => item.includes(char)).join(', ');
 
             if (invalidCharString.length > 0) {
--- a/scripts/apps/authoring/authoring/ValidateCharacters.tsx
+++ b/scripts/apps/authoring/authoring/ValidateCharacters.tsx
@@ -1,35 +1,31 @@
 import React from 'react';
-import ng from 'core/services/ng';
-import {gettext} from 'core/utils';
+import {gettextPlural} from 'core/utils';
+import {appConfig} from 'appConfig';
 
 interface IProps {
     item: string;
     field: string;
-    onChange: (field: string, isError: boolean) => void;
 }
 
 export class ValidateCharacters extends React.PureComponent<IProps> {
     render() {
-        const {item, field, onChange} = this.props;
-        const config = ng.get('$rootScope').config;
+        const {item, field} = this.props;
         let invalidCharError = null;
 
         if (item == null || field == null) {
             return null;
         }
 
-        if (config != null && config.disallowedCharacters != null) {
-            const disallowedCharacters = config.disallowedCharacters.split('');
-            const invalidCharString = disallowedCharacters.filter((char) => item.includes(char)).join(', ');
+        if (appConfig != null && appConfig.disallowed_characters != null) {
+            const invalidCharString = appConfig.disallowed_characters.filter((char) => item.includes(char)).join(', ');
 
             if (invalidCharString.length > 0) {
-                onChange(field, true);
-                invalidCharError = gettext(
+                invalidCharError = gettextPlural(
+                    invalidCharString.length,
                     'Character {{chars}} not allowed in the {{field}}.',
+                    'Characters {{chars}} not allowed in the {{field}}.',
                     {chars: invalidCharString, field: field},
                 );
-            } else {
-                onChange(field, false);
             }
         }
 
--- a/scripts/apps/authoring/authoring/ValidateCharacters.tsx
+++ b/scripts/apps/authoring/authoring/ValidateCharacters.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import ng from 'core/services/ng';
+import {gettext} from 'core/utils';
+
+interface IProps {
+    item: string;
+    field: string;
+    onChange: (field: string, isError: boolean) => void;
+}
+
+export class ValidateCharacters extends React.PureComponent<IProps> {
+    render() {
+        const {item, field, onChange} = this.props;
+        const config = ng.get('$rootScope').config;
+        let invalidCharError = null;
+
+        if (item == null || field == null) {
+            return null;
+        }
+
+        if (config != null && config.disallowedCharacters != null) {
+            const disallowedCharacters = config.disallowedCharacters.split('');
+            const invalidCharString = disallowedCharacters.filter((char) => item.includes(char)).join(', ');
+
+            if (invalidCharString.length > 0) {
+                onChange(field, true);
+                invalidCharError = gettext(
+                    'Character {{chars}} not allowed in the {{field}}.',
+                    {chars: invalidCharString, field: field},
+                );
+            } else {
+                onChange(field, false);
+            }
+        }
+
+        return invalidCharError != null ? <span>{invalidCharError}</span> : null;
+    }
+}
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
@@ -112,6 +112,7 @@ export function AuthoringDirective(
             $scope.mediaFieldVersions = {};
             $scope.refreshTrigger = 0;
             $scope.isPreview = false;
+            $scope.fieldsWithInvalidChar = [];
 
             $scope.$watch('origItem', (newValue, oldValue) => {
                 $scope.itemActions = null;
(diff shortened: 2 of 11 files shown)


scripts/apps/authoring/styles/themes.scss
```diff
@@ -511,6 +511,14 @@ body, html {
             }
         }
     }
+    .disallowed-char-error {
```
petrjasek: I'm not sure why we would need special styles for char error, but I'm pretty sure we don't need it 2x.. :wink:

GyanP: At first, I thought to use the `char-count error` class to make it look like the one shown in the above attachment but later on, decided to make a new one. Using it 2X because slugline could be in `authoring-header` as well as in the `content-fields`.

petrjasek: sure, but you don't have to define the class inside any parent to make it work in both

tomaskikutis:
Why not implement this as a feature for any text field instead of it being specific to slugline? It doesn't seem it would be that hard to do.

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -564,6 +565,13 @@ export function AuthoringDirective(
              * in $scope.
              */
             $scope.publish = function() {
+                if ($scope.fieldsWithInvalidChar.length > 0) {
```
tomaskikutis: Can you handle publishing validation on the backend?

scripts/apps/authoring/authoring/ValidateCharacters.tsx
```diff
@@ -0,0 +1,38 @@
...
+    render() {
+        const {item, field, onChange} = this.props;
+        const config = ng.get('$rootScope').config;
+        let invalidCharError = null;
+
+        if (item == null || field == null) {
+            return null;
+        }
+
+        if (config != null && config.disallowedCharacters != null) {
```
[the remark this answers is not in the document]
karelpetrak: Hi @tomaskikutis disallowed characters for a field are global, that's why we wanted to have them in one place to avoid the need to update all content profiles when you decide to update the list of characters.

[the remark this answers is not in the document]
tomaskikutis: :+1: let's keep it in the config then

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
GyanP: And if we keep it like the way it is we can't perform validations in the backend as we don't know which field to validate i.e. which field has this character validation on.

tomaskikutis: That's a good point. Let's then add a checkbox to the content profile, whether the field has to be validated or not and keep the validation pattern common. You can put the pattern in the server configs and it will get downloaded to the client.

scripts/apps/authoring/authoring/ValidateCharacters.tsx
```diff
@@ -0,0 +1,38 @@
...
+            const disallowedCharacters = config.disallowedCharacters.split('');
+            const invalidCharString = disallowedCharacters.filter((char) => item.includes(char)).join(', ');
```
petrjasek: I wonder why it should do the validation in the render of a component, shouldn't it rather be done in the authoring directive and here only render the error messages if any?

GyanP: I don't think we need to do that, it's just like we do for char-count or line count.

scripts/apps/authoring/authoring/ValidateCharacters.tsx
```diff
@@ -1,35 +1,31 @@
...
         let invalidCharError = null;
 
         if (item == null || field == null) {
             return null;
         }
 
-        if (config != null && config.disallowedCharacters != null) {
-            const disallowedCharacters = config.disallowedCharacters.split('');
-            const invalidCharString = disallowedCharacters.filter((char) => item.includes(char)).join(', ');
+        if (appConfig != null && appConfig.disallowed_characters != null) {
```
tomaskikutis: Since TypeScript 3.7 it's possible to use nullish coalescing for deep access:

```if (appConfig?.disallowed_characters != null) {...}```

File: scripts/core/superdesk-api.d.ts [hunk 6]
```diff
@@ -879,6 +879,7 @@ declare module 'superdesk-api' {
         saml_auth: any;
         google_auth: any;
         saml_label: any;
+        disallowed_characters: any;
```
tomaskikutis: Always define proper types when writing new code. There are still many places where it's set to `any`. It's because we did it for most properties when migrating and didn't update it yet. We shouldn't use `any` for new code unless there are incompatibilities when integrating something will older code.

scripts/apps/authoring/views/article-edit.html
```diff
@@ -7,6 +7,10 @@
 
     <label class="field__label">{{label('headline')}}</label>
     <span ng-if="schema.headline.maxlength" sd-character-count data-item="item.headline" data-html="true" data-limit="schema.headline.maxlength"></span>
+    <sd-validate-characters ng-if="schema.headline.validate_characters"
+        class="disallowed-char-error"
```
tomaskikutis: Can you define the class name inside the React component?

scripts/apps/workspace/content/views/schema-editor.html
```diff
@@ -62,6 +62,11 @@
                         <div sd-check ng-model="model.editor[id].cleanPastedHTML"></div>
                     </div>
 
+                    <div class="field" ng-if="canValidateChars && model.schema[id].type === 'string' || fields[id].field_type === 'text'">
+                        <label translate>Validate Chars</label>
```
tomaskikutis: Don't use abbreviations for text which will be displayed to the user.

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
```diff
@@ -9,6 +10,7 @@ interface IScope extends ng.IScope {
     formattingOptions: Array<string>;
     schemaKeysOrdering: any;
     schemaKeysDisabled: any;
+    canValidateChars: boolean;
```
tomaskikutis: I suggest calling the variable `characterValidationEnabled`. It's technically possible to validate characters even if it's not enabled so it's clearer to refer to it being enabled rather than being possible in the variable name.

scripts/apps/authoring/views/article-edit.html
GyanP: I did that first, but it due to this style was breaking so removed that.
validate_char

But this should look like this.
valid_fromat

tomaskikutis: :+1: keep it as is then

scripts/apps/authoring/authoring/ValidateCharacters.tsx
tomaskikutis: FYI it's actually called optional chaining. I mixed up the names.

scripts/apps/workspace/content/controllers/ContentProfileFields.ts
```diff
@@ -118,7 +118,7 @@ export default function ContentProfileFields($scope, content, vocabularies, meta
                 this.sections[section].enabled.push(key);
                 this.model.editor[key].order = index + 1; // keep order in sync
                 // keep required value of editor fields in sync with the schema fields SDNTB-615
-                this.model.editor[key].required = this.model.schema[key].required;
+                this.model.editor[key].required = this.model?.schema[key]?.required ?? false;
```
tomaskikutis: I see `this.model.schema` used to be accessed unconditionally. Can `this.model` be null? Optional chaining is a great feature, but we should not overuse it.

GyanP: I'm not sure about `this.model.schema` but `key` is definitely missing in some cases.

tomaskikutis: Is there a reason to be missing? It doesn't look like it to me. Can you debug and see if you can find why it's happening?

GyanP: Ok, so the issue is with the content profiles that are prepopulated, for ex: belga_text CP has 38 fields in the `editor` and 29 fields in the `schema` field. Although no such case with the ones created from the UI.

tomaskikutis: > belga_text CP has 38 fields in the editor and 29 fields in the schema field.

Can you investigate it further to see what's the difference between `schema` and `editor` and whether we should support having a different number of fields inside them?

With my limited understanding, it looks like the fields in `schema` and `editor` should be identical. If that ends up being the case, update the prepopulation data and throw an error on the client if the key is missing.

GyanP: There are certain CVs that will look like a separate field in the editor but their value will be stored inside the different fields (ex:`news_service` inside `subject`).  So in the content profile, `news_service` will be declared as a separate field inside the editor, and will be declared inside `subject` in case of `schema`. So, we will have a different number of fields inside the editor and schema.

belga_text

tomaskikutis: I see. Let's keep the key optional then and add a comment explaining it in the code.

scripts/core/superdesk-api.d.ts
```diff
@@ -889,6 +889,7 @@ declare module 'superdesk-api' {
     export interface ISuperdeskGlobalConfig {
         // FROM SERVER
         default_language: string;
+        disallowed_characters: Array<string>;
```
tomaskikutis: Can you add a comment that it only applies to `slugline`? It's a little bit configurable, but still hardcoded.

PR #3222, opened 2019-11-20
Allow duplicated text annotations

pablopunk:
SDESK-4836

Fixes an issue where the user wasn't allowed to create an annotation if another one existed in the database for the same text

### DIFF
--- a/scripts/core/editor3/components/annotations/AnnotationInput.tsx
+++ b/scripts/core/editor3/components/annotations/AnnotationInput.tsx
@@ -119,7 +119,7 @@ class AnnotationInputBody extends React.Component<IProps, IState> {
      * @description onSubmit is called when the user clicks the Submit button in the UI.
      * Consequently, it calls the `onSubmit` prop, passing it the value of the text input.
      */
-    onSubmit() {
+    onSubmit(_event, reusingAnnotationFromExtensions = false) {
         const {body, type} = this.state;
         const _hidePopups = this.props.hidePopups;
         const {highlightId} = this.props.data;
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/components/annotations/AnnotationInput.tsx
```diff
@@ -129,13 +129,15 @@ class AnnotationInputBody extends React.Component<IProps, IState> {
...
 
             if (highlightId === undefined) {
-                this.annotationInputTabsFromExtensions.forEach((tab) => {
-                    tab.onAnnotationCreate(
-                        this.props.language,
-                        this.getAnnotatedText(),
-                        stateToHTML(convertFromRaw(body)),
-                    );
-                });
+                if (!reusingAnnotationFromExtensions) {
```
tomaskikutis: Can you leave `AnnotationInput.tsx` unchanged and handle it all inside the extension?

I'm not getting what `reusingAnnotationFromExtensions` should do. What do you mean by reusing annotations?

pablopunk: At first I tried to do it inside the extension, but in order to check if the annotation already exists I need to query the server for both the text and annotation description ([which is html](https://github.com/superdesk/superdesk-client-core/blob/4ca4d5a62c84b5120df67ec89ee6ce917cafcb53/scripts/core/ui/components/generic-form/generate-filter-for-server.tsx#L20-L23)).

So now, I'm passing the argument `reusingAnnotationFromExtensions` to the `onSubmit` function, but I'm only doing that [when we render the extension tabs](https://github.com/superdesk/superdesk-client-core/blob/4ca4d5a62c84b5120df67ec89ee6ce917cafcb53/scripts/core/ui/components/generic-form/generate-filter-for-server.tsx#L20-L23), so it makes sure that whenever we reuse an annotation (clicking on one on the existing annotations tab) we don't create a duplicate of that one.

tomaskikutis: > At first I tried to do it inside the extension, but in order to check if the annotation already exists I need to query the server for both the text and annotation description

This approach sounds better to me. An extra request won't hurt and you can query by HTML using network helpers like `httpRequestJsonLocal` which aren't in any way constrained by `CrudManager`. What do you think?

pablopunk: no I meant I'd have to include the html text in the current request [we already do](https://github.com/superdesk/superdesk-client-core/blob/91345e95ea063c1f18b769421217a7ff06ea53ff/scripts/extensions/annotationsLibrary/src/extension.tsx#L18), and that was throwing [this error](https://github.com/superdesk/superdesk-client-core/blob/4ca4d5a62c84b5120df67ec89ee6ce917cafcb53/scripts/core/ui/components/generic-form/generate-filter-for-server.tsx#L20-L23) so I guessed it was a bad practice to filter by html.

PR #3223, opened 2019-11-21
implement autocomplete for slugline

petrjasek:
SDCP-81


scripts/core/superdesk-api.d.ts
```diff
@@ -903,6 +903,7 @@ declare module 'superdesk-api' {
         saml_auth: any;
         google_auth: any;
         saml_label: any;
+        slugline_autocomplete: boolean;
```
tomaskikutis: Features should not be associated with specific field types. If you want to display whitespace for `body_html` field, you don't tie it with `body_html` but implement it in editor3 and all fields can use it. Same here. The implementation of autocomplete for `slugline` is the same as it would be for any other plain text field.

Can you make it generic & configurable for any plain text field?

scripts/apps/authoring/authoring/directives/sd-static-autocomplete.ts
```diff
@@ -0,0 +1,125 @@
+import {debounce} from 'lodash';
```
tomaskikutis: > `sd-static-autocomplete.ts`

How is it static? How would a dynamic autocomplete work?

scripts/apps/authoring/authoring/directives/sd-static-autocomplete.ts
```diff
@@ -0,0 +1,125 @@
...
+    value: string;
+    available: Array<string>;
+    suggestions: Array<string>;
+    activeSuggestion?: number;
+    select: (suggestion: string) => void;
+    onSelect: ({suggestion}: {suggestion: string}) => void;
+    suggested?: string;
+}
+
+export const sdStaticAutocompleteDirective = () => ({
```
tomaskikutis: Why not add text autocomplete field to UI framework instead?

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -1215,6 +1215,14 @@ export function AuthoringDirective(
             };
 
             $scope.refresh = () => $scope.refreshTrigger++;
+
+            $scope.sluglines = [];
+            if (appConfig.slugline_autocomplete) {
+                api.query('slugline_autocomplete').then((response) => {
```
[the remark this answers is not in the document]
petrjasek: it's not

scripts/apps/authoring/authoring/directives/sd-static-autocomplete.ts
petrjasek: it would query server or something on every value change, here directive works with the list it gets upfront

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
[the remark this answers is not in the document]
tomaskikutis: How come? Isn't API limited to 200 by default?

scripts/apps/authoring/authoring/directives/sd-static-autocomplete.ts
tomaskikutis: If it's being autocompleted from published sluglines(or is it from another place?) it will be thousands of items in which case it would start lagging at some point. Wouldn't it be better to use a server filtered autocomplete instead?

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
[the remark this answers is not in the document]
petrjasek: it's not listing items from db query, but using db `.distinct()` method which gets the list and has no limit

scripts/apps/authoring/authoring/directives/sd-static-autocomplete.ts
petrjasek: there is a server config to set how long in history it should look for values. usually it will check for last 1-2 days

scripts/core/superdesk-api.d.ts
petrjasek: would there be a usecase for another field @karelpetrak ?

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -1215,6 +1215,14 @@ export function AuthoringDirective(
...
+                    $scope.sluglines = response._items.map((_item) => _item.slugline);
+                    $scope.sluglines.sort();
```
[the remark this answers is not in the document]
petrjasek: good point, will do it on server

scripts/core/superdesk-api.d.ts
tomaskikutis: It's not only about future requirements but about code quality too. If there aren't functional differences it shouldn't be a problem to make it work for all text fields. If it is a problem, it's a good opportunity to discover and improve architectural issues.

scripts/apps/authoring/authoring/directives/sd-static-autocomplete.ts
petrjasek: I'm not sure actually where this should be done, but it should eventually work for react components. just wanted to make it work with current slugline field which has some integration with tansa spellcheck

PR #3224, opened 2019-11-22
reload video in preview on item change

petrjasek:
if the video was started changing it's source didn't make
it reload, so force reload on item change.

SDCP-66

### DIFF
--- a/scripts/core/ui/components/video.tsx
+++ b/scripts/core/ui/components/video.tsx
@@ -22,7 +22,7 @@ export class VideoComponent extends React.PureComponent<IProps> {
             .filter((rendition) => rendition.mimetype != null && rendition.mimetype.startsWith('video'));
 
         return (
-            // using key to force reload video on item change
+            // using key to force reload video on selecting different item for preview
             <video key={item.guid} controls preload="metadata" poster={poster}>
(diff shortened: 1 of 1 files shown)


scripts/core/ui/components/video.tsx
```diff
@@ -22,7 +22,8 @@ export class VideoComponent extends React.PureComponent<IProps> {
             .filter((rendition) => rendition.mimetype != null && rendition.mimetype.startsWith('video'));
 
         return (
-            <video controls preload="metadata" poster={poster}>
+            // using key to force reload video on item change
+            <video key={item.guid} controls preload="metadata" poster={poster}>
```
tomaskikutis: Why is the key referring to `item.guid`? Shouldn't GUID be the same after modifying the entity? Why not refer to last updated date instead?

petrjasek: because I need it to reload when we switch items in preview, there is no video editing possible on video items so it won't change there.

tomaskikutis: Got it. What I didn't understand was why would the item change since you aren't allowed to edit the item in the preview. Can you clarify the comment so it's clear that another item is selected for preview and it's not the current one that changes?

PR #3226, opened 2019-11-26
Fix string type dropdown

pablopunk:
SDESK-4815

Server was throwing errors for some fields like 'usageterms' because we were saving the object instead of the string.

### DIFF
--- a/scripts/apps/authoring/metadata/metadata.ts
+++ b/scripts/apps/authoring/metadata/metadata.ts
@@ -366,6 +366,7 @@ function MetaDropdownDirective($filter) {
                 if (item) {
                     if (!scope.multiInputFields.includes(scope.field)) {
                         // single input field
+                        // we use 'name' attribute for string fields and the whole object for other fields
                         fieldObject[scope.field] = item.name || item;
                     } else if (scope.cv && scope.cv._id != null) {
                         // if there is cv as well as field, store cv._id as scheme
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -366,7 +366,7 @@ function MetaDropdownDirective($filter) {
                 if (item) {
                     if (!scope.multiInputFields.includes(scope.field)) {
                         // single input field
-                        fieldObject[scope.field] = item;
+                        fieldObject[scope.field] = item.name || item;
```
tomaskikutis: If saving an object is a problem, why do you keep it as a fallback?

pablopunk: That's the case for `usageterms` but for other items they can be an object also, and it's displayed anyway in the dropdown:

https://github.com/superdesk/superdesk-client-core/blob/e7410b0c826db4537d04cf816034c36456a6dbcc/scripts/apps/authoring/metadata/views/metadata-dropdown.html#L4

tomaskikutis: I don't like that we are handling cases specific to some data structures while there's no way to see what are the possible data structures by following the code.

Can you add a comment that it's specific to usageterms(or any other fields)?

pablopunk: it's not really specific for usageterms, it's just single input fields (I commented that already).

Yeah I agree that component it's pretty cluttered with several render logics that are not clear at all

tomaskikutis: The comment is referring to all the code inside the `if` branch. What I was referring to, is to comment on when `item.name` or `item` would be used.

PR #3228, opened 2019-11-26
fix bug when video in editor was not playable at some point

petrjasek:
the actuall fix is to use key on video element because otherwise
it does not force reload.

SDESK-4817


petrjasek:
the fix is in https://github.com/superdesk/superdesk-client-core/pull/3224 so it will get there only when it's merged

PR #3230, opened 2019-11-27
Add support for 'application/x-mpegurl' HLS video stream for the built-in video player.

ride90:
This PR adds the ability to render an enhanced `video` element for `VideoComponent` if item has an `application/x-mpegurl` mime-type. 
In most cases it will stream video in [m3u8](https://en.wikipedia.org/wiki/M3U)

### DIFF
--- a/scripts/core/ui/components/video-hls.tsx
+++ b/scripts/core/ui/components/video-hls.tsx
@@ -1,12 +1,10 @@
 import React from 'react';
 import Hls from 'hls.js';
 
-
 interface IProps {
     poster: string;
-    streamUrl: string
+    streamUrl: string;
 }
-
 
 export class HLSVideoComponent extends React.PureComponent<IProps> {
     videoElement: HTMLElement;
--- a/scripts/core/ui/components/video-hls.tsx
+++ b/scripts/core/ui/components/video-hls.tsx
@@ -1,0 +1,52 @@
+import React from 'react';
+import Hls from 'hls.js';
+
+
+interface IProps {
+    poster: string;
+    streamUrl: string
+}
+
+
+export class HLSVideoComponent extends React.PureComponent<IProps> {
+    videoElement: HTMLElement;
+    hls: Hls;
+
+    _initHLS = () => {
+        if (Hls.isSupported() && this.videoElement) {
+            this.hls = new Hls();
+            this.hls.loadSource(this.props.streamUrl);
+            this.hls.attachMedia(this.videoElement);
+        }
+    };
+
+    _destroyHLS = () => {
+        this.hls.stopLoad();
+        this.hls.destroy();
+    };
+
+    componentDidUpdate(prevProps: Readonly<IProps>, prevState: Readonly<{}>, snapshot?: any): void {
+        if (this.props.streamUrl !== prevProps.streamUrl) {
+            this._destroyHLS();
+            this._initHLS();
+        }
+    }
+
+    componentDidMount(): void {
+        this._initHLS();
+    }
+
+    componentWillUnmount(): void {
+        this._destroyHLS();
+    }
+
+    render() {
+        return (
+            <video controls
+                preload="metadata"
+                ref={(el) => this.videoElement = el}
+                poster={this.props.poster}
+            />
+        );
+    }
+}
--- a/scripts/core/ui/components/video-hls.tsx
+++ b/scripts/core/ui/components/video-hls.tsx
@@ -8,6 +8,10 @@ interface IProps {
     height?: string;
 }
 
+/**
+ * HLSVideoComponent is rendered by VideoComponent when item's rendition has an 'application/x-mpegurl' mimetype.
+ * NOTE: It can handle only one stream url.
+ */
 export class HLSVideoComponent extends React.PureComponent<IProps> {
     videoElement: HTMLElement;
     hls: Hls;
--- a/scripts/core/ui/components/video-hls.tsx
+++ b/scripts/core/ui/components/video-hls.tsx
@@ -4,6 +4,8 @@ import Hls from 'hls.js';
 interface IProps {
     poster: string;
     streamUrl: string;
+    width?: string;
+    height?: string;
 }
 
 export class HLSVideoComponent extends React.PureComponent<IProps> {
@@ -39,6 +41,8 @@ export class HLSVideoComponent extends React.PureComponent<IProps> {
                 preload="metadata"
                 ref={(el) => this.videoElement = el}
                 poster={this.props.poster}
+                width={this.props.width}
+                height={this.props.height}
             />
         );
     }
--- a/scripts/core/ui/components/video-hls.tsx
+++ b/scripts/core/ui/components/video-hls.tsx
@@ -12,32 +12,27 @@ export class HLSVideoComponent extends React.PureComponent<IProps> {
     videoElement: HTMLElement;
     hls: Hls;
 
-    _initHLS = () => {
+    private initHLS = () => {
         if (Hls.isSupported() && this.videoElement) {
             this.hls = new Hls();
             this.hls.loadSource(this.props.streamUrl);
             this.hls.attachMedia(this.videoElement);
         }
-    };
+    }
 
-    _destroyHLS = () => {
-        this.hls.stopLoad();
-        this.hls.destroy();
-    };
-
-    componentDidUpdate(prevProps: Readonly<IProps>, prevState: Readonly<{}>, snapshot?: any): void {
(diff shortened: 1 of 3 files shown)


scripts/core/ui/components/video-hls.tsx
```diff
@@ -0,0 +1,53 @@
...
+    streamUrl: string
+}
+
+
+export class HLSVideoComponent extends React.PureComponent<IProps> {
+
+    videoElement: HTMLElement;
+    hls: Hls;
+
+    _initHLS = () => {
```
tomaskikutis: If you want private members you can use `private initHLS() {...}` and the compiler will not allow accessing it outside. If you apply the `key` trick I suggested in another comment, these methods will only be called from one place so you may also inline the code in `componentDidMount` / `componentWillUnmount`

scripts/core/ui/components/video.tsx
```diff
@@ -19,12 +21,18 @@ export class VideoComponent extends React.PureComponent<IProps> {
         }
 
         const videoRenditions = Object.values(item.renditions)
-            .filter((rendition) => rendition.mimetype != null && rendition.mimetype.startsWith('video'));
+            .filter(
+                (rend) => rend.mimetype != null && (rend.mimetype.startsWith('video') || rend.mimetype.startsWith('application/x-mpegurl'))
+            );
+
+        if (videoRenditions.length > 0 && videoRenditions[0].mimetype === 'application/x-mpegurl') {
+            return <HLSVideoComponent poster={poster} streamUrl={videoRenditions[0].href}/>
```
tomaskikutis: If you add a `key` when rendering a component and that key changes, the component will fully remount and trigger all the usual lifecycle methods including `componentWillUnmount` and `componentDidMount`. If you set `videoRenditions[0].href` as a key, you can get rid off `HLSVideoComponent.componentDidUpdate` method.

tomaskikutis:
Don't rebase branches after submitting a PR. It rewrites the git history and I can't see the exact changes you made since the previous review.

ride90:
@tomaskikutis , I will not do it since you've started a review. 
This pr was marked as WIP, so that's why I was using rebase.

tomaskikutis:
Great, thanks. That's my bad for reviewing a WIP PR. By the way, you can run `npm run test` locally to check failing tests/lint issues.

scripts/core/ui/components/video.tsx
```diff
@@ -19,12 +20,23 @@ export class VideoComponent extends React.PureComponent<IProps> {
         }
 
         const videoRenditions = Object.values(item.renditions)
-            .filter((rendition) => rendition.mimetype != null && rendition.mimetype.startsWith('video'));
+            .filter(
+                (rend) => rend.mimetype != null && (rend.mimetype.startsWith('video') ||
+                                                    rend.mimetype.startsWith('application/x-mpegurl')),
+            );
+
+        if (videoRenditions.length > 0 && videoRenditions[0].mimetype === 'application/x-mpegurl') {
```
tomaskikutis: > videoRenditions[0].mimetype === 'application/x-mpegurl'

Why are you checking only the first one? Why not map over all renditions and return the player depending on the mime type?

```
renditions.map((rendition) => {
    if (rendition.mimetype === 'application/x-mpegurl') {
        return (
            <HLSVideoComponent ... />
        );
    } else {
        return (
            <video ... />;
        );
    }
})
```

ride90: The logic of how `HLSVideoComponent` works is different from the default `video`, there is no `source`(s) and it can handle only one stream URL at a period of time. Most probably API will return only one rendition with `application/x-mpegurl` and the rest are images.  For example, Talpa's API returns it like that.

ride90: @tomaskikutis later I'll create a fake PR to trigger CI to build a test instance, so you can test it with a proper backend.

ride90: @tomaskikutis actually, I think this part must be improved, theoretically item's renditions might contain mp4 and m3u8 and in this case, we will show whether `default` player or `HLSVideoComponent ` depends on index order which is wrong.

tomaskikutis: > it can handle only one stream URL at a period of time

I see. That's an important detail. How about handling it in the code so it's clear? You can filter out the renditions so it includes only one `'application/x-mpegurl'` and proceed with mapping it as I suggested previously.

```
let hlsVideoCount = 0;

renditions
    .filter((rendition) => {
        if (rendition.mimetype === 'application/x-mpegurl') {
            hlsVideoCount++;
            
            if (hlsVideoCount === 1) {
                return true;
            } else {
                logger.error(new Error('HLSVideoComponent can handle only one stream URL at a period of time.'));
                return false;
            }
        } else {
            true;
        }
    })
    .map((rendition) => {
        if (rendition.mimetype === 'application/x-mpegurl') {
            return (
                <HLSVideoComponent ... />
            );
        } else {
            return (
                <video ... />;
            );
        }
    })
```

scripts/core/ui/components/video.tsx
```diff
@@ -20,8 +26,23 @@ export class VideoComponent extends React.PureComponent<IProps> {
...
 
-        const videoRenditions = Object.values(item.renditions)
-            .filter((rendition) => rendition.mimetype != null && rendition.mimetype.startsWith('video'));
+        let videoRenditions: Array<IVideoRenditionItem> = [];
+
+        for (const rend of Object.values(item.renditions)) {
+            if (rend.mimetype == null) {
+                continue;
+            } else if (rend.mimetype.toLowerCase() === 'application/x-mpegurl') {
+                return <HLSVideoComponent
```
tomaskikutis: > HLSVideoComponent can handle only one stream URL at a period of time

Can you add this as a comment here and in the component definition? This is a very important piece of knowledge that's not mentioned anywhere in the code.

scripts/core/ui/components/video.tsx
```diff
@@ -20,8 +26,23 @@ export class VideoComponent extends React.PureComponent<IProps> {
...
+                    poster={poster}
+                    streamUrl={rend.href}
+                    key={rend.href}
+                    width={this.props.width}
+                    height={this.props.height}
```
petrjasek: I think it should get width/height from rendition, we might use some preview and not the original size which would be set on the item

ride90: `HLSVideoComponent` is a custom render for `VideoComponent`, in this case, width and height determined by parent (VideoComponent) as for `video`.

petrjasek: yep right, my bad

scripts/core/ui/components/video.tsx
```diff
@@ -32,6 +32,7 @@ export class VideoComponent extends React.PureComponent<IProps> {
             if (rend.mimetype == null) {
                 continue;
             } else if (rend.mimetype.toLowerCase() === 'application/x-mpegurl') {
+                // HLSVideoComponent can handle only one stream url
```
tomaskikutis: Sorry to bother you, but could you clarify the comment to state that "Only one instance of `HLSVideoComponent` can be rendered in the page at any given moment.`?

When you refer to the component being able to only handle one stream URL it wouldn't be incorrect to assume that you can render 5 instances and handle 5 streams this way.

ride90: You can render 5 HLSVideoComponent and theoretically, they will work simultaneously. I've tried with 2. 
When I said `it can handle only one stream URL at a period of time` I did not know that it can actually do this :) But you can provide only one source url to the component which is documented. 
P.S. You are not bothering me :) It's good to have such moments to be clear for other devs.

tomaskikutis: If multiple `HLSVideoComponent`s can work simultaneously, will you then remove `videoRenditions` variable and implement the `if` for deciding which video component to use at line 57 so multiple instances of `HLSVideoComponent` could be used at once?

ride90: I'm not sure if we actually need this. In this case, it will render more than one `HLSVideoComponents`. 
If video is not a stream we render only one player even if we have multiple renditions. 
The difference is that for default video (not stream) we don't use `<video src=".."/>` we use 
```
<video>
<source ..>
<source ..>
</vidoe>
```
which at the end renders only one instance of player.

tomaskikutis: Can you add a comment that `VideoComponent` is only used when `item.type === 'video'`?

I thought it could be used to display multiple different videos in an article, that's why I was suggesting the mapping approach.

ride90: done

PR #3233, opened 2019-11-28
Use '_fetchable' flag to control if item is fetchable or not.

ride90:
Tested with body html, feature media, and related item fields.

### DIFF
--- a/scripts/apps/ingest/index.ts
+++ b/scripts/apps/ingest/index.ts
@@ -136,15 +136,11 @@ angular.module('superdesk.apps.ingest', [
                 }],
                 filters: [{action: 'list', type: 'externalsource'}],
                 privileges: {fetch: 1},
-                additionalCondition: ['desks', 'item', function(desks, item) {
-                    if (item._fetchable !== undefined) {
-                        return isMediaEditable(item) && !_.isNil(desks.getCurrentDeskId());
-                    }
-                    // Fetching to 'personal' desk is not allowed
-                    return appConfig.features != null
-                        && appConfig.features.editFeaturedImage
-                        && !_.isNil(desks.getCurrentDeskId());
-                }],
+                additionalCondition: ['desks', 'item', (desks, item) =>
+                    isMediaEditable(item)
+                    // fetching to 'personal' desk is not allowed
+                    && !_.isNil(desks.getCurrentDeskId()),
+                ],
             })
 
             .activity('externalsource', {
@@ -154,15 +150,11 @@ angular.module('superdesk.apps.ingest', [
                 controller: ctrl.ExternalSourceController,
                 filters: [{action: 'list', type: 'externalsource', id: 'fetch-externalsource'}],
                 privileges: {fetch: 1},
-                additionalCondition: ['desks', 'item', function(desks, item) {
-                    if (item._fetchable !== undefined) {
-                        return isMediaEditable(item) && !_.isNil(desks.getCurrentDeskId());
-                    }
-                    // Fetching to 'personal' desk is not allowed
-                    return appConfig.features != null
-                        && appConfig.features.editFeaturedImage
-                        && !_.isNil(desks.getCurrentDeskId());
-                }],
+                additionalCondition: ['desks', 'item', (desks, item) =>
+                    isMediaEditable(item)
+                    // fetching to 'personal' desk is not allowed
+                    && !_.isNil(desks.getCurrentDeskId()),
+                ],
             });
 
         workspaceMenuProvider.item({
--- a/scripts/apps/ingest/index.ts
+++ b/scripts/apps/ingest/index.ts
@@ -135,7 +135,10 @@ angular.module('superdesk.apps.ingest', [
                 }],
                 filters: [{action: 'list', type: 'externalsource'}],
                 privileges: {fetch: 1},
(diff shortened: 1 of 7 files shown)


scripts/apps/ingest/index.ts
```diff
@@ -135,7 +135,10 @@ angular.module('superdesk.apps.ingest', [
                 }],
                 filters: [{action: 'list', type: 'externalsource'}],
                 privileges: {fetch: 1},
-                additionalCondition: ['desks', function(desks) {
+                additionalCondition: ['desks', 'item', function(desks, item) {
+                    if (item._fetchable !== undefined) {
+                        return item._fetchable;
```
petrjasek: I think it should still not allow fetching to personal desk

scripts/apps/ingest/index.ts
```diff
@@ -135,7 +136,10 @@ angular.module('superdesk.apps.ingest', [
                 }],
                 filters: [{action: 'list', type: 'externalsource'}],
                 privileges: {fetch: 1},
-                additionalCondition: ['desks', function(desks) {
+                additionalCondition: ['desks', 'item', function(desks, item) {
+                    if (item._fetchable !== undefined) {
+                        return isMediaEditable(item) && !_.isNil(desks.getCurrentDeskId());
```
petrjasek: I think you can change this whole function to be just this line, it checks `appConfig` features if `_fetchable` is not set on item

petrjasek: so it would be:
```ts
additionalCondition: ['desks', 'item', (desks, item) => isMediaEditable(item) && !_.isNil(desks.getCurrentDeskId()]
```

scripts/core/superdesk-api.d.ts
```diff
@@ -284,6 +284,7 @@ declare module 'superdesk-api' {
             archive?: boolean;
             externalsource: boolean;
         };
+        _fetchable?: Boolean
```
tomaskikutis: It should be `boolean` with lower-case `b`.

scripts/apps/ingest/index.ts
tomaskikutis: I agree. I'd also add a comment that desk is checked in order to prevent fetching to personal space. It's clear now, but might not be in the future.

ride90: I kept that comment.

ride90:
⚠️ don't merge it
I noticed that on the double click it tries to edit an item. This is something I need to handle as well.

scripts/apps/search/components/Item.tsx
```diff
@@ -198,6 +199,10 @@ export class Item extends React.Component<IProps, IState> {
     }
 
     dbClick(event) {
+        if (!isMediaEditable(this.props.item)) {
```
tomaskikutis: What happens if this function is not here?

ride90: Then it will be an attempt to fetch and edit the item. I think it makes sense to put it a little bit lower, to prevent `open` item on double click.

tomaskikutis: It'd be good to move it closer to the actual logic where fetching/editing actually happens.

ride90: Got it. I've put it before actually to `ItemList.dbClick`, but after, I somehow decided that  `Item.dbClick` is a better candidate for it. Can you explain to me why it's better to move it to `ItemList.dbClick`? In this case, It would be logical to move this condition too:
```
if (!this.props.item.gone) {
        this.props.onDbClick(this.props.item);
 }
```

tomaskikutis: > Can you explain to me why it's better to move it to `ItemList.dbClick` ?

Because if someone called `ItemList.dbClick` and forgot to do the validation, we'd have a bug. It's not too likely at the moment that someone will call it from another place, but when time passes, code gets renamed and restructured and it becomes not that unlikely anymore.

You should probably move the check not to the top of `ItemList.dbClick`, but where the fetching actually happens which is probably [here](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/search/components/ItemList.tsx#L242)

I agree `this.props.item.gone` check should probably be moved too.

scripts/core/superdesk-api.d.ts
```diff
@@ -284,6 +284,7 @@ declare module 'superdesk-api' {
             archive?: boolean;
             externalsource: boolean;
         };
+        _fetchable?: boolean;
```
tomaskikutis: What's the default? If it's not present in the payload can the item be fetched?

ride90: There is no default value.
If it's not in the payload, then a fetching condition is determined by editFeaturedImage config.

tomaskikutis: Either naming or logic doesn't make sense for me. If I have a text item, why would `editFeaturedImage` be used to determine if it can be fetched?

What does `editFeaturedImage` mean originally? Is it referring to editing the image itself like crops/rotation? Why would it be related to fetching?

ride90: Tomas, I think exactly the same :) 
`editFeaturedImage` imo must be renamed/removed and all that logic must be refactored.

PR #3234, opened 2019-11-28
Focus mark for user input on render

pablopunk:
SDBELGA-237


scripts/core/ui/components/SelectUser.tsx
```diff
@@ -72,6 +72,7 @@ export class SelectUser extends React.Component<IPropsSelectUser, IState> {
 
         return (
             <Select2
+                focus={true}
```
tomaskikutis: Can you call it `autoFocus` as in native input fields?

pablopunk: sure

PR #3235, opened 2019-11-28
avoid opening preview when user selects item via checkbox

petrjasek:


scripts/apps/search/components/ItemList.tsx
```diff
@@ -121,12 +121,13 @@ export class ItemList extends React.Component<any, IState> {
 
         $timeout.cancel(this.updateTimeout);
 
-        const showPreview = (event == null || event.target == null) ||
+        const showPreview = event == null || event.target == null ||
```
tomaskikutis: If I'm not wrong, the issue is that it checks for the parent to have that class name, but doesn't check itself which is the problem.

I'd expect you'd either add a check for self like `event.target.classList.contains('sd-monitoring-item-multi-select-checkbox')` or refactor it to use a parameter and remove the class name approach. Why are you keeping both?

petrjasek: it actually works but in photo thumbnail view it didn't have that class

tomaskikutis: Can you add a check to self too?(`event.target.classList.contains('sd-monitoring-item-multi-select-checkbox')`) If you click 2 pixels outside of the checkbox it still triggers the preview which would be nice to avoid too by disabling preview even if checkbox wrapper is clicked.

PR #3240, opened 2019-12-03
Fix: Right after updating a template, the old one is still used. [SDESK-4828]

GyanP:


scripts/apps/workspace/content/services/ContentService.ts
```diff
@@ -135,24 +135,6 @@ export function ContentService(api, templates, desks, packages, archiveService,
         });
     };
 
-    /**
-     * Create new item using given content type
-     *
-     * @param {Object} contentType
-     * @return {Promise}
-     */
-    this.createItemFromContentType = function(contentType) {
```
GyanP: @petrjasek I think we don't need this anymore. I couldn't find it's being used anywhere else now. This was the [PR](https://github.com/superdesk/superdesk-client-core/commit/29dd7b3b97d9d08c14ecd3ddbb8ec36e81685902) where we removed its use.

tomaskikutis:
What was the issue? You already have `template` as an argument in `createFromTemplate`. Why doesn't that work?

GyanP:
> What was the issue?

When we update an existing template and click on the `save as template` button in the monitoring, the new changes get updated but recent templates dropdown still contains the previous template value. So, one has to refresh the page to get updated values. So, to make sure that we create the new item from the last updated template only we fetch it from the backend and then create.

tomaskikutis:
What happens when you add/remove templates? I think it would be better to add a websocket listener and update the item object whenever it changes.

GyanP:
> What happens when you add/remove templates?

@tomaskikutis sorry for the late reply, I was busy in another issue.
Actually, we can create templates in two ways one from `settings->tepmplates` another from the monitoring using `save as template`.  Whenever we add/remove/update a template in the settings we re-fetch all the templates after every add/remove/update operation. We don't have any WebSocket notification.

tomaskikutis:
If we don't have websocket notifications for templates, how do they get updated if another user modifies it? I assume they don't?

Could you add the websocket notification on the server and make it work on the client in settings and monitoring view? It should fix the current issue too.

GyanP:
> If we don't have websocket notifications for templates, how do they get updated if another user modifies it? I assume they don't?

Yes, if one user updates a template another user still gets the old one if he doesn't refresh or change tabs. I tested it. I will add one then.

scripts/apps/workspace/content/directives/ContentCreateDirective.ts
```diff
@@ -56,23 +56,24 @@ export function ContentCreateDirective(
                 return $location.path() !== '/workspace/personal';
             };
 
+            scope.$on('template:update', (e, data) => {
+                if (data?.desks.includes(desks.activeDeskId)) {
+                    getRecentTemplates(desks.activeDeskId);
```
tomaskikutis: When can `data.desks` not be present? Shouldn't templates be updated anyway?

GyanP: >When can data.desks not be present?

Will always be there, will fix it.

Templates could be accessed only from the desk they are allowed to. So, why call `getRecentTemplates` in a desk if a template of another desk is updated?

tomaskikutis: > Templates could be accessed only from the desk they are allowed to. So, why call `getRecentTemplates` in a desk if a template of another desk is updated?

If it's cheap from a performance point of view to update templates unconditionally, I'd do that. We never know what kind of feature requests will come our way so it makes sense not to constrain ourselves too much when possible. For example, a requirement might come in to also display the templates of routed desks and we would get a bug if we forget to change this update logic too.

PR #3243, opened 2019-12-04
support elastic 7

petrjasek:


tomaskikutis:
Don't you think we should decouple front-end from elastic search entirely?

petrjasek:
@tomaskikutis it would make sense to avoid passing some params from client like when doing full text search, not sure about the filters there it's pretty descriptive

tomaskikutis:
I'm not saying they are not descriptive. I'm saying that the client shouldn't depend on storage/search engine specific formats. I suggest creating or picking a subset of an existing query language and using that as "superdesk query format" so it wouldn't matter what's being used under the hood and the same query language would work for mongo, elastic and could even work on SQL if we added an endpoint for example. I'd apply this to the server as well so we don't have infrastructure details mixed with business logic.

petrjasek:
yep it would make sense so client wouldn't have to care if resource is using just mongo or elastic

PR #3245, opened 2019-12-09
Filters for related items

pablopunk:
SDBELGA-226

I made it work just like `allowed_types` but I called it `allowed_workflows`, and both types (in progress and published) are enabled by default.


scripts/apps/vocabularies/controllers/VocabularyEditController.ts
```diff
@@ -10,16 +12,51 @@ VocabularyEditController.$inject = [
...
-export function VocabularyEditController($scope, notify, api, vocabularies, metadata, cvSchema) {
+export function VocabularyEditController(
+    $scope: IScope, notify, api, vocabularies, metadata, cvSchema, relationsService,
+) {
     var origVocabulary = _.cloneDeep($scope.vocabulary);
 
     $scope.idRegex = idRegex;
     $scope.selectionTypes = VOCABULARY_SELECTION_TYPES;
 
+    if ($scope.matchFieldTypeToTab?.('related-content-fields', $scope.vocabulary.field_type)) {
```
tomaskikutis: `matchFieldTypeToTab` looks to be [defined unconditionally](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/vocabularies/controllers/VocabularyConfigController.ts#L86). Why are you checking if it's defined? Can it be undefined?

scripts/apps/vocabularies/controllers/VocabularyEditController.ts
```diff
@@ -10,16 +12,51 @@ VocabularyEditController.$inject = [
...
+        // Insert default allowed workflows
+        if ($scope.vocabulary.field_options == null) {
+            $scope.vocabulary.field_options = {allowed_workflows: relationsService.getDefaultAllowedWorkflows()};
```
tomaskikutis: When I open an existing field for editing, I see the options selected. They couldn't have been selected, because it's the first time I'm opening it for editing after checking out this PR and I haven't run upgrade scripts either.

scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -59,24 +76,33 @@ export function RelatedItemsDirective(authoringWorkspace: AuthoringWorkspaceServ
...
 
                     const itemAlreadyAddedAsRelated = relatedItemsForCurrentField.some(
                         (relatedItem) => relatedItem._id === item._id,
                     );
 
+                    const isWorkflowAllowed = relationsService.itemHasAllowedStatus(item, scope.field);
+
+                    if (!isWorkflowAllowed) {
+                        notify.error(
+                            gettext('The following status is not allowed in this field: ') + `"${item.state}"`,
```
tomaskikutis: Include the status in the translation string so it can appear in different positions in translations:

`gettext('The following status is not allowed in this field: {{status}}', {status: item.state})`

It would also help to display which statuses are allowed.

scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -144,30 +170,6 @@ export function RelatedItemsDirective(authoringWorkspace: AuthoringWorkspaceServ
             };
             scope.refreshRelatedItems();
 
-            /**
-             * Reorder related items on related items list
-             *
-             * @param {int} start
-             * @param {int} end
-             */
-            scope.reorder = (start, end) => {
```
tomaskikutis: was this unused or did you remove the feature in this PR?

scripts/superdesk-interfaces/Vocabulary.ts
```diff
@@ -10,20 +10,30 @@ export interface IVocabulary extends IBaseRestApiResponse {
...
-    field_type: 'text' | 'media' | 'date' | 'embed' | 'related_content' | 'custom';
-    field_options?: {};
+    field_type:
+        | 'text'
+        | 'media'
+        | 'date'
+        | 'embed'
+        | 'related_content'
+        | 'custom';
+    field_options?: {
```
tomaskikutis: Is `field_options` specific to related items field? If so, could you rename it to `related_items_options`? We'll probably move it to be a custom field later, but until we do I think it's best to document that it's specific to related items so no other properties get added there which don't belong to related items field.

scripts/apps/vocabularies/controllers/VocabularyEditController.ts
pablopunk: a lot of tests were failing because they create the scope but don't provide this function (and they don't really need it)

scripts/apps/vocabularies/controllers/VocabularyEditController.ts
pablopunk: but for existing fields, it makes sense to have all options selected, as it's the current behavior they have (they allow any workflow state)

scripts/apps/relations/directives/RelatedItemsDirective.ts
pablopunk: great

scripts/apps/vocabularies/controllers/VocabularyEditController.ts
tomaskikutis: Can you pass a dummy function from tests instead? We need to have the right structure so testing is easy but we shouldn't be adding if statements to the components/pages specifically for tests.

scripts/apps/relations/directives/RelatedItemsDirective.ts
pablopunk: oops, I don't remember why I deleted this but it's not related to my PR, it should be there.

scripts/apps/vocabularies/controllers/VocabularyEditController.ts
tomaskikutis: Right, I thought they were limiting the options. In that case, it's probably fine, even though it's strange that boxes are ticked when there is nothing in the database.

scripts/superdesk-interfaces/Vocabulary.ts
pablopunk: It seems it is, but I can see it as a generic type for links between items. Maybe we could do some renaming. I can add now a comment that's only being used for related items

scripts/apps/vocabularies/controllers/VocabularyEditController.ts
```diff
@@ -48,7 +48,7 @@ export function VocabularyEditController(
     $scope.idRegex = idRegex;
     $scope.selectionTypes = VOCABULARY_SELECTION_TYPES;
 
-    if ($scope.matchFieldTypeToTab?.('related-content-fields', $scope.vocabulary.field_type)) {
+    if ($scope.matchFieldTypeToTab.('related-content-fields', $scope.vocabulary.field_type)) {
```
tomaskikutis: You forgot to remove [the dot](https://user-images.githubusercontent.com/2076953/70918546-7eb3a980-201f-11ea-9008-54156bc0c407.png) from the function call.

You should probably also pass the [same thing](https://user-images.githubusercontent.com/2076953/70918803-fa155b00-201f-11ea-8e7e-3f27104376da.png) from the tests

pablopunk: Yeah I already fixed the dot. And you're right about the noop fn, I did mean to have the same thing in both

PR #3246, opened 2019-12-09
avoid opening preview on select with shift

petrjasek:
SDANSA-356

### DIFF
--- a/scripts/apps/search/components/ItemList.tsx
+++ b/scripts/apps/search/components/ItemList.tsx
@@ -84,7 +84,7 @@ export class ItemList extends React.Component<any, IState> {
             });
         });
 
-        this.select(_.last(items), event);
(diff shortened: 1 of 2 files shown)


scripts/apps/search/components/ItemList.tsx
```diff
@@ -102,7 +102,7 @@ export class ItemList extends React.Component<any, IState> {
         this.setState({narrow: setNarrow});
     }
 
-    select(item, event) {
+    select(item, event, {preview}: {preview?: boolean} = {preview: null}) {
```
tomaskikutis: Would it work removing `CHECKBOX_PARENT_CLASS` so we only have one way of doing the same thing?

petrjasek: actually I don't think it should call that select at all when selecting multiple items

PR #3249, opened 2019-12-10
feat(authoring): Added unpublish button in authoring topbar

pavlovicnemanja:


spec/helpers/authoring.ts
```diff
@@ -194,7 +194,7 @@ class Authoring {
         this.save_button = element(by.buttonText('Save'));
         this.edit_button = element(by.id('Edit'));
         this.update_button = element(by.buttonText('Update'));
-        this.edit_correct_button = element(by.buttonText('Correct'));
+        this.edit_correct_button = element(by.css('[title="Correct"]'));
         this.edit_kill_button = element(by.buttonText('Kill'));
```
petrjasek: seems like kill has the same issue

scripts/apps/authoring/authoring/services/AuthoringService.ts
```diff
@@ -264,6 +266,32 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
             );
     };
 
+    this.unpublish = function unpublish(item) {
```
tomaskikutis: Can you add a type `unpublish(item: IPublishedArticle)` and validation to ensure that `item` is really published?

spec/helpers/authoring.ts
pavlovicnemanja: It was not failing first time. I added change for all buttons that are in header

scripts/apps/authoring/views/authoring-topbar.html
```diff
@@ -98,22 +98,42 @@
...
       <button class="btn btn--primary" ng-if="item._type !== 'legal_archive' && userHasPrivileges({takedown: 1}) === true"
         ng-show="(action === 'view' || !item._editable) && itemActions.takedown"
         ng-click="openAction('takedown')"
-        translate>Takedown</button>
+        title="{{gettext('Takedown')}}">
+            <span class="btn__text" translate>Takedown</span>
+            <span class="btn__text--short" translate>T</span>
+      </button>
+
+      <button class="btn btn--primary" ng-if="item._type !== 'legal_archive' && userHasPrivileges({takedown: 1}) === true"
```
petrjasek: it should check for `unpublish` privilege

scripts/apps/authoring/authoring/services/AuthoringService.ts
pavlovicnemanja: Which validation? Before everything?

pavlovicnemanja: Using this function before? `isIPublishedArticle()`

tomaskikutis: Yep, you can use this function to. It would also be good to log the error like

```
if(isIPublishedArticle(item) !== true) {
    logger.warn(...);
    return;
}

... unpublishing logic
```

PR #3251, opened 2019-12-11
Mark for user in preview

tomaskikutis:
SDBELGA-228


scripts/apps/monitoring/directives/ItemActionsMenu.ts
```diff
@@ -59,43 +85,101 @@ export function ItemActionsMenu(superdesk, activityService, workflowService, arc
...
+                                        .map((activity) => ({kind: 'activity-based', activity: activity})),
+                                });
+                            }
+                        });
+
+                        // handle actions from extensions
+                        let extensionActionsByGroupName: {[groupName: string]: Array<IArticleAction>} = {};
+
+                        for (const action of actionsFromExtensions) {
+                            let name = action.groupId ?? 'default';
```
petrjasek: just wondering why using `let` here and not `const`?

tomaskikutis: I wasn't sure if `const` variables are block or function scoped and I knew that `let` is block scoped. ;D I've looked it up, `const` is block-scoped too, will update it.

PR #3252, opened 2019-12-12
Disable save button until api response is received. [SDESK-4870]

GyanP:
Fixes the issue with the save button getting enabled without making changes in some cases.

See [here](https://drive.google.com/file/d/12FRcfU4smDsPm6fam5KyoqPr6rM92Tbt/view), after clicking the save button is enabled again due to delay in the API response.

Create an article and add a picture(can have a number of possible crops) to its body, 
1. don't click save button, 
2. goto edit crops, 
3. change the POI, 
4. apply changes and click done. 
5. Notice save button is enabled which is expected behavior as the button was enabled on adding the picture, 
6. click save and after few seconds save button will be enabled again because the response of `picture_crop`  API has arrived.

This won't happen if the save button is disabled before going to edit crops.

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
@@ -240,7 +240,7 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, $q) {
                     return;
                 }
             }
-            $scope.resolve({
+            const data = {
                 cropData: $scope.data.cropData,
                 metadata: _.pick($scope.data.metadata, [
                     ...EDITABLE_METADATA,
@@ -248,7 +248,9 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, $q) {
                     'renditions',
                     '_etag',
                 ]),
-            });
+            };
+
+            generateCrops(data);
         } else {
             $scope.reject({done: true});
         }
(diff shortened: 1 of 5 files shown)


tomaskikutis:
If I understand correctly, the problem was that the "save" button in authoring was being clicked before the request to change the point of interest has finished.

I might be wrong, but it looks like you made the save button behave correctly, but the data would still be incorrect because when you click "save" before network requests had finished, you are sending data that is not patched yet. It only gets patched after the `savingImagePromises` are resolved.

Can you instead only close the image editing modal after the requests had finished after clicking "done"? You can display a loading indicator so it's clear to the user.

GyanP:
Save button will be enabled only after the `savingImagePromises` promise is resolved, so the updated data will be patched.

> Can you instead only close the image editing modal after the requests had finished after clicking "done"?

At first, I thought of doing the same, but with the way things are currently, I couldn't make it work. 
Clicking on `done` will call [this](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/activity/activity-modal-directive.ts#L20) method.
This is how the modal is [opened](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/activity/views/activity-modal.html#L1), and [this](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/activity/activity-modal-directive.ts#L9) is how it's closed.

tomaskikutis:
I see. How about moving  `savingImagePromises` to `ChangeImageController.ts` and only [finishing the activity](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts#L243) after the promises are resolved? You should be able to pass any data which needs to be patched the same way `cropData` and `metadata` are passed.

PR #3255, opened 2019-12-13
View configuration for related items

pablopunk:
SDESK-4874

image

### DIFF
--- a/scripts/apps/relations/components/RelatedItemInList.tsx
+++ b/scripts/apps/relations/components/RelatedItemInList.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import {appConfig} from 'appConfig';
 import {DEFAULT_RELATED_ITEMS_LIST_CONFIG} from 'apps/search/constants';
 import {renderArea} from 'apps/search/helpers';
-import {IArticle, IDesk} from 'superdesk-api';
+import {IArticle, IDesk, ISuperdeskGlobalConfig} from 'superdesk-api';
 import {dataApi} from 'core/helpers/CrudManager';
 
 interface IProps {
@@ -14,7 +14,7 @@ interface IState {
 }
 
 export class RelatedItemInListComponent extends React.PureComponent<IProps, IState> {
-    listConfig: any;
+    listConfig: ISuperdeskGlobalConfig['list']['relatedItems'];
 
     constructor(props: IProps) {
         super(props);
--- a/scripts/apps/relations/components/RelatedItemInList.tsx
+++ b/scripts/apps/relations/components/RelatedItemInList.tsx
@@ -19,7 +19,7 @@ export class RelatedItemInListComponent extends React.PureComponent<IProps, ISta
     constructor(props: IProps) {
         super(props);
 
-        this.listConfig = appConfig.list || DEFAULT_RELATED_ITEMS_LIST_CONFIG;
+        this.listConfig = appConfig.list?.relatedItems || DEFAULT_RELATED_ITEMS_LIST_CONFIG;
 
         this.state = {
             desk: null,
@@ -38,13 +38,14 @@ export class RelatedItemInListComponent extends React.PureComponent<IProps, ISta
 
         const {listConfig} = this;
         const itemProps = {item, desk, listConfig};
+        const elemProps = {className: 'line line--no-margin'};
 
         return (
             <div className="mlist-view mlist-view--no-shadow">
                 <div className="media-box media-box--no-padding">
                     <div className="item-info">
-                        {renderArea('firstLine', itemProps)}
-                        {renderArea('secondLine', itemProps)}
+                        {renderArea('firstLine', itemProps, elemProps)}
+                        {renderArea('secondLine', itemProps, elemProps)}
                     </div>
                 </div>
             </div>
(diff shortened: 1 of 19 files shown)


scripts/core/superdesk-api.d.ts
```diff
@@ -220,7 +220,7 @@ declare module 'superdesk-api' {
         sign_off: string;
         feature_media?: any;
         media_description?: string;
-        associations?: { string: IArticle };
+        associations?: any;
```
tomaskikutis: It should probably be `IRelatedArticle`. Even though the entire `IArticle` is stored in some cases, I think for items which aren't locally in superdesk.

scripts/apps/relations/components/RelatedItemInList.tsx
```diff
@@ -0,0 +1,54 @@
...
+interface IProps {
+    item: IArticle;
+}
+
+interface IState {
+    desk: IDesk;
+}
+
+export class RelatedItemInListComponent extends React.PureComponent<IProps, IState> {
+    listConfig: any;
```
tomaskikutis: Can you specify the type?

scripts/apps/search/components/fields/state.tsx
```diff
@@ -5,7 +5,7 @@ import {removeLodash} from 'core/filters';
 import {IPropsItemListInfo} from '../ListItemInfo';
 
 export const state: React.StatelessComponent<Pick<IPropsItemListInfo, 'item' | 'svc'>> = (props) => {
-    const datetime = props.svc.datetime;
+    const datetime = props.svc?.datetime;
```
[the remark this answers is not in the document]
pablopunk: It doesn't look like a good idea to export an angular service, I'd rather have something like `ng.getService('datetime')` but that returns a promise so I'd have to change this component not to be stateless (which is fine I guess)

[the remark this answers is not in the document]
tomaskikutis: I'm suggesting to export a function, not an angular service. First, extract it from the service, export it, use it here and in the original service you extracted it from.

tomaskikutis:
Can you specify a type for related items? https://github.com/superdesk/superdesk-client-core/pull/3255#discussion_r358659044

I think it should be `relatedItem: IRelatedArticle | IArticle`

scripts/core/superdesk-api.d.ts
pablopunk: oops, forgot about that

scripts/core/superdesk-api.d.ts
```diff
@@ -226,8 +226,15 @@ declare module 'superdesk-api' {
         sign_off: string;
         feature_media?: any;
         media_description?: string;
-        associations?: any;
-        type: 'text' | 'picture' | 'video' | 'audio' | 'preformatted' | 'graphic' | 'composite';
+        associations?: IArticle | IRelatedArticle;
```
tomaskikutis: `IRelatedArticle` needs to be exported. This is probably the only file that's not being typed checked properly. Can you also add a comment that `IRelatedArticle` is for internal items and `IArticle` for external like from search providers?

https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/relations/directives/RelatedItemsDirective.ts#L222

pablopunk: `IRelatedArticle` is already being exported in this file. About the comment, it's not like that, here's an example of associations:

image

All of those are internal (i.e they are user created items) and yet they are `IRelatedArticle` and `IArticle`

tomaskikutis: You are right about the `IRelatedArticle` being exported in this file.

I'm not sure about all associations, but related items should defintely be only stored in `IRelatedArticle` format. See `SDESK-3681`.

@petrjasek do you know what are all the cases when entire objects are stored? Items from custom search providers? What more? Custom media, galleries?

petrjasek: custom media, galleries. we only use reference for related items, otherwise there is full object.

scripts/apps/relations/views/related-items.html
```diff
@@ -1,26 +1,25 @@
-<button class="item-association item-association--no-click" 
-        ng-if="isEmptyRelatedItems(field._id)"
-        ng-class="{'item-association--loading': loading}">
-        <span class="item-association__text-label">{{'Drop items here' | translate}}</span>  
+<button class="item-association item-association--no-click" ng-if="isEmptyRelatedItems(field._id)"
+                                                            pen
```
petrjasek: :eyes:

pablopunk: I'll blame it on the conflicts 💃

scripts/core/datetime/datetime.ts
```diff
@@ -1,7 +1,6 @@
 import _ from 'lodash';
-import moment from 'moment';
 import {gettext} from 'core/utils';
-import momentTimezone from 'moment-timezone';
+import moment from 'moment-timezone';
```
tomaskikutis:
```
import moment from 'moment';
import momentTimezone from 'moment-timezone';
```

These are different packages, it would be good not to mix up the names.

pablopunk: but 'moment-timezone' it's exporting everything 'moment' does, but adding timezone functions, it's meant to replace it

tomaskikutis: Ok, I guess it's not a big deal then :)

PR #3259, opened 2019-12-18
Change metadata on profile change based on the config. [SDBELGA-230].

GyanP:
required for https://github.com/superdesk/superdesk-belga/pull/111

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
@@ -852,16 +852,18 @@ export function AuthoringDirective(
                 return coreApplyMiddleware(onChangeMiddleware, {item: $scope.item, original: $scope.origItem}, 'item')
                     .then(() => {
                         const onUpdateFromExtensions = Object.values(extensions).map(
-                            (extension) => extension.activationResult?.contributions?.authoring)
-                            .filter((updates) => updates != null);
+                            (extension) => extension.activationResult?.contributions?.authoring?.onUpdate,
+                        ).filter((updates) => updates != null);
 
                         const reducerFunc = (current, next) => current.then(
-                            (result) => next.onUpdate($scope.origItem._autosave ?? $scope.origItem, result));
+                            (result) => next($scope.origItem._autosave ?? $scope.origItem, result),
+                        );
 
                         (
                             onUpdateFromExtensions.length < 1
                                 ? Promise.resolve(item)
-                                : onUpdateFromExtensions.reduce(reducerFunc, Promise.resolve($scope.item))
+                                : onUpdateFromExtensions
+                                    .reduce(reducerFunc, Promise.resolve($scope.item))
                                     .then((nextItem) => angular.extend($scope.item, nextItem))
                         ).then(() => {
                             var autosavedItem = authoring.autosave($scope.item, $scope.origItem, timeout);
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
@@ -846,40 +846,31 @@ export function AuthoringDirective(
                 $scope.autosave(item);
             };
 
-            Object.values(extensions).forEach((extension) => {
-                if (extension.activationResult?.contributions?.authoring != null) {
-                    $scope.onUpdateFromExtension = extension.activationResult.contributions.authoring.onUpdate;
-                }
-            });
-
-            function performOnUpdateFromExtensionAndAutosave(item, origItem, timeout) {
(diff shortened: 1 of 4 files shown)


tomaskikutis:
What about if I want to say change priority when genre changes? It's the same problem, except with a different field. Can you implement a generic solution?

You could add an extension point
```
IExtensionActivationResult.contributions.authoring {
    onUpdate(current: IArticle, next: IArticle): Promise<IArticle>;
}
```

And add custom code to client extension:

```
function onUpdate(current, next) {
    if (current.profile !== next.profile && next.profile === 'jh43h5ghj43k6gt') {
        Promise.resolve({
            ...next,
            priority: 5,
            distribution: 'out',
        });
    } else {
        Promise.resolve(next);
    }
}
```

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -846,18 +846,40 @@ export function AuthoringDirective(
                 $scope.autosave(item);
             };
 
+            Object.values(extensions).forEach((extension) => {
+                if (extension.activationResult?.contributions?.authoring != null) {
+                    $scope.onUpdateFromExtension = extension.activationResult.contributions.authoring.onUpdate;
```
tomaskikutis: It should support defining `authoring.onUpdate` in multiple extensions. As you've implemented it now, it will only call `onUpdate` for one extension - the one which ends up being the last iteration in `Object.values(extensions).forEach`.

It should also pipe values through extensions so if `next.priority` is 3, but first extension returns `Promise.resolve({...item, priority: 5})`, the second extension should receive 5 as `priority` in its `next` argument.

See an example here https://github.com/superdesk/superdesk-client-core/pull/3262/files#diff-374a2e82b3cfbcea00b03816fc47b614R120

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -846,18 +846,40 @@ export function AuthoringDirective(
...
+                        } else {
+                            var autosavedItem = authoring.autosave($scope.item, $scope.origItem, timeout);
 
-                        initMedia();
-                        updateSchema();
-                        return autosavedItem;
+                            authoringWorkspace.addAutosave();
+                            initMedia();
+                            updateSchema();
+                            return autosavedItem;
```
tomaskikutis: You are duplicating this code in `performOnUpdateFromExtensionAndAutosave`. Actually, you don't need neither `performOnUpdateFromExtensionAndAutosave` as a separate function nor `onUpdateFromExtension` variable in the scope.

You can do it like this

```
const onUpdateFromExtensions = Object.values(extensions).map(...);

(
    onUpdateFromExtensions.length < 1
    ? Promise.resolve(item)
    : onUpdateFromExtensions.reduce(...).then((nextItem) => {
        angular.extend($scope.item, nextItem);
    })
).then(() => {
    var autosavedItem = authoring.autosave($scope.item, $scope.origItem, timeout);

    authoringWorkspace.addAutosave();
    initMedia();
    updateSchema();
    return autosavedItem;
});
```

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -846,40 +846,31 @@ export function AuthoringDirective(
...
-
             $scope.autosave = function(item, timeout) {
                 $scope.dirty = true;
                 angular.extend($scope.item, item); // make sure all changes are available
                 return coreApplyMiddleware(onChangeMiddleware, {item: $scope.item, original: $scope.origItem}, 'item')
                     .then(() => {
-                        if ($scope.onUpdateFromExtension != null) {
-                            performOnUpdateFromExtensionAndAutosave($scope.item, $scope.origItem, timeout);
-                        } else {
+                        const onUpdateFromExtensions = Object.values(extensions).map(
```
tomaskikutis: Can you fix the [alignment of parens](https://user-images.githubusercontent.com/2076953/72155357-42bbfc00-33b3-11ea-9b4b-488e44349996.png)? It's hard to see where `map` call ends. Can you put the closing parent on a new line?

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -846,40 +846,31 @@ export function AuthoringDirective(
...
+                            (extension) => extension.activationResult?.contributions?.authoring)
```
tomaskikutis: If you need the `onUpdate` function, you should get it when mapping - `contributions?.authoring?.onUpdate`. As you do it now, you never check if `onUpdate` is `null`. You only check `contributions.authoring`

PR #3264, opened 2019-12-23
Do not show mark for user option for locked items in authoring

tomaskikutis:
SDESK-4878


scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -66,25 +72,15 @@ export function AuthoringTopbarDirective(
...
-            );
+            scope.$watch('item', () => {
+                setActionsFromExtensions();
+            }, true);
 
-            Promise.all(authoringActionsFromExtensions.map((getPromise) => getPromise(scope.item)))
-                .then((actions) => {
-                    scope.authoringActionsFromExtensions = flatMap(actions);
-                });
+            setActionsFromExtensions();
```
petrjasek: you can avoid this call, `scope.$watch` will call it once on init. so there could be just:
```ts
scope.$watch('item', setActionsFromExtensions, true);`
```

PR #3266, opened 2019-12-25
Add URL field in the search provider settings. [SDESK-4905]

GyanP:


scripts/apps/search-providers/views/search-provider-config.html
```diff
@@ -1,7 +1,7 @@
 <div class="sd-page__flex-helper" ng-hide="hideList">
     <div class="sd-page__header">
         <span class="sd-page__element-grow"></span>
-        <button id="addSearchProvider" class="btn btn--primary" ng-click="edit()" ng-disabled="noProvidersAllowed" title="{{ noProvidersAllowed ? 'There are no providers available.' : 'Add New Search Provider' | translate }}">
+        <button id="addSearchProvider" class="btn btn--primary" ng-click="edit()" ng-disabled="noProvidersAllowed" title="{{gettext(noProvidersAllowed ? 'There are no providers available.' : 'Add New Search Provider')}}">
```
petrjasek: not sure if it could extract such strings, it would be probably better like:
```js
noProvidersAllowed ? gettext('There are no providers available.') : gettext('Add New Search Provider')
```
same for closed/default checkbox

PR #3268, opened 2019-12-30
Improve the Unpublish behaviour. [SDESK-4891]

GyanP:


scripts/apps/authoring/views/authoring-topbar.html
```diff
@@ -86,7 +86,7 @@
     </div>
     <div class="subnav__button-stack subnav__button-stack--default">
       <button id="Edit" class="btn btn--medium btn--primary" ng-if="item._type !== 'legal_archive'"
-        ng-show="action === 'view' && !item._editable && itemActions.edit"
+        ng-show="action === 'view' && !item._editable && (itemActions.edit || item.state === 'unpublished')"
```
petrjasek: I think we should rather set `itemActions.edit` when state is unpublished rather than dealing with that logic here in the template

GyanP: @petrjasek setting `itemActions.edit` to true will not show the edit button, but it would rather open the article in edit mode directly. Will that be ok?

petrjasek: not sure about that. @karelpetrak ?

karelpetrak: @GyanP @petrjasek Does it mean that whenever a user doubleclicks an unpublished story they get directly to the edit mode? Even if the do it on the story's "instance" which has "published" label?

GyanP: @petrjasek Karel, have agreed to how it behaves now.

PR #3270, opened 2020-01-07
display cv values in preview if enabled in profile

petrjasek:
SDANSA-362


scripts/apps/archive/views/preview.html
```diff
@@ -45,6 +45,9 @@
               <span ng-if="item._type === 'archived'" class="stage" tooltip="{{:: 'Archived from' | translate}} {{deskName}} / {{ stage }}" tooltip-placement="bottom" ng-show="stage"><b>{{:: 'Archived from' | translate}} {{deskName}}</b> / {{ stage }}</span>
               <span ng-if="item._type !== 'archived'" class="stage" tooltip="{{deskName}} / {{ stage }}" tooltip-placement="bottom" ng-show="stage"><b>{{deskName}}</b> / {{ stage }}</span>
             </div>
+            <div class="preview-header__flex-row preview-header__flex-row--single-line" ng-if="previewFields.length">
```
tomaskikutis: It's a bit strange that `previewFields.length` is used to check if the subject fields are to be displayed. It's called `previewFields`, not `subjectPreviewFields` so I assume there might be more of them of other types. You should probably be either more clear in the naming that it's only for subject fields or include the field type in the `if` condition instead of only chekcking the length of an array.

petrjasek: yep it works now because you can only enable it for custom subject fields, but eventually this wouldn't work. will fix

petrjasek:
close in favor of https://github.com/superdesk/superdesk-client-core/pull/3278

PR #3271, opened 2020-01-08
Feature: filter content in Monitoring view by content profile. [SDESK-4922]

GyanP:
monitoring

### DIFF
--- a/scripts/apps/monitoring/controllers/AggregateCtrl.ts
+++ b/scripts/apps/monitoring/controllers/AggregateCtrl.ts
@@ -303,7 +303,7 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
             }
         });
 
-        _.each(spikeDesks, (item) => {
+        _.each(spikeDesks, (item: any) => {
             if (item._id === 'personal') {
                 self.spikeGroups.push({_id: item._id, type: 'spike-personal', header: item.name});
             } else {
--- a/scripts/apps/monitoring/controllers/AggregateCtrl.ts
+++ b/scripts/apps/monitoring/controllers/AggregateCtrl.ts
@@ -372,8 +372,8 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
         if (filterType === 'contentProfile') {
             if (!this.activeFilters.contentProfile.includes(filterValue._id)) {
                 this.activeFilters.contentProfile.push(filterValue._id);
-                const tag = {'key': filterValue._id, 'label': filterValue.label};
-                const type = 'Content profile';
+                const tag = {'key': filterValue._id, 'label': gettext(filterValue.label)};
+                const type = gettext('Content profile');
 
                 if (Array.isArray(this.activeFilterTags[type])) {
                     this.activeFilterTags[type].push(tag);
--- a/scripts/apps/monitoring/controllers/AggregateCtrl.ts
+++ b/scripts/apps/monitoring/controllers/AggregateCtrl.ts
@@ -5,7 +5,6 @@ import {appConfig} from 'appConfig';
 
 AggregateCtrl.$inject = ['$scope', 'desks', 'workspaces', 'preferencesService', 'storage',
     'savedSearch', 'content'];
-
 export function AggregateCtrl($scope, desks, workspaces, preferencesService, storage,
     savedSearch, content) {
     var PREFERENCES_KEY = 'agg:view';
@@ -31,10 +30,10 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
 
     this.activeProfiles = [];
     this.activeFilters = {
-        profile: $scope.type === 'monitoring' ? storage.getItem('profile') || [] : [],
+        contentProfile: $scope.type === 'monitoring' ? storage.getItem('contentProfile') || [] : [],
         fileType: $scope.type === 'monitoring' ? storage.getItem('fileType') || [] : [],
     };
-    this.activeFilterTags = [];
+    this.activeFilterTags = {};
 
     desks.initialize()
         .then(angular.bind(this, function() {
@@ -60,9 +59,8 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
                 .then(angular.bind(this, function(settings) {
                     initGroups(settings);
                     setupCards();
-                    getActiveProfiles();
-                    this.loading = false;
                     this.settings = settings;
+                    getActiveProfiles();
                 }));
         }));
 
@@ -305,7 +303,7 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
             }
         });
 
-        _.each(spikeDesks, (item: any) => {
+        _.each(spikeDesks, (item) => {
             if (item._id === 'personal') {
                 self.spikeGroups.push({_id: item._id, type: 'spike-personal', header: item.name});
             } else {
@@ -354,7 +352,7 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
      * @return [{string}] fileType
      */
     this.getSelectedFileTypes = function() {
-        return this.activeFilters.fileType.length === 0 ? null : JSON.stringify(this.this.activeFilters.fileType);
+        return this.activeFilters.fileType.length === 0 ? null : JSON.stringify(this.activeFilters.fileType);
     };
 
     function updateFilteringCriteria() {
@@ -371,10 +369,17 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
     }
 
     this.setFilterType = function(filterType, filterValue, $event?) {
-        if (filterType === 'profile') {
-            if (!this.activeFilters.profile.includes(filterValue._id)) {
-                this.activeFilters.profile.push(filterValue._id);
-                this.activeFilterTags.push({'key': filterValue._id, 'label': filterValue.label});
+        if (filterType === 'contentProfile') {
+            if (!this.activeFilters.contentProfile.includes(filterValue._id)) {
+                this.activeFilters.contentProfile.push(filterValue._id);
+                const tag = {'key': filterValue._id, 'label': filterValue.label};
+                const type = 'Content profile';
+
+                if (Array.isArray(this.activeFilterTags[type])) {
+                    this.activeFilterTags[type].push(tag);
+                } else {
+                    this.activeFilterTags[type] = [tag];
+                }
             }
         } else if (filterType === 'file') {
             if (filterValue === 'all') {
@@ -398,13 +403,14 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
         }
     };
 
-    this.removeFilter = (filter) => {
+    this.removeFilter = (filter, type?) => {
         if (filter == null) {
-            this.activeFilters.profile = [];
-            this.activeFilterTags = [];
+            this.activeFilters.contentProfile = [];
+            this.activeFilterTags = {};
         } else {
-            this.activeFilters.profile = this.activeFilters.profile.filter((profile) => profile !== filter);
-            this.activeFilterTags = this.activeFilterTags.filter((tags) => tags.key !== filter);
+            this.activeFilters.contentProfile = this.activeFilters.contentProfile
+                .filter((profile) => profile !== filter);
+            this.activeFilterTags[type] = this.activeFilterTags[type].filter((tags) => tags.key !== filter);
         }
 
         updateFilterInStore();
@@ -415,7 +421,7 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
     function updateFilterInStore() {
         if ($scope.type === 'monitoring') {
             storage.setItem('fileType', self.activeFilters.fileType);
-            storage.setItem('profile', self.activeFilters.profile);
+            storage.setItem('contentProfile', self.activeFilters.contentProfile);
         }
     }
 
@@ -556,11 +562,12 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
 
     function getActiveProfiles() {
         content.getTypes(false).then((profiles) => {
+            self.loading = false;
             self.activeProfiles = profiles;
 
             // initialize the activeFilterTags once the activeProfiles are available
-            if (self.activeFilters.profile.length > 0) {
-                self.activeFilterTags = self.activeFilters.profile.map((filter) => {
+            if (self.activeFilters.contentProfile.length > 0) {
+                self.activeFilterTags['Content profile'] = self.activeFilters.contentProfile.map((filter) => {
                     const profile = profiles.find((p) => p._id === filter);
 
                     return {key: profile._id, label: profile.label};
--- a/scripts/apps/monitoring/controllers/AggregateCtrl.ts
+++ b/scripts/apps/monitoring/controllers/AggregateCtrl.ts
@@ -4,10 +4,10 @@ import {SCHEDULED_OUTPUT, DESK_OUTPUT} from 'apps/desks/constants';
 import {appConfig} from 'appConfig';
 
 AggregateCtrl.$inject = ['$scope', 'desks', 'workspaces', 'preferencesService', 'storage',
-    'savedSearch'];
+    'savedSearch', 'content'];
 
 export function AggregateCtrl($scope, desks, workspaces, preferencesService, storage,
-    savedSearch) {
+    savedSearch, content) {
     var PREFERENCES_KEY = 'agg:view';
     var defaultMaxItems = 10;
     var self = this;
@@ -25,11 +25,16 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
     this.stageLookup = {};
     this.fileTypes = ['all', 'text', 'picture', 'graphic', 'composite',
         'highlightsPackage', 'video', 'audio'];
-    this.selectedFileType = $scope.type === 'monitoring' ? storage.getItem('selectedFileType') || [] : [];
     this.monitoringSearch = false;
     this.searchQuery = null;
-
     this.isOutputType = desks.isOutputType;
+
+    this.activeProfiles = [];
+    this.activeFilters = {
+        profile: $scope.type === 'monitoring' ? storage.getItem('profile') || [] : [],
+        fileType: $scope.type === 'monitoring' ? storage.getItem('fileType') || [] : [],
+    };
+    this.activeFilterTags = [];
 
     desks.initialize()
         .then(angular.bind(this, function() {
@@ -55,6 +60,7 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
                 .then(angular.bind(this, function(settings) {
                     initGroups(settings);
                     setupCards();
+                    getActiveProfiles();
                     this.loading = false;
                     this.settings = settings;
                 }));
@@ -258,7 +264,7 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
             }
         }
         initSpikeGroups(settings.type === 'desk');
-        updateFileTypeCriteria();
+        updateFilteringCriteria();
         self.search(self.searchQuery);
     }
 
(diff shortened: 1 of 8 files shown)


scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -338,55 +344,81 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
...
+        return this.activeFilters.fileType.includes(fileType);
     };
 
     /**
      * Return selected file types if the 'fileType' filter(s) is selected
      * @return [{string}] fileType
      */
     this.getSelectedFileTypes = function() {
-        return this.selectedFileType.length === 0 ? null : JSON.stringify(this.selectedFileType);
+        return this.activeFilters.fileType.length === 0 ? null : JSON.stringify(this.this.activeFilters.fileType);
```
tomaskikutis: Did you really mean `this.this.activeFilters`?

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -338,55 +344,81 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
...
-        if (fileType === 'all') {
-            this.selectedFileType = [];
-        } else {
-            var index = this.selectedFileType.indexOf(fileType);
-
-            if (index > -1) {
-                this.selectedFileType.splice(index, 1);
+    this.setFilterType = function(filterType, filterValue, $event?) {
+        if (filterType === 'profile') {
+            if (!this.activeFilters.profile.includes(filterValue._id)) {
```
tomaskikutis: It would be clearer if it was named `activeFilters.contentProfile` instead of just `profile`, because I was thinking about user profiles.

scripts/apps/monitoring/services/CardsService.ts
```diff
@@ -180,6 +181,14 @@ export function CardsService(search, session, desks) {
         }
     }
 
+    function filterQueryByContentProfile(query, card: ICard) {
+        if (card.profile) {
+            var termsProfileType: any = {terms: {profile: JSON.parse(card.profile)}};
```
tomaskikutis: I wouldn't use an intermediary variable here.

scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -167,12 +182,22 @@ <h3 class="subnav__page-title sd-flex-no-grow sd-empty">
...
                             </ul>
                         </div>
-            
+                    </div>
+                    <!-- Active filter tags -->
+                    <div class="search-tags" ng-if="aggregate.activeFilterTags.length > 0">
+                        <ul class="search-tags__list">
+                            <li ng-repeat="tag in aggregate.activeFilterTags">
+                                <div class="tag-label tag-label--highlight2">
+                                    <span>{{ gettext('Cp:') }}</span>{{ tag.label }}<button class="tag-label__remove" ng-click="aggregate.removeFilter(tag.key)"><i class="icon-close-small"></i></button>
```
tomaskikutis: I'd avoid abbreviations and instead use one full label for the [entire group](https://user-images.githubusercontent.com/2076953/72244220-16d88a80-35ee-11ea-82db-176813c00fee.png)

scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -167,12 +182,22 @@ <h3 class="subnav__page-title sd-flex-no-grow sd-empty">
...
+                                </div>
+                            </li>
+                            <li><a ng-click="aggregate.removeFilter(null)" translate>Clear filters</a></li>
```
tomaskikutis: When hovered it should display a pointer cursor instead of text selection. That would work automatically if you used a button here.

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -4,10 +4,10 @@ import {SCHEDULED_OUTPUT, DESK_OUTPUT} from 'apps/desks/constants';
 import {appConfig} from 'appConfig';
 
 AggregateCtrl.$inject = ['$scope', 'desks', 'workspaces', 'preferencesService', 'storage',
-    'savedSearch'];
+    'savedSearch', 'content'];
```
tomaskikutis: Can you add TypeScript types to scope and places where you changed things? It would also help if you could add a few tests since this view has long gotten messier than it should be.

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -338,55 +342,89 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
...
-        if (fileType === 'all') {
-            this.selectedFileType = [];
-        } else {
-            var index = this.selectedFileType.indexOf(fileType);
+    this.setFilterType = function(filterType, filterValue, $event?) {
+        if (filterType === 'contentProfile') {
+            if (!this.activeFilters.contentProfile.includes(filterValue._id)) {
+                this.activeFilters.contentProfile.push(filterValue._id);
+                const tag = {'key': filterValue._id, 'label': filterValue.label};
+                const type = 'Content profile';
```
tomaskikutis: The label should be translated

scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -5,15 +5,31 @@
...
                     <div class="subnav" ng-if="hideMonitoringToolbar1 !== true">
                         <div class="flat-searchbar monitoring-backlink" ng-if="monitoring.singleGroup">
-                            <button sd-tooltip="{{ :: 'Back to original view' | translate }}" flow="right" class="navbtn navbtn--left" ng-click="monitoring.viewMonitoringHome()">
+                            <button sd-tooltip="{{ gettext('Back to original view') }}" flow="right" class="navbtn navbtn--left" ng-click="monitoring.viewMonitoringHome()">
                                 <i class="icon-arrow-left"></i>
                             </button>
                         </div>
-                                        
+
+                        <!-- Filter content using content profile -->
```
tomaskikutis: can you make it [scrollable](https://user-images.githubusercontent.com/2076953/72523216-a173f080-385f-11ea-9e46-c5d0dfe9e2b2.png)

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -372,8 +372,8 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
         if (filterType === 'contentProfile') {
             if (!this.activeFilters.contentProfile.includes(filterValue._id)) {
                 this.activeFilters.contentProfile.push(filterValue._id);
-                const tag = {'key': filterValue._id, 'label': filterValue.label};
-                const type = 'Content profile';
+                const tag = {'key': filterValue._id, 'label': gettext(filterValue.label)};
```
tomaskikutis: Only static strings can be passed to `gettext`. Don't attempt translating `filterValue.label` here. Instead, translate it at the point where it's a static string, so when it comes to this variable it's already translated.

scripts/apps/monitoring/tests/monitoring.spec.ts
```diff
@@ -218,6 +218,16 @@ describe('monitoring', () => {
             });
         }));
 
+        it('can get criteria for content type filter', inject((cards, session) => {
```
tomaskikutis: These tests are only testing what query gets generated. The dropdown could be removed and the tests would still pass. Could you instead replace these with e2e tests where you would test if all content profiles appear in the dropdown and if correct items get filtered?

spec/monitoring_spec.ts
```diff
@@ -375,6 +377,51 @@ describe('monitoring', () => {
         expect(monitoring.getGroupItems(4).count()).toBe(1);
     });
 
+    it('can filter content by content profile', () => {
+        contentProfiles.openContentProfileSettings();
```
tomaskikutis: Can you add it to prepopulate test data instead?

spec/monitoring_spec.ts
```diff
@@ -375,6 +377,51 @@ describe('monitoring', () => {
...
+        templates.selectDesk('Politic Desk');
+        templates.selectDesk('Sports Desk');
+        templates.save();
+
+        monitoring.openMonitoring();
+        workspace.selectDesk('Sports Desk');
+        authoring.createTextItemFromTemplate('Simple');
+        authoring.setHeaderSluglineText('Story1 slugline');
+        authoring.getSubjectMetadataDropdownOpened();
+        browser.sleep(100);
```
tomaskikutis: Avoid sleep calls. Instead, use `browser.wait` like `browser.wait(ECE.presenceOf(el('clear-filters')));`

spec/helpers/content_profiles.ts
```diff
@@ -153,6 +158,49 @@ class ContentProfiles {
...
+        */
+        this.selectContentProfile = () => {
+            this.totalProfiles().then(selectProfile);
+        };
+
+        /**
+         * Remove a content profile filter.
+        */
+        this.removeFilter = () => {
+            element(by.className('tag-label__remove')).click();
```
tomaskikutis: I would not use many abstractions in tests. If you want to clear the filters, you do `el('clear-filters').click()` directly from the spec, I don't see a benefit to wrap it in the object. That's the purpose of `data-test-id` tags: even if you change the markup or migrate the component from angular to react, the test tag will still be there. It might be useful to abstract reusable UI components that require multiple steps to do an action like dropdowns/autocomplete/multi-select, but there's not much use abstracting page-specific things that aren't queried from elsewhere.

spec/helpers/content_profiles.ts
```diff
@@ -153,6 +158,49 @@ class ContentProfiles {
...
+            }
+        }
+
+        /**
+         * Returns the list of content profiles.
+        */
+        this.totalProfiles = () => {
+            const dropdownMenu = element(by.id('select-content-profile'));
+
+            return dropdownMenu.all(by.repeater('profile in aggregate.activeProfiles'));
```
tomaskikutis: Only query DOM elements using `data-test-id` markers for testing. Don't query by ids, class names, repeaters, etc. See testing section at https://github.com/superdesk/superdesk-client-core/blob/develop/CONTRIBUTING.md

There are also helpers easily get elements by test-id like `el(['monitoring', 'clear-filters]).click()`. You can take a look at `spec/internal_destinations_spec.ts` for examples of how to use it.

spec/monitoring_spec.ts
```diff
@@ -403,20 +402,24 @@ describe('monitoring', () => {
...
         expect(monitoring.getTextItem(4, 0)).toBe('item4');
-        contentProfiles.openContentProfileDropdown();
-        expect(contentProfiles.totalProfiles().count()).toBe(2);
-        contentProfiles.selectContentProfile();
-        contentProfiles.clearFilters();
+
+        el(['content-profile-dropdown']).click();
+        browser.wait(ECE.hasElementCount(items, 2));
+        items.get(1).click();
+        expect(monitoring.getTextItemBySlugline(0, 0)).toBe('STORY1 SLUGLINE');
```
tomaskikutis: I'm not getting your test. You apply filters, check for condition, remove filters and check for the same condition to be true. Shouldn't the results be different with/without filters?

spec/monitoring_spec.ts
```diff
@@ -403,20 +402,24 @@ describe('monitoring', () => {
...
         expect(monitoring.getTextItem(2, 0)).toBe('item3');
         expect(monitoring.getTextItem(4, 0)).toBe('item4');
-        contentProfiles.openContentProfileDropdown();
-        expect(contentProfiles.totalProfiles().count()).toBe(2);
-        contentProfiles.selectContentProfile();
-        contentProfiles.clearFilters();
+
+        el(['content-profile-dropdown']).click();
+        browser.wait(ECE.hasElementCount(items, 2));
+        items.get(1).click();
```
tomaskikutis: Are you selecting the same item as on line 409? It'd be clearer if you selected by button text. like

`el(['content-profiles'], by.buttonText('content profile 1')).click()`

scripts/apps/monitoring/tests/monitoring.spec.ts
tomaskikutis: Can you remove these since you added e2e tests?

scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -65,9 +81,9 @@ <h3 class="subnav__page-title sd-flex-no-grow sd-empty">
                             <span class="stage-group__desk">{{ monitoring.singleGroup.header }}</span>
                             <span class="stage-group__split">/</span>
                             <div class="dropdown" dropdown>
-                                <button class="dropdown__toggle navbtn navbtn--text-only navbtn--padding-s" dropdown-toggle>{{ monitoring.singleGroup.subheader}} {{ monitoring.singleGroup.type | splitText | translate }}<span class="dropdown__caret"></span></button>
+                                <button class="dropdown__toggle navbtn navbtn--text-only navbtn--padding-s" dropdown-toggle>{{ monitoring.singleGroup.subheader}} {{ gettext(monitoring.singleGroup.type) | splitText }}<span class="dropdown__caret"></span></button>
```
tomaskikutis: Don't use `gettext` on a dynamic value. It should come here already translated. Translate it where it's being set to a static string.

scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -109,35 +125,34 @@ <h3 class="subnav__page-title sd-flex-no-grow sd-empty">
...
                     </div>
-    
+
                     <div ng-hide="hideMonitoringToolbar2 === true" sd-media-query min-width="700" class="subnav">
             
                         <!-- filtering by file type not implemented for usage with custom data source -->
                         <div ng-if="(!elementState || elementState === 'comfort') && customDataSource == null" class="button-list button-list--padded">
-                            <a href="" ng-repeat="fileType in aggregate.fileTypes" ng-click="aggregate.setFileType(fileType)" class="toggle-button" ng-class="{'toggle-button--active': aggregate.hasFileType(fileType)}" title="{{fileType| translate}}">
-                                <span ng-if="fileType === 'all'" class="toggle-button__text toggle-button__text--all">{{:: 'all' | translate}}</span>
+                            <a href="" ng-repeat="fileType in aggregate.fileTypes" ng-click="aggregate.setFilterType('file', fileType)" class="toggle-button" ng-class="{'toggle-button--active': aggregate.hasFileType(fileType)}" title="{{ gettext(fileType) }}">
```
tomaskikutis: don't pass dynamic values to `gettext`. I know it's been here like that, but if you are working on this part, it'd be good to fix it

scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -109,35 +125,34 @@ <h3 class="subnav__page-title sd-flex-no-grow sd-empty">
...
+                        <div class="dropdown dropdown--align-left dropdown--compact-state" dropdown  ng-if="elementState === 'compact'" sd-tooltip="{{ gettext('Filter file types') }}" flow="right">
                                 <button class="navbtn dropdown__toggle" dropdown__toggle><i class="icon-dots"></i></button>
                                 <ul class="dropdown__menu">
-                                    <li><span class="dropdown__menu-label">{{:: 'File types' | translate}}</span></li>
+                                    <li><span class="dropdown__menu-label">{{gettext('File types') }}</span></li>
                                     <li ng-repeat="fileType in aggregate.fileTypes"
                                         ng-click="aggregate.setFileType(fileType)"
                                         ng-class="{'dropdown__menu-item--active': aggregate.hasFileType(fileType)}"
-                                        title="{{fileType| translate}}">
+                                        title="{{ gettext(fileType) }}">
```
tomaskikutis: don't pass a dynamic value to `gettext`

spec/monitoring_spec.ts
```diff
@@ -402,24 +402,27 @@ describe('monitoring', () => {
...
+        expect(monitoring.getAllItems().count()).toBe(3);
         el(['content-profile-dropdown']).click();
-        var items = els(['content-profiles']);
-
-        browser.wait(ECE.hasElementCount(items, 2));
-        items.get(1).click();
+        browser.wait(ECE.hasElementCount(els(['content-profiles']), 2));
+        el(['content-profile-dropdown'], by.buttonText('Simple')).click();
+        expect(monitoring.getAllItems().count()).toBe(1);
         expect(monitoring.getTextItemBySlugline(0, 0)).toBe('STORY1 SLUGLINE');
```
tomaskikutis: Looks like this story is always there. It'd be more useful to check the presence of an item which is supposed to be filtered out.

spec/monitoring_spec.ts
```diff
@@ -408,6 +408,8 @@ describe('monitoring', () => {
         el(['content-profile-dropdown'], by.buttonText('Simple')).click();
         expect(monitoring.getAllItems().count()).toBe(1);
         expect(monitoring.getTextItemBySlugline(0, 0)).toBe('STORY1 SLUGLINE');
+        expect(browser.isElementPresent(monitoring.getGroupItems(2).get(0))).toBe(false);
```
tomaskikutis: What is this for? The test could be simpler and more readable if you were more explicit and queried items by some identifier like slugline instead of index.

For example, there are `data-test-id="article-item"` and `data-test-id="field--slugline"` which you can use.

GyanP: @tomaskikutis  it's there to check there are no items present in rest of the groups other than the one with slugline `STORY1 SLUGLINE`. There are existing items, which does not contain a slugline, so it's hard to check there presence.

tomaskikutis: Could you use a helper like `monitoring.isGroupEmpty`? It would make your intention clearer.

spec/helpers/monitoring.ts
```diff
@@ -457,7 +458,10 @@ class Monitoring {
             var itemElem = this.getSpikedItem(item);
 
             browser.actions().mouseMove(itemElem).perform();
-            itemElem.element(by.className('icon-dots-vertical')).click();
+            var dotsElem = itemElem.element(by.className('icon-dots-vertical'));
```
tomaskikutis: Use `data-test-id` for selectors.

Can you also update `executeContextMenuAction` function and use it here?

scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -175,12 +190,24 @@ <h3 class="subnav__page-title sd-flex-no-grow sd-empty">
...
                         </div>
-            
+                    </div>
+                    <!-- Active filter tags -->
+                    <div class="search-tags" ng-if="aggregate.activeFilters.contentProfile.length > 0">
+                        <ul class="search-tags__list">
+                            <li ng-repeat="(type, keys) in aggregate.activeFilterTags">
+                            <span>{{ type | translate }}:
+                                <div ng-repeat="key in keys" class="tag-label tag-label--highlight2">
+                                    </span>{{ key.label }}
```
petrjasek: the issue I can see locally is this markup - you open span, then open div, then close span, and then div continues.
you can see the wrong output if you run `grunt build` rather than `grunt server` or on [PR test instance](https://sdcpr-3271.test.superdesk.org/)

scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -175,12 +194,28 @@ <h3 class="subnav__page-title sd-flex-no-grow sd-empty">
...
                                 </li>
                             </ul>
                         </div>
-            
+                    </div>
+                    <!-- Active filter tags -->
+                    <div class="search-tags" ng-if="aggregate.activeFilters.contentProfile.length > 0">
+                        <ul class="search-tags__list">
+                            <li ng-repeat="(type, keys) in aggregate.activeFilterTags">
+                                <span>{{ type | translate }}:</span>
```
petrjasek: not sure about this `| translate` here

tomaskikutis: It'd be good to track down where it's coming from, ensure that the string is getting extracted there and remove the filter here.

It's possible to do `var notExtractedHere = 'test str'` and `<div>{{notExtractedHere | translate}}</div>` which would do the translation only if the string is extracted in another place which we shouldn't be relying on.

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -521,4 +559,20 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
...
     });
+
+    function getActiveProfiles() {
+        content.getTypes(false).then((profiles) => {
+            self.loading = false;
+            self.activeProfiles = profiles;
+
+            // initialize the activeFilterTags once the activeProfiles are available
+            if (self.activeFilters.contentProfile.length > 0) {
+                self.activeFilterTags['Content profile'] = self.activeFilters.contentProfile.map((filter) => {
```
petrjasek: here the Content profile is not translated while it is previously, would be good to store it as constant translated and use that

PR #3272, opened 2020-01-08
let user skip ingest config test if needed

petrjasek:
by default the switch is not visible, but once there is
an error or it was set it stays there.

SDESK-4853


MarkLark86:
@petrjasek This is much better thanks. Although, why would this toggle be hidden by default? If a user is setting up an ingest source they know will fail the config test they will need to:

1. click `SAVE` button
2. then click the `Skip config test` toggle
3. then click the `SAVE` button again

I see this happening when setting up file feed where mount is not accessible on the API server, or preparing for an API/FTP ingest where the API or FTP source is not currently available (where the user can enable the ingest/config test when they're ready to go live with the ingest)

petrjasek:
yep exactly, I wanted to make it a bit harder to skip it so users won't just check it right away. but not sure about the usability aspect of it, maybe will just keep it there somewhere under the config

PR #3275, opened 2020-01-10
FIX: The closing 'x' button in 'edit metadata' controls does not work. [SDESK-4919]

GyanP:


scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -73,6 +73,9 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, $q) {
     $scope.crops = {
         isDirty: false,
     };
+    $scope.toggleMetadataPanel = (value?) => {
```
petrjasek: would rather call it `toggleShowMetadata` so it's more obvious what it toggles

scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -73,6 +73,9 @@ export function ChangeImageController($scope, notify, _, api, $rootScope, $q) {
     $scope.crops = {
         isDirty: false,
     };
+    $scope.toggleShowMetadata = (value?) => {
+        $scope.showMetadata = value != null ? value : !$scope.showMetadata;
```
tomaskikutis: Any reason why conditional handling exists here? The places that call this function without value have "Close details" /  "Open metadata" tooltips. Why not pass true/false and get rid of the conditional here?

PR #3276, opened 2020-01-13
User activity widget

pablopunk:
SDBELGA-206

- [x] Hide configuration button for the widget
- [x] SVG for the widget
- [x] Filters for created items:
  - [x] Moved by the user to an incoming stage
- [x] Filters for 'moved to a working stage by the user':
  - [x] Moved to another stage afterwards
- [x] Filters for both above:
  - [x] Locked by another user
  - [x] Marked for another user
  - [x] Marked for a desk
  - [x] Published or scheduled for publishing


scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -0,0 +1,411 @@
...
+            this.setState({groupsData: data, loading: false});
+        });
+    }
+
+    renderGroup(group: IGroup) {
+        const {groups, groupsData} = this.state;
+        const data = groupsData.find((g) => g.id === group.id);
+
+        if (!data) {
+            console.warn(
```
tomaskikutis: It shouldn't happen, right? If so, use `logger.warn` instead, so it gets sent to sentry.

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -0,0 +1,411 @@
...
+                `Tried to render group '${group.id}' but no data was found`,
+            );
+            return null;
+        }
+
+        return (
+            <>
```
tomaskikutis: What is this for?

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -0,0 +1,411 @@
...
+                <div className="stage">
+                    <div className="stage-header">
+                        <button
+                            className={`stage-header__toggle ${
+                                group.collapsed ? 'closed' : ''
+                            }`}
+                            onClick={() => {
+                                const newGroups = groups.map((g) => {
+                                    if (g.id === group.id) {
+                                        g.collapsed = !g.collapsed;
```
tomaskikutis: don't mutate React state

```return {...g, collapsed: !g.collapsed}```

File: scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx [hunk 4]
```diff
@@ -0,0 +1,411 @@
...
+                        </button>
+                        <span className="stage-header__name">
+                            {group.label}
+                        </span>
+                        <div className="stage-header__line" />
+                        <span className="label-total stage-header__number">
+                            {data.itemIds.length}
+                        </span>
+                    </div>
+                    {group.collapsed || (
```
tomaskikutis: should it render when collapsed or not collapsed? Can you be more explicit instead?

```group.collapsed === true ? null : (<div>...</div>)```

File: scripts/apps/search/components/WidgetItemList.tsx [hunk 5]
```diff
@@ -1,86 +1,75 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import {WidgetItem} from './index';
 import {gettext} from 'core/utils';
+import {IArticle} from 'superdesk-api';
+
+interface IProps {
+    allowed?: boolean;
```
tomaskikutis: what is allowed?

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -0,0 +1,411 @@
...
+                                        },
+                                    );
+                                }}
+                            />
+                        </form>
+                    </div>
+                    {
+                        loading && <div className="item-group__loading" />
+                    }
+                    {this.state.user && this.state.groupsData && (
```
tomaskikutis: You shouldn't display old data while new data is loading and a different user is already selected.

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -0,0 +1,411 @@
...
+    refreshItems() {
+        if (this.state.user) {
+            this.fetchGroupsData();
+        }
+    }
+
+    async fetchItems(group: IGroup) {
+        const {api, search} = this.services;
+        const {_items} = typeof group.query === 'function'
+            ? await group.query((repo, filters) => genericFetch({search, api}, repo, filters))
```
tomaskikutis: I see you are trying to sneak in async/await again :smile: We aren't using async/await in the codebase yet and I think it's simpler to keep it this way. We could consider it if it added significant value like flattened a complex promise chain, but it's [almost the same in this case](https://user-images.githubusercontent.com/2076953/72738463-ea9aac00-3ba1-11ea-8753-ccf0f40f511c.png)

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -0,0 +1,411 @@
...
+
+    criteria.repo = repo;
+
+    return api.query('search', criteria);
+}
+
+function getQueryMarkedForUser(userId) {
+    return {
+        term: {
+            marked_for_user: userId,
```
tomaskikutis: Up until now everything related to mark for user feature was contained in the extension and I'd like to keep it contained as much as possible.

Ideally, this widget would be an extension(I wouldn't convert it yet if it's a lot of work) and would expose a method like
```registerActivityType('Marked for user', () => {marked_for_user: user._id})```
which we could call when mark for user extension is initialized and it would register the filter.

The problem is `'Created by this user'` condition which is combining a bunch of other different conditions. In this case, could you try [exposing the queries from the extension](https://user-images.githubusercontent.com/2076953/72739943-fcca1980-3ba4-11ea-861b-a0210f93ee09.png)?


You'd need to modify [the IExtension type](https://user-images.githubusercontent.com/2076953/72740070-3c910100-3ba5-11ea-8cf0-85ba26de1380.png).

File: scripts/apps/monitoring/directives/WidgetGroup.ts [hunk 9]
```diff
@@ -288,7 +288,7 @@ export function WidgetGroup(search, api, superdesk, desks, cards, $timeout, $q,
                 }
 
                 if (container.scrollTop + container.offsetHeight >= container.scrollHeight - 3 &&
-                        lastScrollTop <= container.scrollTop) {
+                    lastScrollTop <= container.scrollTop) {
```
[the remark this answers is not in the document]
pablopunk: sure, I didn't do this, I guess it was my code formatter? I'll move the bracket

scripts/core/ui/components/select2.tsx
```diff
@@ -103,8 +103,8 @@ export class Select2<T> extends React.Component<IProps<T>, IState> {
                                 this.props.loading === true
                                     ? <div style={{padding: 10}}>{gettext('Loading...')}</div>
                                     : items.length < 1
-                                    ? <div style={{padding: 10}}>{gettext('No items found.')}</div>
-                                    : items
+                                        ? <div style={{padding: 10}}>{gettext('No items found.')}</div>
```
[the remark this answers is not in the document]
pablopunk: Also my formatter....

scripts/core/ui/components/SelectUser.tsx
```diff
@@ -72,7 +72,7 @@ export class SelectUser extends React.Component<IPropsSelectUser, IState> {
 
         return (
             <Select2
-                onFocus={true}
+                onFocus={this.props.focus != null ? this.props.focus : true}
```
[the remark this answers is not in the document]
pablopunk: I think I'm gonna ditch this and merge your changes into my PR

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -0,0 +1,411 @@
...
+    componentDidMount() {
+        [
+            'item:lock',
+            'item:unlock',
+            'item:spike',
+            'item:move',
+            'item:publish',
+            'item:update',
+        ].map((event) =>
+            this.services.$rootScope.$on(event, this.refreshItems.bind(this)),
```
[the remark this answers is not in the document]
pablopunk: I thought about it, but not sure how to do it. Would this work?

```ts
$rootScope.$off(event, this.refreshItems)
```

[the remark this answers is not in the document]
tomaskikutis: [screenshot_18](https://user-images.githubusercontent.com/2076953/72790442-e3bd7900-3c35-11ea-9bdd-db8d481570d1.png) yep. You need to do it the same way for every event, except you have to move out the binding because biding creates a new function reference and removing existing listeners wouldn't work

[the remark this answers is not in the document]
tomaskikutis: woops, I forgot it's an angular scope, not js events. I've looked at the docs and didn't see $off method. Looks like calling $on returns a function which you need to store and call when you want to remove the listener https://stackoverflow.com/a/14898795/1175593

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
pablopunk: I'll get rid of it, I had it when I had two divs as the return (so I had to group them into one component)

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
pablopunk: oops! It wasn't part of the state before, but now you're right

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
tomaskikutis: I see where you are coming from and even half-agree, but in the context that we are using asynchronous code*, the difference is quite small, plus we are more familiar to promises so I'd stick to it until we bump into use cases where using fetch would make a significant positive difference.

\* async/await probably helps more on the server-side due to the nature of the domain

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
tomaskikutis: It would be coupled to the extension, but it would be coupled explicitly and only to methods exposed by the extension. It would make discovering dependencies easy as well as removing the feature without leaving dead code since the compiler would raise errors.

It'd be a much better situation than it is with current features. If I asked you to remove "mark for desk" feature and make sure you don't leave dead code it would be quite a grep challenge with a "maybe" outcome. This is what I want to prevent by making dependencies explicit.

package.json
```diff
@@ -161,6 +161,7 @@
     "start-test-server": "cd test-server && python3 -m venv env && . env/bin/activate && pip install -Ur requirements.txt && honcho start",
     "protractor": "protractor protractor.conf.js",
     "webdriver-manager": "webdriver-manager update --gecko false --standalone false",
-    "server": "grunt server"
+    "server": "grunt server",
```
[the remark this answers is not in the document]
pablopunk: Not really, but `npm run dev` is a standard-ish for JS devs, that's why I added it. I didn't remove `npm run server` cause I guess someone else is using it

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -0,0 +1,411 @@
...
+import {IArticle, IUser, IStage} from 'superdesk-api';
+import {AuthoringWorkspaceService} from 'apps/authoring/authoring/services';
+import ng from 'core/services/ng';
+import {gettext} from 'core/utils';
+import {SelectUser} from 'core/ui/components/SelectUser';
+
+interface IGroup {
+    id: string;
+    label: string;
+    repo?: string; // not used if self.query is a function
```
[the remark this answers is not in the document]
pablopunk: but you __can__ have both the repo and the query object. Like [here](https://github.com/superdesk/superdesk-client-core/pull/3276/files#diff-d1f1fc0ae3a5fe36b6f66f5c612035abR119-R120)

scripts/apps/search/components/WidgetItemList.tsx
pablopunk: I didn't change anything about that functionality, I just declared it as a type. We might run into errors if I don't find all the references to it (and it's not on the scope of this task)

package.json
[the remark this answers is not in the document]
tomaskikutis: Can you make the commands identical so it's clear that they are aliases?

scripts/apps/search/components/WidgetItemList.tsx
tomaskikutis: What do you suggest, to create a JIRA task for renaming a variable? There are only 2 usages of the component, you should be able to fix it in 10 minutes or so.

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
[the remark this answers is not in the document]
tomaskikutis: If the `repo` is not used when `query` is defined, why allow setting both at the same time? Shouldn't it be either `repo` or `query` but not both?

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -224,21 +225,27 @@ export default class UserActivityWidget extends React.Component<{}, IState> {
...
             'item:publish',
             'item:update',
         ].map((event) =>
-            this.services.$rootScope.$on(event, this.refreshItems.bind(this)),
+            this.services.$rootScope.$on(event, this.refreshItems),
         );
     }
 
+    componentWillUnmount() {
+        this.unbindListeners.map((unbind) => unbind());
```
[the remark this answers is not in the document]
pablopunk: I know but is there any drawback for using map? I like to stick to just one  method

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -224,21 +225,27 @@ export default class UserActivityWidget extends React.Component<{}, IState> {
             groupsData: null,
             loading: false,
         };
+
+        this.refreshItems = this.refreshItems.bind(this);
     }
 
     componentDidMount() {
-        [
+        this.unbindListeners = [
```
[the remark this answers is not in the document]
pablopunk: Sure. I guess I used unbind because is also used in event handling in libraries like jQuery

package.json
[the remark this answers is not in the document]
pablopunk: I made it so one calls the other so in case we change "server", "dev" gets the changes too

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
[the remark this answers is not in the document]
pablopunk: `repo` is not used when `query` is a function. `query` should always be defined, but when it's a function, the `repo` value is ignored.

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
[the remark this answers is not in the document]
tomaskikutis: There is, you are making your intention in the code less clear. It's like checking if the string is empty by using `!!myVar` and not indicating that a variable is a string.

When I find such code(someone using map instead of forEach) I start wondering whether the return value was used before and someone accidentally removed it.

Sticking to one method is the silliest reason I've heard :smile:

scripts/apps/search/components/WidgetItemList.tsx
pablopunk: well there are 2 react components (this one and the child) that have that prop. I can change those easily. I meant that value is coming from a `scope.allowed` and that can get messy. But I'll change the components tho

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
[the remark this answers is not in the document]
pablopunk: Well in programming using the result of a expression or not is independent of what the function does, and for me they both do the same. Sometimes I do want the value of the expression and sometimes I don't. And `forEach` only gives you the functionality and not the value, that's what I meant by sticking to one method, because it does both. Anyway I can change it to forEach cause I don't care that much to argue about it

scripts/apps/search/components/WidgetItemList.tsx
tomaskikutis: My bad, I didn't notice `scope.allowed`. I've just checked the angular part and found that  `WidgetGroup` is getting registered as a directive which is used in only 3 places and doesn't have values coming from another variable with the same name.

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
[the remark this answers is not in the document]
tomaskikutis: > Well in programming using the result of a expression or not is independent of what the function does

True, in terms of the output using `map` will get the same result, but it's equally as important that humans understand the code, not only machines. For example, I was wondering whether by using `map` you intended to reassign returned value `this.removeListeners`, but forgot or something. I see that the component is getting destroyed and reassigning wouldn't do anything, but I can't be sure that you noticed the same. I can ask you questions now, but after some time you might not even remember it yourself.

My point is not to argue but to try to convince you that it's an issue that causes confusion. This one is pretty minor, but following the same pattern, one could also argue for depending on the side-effects of functions using the same "it works anyway" argument.

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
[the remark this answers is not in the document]
tomaskikutis: We are going in circles :slightly_smiling_face:

```
{
    id: 'locked',
    label: gettext('Locked by this user'),
    repo: 'archive',
    query: {...getQueryLockedByUser(userId)},
},

If it has a `query` and `repo` won't be used, why define it?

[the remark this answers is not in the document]
Comment (pablopunk) [reply scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx]: Right there, `repo` is not ignored because `query` is an object, not a function.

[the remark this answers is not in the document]
Comment (tomaskikutis) [reply scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx]: I get it now. How about `dataSource: {repo: string; query: object} | {query: () => Promise<SomeType>}`. You can use the help of the compiler to enforce your logic so it's not possible to pass an invalid combination like a function for query and a repo too.

File: scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx [hunk 17]
```diff
@@ -6,12 +6,14 @@ import ng from 'core/services/ng';
...
 import {SelectUser} from 'core/ui/components/SelectUser';
 import {logger} from 'core/services/logger';
+import {extensions} from 'appConfig';
 
 type FetchFunction = (repo: string, criteria: any) => Promise<any>;
 
 interface IGroup {
     id: string;
     label: string;
+    precondition?: () => boolean;
```
tomaskikutis: Is there a reason you need a precondition? What about using an `if` statement inside `GET_GROUPS`?

```
const GET_GROUPS = () => {
    const markForUserExtensionEnabled = extensions['markForUser'] != null;

    const groups = [
        {
            id: 'created',
            label: gettext('Created by this user'),
            dataSource(fetchFn) {
                return fetchFn('archive', {
                    bool: {
                        must: [
                            // ...
                            {...(markForUserExtensionEnabled ? getMarkedQuery() : {})},
                        ],
                    },
                });
            },
        },
    ];

    if (markForUserExtensionEnabled) {
        groups.push({
            id: 'marked',
            label: gettext('Marked for this user'),
            // ...
        });
    }

    return groups;
};
```

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -126,21 +102,34 @@ const GET_GROUPS = (userId, services: any): Array<IGroup> => {
         {
             id: 'marked',
             label: gettext('Marked for this user'),
-            dataSource: {
-                repo: 'archive',
-                query: {...getQueryMarkedForUser(userId)},
+            precondition() {
+                return Object.values(extensions)
+                    .some(({extension}) =>
```
tomaskikutis: You should be getting the extension by ID and operating on that single extension. It's not valid for other extensions to expose mark-for-user related methods, but by reading your code one would think that it is, since you are reading the method from all the extensions.

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
pablopunk: I think it's cleaner having a precondition. You could have another group that you don't want to show for example depending on if the user is admin, or something like that. Then I would add all the logic also in GET_GROUPS which IMO is not as clean.

tomaskikutis: You can check for admin rights or permissions using another conditional inside `GET_GROUPS`. How does a precondition make it cleaner?

I'd argue it makes it dirtier. Instead of handling it in a single place, you are now handling it in 2 and have to introduce another member to the interface which you wouldn't need otherwise without no benefits I can see.

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
```diff
@@ -114,7 +114,10 @@ const GET_GROUPS = (userId, services: any): Array<IGroup> => {
             id: 'created',
             label: gettext('Created by this user'),
             dataSource(fetchFn) {
-                const markedQuery = extensions['markForUser'].extension.exposes.getQueryNotMarkedForAnyoneOrMarkedForMe(userId);
+                const markedQuery = extensions['markForUser']
```
tomaskikutis: what if the extension is not registered, won't it crash with `exposes is not a function of undefined`?

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
pablopunk: That check is tight to one group. I'm keeping that group's logic inside the interface. If I add another group with another precondition I just need to add another object, not any code in the GET_GROUPS which is just a function to initialize the groups. Your solution can clutter the GET_GROPS function with logic of each group you add. I don't see any benefits but I see drawbacks.

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
pablopunk: Right, forgot the optional chaining

scripts/apps/dashboard/user-activity/components/UserActivityWidget.tsx
tomaskikutis: The benefit of using an `if` statement is that the logic is kept in a single place and other parts of the code don't know about it. It translates to being easier to understand and faster to change the code. Not by much too be fair since the file is quite small.

> That check is tight to one group. I'm keeping that group's logic inside the interface.

I get your point about keeping related code together, but it's also kept together when using `ifs`. If you needed to repeat the same if somewhere else, I'd suggest going with the precondition.

> If I add another group with another precondition I just need to add another object, not any code in the GET_GROUPS

I don't see much difference between having to [add an `if` inside the function or an inline function to the group object](https://user-images.githubusercontent.com/2076953/73192409-4678ae00-4129-11ea-98ba-ddb762202d94.png).

> Your solution can clutter the GET_GROPS function with logic of each group you add.

Adding an `if` is 2 lines per group which isn't that different to space that would be required for precondition function.

PR #3280, opened 2020-01-15
Remove edit button from related items. [SDESK-4916]

GyanP:


scripts/apps/archive/directives/RelatedItemsPreview.ts
```diff
@@ -16,6 +16,7 @@ export function RelatedItemsPreview(relationsService) {
         scope: {
             item: '=',
             field: '<',
+            showActionMenu: '=?',
```
tomaskikutis: when should it not be shown?

GyanP: When you open the preview of an article.

tomaskikutis: why not show it in the preview?

GyanP: Well, I did ask Karel, and he said:
> nope, the preview is a preview. I think there is no need to have the actions there

tomaskikutis: Can you, in this case, pass a `preview: boolean` instead of `showActionMenu` and hide the menu based on that? It's easier to understand the reasoning this way without digging too much into the code.

GyanP:
Merging does work, but I usually do rebase just to keep my commits together on top.

tomaskikutis:
You can use `--squash` option so merge changes get added as a new commit at the bottom. Rebasing rewrites history so I can no longer differentiate between changes I have reviewed and new changes. I understand that you left the same commits, but a mistake could have been made and new code added under the seemingly old commits. Try only using rebasing as a last resort for when it's hard to resolve conflicts.

PR #3281, opened 2020-01-15
Change UI config for related items

pablopunk:
Only for QA purposes, don't review/merge


superdesk.config.js
```diff
@@ -30,5 +30,50 @@ module.exports = function(grunt) {
...
+                'updated',
+                'category',
+                'provider',
+                'expiry',
+                'desk',
+                'fetchedDesk',
+                'nestedlink',
+                'associatedItems',
+            ],
+            relatedItems: {
```
tomaskikutis: I assume `relatedItems` section doesn't impact the general list. Can you move it up in the config hierarchy so it's at the same level as `list`?

pablopunk: This PR is only for QA purposes, I'll take that in mind if I need to make a PR to fix something in the task

PR #3282, opened 2020-01-15
Make content filters searchable in internal destinations

tomaskikutis:
SDESK-4697


scripts/core/ui/components/select2.tsx
```diff
@@ -86,112 +89,166 @@ export class Select2<T> extends React.Component<IProps<T>, IState> {
         this.search = throttle(searchFn, 300, {leading: false});
     }
 
+    componentDidMount() {
+        // capture all scroll events and close autocomplete on scroll
+        // unless that scroll event is coming from the autocomplete itself
+        window.addEventListener('scroll', (e) => {
```
petrjasek: can't find where is this listener removed

scripts/core/ui/components/generic-form/input-types/select_single_value_autocomplete.tsx
```diff
@@ -0,0 +1,101 @@
...
+    item?: any;
+}
+
+export function getSelectSingleValueAutoComplete(
+    query: (searchString: string, props: IProps) => Promise<IRestApiResponse<any>>,
+    queryById: (id: string) => Promise<any>,
+    getPlaceholder: (props: IProps) => string,
+    getLabel: (item) => string,
+    getDependentFields?: (props: IProps) => Array<string>,
+) {
```
petrjasek: just wondering why having this factory instead of those being props of a component?

tomaskikutis: because I need the component to conform to a [common interface](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/ui/components/generic-form/form-field.tsx#L17)

scripts/core/ui/components/select2.tsx
tomaskikutis: forgot it :upside_down_face:

PR #3284, opened 2020-01-16
fetch all internal destinations

tomaskikutis:
SDESK-4697


tomaskikutis:
`fetchPage` function is internal and is not being exported. It's a part of the implementation so it makes sense to me to include implementation-specific details in its name.

Regarding `fetchAllPages`, handling pages is the main objective of the function so having it in the name makes it easier to understand in my opinion.

petrjasek:
`fetchPage` is fine, but `fetchAllPages` seems to me like something to fetch all items for a resource, pagination handling is implementation detail, if api returns everything via single call would we rename it everywhere?

tomaskikutis:
If API returns everything via a single call we delete both functions and use `dataApi`. Handling pagination is the only point of the code.

Let me know if you aren't convinced and I'll rename it :)

petrjasek:
I'm still not convinced, I would probably use some param on the query to fetch everything to make it explicit, so would keep such method. there will be always pagination for certain resources so will be good to distinguish these

tomaskikutis:
I would add the parameter to [dataApi](https://user-images.githubusercontent.com/2076953/72624096-95af2980-3946-11ea-90a6-a47e337830d3.png), but I don't see issues with your approach either :slightly_smiling_face:

PR #3286, opened 2020-01-16
[SDBELGA-255] More visual indication when items are marked for me

fritzSF:


scripts/core/editor3/styles.scss
```diff
@@ -475,15 +475,43 @@ $editor-styleButton-size: 3rem;
 	padding: 10px 20px;
 }
 
+:root {
+	--preCharacterLimitt:40ch;
```
tomaskikutis: You should scope the variable to the component instead of having it global. Since there is only one usage I wouldn't have it as a variable at all and put it inline. By the way, there's a typo in the word `limit`.

scripts/core/editor3/styles.scss
```diff
@@ -475,15 +475,43 @@ $editor-styleButton-size: 3rem;
...
 	}
+	pre {
+		white-space: pre !important;
+		padding-right: 2ch;
+		.public-DraftStyleDefault-block {
+			white-space: inherit;
+		}
+	}
+	&.showCharacterLimit {
+		&::after {
```
tomaskikutis: It should probably be only a single colon for pseudo-selectors like `&:after {}`, `&:before {}` shouldn't it? You can also write it inline to avoid nesting `&.showCharacterLimit:after {}`

scripts/core/menu/styles/menu.scss
```diff
@@ -15,18 +15,27 @@
...
         &.stable {
-            background: #111;
-            color: #777;
+            color: rgba(255,255,255,0.4);
+            border-color: rgba(255,255,255,0.2);
         }
         &.beta {
-            background: #009360;
-            color: #fff;
+            //background: #009360;
```
tomaskikutis: do you need this?

scripts/core/menu/styles/menu.scss
```diff
@@ -323,6 +340,42 @@
...
+    padding-left: 1.6rem;
+    padding-right: 1.6rem;
+    z-index: 2;
+    color: $white;
+    background-color: rgba(255,255,255,0.06);
+    transition: background-color ease-in-out 0.2s;
+    &:hover {
+        background-color: rgba(255,255,255,0.1);
+    }
+    &--active {
```
tomaskikutis: Can you avoid dynamic class name generation? It's much easier to find usages of class names when you can search for the class name you see in the markup.

scripts/core/menu/styles/menu.scss
fritzSF: No. Removed it

tomaskikutis:
@fritzSF don't rebase your branch after you have submitted the pull request, because it makes it impossible to see the changes you did since I last reviewed the PR

scripts/core/editor3/styles.scss
fritzSF: I'm leaving the variable as it is. We might need it globally as it should be configurable through a config. If it will be implemented differently we'll get rid of the variable completelly.

scripts/core/menu/styles/menu.scss
fritzSF: We use BEM when writing CSS classes so it's always possible to search by block name (e.g. top-menu-dropdown-button in this case)

tomaskikutis: Checking by block makes lookup take a few minutes instead of a few seconds. While it's possible, it's really inefficient.

PR #3287, opened 2020-01-17
[SDESK-4861] Migrate functionPoints to extensions API

MarkLark86:


MarkLark86:
Creating another PR to fix this issue

PR #3288, opened 2020-01-17
Improve management of vocabularies

tomaskikutis:
SDESK-4832


pablopunk:
I think there's an issue with the styles for the tabs

Kapture 2020-01-24 at 12 17 59

tomaskikutis:
Yeah, there are many things wrong with styles, like the toolbar for searching/sorting/filtering. I only implemented the functional part. Nemanja will be styling it. Forgot to add a WIP label.

PR #3290, opened 2020-01-20
initialize selecting users with options collapsed in order not to cover modal footer

tomaskikutis:
SDESK-4877

### DIFF
--- a/scripts/core/ui/components/select2.tsx
+++ b/scripts/core/ui/components/select2.tsx
@@ -15,7 +15,9 @@ interface IProps<T> {
     getItemValue(item: T): string;
     onSelect(value: string): void;
     onSearch?(search: string): Promise<any>;
-    onFocus?: boolean | {hideOptions: boolean};
+
+    // dropdown may be hidden until user starts typing in order to prevent it covering other UI elements
+    autoFocus?: boolean | {initializeWithDropdownHidden: boolean};
     'data-test-id'?: string;
 }
 
@@ -77,7 +79,7 @@ export class Select2<T> extends React.Component<IProps<T>, IState> {
 
         this.state = {
             search: '',
-            isOpen: this.props.onFocus != null,
+            isOpen: this.props.autoFocus != null,
             justInitialized: true,
         };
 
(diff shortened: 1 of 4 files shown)


pablopunk:
I did something similar on #3276, I guess we can get this one merged and I'll undo the changes and merge develop once this is merged

tomaskikutis:
I think yours was about improving the styles like making it stretch to `100%`. We'll probably need to combine the changes.

pablopunk:
I meant for the SelectUser component https://github.com/superdesk/superdesk-client-core/pull/3276/files#diff-826f0d1d82996f6fa51ac07311b81b56R75

I made a similar change

scripts/extensions/markForUser/src/get-mark-for-user-modal.tsx
```diff
@@ -65,6 +65,7 @@ export function getMarkForUserModal(
                             disabled={locked}
                             onSelect={(selectedUser) => this.setState({selectedUserId: selectedUser._id})}
                             selectedUserId={this.state.selectedUserId}
+                            onFocus={{hideOptions: true}}
```
pablopunk: Can't we make it so it's open by default and closed when the item is already marked for a user?

tomaskikutis: Nope. Users want to be able to quickly select another user to mark for when they click the avatar or "mark for another user". I've added an option to `SelectUser` so you can choose whether it auto-expands or not depending on where you are rendering it.

tomaskikutis:
> I meant for the SelectUser component https://github.com/superdesk/superdesk-client-core/pull/3276/files#diff-826f0d1d82996f6fa51ac07311b81b56R75
> 
> I made a similar change

Sure then :)

scripts/core/ui/components/select2.tsx
```diff
@@ -14,14 +14,15 @@ interface IProps<T> {
     renderItem(item: T): JSX.Element;
     getItemValue(item: T): string;
     onSelect(value: string): void;
-    onSearch?(search: string): void;
-    onFocus: boolean;
+    onSearch?(search: string): Promise<any>;
+    onFocus?: boolean | {hideOptions: boolean};
```
petrjasek: imo it can use some renaming, we usually have callbacks named `onSomething` and here it's bool or dict, and I'm not exactly sure what they do

PR #3291, opened 2020-01-20
SDBELGA-267 [EDITOR3] Related images from Belga Image database shouldn't be editable for the users.

ride90:
⚠️ editor2 logic was not touched

### DIFF
--- a/scripts/apps/authoring/views/item-association.html
+++ b/scripts/apps/authoring/views/item-association.html
@@ -49,7 +49,7 @@
                             {{related.headline || ('[No Value]' | translate)}}
                         </span>
                     </div>
-                    <div ng-hide="related._type === 'externalsource' || !associations.isMediaEditable()" class="item-association__icons-block">
+                    <div ng-hide="!associations.isMediaEditable(related)" class="item-association__icons-block">
                         <a ng-if="related && editable" class="item-association__image-action" ng-click="associations.isMediaEditable() && associations.edit(this, related, {defaultTab: 'view'}); $event.stopPropagation()" sd-tooltip="Edit metadata"><i class="icon-pencil"></i></a>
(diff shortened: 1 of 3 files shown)


scripts/core/config.ts
```diff
@@ -3,8 +3,12 @@ import {appConfig} from 'appConfig';
 import {IArticle} from 'superdesk-api';
 
 export function isMediaEditable(item?: IArticle) {
-    if (item !== undefined && item._fetchable !== undefined) {
-        return item._fetchable;
+    if (item !== undefined) {
```
tomaskikutis: I'd use `!= null`. It checks for `null` and `undefined` at the same time(because of single equals, as opposed to `!==`).

scripts/apps/authoring/views/item-association.html
```diff
@@ -49,7 +49,7 @@
                             {{related.headline || ('[No Value]' | translate)}}
                         </span>
                     </div>
-                    <div ng-hide="related._type === 'externalsource' || !associations.isMediaEditable()" class="item-association__icons-block">
+                    <div ng-hide="!associations.isMediaEditable(related)" class="item-association__icons-block">
```
petrjasek: might be better readable with `ng-show="associations.isMediaEditable(related)"`

scripts/core/config.ts
```diff
@@ -3,8 +3,12 @@ import {appConfig} from 'appConfig';
 import {IArticle} from 'superdesk-api';
 
 export function isMediaEditable(item?: IArticle) {
-    if (item !== undefined && item._fetchable !== undefined) {
-        return item._fetchable;
+    if (item != null) {
+        if (item._type != null) {
+            return item._type !== 'externalsource';
+        } else if (item._fetchable != null) {
+            return item._fetchable;
```
petrjasek: shouldn't it be more like:
```ts
if (item != null) {
  if (item._fetchable != null) {
    return item._fetchable;
  } else if (item._type !== 'externalsource') {
    return true;
  }
}
...
```

petrjasek: because I think in general external items will have `_type` set so it wouldn't check `_fetchable` flag

PR #3294, opened 2020-01-21
add create content button to bottom bar

petrjasek:
SDESK-4942


scripts/apps/authoring/workqueue/workqueue.ts
```diff
@@ -283,6 +283,13 @@ function WorkqueueCtrl(
             return $rootScope.link('authoring', item);
         }
     };
+
+    $scope.$on('$locationChangeSuccess', () => {
+        $scope.isMultiedit = $route.current && $route.current._id === 'multiedit';
+        if ($scope.isMultiedit) {
+            $scope.active = null;
```
tomaskikutis: What is `active` referring to? Can you rename it?

petrjasek: it's the item which is opened in authoring. any suggestions?

tomaskikutis: I call it `articleInEditMode` in the API

https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L894

PR #3297, opened 2020-01-22
Do not show warning message for unpublished related items if PUBLISH_ASSOCIATED_ITEMS set to True. [SDESK-4946]

GyanP:


tomaskikutis:
Can you move validation to the server?

GyanP:
Which validations? The message is used just to warn the user and all the validation for publishing is in the backend itself.

tomaskikutis:
I meant the warning. Can you move the warning to the back-end?

GyanP:
If you look, it's kind of confirmation from the user before actually publishing the article. So, user have two choices either to publish the related items first and then publish the main article or proceed without publishing the related items. It's not a good idea to move this to the backend.

tomaskikutis:
It's interactive validation, but a type of validation nonetheless and a part of business logic. If you build another front end for the API, you shouldn't have to reimplement it.

It's also safer in terms of failure. If you have multiple front-ends and one forgets to implement the prompt, the article should not be getting published without it. It's better to have the front-end crash if it's consuming the API incorrectly than to allow breaking business rules.

The API could return something like this:
```
{
    "confirm": {
        "id": "publish-without-related",
        "message": "There are unpublished related items that won't be sent out as related items. Do you want to publish the article anyway?"
    }
}
```

and then client could repeat the request with `?publish-without-related=true` appended to the URL.

petrjasek:
just had some similar requirement from ansa regarding publishing and warnings, would be good to have support for that - not sure if we need to have ids and confirm each separately, either user confirms all or the publishing won't work, so single argument to confirm would be enough imo.

tomaskikutis:
Yep, I agree, we should display all messages at once. We don't need the ID either. We could use a flag like `?publishing_warnings_confirmed=true` instead

karelpetrak:
Just a reminder that the purpose of the original ticket is not to show the existing message if the publish_associatied_items is set to true because the message warns the user that the related items will NOT be published while they actually WILL BE published. 
It would be good to fix the bug first and then we can do the refactoring.

scripts/apps/authoring/authoring/services/AuthoringService.ts
```diff
@@ -277,16 +277,30 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
...
                             if (message != null) {
                                 notify.error(message);
                             }
                         });
+                    } else if (issues?.['validator exception'] && issues?.fields == null) {
+                        const errorObj = JSON.parse(issues?.['validator exception']);
+
+                        if (errorObj.warnings) {
+                            return modal.confirm({
+                                bodyText: errorObj.warnings.toString(),
```
tomaskikutis: Why are you casting warnings to string? Can you add a type for `errorObj`?

Does it support displaying multiple messages properly? If not, you can use [`showModal`](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/services/modalService.tsx#L7) function with a custom React component to do this.

PR #3299, opened 2020-01-23
Show locked items in related field

pablopunk:
Captura de pantalla 2020-01-23 a las 13 18 38

### DIFF
--- a/scripts/apps/relations/views/related-items.html
+++ b/scripts/apps/relations/views/related-items.html
@@ -27,7 +27,7 @@
             id="{{field._id}}"
             ng-repeat="(key, relatedItem) in relatedItems track by key"
             class="sort-item draggable-list__item draggable-list__item--no-padding"
-            ng-class="{locked: relatedItem.lock_session != null}"
+            ng-class="{locked: isLocked(relatedItem)}"
         >
             <div class="sd-list-item sd-list-item--no-hover">
                 <div class="sd-list-item__column"><i class="type-icon" sd-filetype-icon data-item="relatedItem"></i>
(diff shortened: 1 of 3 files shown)


scripts/apps/relations/views/related-items.html
```diff
@@ -23,8 +23,12 @@
 <div class="related-items" ng-if="!isEmptyRelatedItems(field._id)">
     <div ng-if="loading" class="loading-overlay" ng-class="{active: loading}" style="opacity: 0.5;"></div>
     <ul class="groups draggable-list" ng-if="!loading" sd-sort-groups>
-        <li id="{{field._id}}" class="sort-item draggable-list__item draggable-list__item--no-padding"
-            ng-repeat="(key, relatedItem) in relatedItems track by key">
+        <li
+            id="{{field._id}}"
+            ng-repeat="(key, relatedItem) in relatedItems track by key"
+            class="sort-item draggable-list__item draggable-list__item--no-padding"
+            ng-class="{locked: relatedItem.lock_session != null}"
```
tomaskikutis: can you use `lockService.isLocked`?

pablopunk: That won't show me the items that I have locked the current session, not sure if that's okay for them or I need to make another function

pablopunk: The task says "locked by me" so I guess we do need to see the current session locked items

tomaskikutis: There are also `lockService.isLockedInCurrentSession` and `lockService.isLockedByMe`. We should be using one of those instead of accessing field names directly. There are around 5 fields related to locking, so it's better to have the logic in one place.

pablopunk: I know, check my latest commit

PR #3308, opened 2020-01-29
Update related items when [un]locked on another browser

pablopunk:
SDESK-4953

### DIFF
--- a/scripts/apps/relations/directives/RelatedItemsDirective.ts
+++ b/scripts/apps/relations/directives/RelatedItemsDirective.ts
@@ -36,8 +36,8 @@ interface IScope extends IDirectiveScope<void> {
  * add related items by using drag and drop, delete related items and open related items.
  */
 
-RelatedItemsDirective.$inject = ['authoringWorkspace', 'relationsService', 'notify', 'lock'];
-export function RelatedItemsDirective(authoringWorkspace: AuthoringWorkspaceService, relationsService, notify, lock) {
+RelatedItemsDirective.$inject = ['authoringWorkspace', 'relationsService', 'notify', 'lock', '$rootScope'];
+export function RelatedItemsDirective(authoringWorkspace: AuthoringWorkspaceService, relationsService, notify, lock, $rootScope) {
     return {
         scope: {
             item: '=',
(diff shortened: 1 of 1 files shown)


scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -271,6 +283,15 @@ export function RelatedItemsDirective(authoringWorkspace: AuthoringWorkspaceServ
...
                 }
             });
+
+            const removeEventListeners = [
+                'item:lock',
+                'item:unlock',
+            ].map((_) => $rootScope.$on(_, scope.refreshRelatedItems));
+
+            scope.$on('$destroy', () => {
+                removeEventListeners.forEach((_) => _());
```
tomaskikutis: What's up with the underscores? :D The convention is that underscores are used for argument names only when you don't use the argument.

pablopunk: But it looks cool doesn't it? Haha I'll change it 👍

tomaskikutis:
Why do you care about the lock? Shouldn't you only refresh related items every time they were changed? locking is only one type of change and in this case, it seems it wouldn't matter whether it was lock property changed or headline updated

pablopunk:
@tomaskikutis hmmm well the task was only about the lock when someone opens it on another session, that's why I use the lock, not sure if it's necessary to listen to all update events.

tomaskikutis:
Not to all, there should be a single event that fires no matter what changes. Might be `item:update` or `content:update`.

I know the task is about locking, but thinking of related issues is part of all tasks. What if status/headline changes? Would a user expect to see it? I'd say yes.

pablopunk:
@tomaskikutis I included that case and I also included logic to not reload the items if the event doesn't belong to these related items

tomaskikutis:
Yep, I was implicitly suggesting the same, to only reload changed items. Doesn't `item:update` or `content:update` fire when unlocking too? It would be better if one event was enough

pablopunk:
I thought that too but it doesn't. None of both will trigger when locking nor unlocking

tomaskikutis:
@petrjasek can we have a single event that fires whenever anything changes in the article? It's error-prone to have to deal with multiple events. You never know if you aren't missing some.

petrjasek:
yes we can.. would still keep those specific ones but we can have another one which will be always sent as well

PR #3315, opened 2020-01-30
add config to keep meta terms dropdown opened on click

petrjasek:
so user can select multiple items without using keyboard


tomaskikutis:
Why not make it default behavior and drop the config option?

petrjasek:
@karelpetrak any feedback on it? preview:

Peek 2020-01-30 16-27

karelpetrak:
I am for having a config for this. We'd better not change the behaviour of such an important control for everybody.

PR #3316, opened 2020-01-30
make updating editor3 from macro configurable

petrjasek:


tomaskikutis:
What's the purpose of this? To ignore changes? Why would we do that?

petrjasek:
I think it doesn't call onChange because [it calls it later](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/authoring/macros/macros.ts#L198) just once, so macro can update multiple editor instances and only after trigger autosave

tomaskikutis:
I'm not entirely getting it. What problem are you solving? Does it call autosave too often or what's the issues?

petrjasek:
I didn't change anything here, but I'm using the event in client repo where I need the autosave to be called

tomaskikutis:
I was asking about the fix you are doing. What happens without the code in this PR that you want to change? doesn't autosave get called?

petrjasek:
it doesn't get called in that event handler

PR #3320, opened 2020-01-31
add edit thumbnail to associated videos

petrjasek:
SDESK-4926

### DIFF
--- a/scripts/core/ui/components/drop-zone.tsx
+++ b/scripts/core/ui/components/drop-zone.tsx
@@ -91,7 +91,10 @@ export class DropZone extends React.PureComponent<IDropZoneComponentProps, IStat
                         ref={this.input}
                         onChange={(event) => {
                             event.preventDefault();
-                            this.props.onFileSelect(this.input.current.files);
+                            if (this.input.current.files.length) {
+                                this.props.onFileSelect(this.input.current.files);
+                            }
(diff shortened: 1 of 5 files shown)


scripts/core/ui/components/drop-zone.tsx
```diff
@@ -91,7 +91,10 @@ export class DropZone extends React.PureComponent<IDropZoneComponentProps, IStat
                         ref={this.input}
                         onChange={(event) => {
                             event.preventDefault();
-                            this.props.onFileSelect(this.input.current.files);
+                            if (this.input.current.files.length) {
+                                this.props.onFileSelect(this.input.current.files);
+                            }
+                            event.target.value = null; // reset to allow selecting same file again
```
tomaskikutis: Why does it not allow selecting the same file unless the event is modified?

petrjasek: if you select same file `input.value` doesn't change so `onChange` handler is not called

tomaskikutis: I assume you are fixing that issue I showed with changing a video thumbnail. In that case, it doesn't allow uploading the same file when you do fileA > fileB > fileA. Shouldn't selecting fileB had already changed the input so fileA can be selected again?

PR #3327, opened 2020-02-04
[sdesk-4983] Make items with pubstatus withheld unfetchable

marwoodandrew:


tomaskikutis:
@ride90 how about that fetching flag you built, could it be used in such case?

ride90:
@tomaskikutis , yes it will work. Just mark an image on the backend as `'_fetchable': False ` and it should do the trick. Example: https://github.com/superdesk/superdesk-tlp/blob/master/server/anp/search_providers/talpa_video.py#L151

tomaskikutis:
In this case it should depend on `pubstatus` value and we should not set it manually when `pubstatus` changes, since it's likely to get out of sync.

petrjasek:
I think this is not only for search providers but also for ingest, so I wouldn't use the flag. maybe it should check for `pubstatus === 'usable'` so it also handles canceled, but I don't think we show these

marwoodandrew:
I didn't put a PR in against develop as the _fetchable flag will work, just using the pubstatus in version before the _fetchable flag came into existence. The AAP image search provider sets both the pubstatus = withheld and _fetchable=False, We display the pubstatus in our search results pane to provide an indication that the image can't be fetched. (These are images that do not have a high enough resolution to crop, these were sometimes getting fetched and causing angst when theu could not be cropped)
I stuck to using pubstatus == withheld and I couldn't be sure all search providers set the pubstatus or if it always available.

PR #3328, opened 2020-02-04
fix footer buttons not responding when clicked while select is focused

tomaskikutis:
SDESK-4877

This is a strange issue and I couldn't find why is it happening. Event listeners for footer buttons(when changing marked user) wouldn't respond while the select is focused. The first time you click it, blur event would be triggered on select and only future clicks would then trigger button events.

### DIFF
--- a/scripts/core/ui/components/select2.tsx
+++ b/scripts/core/ui/components/select2.tsx
@@ -139,14 +139,17 @@ export class Select2<T> extends React.Component<IProps<T>, IState> {
                         // 2. for event listeners to respond when buttons are clicked outside of a focused select
 
                         const timeout = 200; // smaller values don't work for point 2 above
-                        
+
                         setTimeout(() => {
                             if (this.wrapper != null) {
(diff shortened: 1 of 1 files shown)


scripts/core/ui/components/select2.tsx
```diff
@@ -134,14 +134,24 @@ export class Select2<T> extends React.Component<IProps<T>, IState> {
...
+
+                        const timeout = 200; // smaller values don't work for point 2 above
+
                         setTimeout(() => {
-                            const remainingAtTheBottom =
-                                window.innerHeight - this.wrapper.getBoundingClientRect().bottom - 20;
-                            const oneThirdViewportHeigh = window.innerHeight / 3;
+                            if (this.wrapper != null) {
+                                const remainingAtTheBottom =
+                                    window.innerHeight - this.wrapper.getBoundingClientRect().bottom - 20;
```
petrjasek: not sure about this `20` magic number, can you give it a name?

PR #3329, opened 2020-02-04
Fixed feature media and media gallery issue.

GyanP:
SDESK-4998, SDESK-4999

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/AssociationController.ts
+++ b/scripts/apps/authoring/authoring/controllers/AssociationController.ts
@@ -7,7 +7,7 @@ import {mediaIdGenerator} from '../services/MediaIdGeneratorService';
 
 export function getAssociationsByField(item: IArticle, field: IVocabulary) {
     return Object.keys(item.associations || {})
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/controllers/AssociationController.ts
```diff
@@ -7,7 +7,7 @@ import {mediaIdGenerator} from '../services/MediaIdGeneratorService';
 
 export function getAssociationsByField(item: IArticle, field: IVocabulary) {
     return Object.keys(item.associations || {})
-        .filter((key) => key.startsWith(field._id) && item.associations[key] != null)
+        .filter((key) => key.startsWith(field?._id) && item.associations[key] != null)
```
tomaskikutis: Can a vocabulary not have `_id`? When?

GyanP: Good point, actually in case of featuremedia we don't have field itself as it's not a vocabulary. I will fix this.

PR #3333, opened 2020-02-05
Fix multiselect on authoring

pablopunk:
`this.state.selected` was not being updated so multiselection with SHIFT would select always from the beginning

### DIFF
--- a/scripts/apps/search/components/ItemList.tsx
+++ b/scripts/apps/search/components/ItemList.tsx
@@ -69,22 +69,25 @@ export class ItemList extends React.Component<any, IState> {
         this.unbindActionKeyShortcuts = this.unbindActionKeyShortcuts.bind(this);
     }
 
-    multiSelect(items, selected) {
+    multiSelect(items: Array<IArticle>, selected) {
         const {search, multi} = this.props.svc;
         const {scope} = this.props;
 
         const itemsById = angular.extend({}, this.state.itemsById);
 
+        let selectedId = null;
+
         items.forEach((item) => {
             const itemId = search.generateTrackByIdentifier(item);
 
+            selectedId = itemId;
             itemsById[itemId] = angular.extend({}, item, {selected: selected});
             scope.$applyAsync(() => {
                 multi.toggle(itemsById[itemId]);
             });
         });
 
-        this.setState({itemsById: itemsById});
+        this.setState({itemsById, selected: selectedId});
     }
 
     // Method to check the selectBox of the selected item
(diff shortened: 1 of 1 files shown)


scripts/apps/search/components/ItemList.tsx
```diff
@@ -69,22 +69,26 @@ export class ItemList extends React.Component<any, IState> {
...
         const {scope} = this.props;
+        let {selected: selectedId} = this.state;
 
         const itemsById = angular.extend({}, this.state.itemsById);
 
         items.forEach((item) => {
             const itemId = search.generateTrackByIdentifier(item);
 
+            if (selected) {
+                selectedId = itemId;
```
tomaskikutis: are you sure you want to set this on every iteration? only last iteration result will get used at line 91, no?

pablopunk: yeah, it will only use the last one. Initially, I set it only if `items.length === 1` but I included it inside the loop so the id is not calculated twice.

tomaskikutis: Can you use a function instead? There's no iteration in the function, it's a few milliseconds to execute. It's not worth sacrificing readability over that.

It's confusing when you place code in a loop when you only really want to execute it on a single item.

```
function getActiveId() {
    if (items.length < 1) {
        return undefined;
    }

    const lastItem = items[items.length - 1];

    if (lastItem.active === true) {
        return lastItem._id;
    } else {
        return undefined;
    }
}
```

pablopunk: I can just add a check to the current `if`:

```ts
  if (selected && i === items.length - 1) {
    // Mark last item as selected
    selectedId = itemId;
  }
```

tomaskikutis: That's better, but I'd keep it out of the loop. Why do you want to keep it inside? To save 5 lines of code and 10 processor instructions? :)

pablopunk: if I put it outside I would call `generateTrackByIdentifier` twice for the same item

tomaskikutis: and why do you prefer not to call it twice?

pablopunk: well, that function is doing some conditionals and concatenating strings. It could be worse of course, not a big deal in performance for now.

tomaskikutis: From a performance point of view, that function does so little that you could call it 1000 times and it wouldn't matter. Here's some empirical data:

```
const first = items[0];

console.time('duration');

for (var i = 0; i < 999; i++) {
    search.generateTrackByIdentifier(first);
}

console.timeEnd('duration');
```

```duration: 1.671875ms```

The benchmarking is not the best since I'm calling it on the same item and JS engine is probably optimizing it, but even without the optimizations it wouldn't be much different

pablopunk: yeah yeah when I said what it does I meant it does very little so it won't matter. I just meant it as a general thing to consider, not a big deal here

PR #3334, opened 2020-02-05
Allow marking for user feature to work instantly without having to save the article in edit mode

tomaskikutis:
SDBELGA-295

### DIFF
--- a/scripts/extensions/markForUser/src/common.ts
+++ b/scripts/extensions/markForUser/src/common.ts
@@ -1,0 +1,9 @@
+import {ISuperdesk, IArticle} from 'superdesk-api';
+
+export function updateMarkedUser(superdesk: ISuperdesk, article: IArticle, patch: Partial<IArticle>): void {
+    superdesk.entities.article.patch(
+        article,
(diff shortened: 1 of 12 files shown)


scripts/extensions/markForUser/src/common.ts
```diff
@@ -0,0 +1,9 @@
+import {ISuperdesk, IArticle} from 'superdesk-api';
+
+export function updateMarkedUser(superdesk: ISuperdesk, article: IArticle, patch: Partial<IArticle>): void {
+    superdesk.entities.article.patch(
+        article,
+        patch,
+        {patchDirectlyAndOverwriteAuthoringValues: true},
```
pablopunk: when using `updateMarkedUser` for example without the item open to edit, is this option really necessary?

tomaskikutis: In terms of code execution - no. The option will be ignored if the article is not in edit mode.

I would leave it unconditional through and assign the responsibility of figuring out when to apply it to the `patch` method.

PR #3338, opened 2020-02-05
[SDESK-4903] Add onSend interface to the extensions api

MarkLark86:


MarkLark86:
@tomaskikutis We required this functionality into 1.30, which I have added as a functionPoint (https://github.com/superdesk/superdesk-client-core/pull/3337)

This interface will need to be cherry-picked (back-ported) into 1.31+.
I will create another PR for <1.33 to add functionality into the planning extension once this is approved & merged in all respective version branches

scripts/core/superdesk-api.d.ts
```diff
@@ -75,6 +75,7 @@ declare module 'superdesk-api' {
                     onSpikeMultiple?(items: Array<IArticle>): Promise<onSpikeMiddlewareResult>;
                     onPublish?(item: IArticle): Promise<onPublishMiddlewareResult>;
                     onRewriteAfter?(item: IArticle): Promise<IArticle>;
+                    onSend?(desk: IDesk, itemIds: Array<string>): Promise<void>;
```
tomaskikutis: I'd call it `onSendBefore` so it's clear when the code is getting executed.

I'd also have `itemIds` argument come first in the function signature so it reads more naturally like "send those items to this desk".

What are `itemIds` by the way? Shouldn't it accept a single `item: IArticle` instead?

MarkLark86: This action can be used with multi-select as well, in which case the directive provides the ids of the items.

scripts/core/superdesk-api.d.ts
```diff
@@ -75,6 +75,7 @@ declare module 'superdesk-api' {
                     onSpikeMultiple?(items: Array<IArticle>): Promise<onSpikeMiddlewareResult>;
                     onPublish?(item: IArticle): Promise<onPublishMiddlewareResult>;
                     onRewriteAfter?(item: IArticle): Promise<IArticle>;
+                    onSendBefore?(itemIds: Array<string>, desk: IDesk): Promise<void>;
```
tomaskikutis: Could you make it `items: Array<IArticle>` instead of IDs only? I think the items are already fetched at this point and it'd be faster to resolve the promise when extensions don't have to fetch items themselves.

PR #3341, opened 2020-02-06
Add a new user privilege for "Mark for user" action. [SDBELGA-300]

GyanP:


scripts/core/get-superdesk-api-implementation.tsx
```diff
@@ -118,6 +118,7 @@ export function getSuperdeskApiImplementation(
                 isPersonal: (article) => article.task == null || article.task.desk == null,
                 isLocked: (article) => article['lock_session'] != null,
                 isLockedByCurrentUser: (article) => lock.isLockedInCurrentSession(article),
+                hasMarkForUserPrivilege: () => privileges.userHasPrivileges({mark_for_user: 1}),
```
tomaskikutis: API has `getOwnPrivileges` method - use that or add `hasPrivilege()` method. The core should not have any code related to a particular extension. All your changes for this task should be limited to files in the extension, except maybe if you want to add `hasPrivilege` method to the API.

scripts/core/get-superdesk-api-implementation.tsx
```diff
@@ -258,6 +258,7 @@ export function getSuperdeskApiImplementation(
         },
         privileges: {
             getOwnPrivileges: () => privileges.loaded.then(() => privileges.privileges),
+            hasPrivilege: (privilege: {[key: string]: number}) => privileges.userHasPrivileges(privilege),
```
tomaskikutis: What's the schema for privileges? Can it only be `1` or `0`? If so, it'd be nicer to have the interface be `hasPrivilege(name: string): boolean`

scripts/extensions/markForUser/src/get-article-actions-bulk.tsx
```diff
@@ -14,7 +15,7 @@ export function getActionsBulkInitialize(superdesk: ISuperdesk) {
 
         const someItemsSpiked = articles.some(({state}) => state === 'spiked');
 
-        if (articles.some(isPersonal) || someItemsLocked || someItemsSpiked) {
+        if (articles.some(isPersonal) || someItemsLocked || someItemsSpiked || !hasPrivilege({mark_for_user: 1})) {
```
tomaskikutis: I'd put `hasPrivilege` check as the first condition in the `if` since it's the most important and would also terminate the `if` sooner when the privilege is not there without having to iterate the items and check their state.

scripts/core/get-superdesk-api-implementation.tsx
GyanP: @tomaskikutis yes, the schema for privileges will have only 0 or 1,  for ex: `{mark_for_user: 1, mark_for_desk: 1, dashboard: 0}`

PR #3343, opened 2020-02-07
fix npm install when package.json is missing for extension

petrjasek:
when you switch to different version or different branch
in superdesk the extension folder stays there with node_modules
but without package.json and install fails then.

### DIFF
--- a/tasks/get-extension-directories-sync.js
+++ b/tasks/get-extension-directories-sync.js
@@ -2,8 +2,6 @@
 
 var fs = require('fs');
 var path = require('path');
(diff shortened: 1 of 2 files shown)


tomaskikutis:
What is reading `--if-present` option? What about using a node function to check if the file exists before running npm install?

petrjasek:
with if present it won't fail on `npm run something` if `something` is not defined in package.json

petrjasek:
and I've added a check for package.json to be there, that's the main thing in this PR

PR #3344, opened 2020-02-07
Added background color to add suggestions

bfkolle:
Apologies for not adding a Jira Id, I could not find out how to do that anywhere.
My news department has had issues with trying to view suggestions that are spaces. Suggestion spaces are invisible because only text is highlighted currently. My change adds a light green background to suggestions so that even spaces will be visible. I previously made this change on our local instance of Superdesk, and it has been working well for 3+ months and resolved major issues with our news process.

**Space Suggestion Without Background**
image

**With Background**
image


petrjasek:
thank you

PR #3348, opened 2020-02-10
handle editor state changes done in macro

petrjasek:
SDBELGA-250

### DIFF
--- a/scripts/core/editor3/reducers/editor3.tsx
+++ b/scripts/core/editor3/reducers/editor3.tsx
@@ -1,4 +1,4 @@
-import {RichUtils, EditorState} from 'draft-js';
+import {RichUtils, EditorState, ContentState} from 'draft-js';
 import {setTansaHtml} from '../helpers/tansa';
 import {addMedia} from './toolbar';
 import {getCustomDecorator, IEditorStore} from '../store';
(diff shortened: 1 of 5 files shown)


scripts/core/editor3/reducers/editor3.tsx
```diff
@@ -16,6 +16,8 @@ const editor3 = (state: IEditorStore, action) => {
     switch (action.type) {
     case 'EDITOR_CHANGE_STATE':
         return onChange(state, action.payload.editorState, action.payload.force, false, action.payload.skipOnChange);
+    case 'EDITOR_PUSH_STATE':
+        return pushState(state, action.payload.contentState);
```
tomaskikutis: What about converting html to editorState and calling `onChange`(as in line 18) instead of adding `pushState`?

petrjasek: macros can handle editor state now so imo it's better to use that rather than converting it to html on backend and then parse it here and hope for the best..

tomaskikutis: Sorry, I thought we are getting `html` from the server after the macro is applied, which is probably not the case. What I was suggesting is to inline `pushState` code instead of creating a 2 line function, but it doesn't really matter.

PR #3350, opened 2020-02-12
Unify avatars

tomaskikutis:
SDBELGA-292

DEPENDS ON: https://github.com/superdesk/superdesk-ui-framework/pull/432


scripts/apps/authoring/views/authoring-topbar.html
```diff
@@ -13,11 +13,10 @@
     ></sd-authoring-topbar-react>
 
     <div ng-show="isLocked()" class="locked-info">
-      <figure class="avatar small">
-          <div class="user-image">
-              <img sd-user-avatar data-user="item.lock_user">
-          </div>
-      </figure>
+
+      <div style="margin-right: 4px; margin-left: -2px; transform: scale(0.9);">
```
fritzSF: Why inline styles? Also, why isn't there a prop for the avatar sizes? We have at least 3 sizes

tomaskikutis:
@fritzSF I moved the avatar component to the framework.

Regarding inline styles, is there a difference whether I use `style="display: flex"` or `class="flex"` and have `.flex {display: flex}` in CSS? For me, the only difference is spending an extra 5 minutes for moving it to the stylesheet.

I try to keep styles contained in logical CSS component groups when possible, but there are cases where parent components have implicit dependencies on children which would take many days if I started refactoring the classes and going over all usages. For example, some components were expecting avatars to be floated to the left. I removed the floats and wrapped the avatars in floated divs when used in those classes in order to preserve the current behaviour and not have to refactor the class.

fritzSF:
@tomaskikutis great regarding the framework.
Regarding the in-line styles: I don't mind if it is used for fixes of the old codebase where CSS is heavily nested and should be refactored. It's more about cases like `style="margin-right: 4px; margin-left: -2px; transform: scale(0.9);"` The size should be solved with a additional class, or a prop in this case, and for margin and padding overrides we have helper classes in the framework. It might not be a solution for all cases but it should work in most.

PR #3351, opened 2020-02-12
Show linebreaks in preview when html is plain text

pablopunk:
SDBELGA-284

If CP has no formatting options for BODY_HTML, then it will be saved as plain text, which makes it not show linebreaks on the html preview (as linebreaks don't really mean anything in html).

### DIFF
--- a/scripts/apps/archive/directives/HtmlPreview.ts
+++ b/scripts/apps/archive/directives/HtmlPreview.ts
@@ -52,9 +52,11 @@ export function HtmlPreview($sce, $timeout) {
         templateUrl: 'scripts/apps/archive/views/html-preview.html',
         link: function(scope, elem, attrs) {
             scope.$watch('sdHtmlPreview', (html) => {
-                scope.html = stringIsHtml(html)
-                    ? $sce.trustAsHtml(html)
-                    : $sce.trustAsHtml(plainTextToHtml(html));
+                if (html && typeof html === 'string') {
+                    scope.html = stringIsHtml(html)
+                        ? $sce.trustAsHtml(html)
+                        : $sce.trustAsHtml(plainTextToHtml(html));
+                }
 
                 if (window.hasOwnProperty('instgrm')) {
                     window.instgrm.Embeds.process();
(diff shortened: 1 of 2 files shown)


scripts/core/editor3/html/to-html/plainTextToHtml.ts
```diff
@@ -0,0 +1,6 @@
+// converts linebreaks in paragraphs
+export const plainTextToHtml = (text: string) =>
+    text.replace(/.*[^\n]/g, '<p>$&</p>');
```
tomaskikutis: how about ```str.split('\n').map((line) => `<p>${line}</p>`).join('')```? Is your regex doing something different?

scripts/apps/archive/directives/HtmlPreview.ts
```diff
@@ -52,9 +52,11 @@ export function HtmlPreview($sce, $timeout) {
         templateUrl: 'scripts/apps/archive/views/html-preview.html',
         link: function(scope, elem, attrs) {
             scope.$watch('sdHtmlPreview', (html) => {
-                scope.html = stringIsHtml(html)
-                    ? $sce.trustAsHtml(html)
-                    : $sce.trustAsHtml(plainTextToHtml(html));
+                if (html && typeof html === 'string') {
```
tomaskikutis: You don't need `html &&` part. You only need to check if it's not null if you accessing object properties.

scripts/core/editor3/html/to-html/plainTextToHtml.ts
```diff
@@ -0,0 +1,6 @@
...
+
+export const stringIsHtml = (text: string) =>
```
tomaskikutis: I'd use `DOMParser` for this https://stackoverflow.com/a/15458968/1175593

I'd also name it `isStringHtml`. Adding `is` in front makes it more of a question which reads better for functions while `stringIsHtml` is more of a statement which is better for booleans: `const stringIsHtml = isStringHtml(str)`.

scripts/core/editor3/html/to-html/plainTextToHtml.ts
pablopunk: Nope, it's just doing that

tomaskikutis: Can you use the JS version then so I don't need a regex manual to understand it? :D

PR #3353, opened 2020-02-13
Publish an update with a publish schedule when original is also published with publish schedule. [SDBELGA-307]

GyanP:
Requires https://github.com/superdesk/superdesk-core/pull/1804

### DIFF
--- a/scripts/apps/authoring/authoring/services/AuthoringService.ts
+++ b/scripts/apps/authoring/authoring/services/AuthoringService.ts
@@ -630,7 +630,7 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         let userPrivileges = privileges.privileges;
 
         action.re_write = (!isReadOnlyState || appConfig.allow_updating_scheduled_items)
-            && _.includes(['text'], currentItem.type)
+            && currentItem.type === 'text'
             && !currentItem.embargo && !currentItem.rewritten_by
             && (!currentItem.broadcast || !currentItem.broadcast.master_id)
             && (
@@ -639,7 +639,7 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
                 ) || appConfig.workflow_allow_multiple_updates)
             );
 
-        action.resend = _.includes(['text'], currentItem.type) &&
+        action.resend = currentItem.type === 'text' &&
             isPublished(currentItem, false);
 
         // mark item for highlights
--- a/scripts/apps/authoring/authoring/services/AuthoringService.ts
+++ b/scripts/apps/authoring/authoring/services/AuthoringService.ts
@@ -629,13 +629,14 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         let isReadOnlyState = this._isReadOnly(currentItem);
         let userPrivileges = privileges.privileges;
 
-        action.re_write = !isReadOnlyState && _.includes(['text'], currentItem.type) &&
-            !currentItem.embargo && !currentItem.rewritten_by &&
-            (!currentItem.broadcast || !currentItem.broadcast.master_id) &&
-            (
-                !currentItem.rewrite_of || (
+        action.re_write = (!isReadOnlyState || appConfig.allow_updating_scheduled_items)
+            && _.includes(['text'], currentItem.type)
+            && !currentItem.embargo && !currentItem.rewritten_by
+            && (!currentItem.broadcast || !currentItem.broadcast.master_id)
+            && (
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/authoring/services/AuthoringService.ts
```diff
@@ -629,13 +629,14 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         let isReadOnlyState = this._isReadOnly(currentItem);
         let userPrivileges = privileges.privileges;
 
-        action.re_write = !isReadOnlyState && _.includes(['text'], currentItem.type) &&
-            !currentItem.embargo && !currentItem.rewritten_by &&
-            (!currentItem.broadcast || !currentItem.broadcast.master_id) &&
-            (
-                !currentItem.rewrite_of || (
+        action.re_write = (!isReadOnlyState || appConfig.allow_updating_scheduled_items)
```
tomaskikutis: `(!isReadOnlyState || appConfig.allow_updating_scheduled_items)`

I see you are using `||` operator. Even if `allow_updating_scheduled_items` is set to `false`, it will still update it if `isReadOnlyState` is also `false`. It should probably not behave like that.

scripts/apps/authoring/authoring/services/AuthoringService.ts
```diff
@@ -629,13 +629,14 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
...
+            && _.includes(['text'], currentItem.type)
```
tomaskikutis: Use native methods when available https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

scripts/apps/authoring/authoring/services/AuthoringService.ts
GyanP: Here I am extending the existing functionality to work in case of `allow_updating_scheduled_items=true` as well. So, if any of the two condition is true it should check the remaining conditions.

tomaskikutis: I'm not sure I understand. If you have `allow_updating_scheduled_items` set to false, it should not update scheduled items under any other conditions. Isn't that right?

GyanP: Yes, and that's how it is behaving currently, there is no way scheduled items can be updated. And my changes will allow the updates in case of scheduled items only if  `allow_updating_scheduled_items` is set to true.

tomaskikutis: Does `isReadOnlyState` mean `isScheduled`? If it does, could you write it like this
```
const isScheduled = isReadOnlyState;

action.re_write = (!isScheduled || appConfig.allow_updating_scheduled_items) ...
```
It's confusing when you are using an `OR` operator against seemingly unrelated property. That's why I was thinking that it may still update the item when `allow_updating_scheduled_items` is set to false.

GyanP: `KILLED | RECALLED | UNPUBLISHED | SPIKED | SCHEDULED` state are included in `ReadOnlyState`. So, I can't really say `const isScheduled = isReadOnlyState;`. But yes, it's confusing.

tomaskikutis: How about putting it like this?
```
action.re_write = (item.state !== 'scheduled' || appConfig.allow_updating_scheduled_items) && !isReadOnlyState && ...
```

GyanP: @tomaskikutis this is how it can be done `action.re_write = (!isReadOnlyState || (item.state === 'scheduled' && appConfig.allow_updating_scheduled_items)) && ...`. Lets say the items which are in in_progress state can also be updated, so basically isReadOnlyState means item is in in_progress state, so we should allow it to get updated.

tomaskikutis: I now see how my previous suggestion wouldn't work. `isReadOnlyState` combines many states and it seems that this feature only requires one, so I'd prefer to use only that one to keep it simpler.

How about converting it to a function then?

```
function canRewrite() {
    if (item.state === 'scheduled' && appConfig.allow_updating_scheduled_items !== true) {
        return false;
    }

    return !isReadOnlyState && ...
}
```

tomaskikutis: Thanks, it's much clearer now

PR #3357, opened 2020-02-17
Fix dragging text on editor3

pablopunk:
SDESK-4956

Toolbar component state was not being computed properly due to toolbar ref being null.


pablopunk:
@petrjasek do you want me to make this change in 1.33 directly?

petrjasek:
no that's fine, I can cherry-pick, it's not so big :smile:

PR #3360, opened 2020-02-18
Fix issues with media carousel videos

pablopunk:
SDESK-5040

* Stream video has no `source` tag so the OWL component was not loaded
(cause it was waiting for the tag to appear)
* Tags like <video> are now rendered in react so they might not be
loaded after angular finishes rendering

### DIFF
--- a/scripts/core/helpers/waitForMediaToBeReady.ts
+++ b/scripts/core/helpers/waitForMediaToBeReady.ts
@@ -4,44 +4,53 @@ import {logger} from 'core/services/logger';
 /**
  * @param elements An array of img, audio and video elements
  */
-export const waitForMediaToLoad = (elements: Array<HTMLImageElement | HTMLAudioElement | HTMLVideoElement>):
-Promise<void> => new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


scripts/core/helpers/waitForMediaToBeReady.ts
```diff
@@ -4,44 +4,53 @@ import {logger} from 'core/services/logger';
...
+            if (isImage(element)) {
+                // check for error in image src
+                element.addEventListener('error', eventHandler, {once: true});
+                element.addEventListener('load', eventHandler, {once: true});
+            } else if (isAudio(element) || isVideo(element)) {
+                // for audio and video check for the error in source
+                const source = element.getElementsByTagName('source')[0];
+
+                if (!source) {
+                    // Live videos don't have a <source> tag
```
tomaskikutis: You probably mean streamed videos.

pablopunk: indeed I do, I always say live but it's not really live

PR #3364, opened 2020-02-20
Do not allow updating of already updated item. [SDBELGA-307]

GyanP:


scripts/apps/authoring/authoring/services/AuthoringService.ts
```diff
@@ -630,7 +630,9 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         let userPrivileges = privileges.privileges;
 
         function canRewrite() {
-            if (currentItem.state === ITEM_STATE.SCHEDULED && appConfig.allow_updating_scheduled_items === true) {
+            if (currentItem.state === ITEM_STATE.SCHEDULED
+                && appConfig.allow_updating_scheduled_items
+                && !currentItem.rewritten_by) {
```
tomaskikutis: Avoid using `!` operators on non-boolean values and be explicit about what are you checking. Do you only care if it's null or is the length of the string also relevant?

Is `rewritten_by` only relevant when `currentItem.state === ITEM_STATE.SCHEDULED`? Why?

If not, I'd write it like this:

```
if (currentItem.rewritten_by != null) {
    return false;
}

if (currentItem.state === ITEM_STATE.SCHEDULED && appConfig.allow_updating_scheduled_items === true) {
    return true;
}
```

PR #3365, opened 2020-02-20
Search providers default view

pablopunk:
SDBELGA-81

### DIFF
--- a/scripts/apps/search-providers/directive.ts
+++ b/scripts/apps/search-providers/directive.ts
@@ -14,7 +14,7 @@ export default function SearchProviderConfigDirective(searchProviderService, not
             $scope.availableListViews = {
                 '': {label: gettext('None')},
                 list: {label: gettext('List View'), icon: 'th-list'},
-                photogrid: {label: gettext('Photo Grid View'), icon: 'th'},
+                grid: {label: gettext('Photo Grid View'), icon: 'th'},
             };
 
             searchProviderService.getAllowedProviderTypes().then((providerTypes) => {
--- a/scripts/apps/search-providers/directive.ts
+++ b/scripts/apps/search-providers/directive.ts
@@ -11,6 +11,12 @@ export default function SearchProviderConfigDirective(searchProviderService, not
             $scope.newDestination = null;
             $scope.gettext = gettext;
 
+            $scope.availableListViews = {
+                '': {label: gettext('None')},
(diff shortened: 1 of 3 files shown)


scripts/apps/search/directives/SearchContainer.ts
```diff
@@ -12,8 +12,20 @@ export function SearchContainer() {
...
-                    if (activeProvider && activeProvider.advanced_search !== undefined) {
-                        this.flags.facets = !!activeProvider.advanced_search;
+                    if (activeProvider) {
+                        if (activeProvider.advanced_search !== undefined) {
+                            this.flags.facets = !!activeProvider.advanced_search;
+                        }
+                        if (activeProvider.config?.default_list_view) {
+                            switch (activeProvider.config.default_list_view) {
+                            case 'list':
+                                $scope.view = 'compact';
```
petrjasek: what about `$scope.view = activeProvider.config.default_list_view;` and use `compact/photogrid` as `viewId`

pablopunk: I thought about it, but that would make the UI and the DB very coupled together. I mean that this naming is used in the search view but also in other places. If we wanna change the UI we would change the data too instead of the code only. What do you think?

tomaskikutis: I agree with Pablo on this. It's easier when all possible js variable values can be found statically from the code. It's also easier to rename and is clear what has to be changed in DB if necessary.

scripts/apps/search-providers/directive.ts
```diff
@@ -11,6 +11,12 @@ export default function SearchProviderConfigDirective(searchProviderService, not
             $scope.newDestination = null;
             $scope.gettext = gettext;
 
+            $scope.availableListViews = {
+                '': {label: gettext('None')},
+                list: {label: gettext('List View'), icon: 'th-list'},
+                photogrid: {label: gettext('Photo Grid View'), icon: 'th'},
```
tomaskikutis: I'd call it simply "grid" since it's not limited to working with photos.

pablopunk: That's how the ticket required the button to be named. I guess they are used to it and also it's very verbose because even if they are not images you do see images in the grid

tomaskikutis: I was referring to the object key name. Labels visible to users are easy to change so I'm not concerned about them.

PR #3366, opened 2020-02-24
Fix save button active after no changes on metadata

pablopunk:
SDESK-4732

Issue: Save an article with an image on it. Open image editor and close it without making changes. Article state is now dirty and the "Save" button is active.

The code producing this error was introduced here https://github.com/superdesk/superdesk-client-core/pull/3117 but I could not reproduce that issue after removing that code, because this way we were passing the current image to the "change crops" function, which would, of course, make the article "dirty".

Also, I included some fixes for errors being thrown from related items
(using src instead of ng-src)


scripts/apps/authoring/authoring/services/RenditionsService.ts
```diff
@@ -91,12 +91,6 @@ export function RenditionsService(metadata, $q, api, superdesk, _) {
                 .then((metaData) => {
                     // apply the metadata changes
                     return _.extend(item, metaData);
-                }).catch((response) => {
-                    // if new crops not generated continue with default one
-                    // see https://github.com/superdesk/superdesk-client-core/pull/3117#discussion_r328440897
```
tomaskikutis: I hope you have checked that you aren't breaking the behaviour from another PR

pablopunk: I'm waiting for Gyan to review the PR. I did check it but I want him to confirm with him

GyanP:
@pablopunk this PR definitely breaks the uploading of media. You can see [here](https://drive.google.com/file/d/14Y2PdM0gDQ3ztzyRnXifZwJhPZck7e0J/view)

pablopunk:
> @pablopunk this PR definitely breaks the uploading of media. You can see [here](https://drive.google.com/file/d/14Y2PdM0gDQ3ztzyRnXifZwJhPZck7e0J/view)

I still couldn't reproduce that issue, [watch this](https://files-gnjpkey8c.now.sh/sd.webm), isn't it the same you're doing?

GyanP:
> I still couldn't reproduce that issue, watch this, isn't it the same you're doing?

Try to add an image which has at least one custom crops, maybe of a bit higher resolution

MarkLark86:
> > I still couldn't reproduce that issue, watch this, isn't it the same you're doing?
> 
> Try to add an image which has at least one custom crops, maybe of a bit higher resolution

I don't think its wise to shift the task of regression testing to the developer(s). This is a task better suited to unit, behave or end-to-end testing. Can we have an e2e test for both of these bugs please.

tomaskikutis:
We do normally rely on tests, but we didn't have one for this case so it makes sense to do a manual test when code that is being removed was initially added for another feature. I agree that we should add tests for these cases.

spec/authoring_spec.ts
```diff
@@ -716,6 +724,62 @@ describe('authoring', () => {
...
+
+        el(['media-metadata-editor', 'field--headline'], by.tagName('input')).sendKeys('image headline');
+        el(['media-metadata-editor', 'field--alt_text'], by.tagName('input')).sendKeys('image alt text');
+        el(['media-metadata-editor', 'field--description_text'], by.tagName('textarea')).sendKeys('image description');
+
+        el(['multi-image-edit--start-upload']).click();
+
+        waitAndClick(el(['change-image', 'done']));
+
+        browser.wait(ECE.hasElementCount(els(['authoring-field--media-gallery', 'media-gallery-image']), 1), 1000);
```
tomaskikutis: Is there a reason for a specific timeout? I always go with the default, because when the element is found, the test will not wait any longer anyway. And if the test fails, it doesn't matter much if it fails after 5 or 10 seconds. Or does it? If you have many tests failing, it would take longer for tests to run, but I'd say a reasonable default solves this.

spec/helpers/utils.ts
```diff
@@ -170,6 +170,17 @@ export function assertToastMsg(type, msg) {
     browser.sleep(500);
 }
 
+// Don't expect message to appear
+export function assertToastMsgNotDisplayed(type, msg) {
+    expect(element.all(by.cssContainingText(`.notification-holder .alert-${type}`, msg)).isPresent()).toBe(false);
```
tomaskikutis: Can you add test IDs for messages? I would exclude the type of the message since in most cases you will either want to check for a message with a specific text or to ensure that there aren't any messages at all. We could add type information if we find that we really need to, but I think it'd be simpler to start without it.

spec/authoring_spec.ts
```diff
@@ -716,6 +724,62 @@ describe('authoring', () => {
         expect(authoring.getAbstractText()).toBe('item6 text');
     });
 
+    it('Not modifying crops will not trigger an article change', () => {
```
tomaskikutis: Good job on writing tests decoupled from the DOM structure. Any feedback on writing them this way with these helpers? Do you have suggestions for improvement?

pablopunk: No suggestions, it was a good experience overall, also they make tests more readable

spec/authoring_spec.ts
pablopunk: I don't remember why I put that there, probably just something I forgot to remove

spec/helpers/utils.ts
pablopunk: I made it work like `assertToastMsg` and that one also has a type arg. Maybe it is a good idea if you don't want to search for text but for the type only (success, error...)

pablopunk: You mean like searching for `'[data-test-id="notification-holder"]'` or something like that? Should I include the `type` in there too?

tomaskikutis: Yep. I don't think we need a "holder" since it's a DOM related wrapper and DOM structure is not that relevant in tests. I'd use `[data-test-id="notification--error"]` `[data-test-id="notification--success"]`, etc.

PR #3367, opened 2020-02-25
run e2e tests on travis

petrjasek:
with test server on docker


petrjasek:
will probably have to make them pass first :sweat_smile:

PR #3369, opened 2020-02-25
when dragging item with custom mimetype use that

petrjasek:
but only for `application/*` mimetypes, not `image/jpeg` etc.
also resize the image thumbnail when dragging

SDBELGA-325

### DIFF
--- a/scripts/apps/archive/services/DragItemService.ts
+++ b/scripts/apps/archive/services/DragItemService.ts
@@ -6,6 +6,11 @@
 
 const DRAG_IMAGE_HOLDER = 'drag-image-holder';
 
+/**
+ * Using div with max width/height to resize thumbnail image.
+ *
+ * link: https://stackoverflow.com/a/26347784
+ */
 const getThumbnailPlaceholder = () => {
     let div = document.getElementById(DRAG_IMAGE_HOLDER);
 
--- a/scripts/apps/archive/services/DragItemService.ts
+++ b/scripts/apps/archive/services/DragItemService.ts
@@ -1,4 +1,26 @@
 /**
+ * NOTICE:
+ * if you want to test drag image in chrome, make sure to close dev tools first,
+ * otherwise this thumbnail image might not be displayed, depends on some performance settings there.
+ */
+
+const DRAG_IMAGE_HOLDER = 'drag-image-holder';
+
+const getThumbnailPlaceholder = () => {
+    let div = document.getElementById(DRAG_IMAGE_HOLDER);
+
+    if (div == null) {
+        div = document.createElement('div');
+        div.id = DRAG_IMAGE_HOLDER;
+        document.body.appendChild(div);
+    } else {
+        div.innerHTML = ''; // removes all children
+    }
+
+    return div;
+};
+
+/**
  * Service to handle item dragging
  */
 export function DragItemService() {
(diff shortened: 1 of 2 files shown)


scripts/apps/archive/services/DragItemService.ts
```diff
@@ -9,18 +31,24 @@ export function DragItemService() {
      * @param {Object} item
      */
     this.start = function(event, item) {
-        var dt = event.dataTransfer || event.originalEvent.dataTransfer;
+        const dt = event.dataTransfer || event.originalEvent.dataTransfer;
+
+        // search providers could specify custom mimetype
```
tomaskikutis: > // search providers ~could~ **can** specify custom mimetype

Or can't they do it anymore?

scripts/apps/search/styles/search.scss
```diff
@@ -541,4 +541,17 @@
...
+
+
+// used to resize thumbnail of item while dragging it
+#drag-image-holder {
+    position: absolute;
+    left: -2000px;
+    float: left;
+    max-width: 200px;
+
+    img {
```
tomaskikutis: Can you post a gif to illustrate what's the problem with dragging? It'd be nice not to rely upon certain elements being inside. Can't a clone of the element be displayed while dragging or doesn't it work for us for some reason?

petrjasek: seems like I can't do it, that dragged image is not visible while doing screen grabbing or screen shot. but you can see the big pictures [here](https://bel-master.test.superdesk.org/#/search?repo=5cac5110405ecc00e4403708). clone doesn't work, it would always use the full size of image, that's why container is needed

scripts/apps/archive/services/DragItemService.ts
petrjasek: they can, my bad

scripts/apps/search/styles/search.scss
petrjasek: @tomaskikutis so did a screenshot with phone..
before:
DSC_0049
after:
DSC_0050

scripts/apps/search/styles/search.scss
```diff
@@ -541,4 +541,17 @@
...
         }
     }
+}
+
+
+// used to resize thumbnail of item while dragging it
+#drag-image-holder {
+    position: absolute;
+    left: -2000px;
+    float: left;
```
tomaskikutis: Do you need a float for some reason? It works fine for me without it, at least in chromium.

You should probably limit `max-height` too since if you have a `200 x 800` image it won't look good either.

scripts/apps/archive/services/DragItemService.ts
```diff
@@ -1,3 +1,25 @@
+/**
+ * NOTICE:
+ * if you want to test drag image in chrome, make sure to close dev tools first,
+ * otherwise this thumbnail image might not be displayed, depends on some performance settings there.
+ */
+
+const DRAG_IMAGE_HOLDER = 'drag-image-holder';
+
+const getThumbnailPlaceholder = () => {
```
tomaskikutis: Can you describe or add a link to this "resizing" hack?

scripts/apps/search/styles/search.scss
tomaskikutis: That's weird. I've tried recording it with the simple screen recorder and it did appear in the video for me.

petrjasek: I guess it wasn't on debian..

petrjasek: and it works on ubuntu I have on other PC actually, probably gnome/wayland version problem

PR #3374, opened 2020-03-09
Make editor3 fields html if they are not single line

pablopunk:
SDESK-5113

This will also revert the changes in https://github.com/superdesk/superdesk-client-core/pull/3351 as they are not needed anymore (because that issue was caused by body_html not being really html)


pablopunk:
I'm gonna mark this as WIP cause I'm not sure if this will impact some current CPs or get old issues back

PR #3375, opened 2020-03-09
Introduce config for minimum size of uploaded/fetched images. [SDESK-5056]

GyanP:


scripts/apps/archive/controllers/UploadController.ts
```diff
@@ -245,16 +246,15 @@ export function UploadController(
             return false;
         }
 
-        let acceptedFiles: Array<{file: File, getThumbnail: (file: File) => Promise<string>}> = [];
+        let acceptedFiles: Array<{ file: File, getThumbnail: (file: File) => Promise<string> }> = [];
         let uploadOfDisallowedFileTypesAttempted: boolean = false;
+        const imageFiles = [];
```
tomaskikutis: Can you use `acceptedFiles` array and avoid adding `imageFiles`?

There's a problem with this approach, that images are iterated synchronously, but making it async is a cleaner solution than adding additional arrays.

It would also be good to implement it in a way that if some images are too small, the ones which aren't would get uploaded and a message would pop up with the list of filenames of images that weren't uploaded and all their dimensions in addition to required dimensions that are currently displayed.

Here's some psuedo-code on converting the iteration to be async:

```
Promise.all(files.map((file) => {
    if (isImage) {
        if (!imagesAllowed) {
            return Promise.resolve(`Images aren't allowed`);
        }

        return fileDimensionsValid(file).then((dimensions) => {
            if (dimensions.valid) {
                return {
                    file: files,
                    getThumbnail: () => Promise.resolve('<i class="icon--2x icon-video"></i>'),
                };
            } else {
                return {error: 'dimensions not valid', dimensions: dimensions};
            }
        });
    } else {
        // ...
    }
}).then((result) => {
    // add files without errors to acceptedFiles

    // upload

    // display error messages
});

```

scripts/apps/archive/controllers/UploadController.ts
```diff
@@ -245,70 +246,122 @@ export function UploadController(
...
 
-            notify.error(message);
-        }
+                notify.error(message);
+            }
+
+            if (invalidFiles.length > 0) {
+                notify.error(
+                    gettext(
+                        `The image you\'re trying to add is smaller than
```
tomaskikutis: How do you know if `invalidFiles` contains images and not other file types? You should either call the variable `invalidImages` or check the type of invalid files.

Can you include the dimensions of the current image to the error message too? `Minimal allowed image size is {{minWidth}}x{{minHeight}}. The size of {{file.name}} is {{file.width}}x{{file.height}}.`

I'd say notifications aren't good for this, because if you are uploading 10 images, the notifications boxes will close sooner than users will be able to read the messages. Can you use a modal with all the messages in one place instead?

scripts/apps/workspace/content/services/ContentService.ts
```diff
@@ -344,21 +344,38 @@ export function ContentService(api, templates, desks, packages, archiveService,
...
-            if (item._type === 'ingest') {
-                return send.one(item);
-            }
+        if (appConfig.pictures != null
+            && (item.renditions.original.width < appConfig.pictures.minWidth
+                || item.renditions.original.height < appConfig.pictures.minHeight)) {
+            notify.error(
+                gettext(
+                    `The image you\'re trying to add is smaller than
+                    {{width}} x {{height}} pixels. Please use another one. File names are: {{filename}}`,
```
tomaskikutis: can you include image dimensions too?

scripts/apps/archive/controllers/UploadController.ts
```diff
@@ -315,7 +322,7 @@ export function UploadController(
                 } else if (file.error.isAllowedFileType === false) {
                     uploadOfDisallowedFileTypesAttempted = true;
                 } else {
-                    invalidFiles.push(file.error.filename);
+                    invalidImages.push(file.error);
```
tomaskikutis: How do you know that `file` is an image at this line? `result` can contain audio and videos too.

scripts/apps/archive/controllers/UploadController.ts
```diff
@@ -328,18 +335,32 @@ export function UploadController(
...
-                            height: appConfig.pictures.minHeight,
-                            filename: invalidFiles.join(', '),
+                            name: image.name,
+                            width: image.width,
+                            height: image.height,
+                        },
+                    );
+                });
+
+                modal.alert({
```
tomaskikutis: Can you use [showModal](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/services/modalService.tsx#L7) when displaying HTML content?

scripts/apps/archive/controllers/UploadController.ts
```diff
@@ -328,18 +335,32 @@ export function UploadController(
...
+                    headerText: gettextPlural(
+                        images.length,
+                        `The image you\'re trying to add is smaller than {{minWidth}}x{{minHeight}}
```
tomaskikutis: It's better to keep titles short and explain the issue in detail in the modal body. Something like "(Some) files were not uploaded" or "Uploading error" would be enough for the title.

It would also be useful to mention that other files were successfully uploaded and only the ones with errors have to be reuploaded, given that's how it works.

scripts/apps/archive/controllers/file-upload-error-modal.tsx
```diff
@@ -0,0 +1,98 @@
...
+                                    'The file was not uploaded',
+                                    'Some files were not uploaded',
+                                )
+                            }
+                        </ModalHeader>
+                        <ModalBody>
+                            <h4>
+                                {
+                                    gettext(
+                                        'Minimum allowed image size is {{minWidth}}x{{minHeight}}',
```
tomaskikutis: Looks like you are displaying the error about incorrect image size unconditionally for all files including non-image files. Could you explain this? When I'm asking a question, it doesn't mean that your solution is incorrect or needs changes. I just want to hear your reasoning behind the decision. This is the same question as these two https://github.com/superdesk/superdesk-client-core/pull/3375#discussion_r392095380 https://github.com/superdesk/superdesk-client-core/pull/3375#discussion_r392095380

scripts/apps/archive/controllers/file-upload-error-modal.tsx
```diff
@@ -0,0 +1,98 @@
...
+                                }
+                            </h4>
+                            <ol>
+                                {
+                                    invalidFiles.map((file, index) => {
+                                        if (file.type !== null) {
+                                            if (file.type.startsWith('image')) {
+                                                return (
+                                                    <li key={index}>
+                                                        {gettext('The size of {{name}} is {{width}}x{{height}}',
```
tomaskikutis: It'd be better to call it `{{filename}}` instead of `{{name}}` so the context of the message is clear for translators.

scripts/apps/workspace/content/services/ContentService.ts
```diff
@@ -344,21 +347,37 @@ export function ContentService(api, templates, desks, packages, archiveService,
...
      */
     this.dropItem = (item: IArticle, {fetchExternal} = {fetchExternal: true}) => {
-        if (item._type !== 'externalsource') {
-            if (item._type === 'ingest') {
-                return send.one(item);
-            }
+        const invalidFiles = [];
+
+        if (appConfig.pictures != null
+            && (item.renditions.original.width < appConfig.pictures.minWidth
```
tomaskikutis: You should probably be using optional chaining when accessing renditions since an item might not have them and an exception would be thrown when you try to access a property of `undefined`

scripts/apps/archive/controllers/UploadController.ts
GyanP: I was returning the error object in case of images only.

scripts/apps/archive/controllers/file-upload-error-modal.tsx
GyanP: yeah, I'll fix this. But the idea is that I'm returning the error object in case of invalid images only. And in case of audio and video will simply show the name of the invalid files as we don't know what could be the error beforehand.

tomaskikutis: > But the idea is that I'm returning the error object in case of invalid images only.

I understand that this "works", but it's a very unreliable solution in the long run because it's implicit and for a developer making future changes there is no way to know that it would cause a bug if an error object was returned if there's an error for a video file for example. Thanks for fixing this.

scripts/core/superdesk-api.d.ts
```diff
@@ -1218,6 +1218,10 @@ declare module 'superdesk-api' {
         editor3: {
             browserSpellCheck: boolean;
         };
+        pictures: {
```
tomaskikutis: shouldn't this be optional? You can do it by adding a question mark `pictures?: {...`

PR #3376, opened 2020-03-10
[SDESK-5073] Bottom bar with open stories tabs stretched to two lines

fritzSF:


scripts/apps/authoring/styles/authoring.scss
```diff
@@ -212,99 +212,145 @@
...
+    background: rgba(20, 20, 20, 0.95);
     z-index: $zindexFixedNavbar;
-
-    .opened-articles__list > li {
-        @include border-box();
-        padding: 0 1.6rem;
-        max-width: 24rem;
+    white-space: nowrap;
+    max-width: 100vw;
+    display: grid;
```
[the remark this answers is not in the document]
fritzSF: I hade problems making it behave like intended in all cases with flex. Grid produced better results when there is a lot of items in the list. And how simpler would it really be?

scripts/apps/authoring/styles/authoring.scss
```diff
@@ -212,99 +212,145 @@
...
-    background: rgba(0,0,0,0.85);
+    background: rgba(20, 20, 20, 0.95);
     z-index: $zindexFixedNavbar;
-
-    .opened-articles__list > li {
-        @include border-box();
-        padding: 0 1.6rem;
-        max-width: 24rem;
+    white-space: nowrap;
+    max-width: 100vw;
```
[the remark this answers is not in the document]
fritzSF: This component should always take the full width of the screen so it should be fine. The result is the same in this case, you are right.

scripts/apps/authoring/styles/authoring.scss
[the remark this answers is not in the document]
tomaskikutis: For one, flexbox is older thus more common and more people know how to use it. For example, I never took the time to learn how CSS grids work since I've never had a problem that couldn't be solved with flexbox.

From my initial impressions, grid also seems to be less flexible.

> grid-template-columns: 48px 1fr auto auto;

For example, here the container is coupled to its children. If I wanted to add another button or two, I'd need to also update the widths in container styles or add more columns. When using flexbox, you define which parts grow and which shrink, but the width of children doesn't need to be specified and adjust automatically. See https://jsfiddle.net/9n6sbuto/

I'm wondering which cases were hard to make behave correctly?

superdesk.config.js
```diff
@@ -7,6 +7,9 @@
 
 module.exports = function(grunt) {
     return {
+        apps: [
```
tomaskikutis: It should be enabled from [main]( https://github.com/superdesk/superdesk-belga/blob/master/client/superdesk.config.js#L11) repositories instead of here

pavlovicnemanja: sorry, forgot to exclude it

scripts/apps/authoring/styles/authoring.scss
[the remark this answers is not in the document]
fritzSF: > For one, flexbox is older thus more common and more people know how to use it. For example, I never took the time to learn how CSS grids work since I've never had a problem that couldn't be solved with flexbox.

There are a quite a few of things grid does that flexbox doesn't. You should take the time as it is fully supported by all major browsers. As an example what it does that flex doesn't take a look at the Photo grid view in Superdesk. 

> For example, here the container is coupled to its children. If I wanted to add another button or two, I'd need to also update the widths in container styles or add more columns. When using flexbox, you define which parts grow and which shrink, but the width of children doesn't need to be specified and adjust automatically. See https://jsfiddle.net/9n6sbuto/

No, you don't have to specify any width. 48px is reserved for the button on the far left, which should always be the same width as the sidebar. The rest is flexible, 1fr stretches the item and auto adapts so it can fit more items inside. The example you provided is pretty much how it was before and didn't behave as desired. 

> I'm wondering which cases were hard to make behave correctly?
With a large number of open articles, the list started to push the Create button outside the viewport. With grid, it was not the case. 

But let's not turn this into a long debate – honestly, it could be just my personal preference and it would still be a perfectly valid way to do it.

scripts/apps/authoring/styles/authoring.scss
```diff
@@ -212,99 +212,150 @@
...
+        flex-shrink: 1;
+        flex-wrap: nowrap;
+        overflow-x: auto;
+        grid-column: 2/3;
+    }
+    .opened-articles-bar__item {
+        flex: 0 1 24rem;
+        border-right: 1px solid $bottom-bar-item-border-color;
+        padding: 0 0.4rem 0 1.2rem;
+        //max-width: 20rem;
```
tomaskikutis: this should probably be removed

scripts/apps/authoring/styles/authoring.scss
```diff
@@ -212,99 +212,150 @@
...
-
-    .opened-articles__list > li {
-        @include border-box();
-        padding: 0 1.6rem;
-        max-width: 24rem;
+    white-space: nowrap;
+    max-width: 100vw;
+    display: grid;
+    grid-template-rows: 32px;
+    grid-template-columns: 48px 1fr auto auto;
```
tomaskikutis: if that `48px` is for aligning the button with the sidebar, can you make the two to use the same variable for the width?

scripts/apps/authoring/styles/authoring.scss
```diff
@@ -212,99 +212,150 @@
...
+            background: $sd-blue;
         }
     }
-    .multiedit-item {
-        float: right;
-        > li {
-            border-left: 1px solid #333;
-            width: 24rem;
+    .opened-articles-bar__multiedit {
+        grid-column: 3/4;
```
tomaskikutis: I'd be good if it were more flexible and adding/removing buttons didn't require stylesheet changes.

fritzSF: It doesn't require changes.

scripts/apps/authoring/styles/authoring.scss
fritzSF: yep, this should be done.

scripts/apps/authoring/styles/authoring.scss
tomaskikutis: I was referring to [action buttons on the right](https://user-images.githubusercontent.com/2076953/76771875-49466580-67a0-11ea-839c-c8f32bbab656.png). If you add another, one gets pushed out. It'd be nice to have a container element where you can put as many buttons as you want and they simply expand the container to fit.

fritzSF: I don't think this is necessary. This element does not have to be flexible that way. All changes that may come in the future should go trough the UI design team anyway and if some CSS adaptation is needed we will do them.

tomaskikutis: Many things aren't necessary if you think about it this way. Using a variable for that 48px value isn't necessary either - worked as good without it. It's a matter of clean code to allow for flexibility even if you don't need it just yet, especially when it's not that hard to do...

PR #3377, opened 2020-03-10
feat(editor3): Added editor3 as default in plain text CP

pavlovicnemanja:
- [SDESK-5032] The Plain text profile should use editor 3


tomaskikutis:
continued in https://github.com/superdesk/superdesk-client-core/pull/3602

PR #3378, opened 2020-03-11
Update incoming stage when correcting related item

pablopunk:
SDESK-5076

The items on Incoming Stage were not being updated cause the `content:update` event was being triggered from the working stage, and checking the stage id wouldn't work there. I added an additional check that triggers the update only when the stage contains one of the updated item ids


pablopunk:
Hey @petrjasek this one needs to go on `1.33.1`. Should I cherry-pick this for `1.33` branch?

petrjasek:
hi, if you push it there now it would make it to 1.33, so lets wait a bit

tomaskikutis:
It'd be nice if we could have some automation to do cherry-picking and developers could simply push everything to develop without having to worry about what goes where. We could do it manually when we are doing bugfixes on released code, but otherwise, it'd save time and effort to have it automated.

pablopunk:
that can be tricky as cherry-picking sometimes causes conflicts

tomaskikutis:
That's why I don't suggest doing auto cherry picking on released code since differences are likely to be larger, but for develop vs a branch that isn't too old, the conflicts should be minimal.

tomaskikutis:
the person that is doing the merging would still have to resolve some conflicts, but there wouldn't be a need to search for commits that need to be merged. I was referring to automate the gathering of such commits, not the entire merging process

petrjasek:
in theory it would be possible if we had good release planning and we would only work on things that go to next release until that release is done, we could just use develop for that. but now while we work on multiple releases at the same time (1.33, 1.33.1, 2.0) I'm not sure how to handle it. it would be simpler if we push changes for 1.33 into `release/1.33` branch and merge that branch into develop rather then the other way around which requires cherry-picking

tomaskikutis:
We may try it for a few next releases, but I suspect we might get into issues when something has to be release specific or we need some dependencies from `develop` which might be hard to merge to the release branch given that QA is done after merging and many tickets end up being resolved via multiple commits.

PR #3380, opened 2020-03-12
Filter marked for me in monitoring

tomaskikutis:
SDBELGA-156

DEPENDS ON: https://github.com/superdesk/superdesk-ui-framework/pull/442


scripts/apps/monitoring/directives/MonitoringFilteringButtons.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+class FilterButton extends React.PureComponent<IPropsFilterButton, IState> {
+    getFilters: () => {[key: string]: any};
+
+    constructor(props: IPropsFilterButton) {
+        super(props);
+
+        this.state = {};
+
+        this.getFilters = () => ({'task.desk': [this.props.deskId], ...this.props.button.query});
```
pablopunk: why is `getFilters` a function instead of a plain object?

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -370,6 +371,26 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
         });
     }
 
+    this.isCustomFilterActive = (filter: IMonitoringFilter) => {
+        return Object.keys(this.activeFilters.customFilters ?? {}).includes(filter.label);
```
pablopunk: what about simply:

```js
return this.activeFilters.customFilters?.[filter.label] != null
```

scripts/apps/monitoring/directives/MonitoringFilteringButtons.tsx
tomaskikutis: because `getFilters` depends on props and is initialized in the constructor. When props change, the constructor is not called again. If it were a plain object it would get stuck with initial values and wouldn't react to prop changes.

scripts/apps/monitoring/controllers/AggregateCtrl.ts
tomaskikutis: it may not work with certain filters like `filter all items which don't have a slugline` which might be `customFilters = {slugline: null}`

pablopunk: You could also use `in` notation:

```js
return filter.label in this.activeFilters.customFilters ?? {}
```

This might be more readable.

tomaskikutis: `in` operator doesn't work the same. It also includes keys from inherited objects which not desired in most cases.

```
Object.keys(a)
[]
'a' in a
true
```

I could use a [helper function](https://user-images.githubusercontent.com/2076953/76621876-a76d1180-6530-11ea-8411-d9d24746c870.png), but it doesn't increase readability to me. What do you think?

pablopunk: nah I think it's good as it is now, readable enough

PR #3382, opened 2020-03-13
Fix ts compiler warning

pablopunk:
Captura de pantalla 2020-03-13 a las 14 31 34

This was preventing users from creating new metadata items due to this error:

Captura de pantalla 2020-03-13 a las 14 32 38


petrjasek:
@pablopunk any idea if that's only affecting develop branch or also 1.33?

pablopunk:
@petrjasek I don't think so, it was introduced here https://github.com/superdesk/superdesk-client-core/commit/4db9a3f18ec37fba8f69ccc8e2282f14e8b1b5bf

PR #3383, opened 2020-03-17
Omit invalid translations on the server before returning the JSON to the client. [SDCP-187]

GyanP:


petrjasek:
that's file generated on transifex but then it's stored as part of client core and fetched by client when user changes language

tomaskikutis:
How does it get here from transifex? We need a convenient place to do processing on it to remove invalid translations.

petrjasek:
there is this `nggettext_compile` grunt task which creates json from po

tomaskikutis:
@GyanP can you add code in that grunt task to automatically strip out invalid translations so they don't get to JSON?

petrjasek:
will merge it as I need it for the release, lets make that check in another PR

PR #3384, opened 2020-03-17
Fix e2e test: wait for attribute

tomaskikutis:
fails on 1.33 release branch, but the fix should be merged to develop too


tomaskikutis:
@pablopunk @petrjasek can you review the code? I know the tests are failing, but it's because a release of e2e testing helpers package is required and I want to do it after the code is reviewed.

PR #3388, opened 2020-03-19
feat(multi): Added duplicate actions in multi action bar

pavlovicnemanja:
[SDANSA-373] - Allow Multiple items duplication

### DIFF
--- a/scripts/apps/monitoring/MultiActionBarReact.tsx
+++ b/scripts/apps/monitoring/MultiActionBarReact.tsx
@@ -133,8 +133,10 @@ export class MultiActionBarReact extends React.Component<IProps, IState> {
                 </div>
             );
         } else {
-            const groups = groupBy(this.state.actions, 'group');
+            const groups = groupBy(this.state.actions, (item) => item.group && item.group.label);
             const groupNames = Object.keys(groups);
+
+            console.log(groupNames);
 
             return (
                 <div data-test-id="multi-actions-inline">
--- a/scripts/apps/monitoring/MultiActionBarReact.tsx
+++ b/scripts/apps/monitoring/MultiActionBarReact.tsx
@@ -136,8 +136,6 @@ export class MultiActionBarReact extends React.Component<IProps, IState> {
             const groups = groupBy(this.state.actions, (item) => item.group && item.group.label);
             const groupNames = Object.keys(groups);
 
-            console.log(groupNames);
-
             return (
                 <div data-test-id="multi-actions-inline">
                     {
--- a/scripts/apps/monitoring/MultiActionBarReact.tsx
+++ b/scripts/apps/monitoring/MultiActionBarReact.tsx
@@ -159,9 +159,10 @@ export class MultiActionBarReact extends React.Component<IProps, IState> {
                                     <button
                                         onClick={onClick}
                                         className="navbtn"
+                                        title={groups[group][0].group.label}
                                         data-test-id="dropdown-toggle"
                                     >
-                                        <Icon className={groups[group][0].icon} size={22} />
+                                        <Icon className={groups[group][0].group.icon} size={22} />
                                     </button>
                                 )}
                                 inline={true}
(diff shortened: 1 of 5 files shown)


scripts/apps/monitoring/MultiActionBarReact.tsx
```diff
@@ -133,22 +133,75 @@ export class MultiActionBarReact extends React.Component<IProps, IState> {
...
+                                    className="navbtn strict"
+                                    title={action.label}
+                                    key={key}
+                                    data-test-id={action.label}
+                                >
+                                    <Icon className={action.icon} size={22} />
+                                </button>
+                            )) :
+                            <DropdownTree
+                                className="dropdown navbtn strict"
```
petrjasek: I wonder if that className + different styles could be named somehow so it would describe what it does..
maybe some `inline={true}` property could trigger that class and style rather than this

pavlovicnemanja: yea, I'm not happy with this neither. Issue is that DropdownTree has some default inline styles that are overriding styles in classes, because of that I needed to add both things (className and stylings). This case is especially for nav btns, and can be applied only in this multiedit bar. We had meeting today and we agreed on creating some unified dropdown component that can cover all the cases.

petrjasek: ok so can you add some comment that this will be handled?

pavlovicnemanja: Hmm, but yes, I think we can go with inline, I just checked other places where this is implemented, it will work in any case

scripts/core/superdesk-api.d.ts
```diff
@@ -46,6 +46,7 @@ declare module 'superdesk-api' {
         priority?: IDisplayPriority;
         label: string;
         icon: string;
+        group?: string;
```
tomaskikutis: Can you rename this to `groupLabel` and output it as a title of the icon for accessibility? It would also be useful to be able to set a group icon. We could have the interface be `group: {label: string: icon: string};`

scripts/core/ui/components/dropdown-tree.tsx
```diff
@@ -47,12 +47,15 @@ export class DropdownTree<T> extends React.PureComponent<IPropsDropdownTree<T>,
...
     render() {
-        const {groups, getToggleElement} = this.props;
+        const {groups, getToggleElement, inline} = this.props;
         const onClick = () => this.setState({open: !this.state.open});
 
         return (
             <div
-                style={{display: 'flex', position: 'relative', lineHeight: 'initial'}}
+                style={inline ?
+                    {display: 'inline-flex', lineHeight: 'initial', verticalAlign: 'bottom'} :
```
tomaskikutis: it'd be better to control the alignment from a parent element since in other cases you might need it aligned to the center for example

tomaskikutis:
Looks like grouping isn't being applied in [minimized view](https://user-images.githubusercontent.com/2076953/77067152-cc57fd80-69e4-11ea-9277-985e75a62cd8.png)

pavlovicnemanja:
@tomaskikutis yep that is on purpose, I don't want to group them on minimized view

scripts/core/ui/components/dropdown-tree.tsx
pavlovicnemanja: This is alignment for button only, not for dropdown menu

scripts/apps/search/directives/MultiActionBar.ts
```diff
@@ -214,7 +214,10 @@ export function MultiActionBar(
                     actions.push({
                         label: gettext('Duplicate To'),
                         icon: 'icon-copy',
-                        group: 'Duplicate',
+                        group: {
+                            label: 'Duplicate',
```
tomaskikutis: don't forget to translate the label and output it in the markup as a title

PR #3389, opened 2020-03-20
Fixes on Canadian French translations

pablopunk:
SDCP-199

It's still missing the translation for "Personal Space". I don't find it on the `.po` file, how are these generated @petrjasek ?

### DIFF
--- a/po/superdesk.pot
+++ b/po/superdesk.pot
@@ -4,82 +4,94 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Project-Id-Version: \n"
 
-#: scripts/core/lang/missing-translations-strings.js:27
+#: scripts/core/lang/missing-translations-strings.ts:32
 msgid "\"GENRE 'null value not allowed'"
 msgstr ""
 
-#: ../superdesk-planning/client/components/Planning/FeaturedPlanning/UnlockFeaturedPlanning.jsx:51
+#: ../superdesk-planning/client/components/Planning/FeaturedPlanning/UnlockFeaturedPlanning.jsx:74
 msgid "'' is currently managing featured stories. This feature can only be accessed by one user at a time."
 msgstr ""
 
-#: scripts/core/lang/missing-translations-strings.js:18
+#: scripts/core/lang/missing-translations-strings.ts:23
 msgid "'ABSTRACT is a required field'"
 msgstr ""
 
-#: scripts/core/lang/missing-translations-strings.js:19
+#: scripts/core/lang/missing-translations-strings.ts:24
 msgid "'ANPA_CATEGORY is a required field'"
 msgstr ""
 
-#: scripts/core/lang/missing-translations-strings.js:18
+#: scripts/core/lang/missing-translations-strings.ts:23
 msgid "'BYLINE is a required field'"
 msgstr ""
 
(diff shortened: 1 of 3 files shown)


petrjasek:
with grunt task `nggettext_extract` it generates pot file which must contain it, then we translate that pot via transifex to get po file

pablopunk:
@petrjasek so should I do that manually and update the PR?

petrjasek:
yes you can update it there as well

scripts/apps/monitoring/views/monitoring-view.html
```diff
@@ -55,7 +55,7 @@ <h3 class="subnav__page-title sd-flex-no-grow sd-empty">
                                 <button class="sd-checkbox sd-checkbox--button-style"
                                     ng-class="{checked: personalShowSent}"
                                     ng-click="togglePersonalShowSent()"
-                                    translate>Sent Items</button>
+                                    translate>{{ gettext('Sent Items') }}</button>
```
petrjasek: this should work via `translate` directive, is it not being extracted or translated?

pablopunk: It was not on the `.po` file 🤔

petrjasek: well it might not be translated on transifex, so better check if it's in pot file after extraction

tomaskikutis: a translate directive should be enough for it to be extracted. I was testing the translations and the ones from `gettext` calls are only extracted from js/ts files, not .html. We shouldn't be using gettext in `.html` files. If it's not extracting it from a HTML file there's likely some parsing error, try removing sections of the file until it gets extracted.

pablopunk: Yeah I updated it and generated the corresponding `.pot`

po/superdesk.pot
```diff
@@ -11517,72 +12609,92 @@ msgstr ""
...
-#: ../superdesk-planning/client/actions/assignments/notifications.js:318
+#: scripts/apps/monitoring/views/aggregate-settings.html:109
+#: scripts/apps/monitoring/views/aggregate-settings.html:88
+msgid "{{ output.listLabel }}"
+msgstr ""
+
+#: scripts/apps/monitoring/views/monitoring-view.html:23
+msgid ""
+"{{ profile.label }}\n"
+"                                        <i ng-show=\"aggregate.activeFilters.contentProfile.includes(profile._id)\" class=\"icon-ok\"></i>"
```
petrjasek: looks like there is some extra `translate`

pablopunk: I'm not sure what you mean

petrjasek: I mean https://github.com/pablopunk/superdesk-client-core/blob/7c080e95badc1aaa6d77cb95d762b2baed01d94c/scripts/apps/monitoring/views/monitoring-view.html#L26-L28 there it `translate>` which makes this whole thing including `<i ng-show...` being string for translation on transifex

pablopunk: ok I can make another PR with a `<span translate>` to get only the label

pablopunk: https://github.com/superdesk/superdesk-client-core/pull/3391

PR #3390, opened 2020-03-23
fix(multiedit): Fixed issue with missing actions on multiedit bar

pavlovicnemanja:
[SDANSA-387] Fixed issue with missing action on published items in multiedit


tomaskikutis:
"Duplicate in place" shouldn't be available for published items. Also when you perform duplication in the working stage, multiselection should disappear after the action is completed like with other options.

PR #3394, opened 2020-03-24
Fix creating users

tomaskikutis:
SDESK-5141

### DIFF
--- a/scripts/apps/users/views/edit-form.html
+++ b/scripts/apps/users/views/edit-form.html
@@ -1,5 +1,5 @@
 <form name="userForm">
-    <div ng-if="user != null">
+    <div>
         <div class="action-bar clearfix" ng-class="{show:dirty}">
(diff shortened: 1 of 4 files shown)


scripts/apps/users/views/edit-form.html
```diff
@@ -1,5 +1,5 @@
 <form name="userForm">
-    <div ng-if="user != null">
```
petrjasek: I think we need this, otherwise if you select a user and then click on creating new one the form stays filled with selected user details.

tomaskikutis: I added the condition only a few weeks ago so the avatar component doesn't crash when the user object is null or some properties are missing so it shouldn't be related with selected user details staying when creating a new one. At least I can't reproduce it.

Peek 2020-03-24 12-41

petrjasek: yep looks ok, I have some weird behaviour if I have developer tools opened, but when I close it it's all good

tomaskikutis: That's interesting. For me, it works the same no matter if I have devtools opened or not. Maybe it's some extension of yours that's messing things up

PR #3397, opened 2020-03-25
Fix infitite loop in user management

tomaskikutis:
SDESK-5144


petrjasek:
seems like now if you select user for preview and then switch to another one the preview doesn't change

PR #3400, opened 2020-03-26
Fix related items gallery order in preview

tomaskikutis:
SDESK-4826


tomaskikutis:
@GyanP check `6ea58ae`. It's a fix for all items having zero set as order on upload to media gallery. Looks like you did the same thing here https://github.com/superdesk/superdesk-client-core/pull/3158, but it didn't work for some reason

GyanP:
> Looks like you did the same thing here #3158, but it didn't work for some reason

That's weird, if it wasn't working then ninjs output would have been wrong as well. Maybe missed testing by uploading multiple images from external folder together.

GyanP:
Not sure but I think you should sort by order in case of full [preview](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/authoring/authoring/directives/FullPreviewDirective.ts#L68) as well.

tomaskikutis:
> That's weird, if it wasn't working then ninjs output would have been wrong as well

The order fields get populated when reordering so maybe that's why you saw it in the output

> Not sure but I think you should sort by order in case of full preview as well.

sure, I forgot that full preview exists :)

scripts/apps/authoring/authoring/controllers/AssociationController.ts
```diff
@@ -134,8 +136,15 @@ export function AssociationController(content, superdesk, renditions, notify) {
         }
 
         if (scope.field != null && scope.field.field_type === 'media' && updated != null && updated.order == null) {
-            // if the field is of type media-gallery, assign order to the item being added
-            updated['order'] = scope.currentIndex;
+            // get greatest order from current items(or -1 if there aren't any items) and add one
+            const nextOrder = (
+                Object.keys(scope.item.associations)
```
petrjasek, line 141: I guess you could use here `getAssociationsByFieldId` here as well:
```ts
const items = getAssociationsByFieldId(...);
const nextOrder = items.length > 0 ? items[items.length - 1].order + 1 : 0;
```

tomaskikutis: I've made it use `getAssociationsByFieldId`, but I am still reordering it in order not to depend on the sort order of `getAssociationsByFieldId`.

PR #3404, opened 2020-03-27
fix user management diffs

tomaskikutis:
SDESK-5145

the problem was partially in the database since it had `dateline_source` set to an empty string when in theory it should only be a non-empty string or `null`. I've adjusted the code to only write `null` if it wasn't an empty string originally and was changed by a user to an empty string.

### DIFF
--- a/scripts/apps/users/directives/UserEditDirective.ts
+++ b/scripts/apps/users/directives/UserEditDirective.ts
@@ -10,6 +10,10 @@ import {noop} from 'lodash';
 // according to the latest data on the server
 let clearOrigUserWatcher = noop;
 
+// only initialize after selecting a user for editing/creation
+// having it running when switcing between users can cause fields to be modified or produce an incorrect dirtiness value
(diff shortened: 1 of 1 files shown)


petrjasek:
so if I revert that revert on `release/1.33` branch and add this it should work?

scripts/apps/users/directives/UserEditDirective.ts
```diff
@@ -10,6 +10,10 @@ import {noop} from 'lodash';
 // according to the latest data on the server
 let clearOrigUserWatcher = noop;
 
+// only initialize after selecting a user for editing/creation
+// having it running when switcing between users can cause fields to be modified or produce an incorrect dirtiness value
+let clearUserWatcher = noop;
```
petrjasek: I think it would be better to have this in the link func, otherwise scope and element would probably stay around even when that directive is destroyed

tomaskikutis:
yep, I've tested it on that tcp uat server

PR #3406, opened 2020-03-27
use 2.1.15 version of ui-framework so planning styles are fixed

tomaskikutis:
SDESK-5140


tomaskikutis:
update to 2.1.16 instead https://github.com/superdesk/superdesk-client-core/pull/3413

PR #3409, opened 2020-03-30
fix unstable e2e test

petrjasek:


spec/content_spec.ts
```diff
@@ -184,7 +184,7 @@ describe('content', () => {
         storyNameEl.sendKeys('package1');
 
         element(by.id('searchItemByNameBtn')).click();
-        browser.sleep(500);
+        browser.sleep(1000);
```
tomaskikutis: it would probably be more stable to wait for a condition instead of setting a timeout. See https://github.com/superdesk/superdesk-client-core/blob/develop/spec/authoring_spec.ts#L770

PR #3411, opened 2020-03-31
feat(editor3): Support tabs

pablopunk:
SDESK-5158

Requires new version of draft-js (https://github.com/tomaskikutis/draft-js/pull/1)


scripts/core/editor3/reducers/editor3.tsx
```diff
@@ -283,11 +289,17 @@ const onTab = (state, e) => {
...
-        newState = EditorState.push(editorState, newContent, 'insert-characters');
+    } else {
+        const tabOption = editorFormat.includes('tab');
+        const spacesOption = editorFormat.includes('tab as spaces');
+
+        if (tabOption && !e.shiftKey) {
+            newState = insertTab(editorState, selection, '\t');
+            e.preventDefault();
+        } else if (spacesOption && e.shiftKey) {
+            newState = insertTab(editorState, selection, '        ');
```
tomaskikutis, line 300: looks like you can insert anything using that function, not only tabs. How about calling it `insertText` instead?

scripts/core/editor3/reducers/editor3.tsx
```diff
@@ -262,6 +262,12 @@ function escapeRegExp(string) {
     return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
 
+function insertTab(editorState: EditorState, selection: SelectionState, tabCharacter: string) {
+    const newContent = Modifier.replaceText(editorState.getCurrentContent(), selection, tabCharacter);
```
tomaskikutis, line 266: why not `Modifier.insertText`?

pablopunk: ha... That's what I meant to put. I guess it was working anyway cause the selection is collapsed

pablopunk: Actually, now that I think about it, a user could select some text, press TAB, and expect the text to be replaced by the tab character... I have some questions for Karel so I'll add this one to the list

tomaskikutis: good point, `replaceText` might be useful in that case

scripts/core/editor3/reducers/editor3.tsx
```diff
@@ -262,6 +269,46 @@ function escapeRegExp(string) {
...
+        selection,
+        text,
+    );
+
+    return EditorState.push(editorState, newContent, 'insert-characters');
+}
+
+function insertTab(state: EditorState, selection: SelectionState, block: ContentBlock, tabString: string) {
+    const indentInsteadOfInsert = !selection.isCollapsed();
+    const selectionForInsert = indentInsteadOfInsert
```
tomaskikutis, line 288: Why aren't you using the original selection?

scripts/core/editor3/reducers/editor3.tsx
```diff
@@ -262,6 +269,46 @@ function escapeRegExp(string) {
...
+        ? (selection.merge({
+            focusOffset: 0,
+            anchorOffset: 0,
+        }) as SelectionState)
+        : selection;
+
+    let newState = replaceText(state, selectionForInsert, tabString);
+
+    if (indentInsteadOfInsert) {
+        const restoreSelection = selectionForInsert.merge({
```
tomaskikutis, line 298: Isn't it possible to replace text without losing the selection?

scripts/core/editor3/reducers/editor3.tsx
```diff
@@ -270,8 +317,30 @@ function escapeRegExp(string) {
...
+    const selection = editorState.getSelection() as SelectionState;
+    const moreThanOneBlockSelected =
+        selection.getStartKey() !== selection.getEndKey();
+    const block = editorState
+        .getCurrentContent()
+        .getBlockForKey(selection.getStartKey()) as ContentBlock;
+    const blockType = block.getType();
+    let newState = editorState;
+
+    if (['unordered-list-item', 'ordered-list-item'].includes(blockType)) {
```
tomaskikutis, line 330: can you explain what happens in both cases?

pablopunk: Draft-js uses tab & shift+tab to indent/outdent an item on a list

scripts/core/editor3/reducers/editor3.tsx
pablopunk: I do use it, but I change it when the selection is collapsed so instead of inserting a tab there I indent the text. Draft-js doesn't support text indentation yet so I just select the first character in the block, and I insert a tab there.

scripts/core/editor3/reducers/editor3.tsx
pablopunk: It is, I explain it in the comment above

scripts/core/editor3/reducers/editor3.tsx
tomaskikutis: How about inserting the tab at the cursor instead of at the first character and leave it up to the user to jump to the beginning of the line beforehand if that's where he wants the tab. That's how it works in most editors. You may avoid having to modify and restore selection this way too.

pablopunk: We thought that when selecting text and pressing tab we could indent the text, as it is how a lot of editors work. But I guess it will be a very minor usage, the main feature (inserting tabs) works anyway, without restoring the selection

tomaskikutis: Which editors indent the entire line when your selection is expanded? I tested it with libre office, geddit, google docs and vscode. google docs and vscode indent the line only if the entire line is selected. Otherwise, the selection gets replaced with a tab in all these editors.

> the main feature (inserting tabs) works anyway, without restoring the selection

when is restoring selection required then?

pablopunk: I can do it when the entire line then, or even when you have several blocks selected I could indent those...

Restoring the selection is only required when I indent (cause I basically make a new selection at the start, insert the text, and restore). If there's no text selected the selection stays the same

pablopunk: Forget it, I think I'll go back and stick to just inserting tabs, cause this is getting complicated and we'll probably have bugs for something is not really required

tomaskikutis: Yep, I agree about keeping it simple. It's easy to get into undo/redo bugs when forcing a selection.

tomaskikutis:
I've tested it locally, but couldn't find the icons buttons in the editor toolbar, shouldn't we have them? 

Also, when I selected `tabs` in content profile, it was working with `shift` only, but `shift + tab` had to be used if "tabs as spaces" is selected.

I'm not sure what to do with the keyboard shortcut, `shift tab` should jump to the previous field the same way `shift` jumps to the next one. If the feature isn't used frequently, keeping only a button in the toolbar and removing the keyboard binding might be the easiest solution.

We could also use `alt + tab`, but I'm not sure if we can prevent the browser/os from reacting to it even if we stop propagation from js

pablopunk:
@tomaskikutis we discussed if we should have the icons and it's likely that we end up doing them but that's gonna be another task.

Yes, `tab`'s shortcut is `TAB` (of course) but `tab as spaces`'s shortcut is `SHIFT+TAB`. This way you can have both options enabled in the editor and both can work independently (as they have different shortcuts).

And yeah, we shouldn't mess with the `ALT+TAB` shortcut as it's the way you switch open apps on Windows.

PR #3413, opened 2020-03-31
update ui framework to 2.1.16

tomaskikutis:
SDESK-5151
SDESK-5140


tomaskikutis:
@petrjasek looks like macro grouping issue is still there. Do we need to do something to update the server?

petrjasek:
I don't think so, it should be using latest server

PR #3415, opened 2020-03-31
SDBELGA-370 Image upload window is stuck

ride90:
The reason is that we send a wrong payload to the server, more concrete example:
```
{
    "language": "nl",
    "source": "Belga",
    "subject": [
        ...
    ],
    "extra": {
        "city": "Brussels"
    },
    "[object Object]": ""
}
```
it should be `keywords`  instead of `[object Object]`

### DIFF
--- a/scripts/apps/authoring/media/views/media-metadata-editor-directive.html
+++ b/scripts/apps/authoring/media/views/media-metadata-editor-directive.html
@@ -51,7 +51,7 @@
             class="data sd-terms--boxed sd-line-input--dark-ui sd-line-input--boxed"
             data-item="item"
             data-field="keywords"
-            data-change="onChange({key: field})"
+            data-change="onChange({key: field.field})"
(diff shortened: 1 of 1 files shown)


tomaskikutis:
by the way, how come the upload window wasn't getting stuck for me? could it be that this PR fixes one use case, but breaks another?

ride90:
Hi @tomaskikutis , did this PR brake something? 🤕 
To reproduce this bug, you must configure a media upload form to have `keywords` field https://github.com/superdesk/superdesk-belga/blob/master/server/picture-profile.json#L172

In case of Belga it was an issue.
Screenshot 2020-04-01 at 12 53 55

tomaskikutis:
It didn't break anything. I was just wondering why wasn't it always breaking. Since it always breaks when the field is enabled I think it's fine and shouldn't cause issues in other places

PR #3417, opened 2020-04-02
Translate default stages

pablopunk:
SDCP-199

Translate default stage names:

- Working Stage
- Incoming Stage
- Desk Output
- Scheduled Desk Output

### DIFF
--- a/scripts/apps/monitoring/controllers/MonitoringController.ts
+++ b/scripts/apps/monitoring/controllers/MonitoringController.ts
@@ -146,14 +146,8 @@ export function MonitoringController($rootScope, $scope, $location, desks, super
         let groupLabel;
 
         if (group.subheader) {
-            const stageLabelObject = getLabelForStageIfExists(group.subheader);
-
-            if (stageLabelObject != null) {
-                groupLabel = stageLabelObject.label;
-            } else {
-                groupLabel = activeWorkspace === 'workspace' ?
-                    group.header + ' ' + group.subheader : group.subheader;
-            }
+            groupLabel = activeWorkspace === 'workspace' ?
+                group.header + ' ' + group.subheader : group.subheader;
         } else {
             groupLabel = group.header + ' ' + $filter('splitText')(group.type);
         }
--- a/scripts/apps/monitoring/controllers/MonitoringController.ts
+++ b/scripts/apps/monitoring/controllers/MonitoringController.ts
@@ -1,6 +1,5 @@
 import _ from 'lodash';
 import {appConfig} from 'appConfig';
-import {getLabelForStageIfExists} from 'apps/workspace/content/constants';
 
 /**
  * @ngdoc controller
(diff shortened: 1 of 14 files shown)


scripts/apps/workspace/content/constants.ts
```diff
@@ -102,6 +102,23 @@ export const GET_LABEL_MAP = () => ({
     usageterms: gettext('Usage Terms'),
 });
 
+const DEFAULT_STAGES_LABELS = [
```
petrjasek, line 105: I don't think it would work now if defined on top level right @tomaskikutis ? should be fine if you define it inside that `getLabelForStageIfExists` func

scripts/apps/monitoring/controllers/MonitoringController.ts
```diff
@@ -145,8 +146,14 @@ export function MonitoringController($rootScope, $scope, $location, desks, super
         let groupLabel;
 
         if (group.subheader) {
-            groupLabel = activeWorkspace === 'workspace' ?
-                group.header + ' ' + group.subheader : group.subheader;
+            const stageLabelObject = getLabelForStageIfExists(group.subheader);
```
tomaskikutis, line 149: see if you can translate the stage names earlier before they are converted into a `group`. `group` is not always equal to a stage. It can also be a saved search or maybe even a workspace group, not sure about that.

It'd be better to call this function only when we are sure that the object is a stage

scripts/apps/workspace/content/constants.ts
pablopunk: you mean `gettext` wouldn't work? `superdesk.pot` seems like it did extract it

tomaskikutis: yep, you're right. You can't call `gettext` from the top level of the code, because top level gets executed before everything else including the loading of translations. Instead, you can define a function and call it when these labels are needed in the UI. UI is only initialized after the translations have been loaded

pablopunk: ok I'll change that

scripts/apps/monitoring/controllers/MonitoringController.ts
pablopunk: good catch

pablopunk:
> Is `stage.name` static? Can you then use `stage-name` since it's the unspoken convention used in all other tests using `data-test-id` ;D

sure

spec/helpers/desks.ts
```diff
@@ -187,7 +187,7 @@ class Desks {
          **/
         this.getStage = function(name) {
             return this.stages.filter((elem, index) =>
-                elem.element(by.css('[data-test-id="stage.name"]'))
+                elem.element(by.css('[data-test-id="stage-name"]'))
```
tomaskikutis, line 190: you could also use `el(['stage-name'])`

PR #3419, opened 2020-04-03
Initialize translations before starting the app

tomaskikutis:
SDESK-5122

I've added [gettext.js](https://github.com/guillaumepotier/gettext.js/) to be used for translations. It's not coupled to angular so we can load translations before loading the angular app. For angular templates, `angular-gettext` is still used.


scripts/appConfig.ts
```diff
@@ -20,5 +20,21 @@ if (appConfig.ui == null) {
...
+export const IDENTITY_KEY = 'sess:user';
+
+export function getUserInterfaceLanguage() {
+    const user: IUser | null = JSON.parse(localStorage.getItem(IDENTITY_KEY));
+    const language = user?.language ?? appConfig.language ?? window.navigator.language ?? 'en';
+
+    if (appConfig.profileLanguages.includes(language)) {
+        return language;
+    } else {
+        return 'en';
```
petrjasek, line 32: there is also [default_language](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L1070) we might use here

tomaskikutis: > language: string; // default client language

we should probably remove one. Do you know what's the difference between `language` and `default_language`?

petrjasek: `default_language` is set on server, so would make sense to use this one

tomaskikutis: I'll drop the `language` from the interface then since I don't see it used anywhere else but in old `gettext` setup I removed. I didn't find any references in `superdesk.config.js` in customer repos either.

tomaskikutis:
unit tests ok for `85cf95045`

PR #3420, opened 2020-04-06
fix macros test

tomaskikutis:


tomaskikutis:
Petr said he added some macros on the server. Previously, there probably weren't any groups so the list was being displayed ungrouped.

pablopunk:
Yeah I heard it, I asked this before the scrum 👍

petrjasek:
fyi the groups are gone (and will be gone by default)

tomaskikutis:
Why? They seemed helpful when there are >10 macros. Should I close the PR then if we are getting rid of grouping?

petrjasek:
by default all those macros are not enabled, but the groups will be there if needed

tomaskikutis:
so if I closed this PR and restarted the tests on other PRs it should now pass since grouping will no longer be enabled by default?

petrjasek:
that's right

PR #3421, opened 2020-04-06
FIX: Spike/unspike issues of related items. [SDESK-5014]

GyanP:
1. In case of unpublished related items, show `Edit` and `Edit in new window` actions only.
2. In case of published related items, show `Open` and `Open in new window` actions only.


scripts/apps/monitoring/directives/ItemActionsMenu.ts
```diff
@@ -17,6 +17,7 @@ interface IScope extends ng.IScope {
     item?: IArticle;
     open: any;
     active: any;
+    isRelatedItem: boolean;
```
tomaskikutis, line 20: Can you make the directive take an optional `allowedActions: Array<string>` parameter so it's not coupled to any particular feature?

scripts/apps/monitoring/directives/ItemActionsMenu.ts
```diff
@@ -145,7 +143,7 @@ export function ItemActionsMenu(superdesk, activityService, workflowService, arc
                         });
 
                         // actions(except viewing an item) are not allowed for items in legal archive
-                        if (item._type !== 'legal_archive' && !scope.isRelatedItem) {
+                        if (item._type !== 'legal_archive' && !scope.allowedActions) {
```
tomaskikutis, line 146: Avoid using `!x` for non-boolean values. Use `x == null` / `x != null` / `x?.length > 0` etc.

PR #3423, opened 2020-04-06
fix scroll position being lost when repositioning highlights popup box

tomaskikutis:
SDESK-4544

resetting `max-height` was causing scroll position to go to the top in chromium, but not firefox


scripts/core/editor3/components/HighlightsPopupPositioner.tsx
```diff
@@ -5,6 +5,57 @@ interface IProps {
     editorNode: any;
 }
 
+// do calculations on a cloned element to avoid scroll position jumping to the top on chrome
+// when setting resetting `max-height` to an empty value
```
pablopunk, line 9: setting or resetting? 🤔

tomaskikutis: Resetting when it already has a value set previously.

When you resize the window, it needs to recalculate the position. In order to calculate it, you need to determine whether there is more space at the top or bottom and that calculation wouldn't work well if there was `max-height` applied.

pablopunk: I just meant you put both verbs there, what about `setting/resetting`?

tomaskikutis: right, I didn't see it, thanks :man_facepalming:

PR #3424, opened 2020-04-06
SDBELGA-377 Fix terms conflict in sd-meta-terms.

ride90:
This fix is to avoid this:
Screenshot 2020-04-03 at 17 27 42
and this:
Screenshot 2020-04-06 at 17 19 35

When 2 different CVs have the same terms(qcodes) they will conflict with each other. It affects selection and search.

After fix:
Screenshot 2020-04-06 at 17 22 04
Screenshot 2020-04-06 at 17 22 19


scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -752,7 +752,13 @@ function MetaTermsDirective(metadata, $filter, $timeout, preferencesService, des
                 scope.activeList = false;
             };
 
-            scope.isSelected = (term) => !!_.find(scope.item[scope.field], term);
+            scope.isSelected = (term) => {
+                if (scope.cv) {
+                    term.scheme = scope.cv._id;
```
tomaskikutis, line 757: is `term.scheme` a name you came up now or is it used elsewhere? If not, I'd say `term.cvId` would be easier to understand

ride90: it's used in `scope.item[scope.field]`

tomaskikutis: only [here](https://user-images.githubusercontent.com/2076953/78659200-80a1c100-78cb-11ea-8cf6-46265f1a52be.png) or anywhere else too?

ride90: There are more cases of it.
Screenshot 2020-04-07 at 12 46 00

tomaskikutis: got it, let's keep the name then. What is a bit strange is why are you writing data in an `isSelected` method which looks like it should only be reading data to determine if an item is selected.

How about writing this scheme field to all terms when initializing all the terms?

ride90: That makes sense, I check if I can achieve it.

scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -656,6 +656,13 @@ function MetaTermsDirective(metadata, $filter, $timeout, preferencesService, des
                 scope.uniqueField = value || 'qcode';
             });
 
+            // add scheme to terms
+            if (scope.list && scope.cv) {
+                angular.forEach(scope.list, (listItem) => {
```
tomaskikutis, line 661: can you use [Array.prototype.forEach()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)?

ride90: yes, I can, but to keep things consistent I need to change all `angular.forEach` in this file or at least in `MetaTermsDirective`, because `angular.forEach` is used there.

ride90: I've changed it in all possible places in the current directive, except the ones where `undefined` might come, unlike  `Array.prototype.forEach()` the `angular.forEach` will not throw an error in this case.

tomaskikutis: > yes, I can, but to keep things consistent I need to change all `angular.forEach` in this file

I don't think there's a need to update it all at once. It's fine if you did, but I don't see any issues if you left other instances as they were.

I would suggest updating it all if it was some complex piece of business logic, but it's a simple helper function that can be understood and refactored in 5 minutes so we wouldn't gain much by refactoring it all upfront.

scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -808,7 +817,7 @@ function MetaTermsDirective(metadata, $filter, $timeout, preferencesService, des
                 var selected = {};
 
                 angular.forEach(scope.item[scope.field], (term) => {
-                    if (term) {
+                    if (term && scope.cv && scope.cv._id == term.scheme) {
```
tomaskikutis, line 820: Can you use triple equals?

PR #3427, opened 2020-04-10
FIX: Save button in search providers config should activate after changes are made. [SDESK-5074]

GyanP:


tomaskikutis:
can you add `editor3 ctrl+z on tables mantains cursor position in the middle`(fails on travis) to `spec/_unstable-tests.txt`?

PR #3428, opened 2020-04-13
FIX: Max number of items for related items widget is not checked when 'create new' is used. [SDESK-5096]

GyanP:


scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -73,7 +73,18 @@ export function RelatedItemsDirective(
                 return lock.isLocked(item) || lock.isLockedInCurrentSession(item);
             };
 
-            scope.canAddRelatedItems = () => scope.field?.field_options?.allowed_workflows?.in_progress === true;
+            scope.canAddRelatedItems = () => {
+                const currentItemsLength = Object.keys(scope.item.associations || {})
```
tomaskikutis, line 77: can you use `getAssociationsByField`?

PR #3434, opened 2020-04-15
Exclude invalid translations from json

tomaskikutis:
SDCP-187

branched from #3419


tasks/compile-translations-po-to-json.js
```diff
@@ -11,6 +11,56 @@ function isDirectory(path) {
...
+
+function removeInvalidTranslations(grunt, jsonFilePath, filename) {
+    var translations = JSON.parse(fs.readFileSync(jsonFilePath, 'utf8'));
+
+    const validTranslations = Object.keys(translations).filter((key) => {
+        if (key === '') { // metadata object added by gettext.js
+            return true;
+        }
+
+        const placeHolders = key.match(/{{.+?}}/g);
```
petrjasek, line 35: just wondering what is that `?` supposed to do?

tomaskikutis: > a {{b}} c {{d}} e

`?` makes it stop at the first occurrence of the closing brackets(if you have multiple placeholders at the same line). If you use `?`, you'll get 2 matches - `{{b}}` and `{{c}}`. If not, you'd get a single match `{{b}} c {{d}}` which would also include text between the placeholders that shouldn't be included.

petrjasek: I see that's non greedy switch, thx

PR #3435, opened 2020-04-16
Dropdowns should show (multilingual) CV values in the language of the story. [SDESK-5182]

GyanP:


scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -412,7 +413,11 @@ function MetaDropdownDirective($filter) {
                 scope.item[scope.field] = scope.item.default;
             }
 
-            scope.findItemByScheme = (item, scheme) => item.find((o) => o.scheme === scheme);
+            scope.findItemByScheme = (item, scheme) => {
```
tomaskikutis, line 416: is `item` an array? How about calling it `items` then? Can you also add type signatures for both `item` and `scheme`?

PR #3438, opened 2020-04-17
FIX: Extra set of tooltips for right side widget panel. [SDESK-5033]

GyanP:


tomaskikutis:
`sd-hotkey` is the directive that is applying the keybindings. you should remove the tooltip instead

PR #3440, opened 2020-04-17
fix directory paths

tomaskikutis:


webpack.config.js
```diff
@@ -2,6 +2,7 @@ var path = require('path');
 var webpack = require('webpack');
 var lodash = require('lodash');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
+var getModuleDir = require('./tasks/get-module-directory').getModuleDir;
```
pablopunk, line 5: you can use destructuring here `const {getModuleDir} = require...` (and also in the task file)

tasks/compile-translations-po-to-json.js
```diff
@@ -82,7 +83,7 @@ function compileTranslationsPoToJson(grunt) {
             return;
         }
 
-        var po2json = path.join(clientCoreRoot, 'node_modules/gettext.js/bin/po2json');
+        var po2json = `${getModuleDir('gettext.js')}/bin/po2json`;
```
petrjasek, line 86: `npx po2json` should handle it no matter where it is

tomaskikutis:
```
execSync(
            `npx gettext.js/bin/po2json ${poFile} ${jsonFile}`,
            {stdio: 'inherit'}
        );
```

> ENOENT: no such file or directory, stat 'gettext.js/bin/po2json'

webpack.config.js
tomaskikutis: I got tired of reverting stuff due to incompatible node versions so I just stick to es5 for all node scripts now

tasks/compile-translations-po-to-json.js
pablopunk: It should be `npx po2json ...`

webpack.config.js
pablopunk: They work since node 6

tasks/compile-translations-po-to-json.js
tomaskikutis: my bad. It still doesn't work though

```
execSync(
            `npx po2json ${poFile} ${jsonFile}`,
            {stdio: 'inherit'}
        );
```

```
grunt gettext:compile
Running "gettext:compile" task
Warning: Cannot read property 'match' of null Use --force to continue.

Aborted due to warnings.
```

webpack.config.js
tomaskikutis: It might be the case for destructuring, but in general, there are different requirements for different features and I'm too lazy to check it. We probably have 500 LOC of node code, so I don't think it makes a difference either.

PR #3442, opened 2020-04-20
FIX: Proper positioning of the 'required' indication for related items. [SDESK-5098]

GyanP:


tomaskikutis:
* When removing class names, check for usages. If there aren't any, remove the CSS rule too.
* "Required" label [isn't displayed](https://drive.google.com/open?id=1ZTcrMzduX-xpJM2fRPDSfeXFu4HWPj_I) for the first time when you open the item.

tomaskikutis:
What was the reason why "required" label didn't appear unless the item was opened for the second time?

GyanP:
Since the label is not present [here](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/authoring/views/article-edit.html#L698) so over [here](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/ui/ui.ts#L1051), it is unable to find the label and doesn't show the required label.

PR #3444, opened 2020-04-21
Fix moving text blocks in editor3

tomaskikutis:
SDESK-5058

### DIFF
--- a/scripts/core/menu/views/superdesk-view.html
+++ b/scripts/core/menu/views/superdesk-view.html
@@ -14,28 +14,23 @@
 
 <div class="loading" ng-if="loading"></div>
 
-<div id="main-container" ng-class="superdesk.flags" ng-if="!popup && superdesk.flags.workspace === true"
-    ngf-drop="superdesk.openUpload($files)"
-    ngf-model-options="{updateOn: 'drop'}"
-    >
+<div
+    id="main-container"
+    ng-class="superdesk.flags"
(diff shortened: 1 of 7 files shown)


scripts/core/menu/views/superdesk-view.html
```diff
@@ -14,28 +14,23 @@
 
 <div class="loading" ng-if="loading"></div>
 
-<div id="main-container" ng-class="superdesk.flags" ng-if="!popup && superdesk.flags.workspace === true"
-    ngf-drop="superdesk.openUpload($files)"
-    ngf-model-options="{updateOn: 'drop'}"
```
tomaskikutis, line 19: @petrjasek any idea why `ngf-drop` stuff was here in the first place? It looks like it's trying to make the entire app as a drop target which doesn't make sense. I've removed this and uploading still works when adding media items to editor3/editor2, via "+" button in authoring as well as uploading avatars. Any ideas on what removing this might break?

petrjasek: I think this is for uploading pictures via drag and drop

petrjasek: I mean to create new items not adding to existing one

tomaskikutis: I tried it again and I found how it works. I was trying to drop images in the top bar previously so it seemed not to work.

PR #3446, opened 2020-04-23
UI framework version update

fritzSF:


MarkLark86:
@petrjasek The bump in ui-framework version fixes the background colour for Planning form and preview components (https://dev.sourcefabric.org/browse/SDESK-5199), so is required for the Planning 1.33 release

PR #3447, opened 2020-04-23
FIX: Undo does not work after an abbreviation is processed. [SDESK-5067]

GyanP:


tomaskikutis:
Can you explain the issue? Why is it a problem if the last operation was "undo"? What about "redo"?

GyanP:
[this](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/editor3/reducers/editor3.tsx#L182) function sets the abbreviations.
When we try to undo [this](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/editor3/reducers/editor3.tsx#L192 ) will return offset and text and if we are getting offset and text then it will every time sets the abbreviations.
At the time of redo [this](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/editor3/reducers/editor3.tsx#L192) will return null so that's why it worked properly at the time of redo.

tomaskikutis:
If I understand it correctly, the redo case only works by accident - it shouldn't apply abbreviations while redoing, but we have no code preventing it and it only doesn't get applied because it can't retrieve abbreviation text.

Can you add an explicit check for "redo" the same as for "undo"? It would also be good to move the conditions up to line 179. There's some processing to do retrieve the word, so it's better if we can return earlier in case of undo/redo, before doing that processing.

PR #3448, opened 2020-04-27
Add an option for content profile fields to be read only

tomaskikutis:
SDESK-5202

### DIFF
--- a/scripts/apps/authoring/attachments/AttachmentsEditorDirective.ts
+++ b/scripts/apps/authoring/attachments/AttachmentsEditorDirective.ts
@@ -4,7 +4,7 @@ export default function AttachmentsEditorDirective() {
         templateUrl: 'scripts/apps/authoring/attachments/attachments.html',
         link: function(scope, elem, attrs, controllers) {
             if (attrs.readOnly != null) {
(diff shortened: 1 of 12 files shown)


scripts/apps/authoring/attachments/AttachmentsEditorDirective.ts
```diff
@@ -2,5 +2,13 @@
 export default function AttachmentsEditorDirective() {
     return {
         templateUrl: 'scripts/apps/authoring/attachments/attachments.html',
+        link: function(scope, elem, attrs, controllers) {
+            if (attrs.readOnly != null) {
+                scope.$watch(() => scope.$eval(attrs.readOnly), (value) => {
```
petrjasek, line 7: just wondering what's the difference to do `scope.$watch(attrs.readOnly, ...)`

tomaskikutis: I don't know, I'm not too savvy in angular details. I though one can only do `scope.$watch(prop)` when `prop` is a part of the scope. In my case, it's coming as an attribute which I needed to evaluate in order to get a value and after I've done that, I simply used the same in the watch. I'll see if the shortcut you're proposing works.

PR #3449, opened 2020-04-27
fix to desk btn making items disappear

petrjasek:
there is `scope.watch` callback running during
middleware resolving which changes the desk/stage
to different values based on where you sent items
last time.

SDCP-215

### DIFF
--- a/scripts/apps/authoring/authoring/directives/SendItem.tsx
+++ b/scripts/apps/authoring/authoring/directives/SendItem.tsx
@@ -255,7 +255,7 @@ export function SendItem($q,
 
                 // save these here, it might get changed on scope while middlewares run
                 const deskId = scope.selectedDesk._id;
-                const stageId = scope.selectedStage._id;
+                const stageId = scope.selectedStage._id || scope.selectedDesk.incoming_stage;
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/directives/SendItem.tsx
```diff
@@ -419,11 +424,12 @@ export function SendItem($q,
...
              * @return {Object} promise
              */
-            function runSend(open, sendAllPackageItems) {
+            function runSend(open, sendAllPackageItems, _deskId, _stageId) {
                 scope.loading = true;
                 scope.item.sendTo = true;
-                var deskId = scope.selectedDesk._id;
-                var stageId = scope.selectedStage._id || scope.selectedDesk.incoming_stage;
+
+                const deskId = _deskId || scope.selectedDesk._id;
```
tomaskikutis, line 431: there's only one place `runSend` is called and `deskId`/`stageId` are passed as an arguments. Is there a reason to have fallbacks?

I see that `stageId` already had a fallback to `scope.selectedDesk.incoming_stage`, but it's not used in scope.send. How about using that fallback there instead?

PR #3450, opened 2020-04-28
Revert "[SDESK-4861] (1.32) Migrate functionPoints to extensions API (#3323)"

tomaskikutis:
This reverts commit 744e4d495e22ffd22a6c0601ab551ab7c5f1d47c.


tomaskikutis:
will use planning 1.33 instead

PR #3452, opened 2020-04-29
Spiking of unpublished items. [SDESK-5104]

GyanP:


tomaskikutis:
Why doesn't spiking work out of the box for unpublished items? It would be better if we could address the root cause instead of adding exceptions on top. After you unpublish the item it should be, for the most part, the same as before publishing and shouldn't require any code changes for any actions that work on non-published items.

scripts/apps/monitoring/services/CardsService.ts
```diff
@@ -242,6 +242,9 @@ export function CardsService(search, session, desks) {
             criteria.repo = card.search.filter.query.repo;
         } else if (desks.isPublishType(card.type)) {
             criteria.repo = 'archive,published';
+            if (card.type === 'deskOutput') {
+                query.filter({not: {term: {state: 'unpublished'}}});
```
tomaskikutis, line 246: What the purpose of this? To exclude unpublished items from being displayed in the output stage? If so, I'd say it'd be better to exclude it in the backend.

GyanP: Both search and deskOutput call the search api so if we filter the unpublish items from the backend then when we try to search also, it will not show the unpublished items in the search.

PR #3459, opened 2020-05-05
Include find/replace on undo stack

pablopunk:
SDESK-4611

Now ctrl+z and ctrl+y can undo/redo changes made with find/replace

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ spec/**/*.js
 npm-debug.log
 index.html
 yarn-error.log
-yarn.lock
 
 # VIM related files
 *.swo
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -17,7 +17,6 @@ spec/**/*.js
 npm-debug.log
 index.html
 yarn-error.log
-yarn.lock
```
tomaskikutis, line 20: why are you removing `yarn.lock` from .gitignore?

pablopunk: it was [duplicated](https://github.com/superdesk/superdesk-client-core/pull/3459/files/012819b2af7004afc3c51ffcbe461e6d75d50e63#diff-a084b794bc0759e7a6b77810e01874f2R3), probably on a merge commit

PR #3463, opened 2020-05-07
Simplify the advanced search panel. [SDANSA-391]

GyanP:


scripts/apps/search/views/search-parameters.html
```diff
@@ -162,16 +169,46 @@
             </div>
         </div>
 
+        <div ng-if="search_config.date_published"
```
tomaskikutis, line 172: There's a loop rendering multiple fields here, so you can't use `ng-if="search_config.date_published"`, because the code isn't specific to this single field. Instead, you should make sure that `dateFilters` is filtered before the view is rendered.

scripts/apps/search/directives/DateFilters.ts
```diff
@@ -16,6 +17,7 @@ export function getDateRangesByKey(): Dictionary<string, IDateRange> {
         elasticSearchDateRange: {
             lte: 'now+1M/d',
         },
+        active: false,
```
tomaskikutis, line 20: We shouldn't have UI state outside of the UI components.

Instead, you could iterate the results of `getDateRangesByKey()` and add the property later, inside the UI view. You can look up how is it implemented in filters view.

scripts/apps/search/directives/SearchParameters.ts
```diff
@@ -38,17 +40,35 @@ export function SearchParameters($location, asset, tags, metadata, common, desks
...
             scope.keyPressed = function(event) {
                 if (event.keyCode === ENTER) {
                     searchParameters();
                     event.preventDefault();
                 }
             };
 
+            scope.toggleDateFilter = function(predefinedFilters) {
+                predefinedFilters.forEach((filters) => {
+                    if (filters.key === scope.fields.firstpublished) {
```
tomaskikutis, line 56: can you remove hardcoded `firstpublished` field?

scripts/apps/search/directives/SearchParameters.ts
```diff
@@ -38,17 +40,35 @@ export function SearchParameters($location, asset, tags, metadata, common, desks
...
+                        filters.active = !filters.active;
+                    } else {
+                        filters.active = false;
+                    }
+                });
+            };
+
+            scope.togglePredefinedDateFilter = function(fieldname, predefinedFilter) {
+                scope.fields.firstpublished = predefinedFilter;
```
tomaskikutis, line 65: can you remove hardcoded `firstpublished` field?

scripts/apps/search/directives/SearchParameters.ts
```diff
@@ -98,6 +118,10 @@ export function SearchParameters($location, asset, tags, metadata, common, desks
                     scope.fields.spike = 'exclude';
                 }
 
+                if (!$location.search().firstpublished) {
```
tomaskikutis, line 121: can you remove hardcoded `firstpublished` field?

scripts/apps/search/constants.ts
```diff
@@ -15,6 +15,9 @@ export const PARAMETERS = Object.freeze({
     ingest_provider: 'Provider',
     featuremedia: 'Associated Feature Media',
     subscriber: 'Subscriber',
+    firstpublished: 'Date published',
+    firstpublishedfrom: 'Published from',
+    firstpublishedto: 'Published to',
```
petrjasek, line 20: I think we should use gettext for these

scripts/apps/search/directives/SearchParameters.ts
```diff
@@ -98,6 +123,21 @@ export function SearchParameters($location, asset, tags, metadata, common, desks
                     scope.fields.spike = 'exclude';
                 }
 
+                scope.dateFilters.forEach((dateFilter) => {
```
tomaskikutis, line 126: Is the purpose of this block to set active filters from the URL? Can you add a comment before the block?

tomaskikutis:
There are still some issues with [the styles](https://user-images.githubusercontent.com/2076953/82194446-20089980-98f7-11ea-9c7f-8068956fec45.png). If you look at the buttons in the filters section, they stretch to fill 100% of the width even if labels are shorter, which is not the case here. Could you use the same styles as in filters section?

There is also [padding](https://user-images.githubusercontent.com/2076953/82193749-20546500-98f6-11ea-9361-c0340208da58.png) on both sides of the panel which gets broken as soon as date filters are expanded.

tomaskikutis:
It looks ok now. Can you reuse the same CSS instead of copying it? I see it is scoped to `.archive-sidebar`, but that scoping serves no purpose. You can move the class out of that scope and let it be global so it can be used in both places.

PR #3470, opened 2020-05-14
FIX: Publish action for pictures available in Personal desk. [SDANSA-397]

GyanP:


scripts/apps/search/controllers/MultiActionBarController.tsx
```diff
@@ -187,6 +187,15 @@ export function MultiActionBarController(
         return canPackage;
     };
 
+    this.canPublishItem = function() {
+        var canPublish = true;
+
+        multi.getItems().forEach((item) => {
+            canPublish = item.state !== 'draft' && $location.path() !== '/workspace/personal';
+        });
+        return canPublish;
```
tomaskikutis, line 196: Looks like a good use case for [Array.prototype.every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every).

One nice thing about `every` is that it stops iteration as soon as `false` is encountered.

PR #3476, opened 2020-05-19
Display feature media on related items only if exists

pablopunk:
SDESK-5013

### DIFF
--- a/scripts/apps/relations/directives/RelatedItemsDirective.ts
+++ b/scripts/apps/relations/directives/RelatedItemsDirective.ts
@@ -1,9 +1,10 @@
 import {getSuperdeskType} from 'core/utils';
 import {gettext} from 'core/utils';
 import {AuthoringWorkspaceService} from 'apps/authoring/authoring/services/AuthoringWorkspaceService';
-import {IArticle, IArticleField} from 'superdesk-api';
+import {IArticle, IArticleField, IRendition} from 'superdesk-api';
 import {IDirectiveScope} from 'types/Angular/DirectiveScope';
 import {getAssociationsByFieldId} from '../../authoring/authoring/controllers/AssociationController';
+import {getViewImage} from 'core/helpers/item';
 
 const ARCHIVE_TYPES = ['archive', 'published'];
 const isInArchive = (item: IArticle) => item._type != null && ARCHIVE_TYPES.includes(item._type);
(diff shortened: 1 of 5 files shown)


petrjasek:
might be good to get rendition via https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/helpers/item.ts so it's consistent within ui

scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -217,15 +219,23 @@ export function RelatedItemsDirective(
...
             };
-            scope.refreshRelatedItems();
+            scope.loadRelatedItems();
+
+            scope.getThumbnailForItem = (item) => {
+                if (!item.associations?.featuremedia) {
+                    return null;
+                }
+
+                return getViewImage(item.associations?.featuremedia as IArticle);
```
tomaskikutis, line 237: Avoid using type casting via `as` where possible. In this case, you can do a check similar to https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/workspace/content/constants.ts#L131

scripts/core/helpers/item.ts
```diff
@@ -1,3 +1,22 @@
...
+
+    if (associationIsArticle(featuremedia)) {
+        return getViewImage(featuremedia);
+    }
+
+    return null;
+};
+
+export function associationIsArticle(a: any): a is IArticle {
+    return a.renditions != null;
```
tomaskikutis, line 20: It doesn't seem correct since if you have a text article, renditions aren't required and it would produce a false result. I would use `a._created != null`

It would also be better to limit the type of `a` only to those two types `associationIsArticle(a: IArticle | IRelatedArticle)` because if `_created` would also be present in all other entities.

pablopunk: imagen

`_created` is not present in the association. Also, the actual purpose of the function is to use it with associations, so checking renditions being null makes more sense. If we want to check if another thing is an article we won't use `associationIsArticle`. What do u think?

tomaskikutis: > `_created` is not present in the association.

@petrjasek can you check if this is a bug?

Pablo, you could then rely on another field like `_etag`?

> Also, the actual purpose of the function is to use it with associations, so checking renditions being null makes more sense.

If we can easily make a generic method `isArticle(a: IArticle | IRelatedArticle): a is IArticle` which does the type checking and works in all cases, not only in associations, I don't see a reason to opt-out for a more limited solution.

PR #3477, opened 2020-05-20
Ensure that block cursor is displayed on all disabled/read only fields

tomaskikutis:
SDESK-5202

DEPENDS ON: https://github.com/superdesk/superdesk-ui-framework/pull/471


scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
```diff
@@ -48,6 +50,10 @@ interface IScope extends ng.IScope {
     toggleSMS(): void;
 }
 
+export function getReadOnlyLabel() {
+    return gettext('The field is read-only');
```
petrjasek, line 54: I would suggest `This field is read-only`

PR #3482, opened 2020-05-26
hide login form if db auth is disabled

petrjasek:
SDNTB-626

### DIFF
--- a/scripts/core/auth/login-modal.html
+++ b/scripts/core/auth/login-modal.html
@@ -1,3 +1,4 @@
+
 <div class="login-screen" ng-show="active">
     <div class="login-form-container" data-test-id="login-page">
         <p ng-if="identity._id" class="session-error">{{ 'Your session has expired.'|translate }}<br />{{ 'Please log in again.'|translate }}</p>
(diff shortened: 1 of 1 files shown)


scripts/core/auth/login-modal.html
```diff
@@ -9,12 +10,12 @@
         <div class="form-handler" ng-show="!changePassword">
             <figure class="avatar" style="background-image:url('{{ identity.Avatar.href || 'images/avatar_200.png' }}')" ng-show="identity.Avatar"></figure>
             <form name="loginForm" ng-submit="authenticate()">
-                <fieldset class="inputs">
+                <fieldset class="inputs" ng-show="features.auth_db">
```
tomaskikutis, line 13: Doesn't `ng-if` work here?

petrjasek: no it doesn't, it's angular thing, the form it sends then is empty

pablopunk: I think ng-if creates a new child scope for everything inside it, that's why

petrjasek:
will be in 1.33.2 and 1.34 eventually

PR #3484, opened 2020-05-27
Send to Panel changes. [SDANSA-388]

GyanP:


scripts/apps/authoring/authoring/directives/SendItem.tsx
```diff
@@ -92,6 +92,10 @@ export function SendItem($q,
             scope.destination_last = {send_to: null, publish: null, duplicate_to: null};
             scope.origItem = angular.extend({}, scope.item);
             scope.subscribersWithPreviewConfigured = [];
+            scope.sendPublishSchedule = appConfig.ui && appConfig.ui.sendPublishSchedule === false
+                ? appConfig.ui.sendPublishSchedule
```
tomaskikutis, line 96: `appConfig.ui.sendPublishSchedule` can only be false at this point so using a variable is a bit confusing.

What about using a shorter version? `appConfig?.ui?.sendPublishSchedule ?? true`

scripts/apps/authoring/authoring/directives/SendItem.tsx
```diff
@@ -920,6 +924,12 @@ export function SendItem($q,
                     stage = _.find(scope.stages, {_id: scope.selectedDesk.incoming_stage});
                 }
 
+                if (appConfig.ui && appConfig.ui.defaultStage != null) {
+                    stage = scope.stages
```
tomaskikutis, line 928: This would overwrite previously set values. Can you restructure the code so it wouldn't do it?

scripts/apps/authoring/authoring/directives/SendItem.tsx
```diff
@@ -909,7 +911,11 @@ export function SendItem($q,
                     scope.currentUserAction === ctrl.userActions.duplicate_to) {
                     var lastDestination = scope.destination_last[scope.currentUserAction];
 
-                    if (lastDestination) {
+                    if (appConfig.ui && appConfig.ui.defaultStage != null) {
+                        stage = scope.stages
```
tomaskikutis, line 915: Can you search the stage by ID instead of the name? Renaming desks/stages/content profiles shouldn't break functionality.

Why do you need this part `&& stg._id !== scope.item.task.stage)`? And this `|| stage` too - at this point it can only be null. Does it matter if it's null or undefined?

Do you know what is this case is for `} else if (scope.item.task && scope.item.task.stage) {`?

It looks to me that it's reading the stage from the item.task which may need to have priority over `appConfig.ui.defaultStage`. I'm not sure about it, just asking.

GyanP: > Can you search the stage by ID instead of the name?

In the config we can set only name of the stage as the ID of the stage is not pre-known. So what I can do is I can fetch stage details by name and we can use that.

> Why do you need this part && stg._id !== scope.item.task.stage) ?

It's actually filtering out the current stage of the item.

> And this || stage too - at this point it can only be null.

I'll remove it.

> Do you know what is this case is for } else if (scope.item.task && scope.item.task.stage) {?

I think we can remove this condition, In this condition what we are doing is we are getting the current stage of item, but we want to send it to different stage not on the same stage.

tomaskikutis: > In the config we can set only name of the stage as the ID of the stage is not pre-known. So what I can do is I can fetch stage details by name and we can use that.

ID is not known only before the stage is created. After that, it can be looked up and used in the config for the next deployment.

It's not nice though. Can you instead add a section in the settings so it's configurable via UI?

GyanP: After discussion with Petr ana and Gideon this is what I have [got](https://sourcefabric.slack.com/archives/G014GQUN2N7/p1590749453010500)

petrjasek: I don't think we should use stage name, but use that flag working/incoming stage and only allow selecting one or the other

scripts/core/superdesk-api.d.ts
```diff
@@ -1204,6 +1204,9 @@ declare module 'superdesk-api' {
             publishEmbargo?: any;
             sendAndPublish?: any;
             italicAbstract?: any;
+            sendPublishSchedule?: boolean;
+            sendEmbargo?: boolean;
+            defaultStage?: string;
```
petrjasek, line 1209: I would suggest `defaultSendStage?: 'working' | 'incoming';`

scripts/apps/authoring/authoring/directives/SendItem.tsx
tomaskikutis: > After discussion with Petr ana and Gideon this is what I have got

You are probably linking to a private conversation, I can't access it.

> I don't think we should use stage name, but use that flag working/incoming stage and only allow selecting one or the other

At least in the UI, it's possible to add many stages and make them all both, working and incoming, so I don't know how that would solve it. Why not have a UI widget in the settings to pick one of the stages?

scripts/core/superdesk-api.d.ts
```diff
@@ -1204,6 +1204,9 @@ declare module 'superdesk-api' {
             publishEmbargo?: any;
             sendAndPublish?: any;
             italicAbstract?: any;
+            sendPublishSchedule?: boolean;
+            sendEmbargo?: boolean;
+            defaultStage?: 'working' | 'incoming';
```
petrjasek, line 1209: I think it should also have that `send` in the name

scripts/apps/authoring/authoring/directives/SendItem.tsx
```diff
@@ -909,7 +911,9 @@ export function SendItem($q,
                     scope.currentUserAction === ctrl.userActions.duplicate_to) {
                     var lastDestination = scope.destination_last[scope.currentUserAction];
 
-                    if (lastDestination) {
+                    if (appConfig.ui && appConfig.ui.defaultStage != null && appConfig.ui.defaultStage === 'working') {
+                        stage = scope.stages.find((stg) => stg.working_stage);
```
petrjasek, line 915: I think there should be also handling for the setting being incoming so if set it behaves consistently:
```typescript
if (appConfig.ui && appConfig.ui.sendDefaultStage != null) {
    stage = scope.stages.find((stg) => appConfig.ui.defaultStage === 'working' ? stg.working_stage : stg.incoming_stage));
} ...
```

scripts/apps/authoring/authoring/directives/SendItem.tsx
petrjasek: you can have only 1 incoming and 1 working stage on a single desk

scripts/apps/authoring/authoring/directives/SendItem.tsx
```diff
@@ -911,8 +911,11 @@ export function SendItem($q,
                     scope.currentUserAction === ctrl.userActions.duplicate_to) {
                     var lastDestination = scope.destination_last[scope.currentUserAction];
 
-                    if (appConfig.ui && appConfig.ui.defaultStage != null && appConfig.ui.defaultStage === 'working') {
-                        stage = scope.stages.find((stg) => stg.working_stage);
+                    if (appConfig.ui
+                        && appConfig.ui.sendDefaultStage != null && appConfig.ui.sendDefaultStage === 'working') {
```
petrjasek, line 915: you check here still for `working` so it won't work for `incoming`

scripts/apps/authoring/authoring/directives/SendItem.tsx
tomaskikutis: Yep, I tested it again and found that the other stage is modified silently if I set working/incoming to true for another stage.

The UI and data structures are messy though. If I can only have one incoming/working stage per desk, it should be a property of a desk, not a stage.

By the way, what is the relationship between incoming/working stages and setting a default "send to" stage? If there are none, we should probably not be coupling these 2 unrelated features.

petrjasek: while I can imagine getting content to a different stage, so far it was always incoming, and in ansa case it would be working. @karelpetrak any feedback on this? would additional flag make sense?

tomaskikutis: > while I can imagine getting content to a different stage, so far it was always incoming

When you open the "send to" panel, you get to choose from all the stages, not only incoming/working. If we want an option to have one of them selected by default when the panel is opened, any stage is a valid option and I don't see a reason to restrict it. The only downside is that it's hard to put a stage ID into a superdesk.config.js file. We should create a section in the UI for such global configs.

I'm not suggesting a flag on the stage itself by the way, because it not a property of the stage, but stage id is a property of the "send to panel" section. I suggest making a separate section for global configs in the database.

petrjasek: > When you open the "send to" panel, you get to choose from all the stages, not only incoming/working.

yep in theory any stage is valid option, in practice it depends on the workflow and is limited.

> The only downside is that it's hard to put a stage ID into a superdesk.config.js file. We should create a section in the UI for such global configs.

we need to define such a stage for each desk so not sure how to handle that via config outside desk settings

tomaskikutis: > yep in theory any stage is valid option, in practice it depends on the workflow and is limited.

Not only in theory. When "send to" panel opens, you can choose any stage - in practice! :)

> we need to define such a stage for each desk so not sure how to handle that via config outside desk settings

It could work similarly as in [internal destinations](https://user-images.githubusercontent.com/2076953/83740125-b7d7e880-a656-11ea-9f39-d7a150ad00a5.png), except would allow many desk-stage pairs.

karelpetrak: IMO having a "default sent to stage" for each desk is an overkill. The incoming/working stage type is an attribute that defines a main purpose of such a stage (default for fetch/create). 
I'd stick with the ticket requirements to have an option to configure which stage type is offered to users as default. 
We can improve it by remembering the last used stage for every desk which should cover 90% of cases I think. Now we remember only the last used (but not the ones before). So if there is not "last used" stored for a given desk, we'd show the default as per config.

GyanP: > We can improve it by remembering the last used stage for every desk which should cover 90% of cases I think

 I think it's the existing functionality.

PR #3485, opened 2020-05-27
Spellchecker for media medatada

pablopunk:
SDESK-4638

Replaces input fields with draft-js editor with the spellchecker decorator. Spellchecker context menu is disabled ATM as it relies on editor3 redux store


scripts/core/ui/components/PlainTextEditor/PlainTextEditor.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+                    this.setState({
+                        editorState: editorStateWithSelection,
+                    });
+                });
+        }
+    }
+
+    componentWillReceiveProps(newProps: IProps) {
+        if (newProps !== this.props) {
```
tomaskikutis, line 107: Why not check only `value`?

scripts/core/ui/components/PlainTextEditor/PlainTextEditor.tsx
```diff
@@ -0,0 +1,154 @@
...
+    language?: string;
+    placeholder?: string;
+}
+
+interface IState {
+    editorState: EditorState;
+    hasFocus: boolean;
+}
+
+export class PlainTextEditor extends React.Component<IProps, IState> {
```
tomaskikutis, line 33: Does the component work well as a standalone react component not connected to angular?

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -197,10 +197,6 @@ export class Editor3Component extends React.Component<IProps, IState> {
     }
 
     spellcheck() {
```
tomaskikutis, line 199: Why are you changing this? Isn't it better to have the condition in one place vs multiple?

scripts/core/ui/components/index.ts
```diff
@@ -25,6 +25,7 @@ import {Dropdown} from './dropdown';
 import {DropdownMenuDivider} from './dropdown-menu-divider';
 import {NavTabs} from './nav-tabs';
 import {DropZone} from './drop-zone';
+import {PlainTextEditor} from './PlainTextEditor/PlainTextEditor';
```
tomaskikutis, line 28: Why are you not importing it directly?

package.json
```diff
@@ -169,6 +169,7 @@
     "start-client-server": "http-server dist -p 9000 -s &",
     "stop-client-server": "fuser -k 9000/tcp",
     "protractor": "protractor protractor.conf.js",
+    "e2e": "npm run compile-end-to-end-tests && npm run protractor",
```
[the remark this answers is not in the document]
pablopunk: I just added it cause it's the command I use to run e2e tests, I can remove it, I didn't mean to make it for everyone

scripts/core/ui/components/Form/TextArea.tsx
```diff
@@ -91,6 +91,7 @@ export class TextArea extends React.Component<any, any> {
 
         return (
             <textarea
+                style={{minHeight: '32px'}}
```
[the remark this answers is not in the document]
pablopunk: I don't use this anymore as I dropped the use of TextArea in favor of a custom editor, so I'll revert it

scripts/core/ui/components/PlainTextEditor/PlainTextEditor.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+        let newState = EditorState.set(editorState, {allowUndo: false});
+
+        newState = EditorState.push(newState, newContent, 'insert-characters');
+
+        return EditorState.set(newState, {allowUndo: true});
+    }
+
+    runSpellchecker() {
+        if (!this.spellcheckInProgress) {
```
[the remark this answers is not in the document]
pablopunk: hmmm that makes sense, I can create a `cancel` function that's triggered by input change

scripts/core/editor3/components/Editor3Component.tsx
pablopunk: I don't use this spellchecker anymore but I think it's a good practice to do this as it can prevent other conditions that affect performance. For example, [this condition](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/editor3/components/Editor3Component.tsx#L495) can be heavy on performance and it might not even be necessary if spellchecker is disabled.

scripts/core/ui/components/PlainTextEditor/PlainTextEditor.tsx
```diff
@@ -0,0 +1,154 @@
...
+import {getSpellchecker} from 'core/editor3/components/spellchecker/default-spellcheckers';
+import {
+    getSpellcheckWarningsByBlock,
+    getSpellcheckingDecorator,
+} from 'core/editor3/components/spellchecker/SpellcheckerDecorator';
+import {getDraftSelectionForEntireContent} from 'core/editor3/helpers/getDraftSelectionForEntireContent';
+
+export interface IProps {
+    value: string;
+    onChange: (newValue: string, props: IProps) => void;
```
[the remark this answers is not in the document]
pablopunk: oops, forgot about that one. Yes, it is mandatory, when we handle editor changes we usually trigger an `onchange` function in angular, and that function usually requires data like the field you're changing ([example](https://github.com/superdesk/superdesk-client-core/blob/59c87dcb54878baa906d3b5fe6bc9f6cc36382f0/scripts/apps/authoring/media/views/media-metadata-editor-directive.html#L33)). It also makes the editor more flexible as it can return any data you pass to it.

scripts/core/ui/components/PlainTextEditor/PlainTextEditor.tsx
pablopunk: It should, I didn't use it without angular but it doesn't rely on anything angular-related.

scripts/core/ui/components/PlainTextEditor/PlainTextEditor.tsx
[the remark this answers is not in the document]
tomaskikutis: A cancel function would work, but throttling would still be good to have so we don't spam the server with spellchecking requests on every keypress.

scripts/core/ui/components/index.ts
pablopunk: what do you mean?

scripts/core/editor3/components/Editor3Component.tsx
tomaskikutis: I get what you're trying to do now. You could keep the condition in one place and have the same performance if you passed the expensive function as a callback which would only get executed if the spellchecker is enabled.

```
spellcheck(() => prevProps.editorState.getCurrentContent() !== this.props.editorState.getCurrentContent());
```

On the other hand, draft-js is using immutable-js, so I expect that when you call `getCurrentContent` it simply returns an reference to the object and doesn't do any processing, no?

scripts/core/ui/components/index.ts
tomaskikutis: You are importing `PlainTextEditor`, are not doing anything with it, and exporting it. Why not import it directly in the file where you are about to use it?

pablopunk:
I discovered an issue with the cursor jumping back to beginning of text so I'll mark this as a WIP until I fix it

scripts/apps/authoring/tests/authoring.spec.ts
```diff
@@ -124,7 +124,7 @@ describe('authoring', () => {
                 expect($scope.dirty).toBe(false);
                 expect(api.save).toHaveBeenCalled();
             })
-            .finally(done);
+            .finally.catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail).catch(done.fail)(done);
```
tomaskikutis: :rofl: you shouldn't run lint with a "fix" option. We should probably drop it from `packgage.json` too.

scripts/index.ts
```diff
@@ -207,6 +209,6 @@ export function startApp(
 // this is a fallback for e2e tests.
 setTimeout(() => {
     if (started !== true) {
-        startApp([], {});
+        startApp([markForUserExtension], {});
```
tomaskikutis, line 212: Extensions should be registered from a client repo.

scripts/apps/authoring/tests/authoring.spec.ts
pablopunk: I know, this was just my editor using the linter to format the file on save... I had to disable it every time I opened that file 😅

PR #3486, opened 2020-05-29
filter out published items from personal space

petrjasek:
SDANSA-346


tomaskikutis:
If we got hit by a bug and got our data into an invalid state, not allowed by our business logic, shouldn't we fix the data, instead of changing the business logic to work with invalid data?

petrjasek:
you're right, it's not so straightforward as users didn't provide the data, but we can guess or ask

PR #3489, opened 2020-06-03
fix(build): Update ui-framework to 2.2.2

MarkLark86:


MarkLark86:
`datetimeField` extension is not building, therefor Superdesk/Planning/Analytics PR checks won't pass

PR #3495, opened 2020-06-08
FIX: Feature media field breaks after an upload. [SDESK-5315]

GyanP:


tomaskikutis:
Why was this a problem? If it's clickable it should probably be a button.

PR #3499, opened 2020-06-12
[SDESK-4694] Improve validation feedback when editing media contacts.

marwoodandrew:


scripts/apps/contacts/components/Form/ProfileDetail.tsx
```diff
@@ -193,6 +193,9 @@ export class ProfileDetail extends React.PureComponent<IProps, IState> {
         const isRequired = get(this.state, 'requiredField', false);
         const MSG_REQUIRED = gettext('This field is required.');
         const isAssignable = this.state.contactType && this.state.contactType.assignable;
+        const minFieldMessage = gettext('At least one of [email, phone, mobile, twitter, ' +
```
tomaskikutis, line 196: Can you rephrase the error message so it doesn't include any fields or if you want to include field names, can you get them from the function doing the validation instead of hardcoding?

scripts/apps/contacts/helpers.ts
```diff
@@ -85,19 +86,29 @@ function contactHasEmailAddress(contact: IContact): boolean {
...
 }
 
+export function getMinRequiredFieldLabels() {
+    return LOOKUP_FIELDS.map((f) => gettext(f.label)).join(', ');
+}
+
+export function getMinRequiredFieldLabel(field) {
+    let thisField = LOOKUP_FIELDS.find((f) => (_.get(f, 'name') === field));
+
+    return _.get(thisField, 'label');
```
tomaskikutis, line 109: You can use (Nullish coalescing operator)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator] instead of lodash functions.

Why do you use "label" as a default instead of logging an error if it can't be found?

marwoodandrew: Sadly I'm not working in the develop branch as I would need to build a new environment. I don't think Nullish coalescing operator is available to me back in 1.31.

"label" is not a default it is the path in the object, I've added an empty string default. I think logging an error is reasonable.

PR #3501, opened 2020-06-12
Content profiles for non text content types

tomaskikutis:
SDESK-5216

DEPENDS ON: SDESK-5217


scripts/apps/workspace/content/components/ContentProfileFieldsConfig.tsx
```diff
@@ -0,0 +1,634 @@
...
+                    if (
+                        querySelectorParent(e.target, 'select', {self: true}) == null
+                        && querySelectorParent(e.target, 'button', {self: true}) == null
+                    ) {
+                        page.startEditing(item._id);
+                    }
+                }}
+            >
+                {
+                    !sortingInProgress
```
pablopunk, line 167: No need for a ternary operator here: `sortingInProgress || (<div />)`

scripts/apps/workspace/content/components/ContentProfileFieldsConfig.tsx
```diff
@@ -0,0 +1,634 @@
...
+                                display: 'flex',
+                                justifyContent: 'center',
+                                width: '100%',
+                                position: 'absolute',
+                                bottom: '-17px',
+                            }}>
+                                {newFieldSelect(index + 1)}
+                            </div>
+                        )
+                        : null
```
pablopunk, line 202: ok I see you use the ternary operator a lot, is there any reason for that? do you think it's more readable or that it's easier to add things in the future?

scripts/apps/workspace/content/components/ContentProfileFieldsConfig.tsx
```diff
@@ -0,0 +1,634 @@
...
+                index={this.props.index}
+            />
+        );
+    }
+}
+
+class ItemsContainerBase extends React.PureComponent {
+    render() {
+        return (
+            <div style={{background: 'beige', marginLeft: -20, marginRight: -20}}>
```
pablopunk, line 225: why negative margin? shouldn't we fix the parent or something to avoid this?

scripts/apps/workspace/content/components/ContentProfileFieldsConfig.tsx
pablopunk: True, but ternary operators use the same boolean logic so they have that issue too. I think they are a bit more readable than ternary if you don't use both cases of the ternary operator. Not a big deal tho

pablopunk:
What's needed on this PR? A lot of tests are failing so I'm guessing there's still some work to do here

tomaskikutis:
I'm waiting for Nemanja to style it.

pablopunk:
Ah that's right, I remember now

pavlovicnemanja:
hey @tomaskikutis everything is fixed now as requested. Can you please check it and merge if it is fine?

tomaskikutis (commented):
I found a few issues with dropdowns:
* Some options are displayed [outside of viewport ](https://user-images.githubusercontent.com/2076953/129022179-cf82adda-5e27-4688-a1cf-fdf53441cbad.png) and it's not possible to select them.
* [In some cases](https://user-images.githubusercontent.com/2076953/129022666-30de1911-afb0-4951-a21b-83f48dc691c5.gif)(<-- always reproduces) the dropdown opens below the modal backdrop and no options can be selected.
* When modal is closed, dropdown options [remain on the screen](https://user-images.githubusercontent.com/2076953/129022918-ccd0cd13-6cb4-46e8-bb8d-73a432f193af.gif). Dropdowns should close if an element outside of dropdown is clicked.

> When there is already one image content profile, and new content profile is being created, image type should be disabled and not available for selection. It is not disabled (at least it's not apparent visually).

This point seems not to have been addressed. I am able to create multiple `image` content profiles.

> Description field can be resized beyond 100%.

This is fixed, but now textarea is visually identical to single line text input. It'd be good to make it higher so it's possible to differentiate them

styles/sass/sf-additional.scss
```diff
@@ -2230,3 +2236,8 @@ sd-multi-image-edit {
 .sd-line-input textarea.sd-line-input__input {
     display: block !important;
 }
+
+// Transition for draggable lists
+.modal__body .sd-list-item {
```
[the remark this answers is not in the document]
pavlovicnemanja: This is case only for modals, because lists embeded in modal, need to have extra z-index.

pavlovicnemanja:
`When there is already one image content profile, and new content profile is being created, image type should be disabled and not available for selection. It is not disabled (at least it's not apparent visually).` I wasn't able to reproduce this, can you do some screen recording to show where this is problem?

styles/sass/sf-additional.scss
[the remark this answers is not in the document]
tomaskikutis: What I meant is to make it more specific, so it applies not to _all lists in modals_, but only to sortable lists(no matter if they are in the modal or not).

The second point is more important though, to not apply a second overwrite and use the transform that is coming from the library.

tomaskikutis:
> Textareas got auto height, so if you type in two lines, it will get heigher.

The problem is that it's not possible to visually see which component it is, without trying it out. If default height was 2 lines, it would be obvious then. It's not an important issue, we can leave it if it's not easy to fix.

> I wasn't able to reproduce this, can you do some screen recording to show where this is problem?

Upon checking again I can't either. Maybe it was my messy database.

tomaskikutis:
> I wasn't able to reproduce this, can you do some screen recording to show where this is problem?

I found a similar issue though, that image option is disabled, but is shown as [enabled in the UI](https://user-images.githubusercontent.com/2076953/129208488-eb1bc105-c277-4508-84f6-737f54f8c415.gif) until some other option is clicked.

petrjasek:
lets not merge this now as we should only make this part of 2.5

PR #3505, opened 2020-06-15
Allow translations of timezone names

pablopunk:
SDESK-5309


scripts/apps/dashboard/world-clock/timezones-all-labels.ts
```diff
@@ -0,0 +1,95 @@
+import {gettext} from 'core/utils';
+
+export const labels = {
```
petrjasek, line 3: this won't fly as it's on top level, we would need a callback so it can load those translations meanwhile

scripts/apps/dashboard/world-clock/configuration.html
```diff
@@ -19,7 +19,7 @@
                 <div class="grid__item grid__item--col-12">
                     <ul class="list-group">
                         <li class="list-group-item" ng-repeat="zone in configuration.zones">
-                            <a href="">{{ zone.replace('_', ' ')}}</a>
+                            <a href="">{{ zoneLabels[zone].replace('_', ' ')}}</a>
```
tomaskikutis, line 22: Why are you replacing the underscore? It shouldn't be in the translations strings in the first place, should it?

scripts/apps/dashboard/world-clock/configuration.html
```diff
@@ -32,17 +32,23 @@
...
+                            placeholder="{{ 'Search for clock'|translate }}"
+                        >
+                        <button class="btn" ng-disabled="!search" ng-click="search = ''; searchZones('')" translate>Clear</button>
                     </div>
 
                     <ul class="list-group">
-                        <li class="list-group-item" ng-repeat="zone in availableZones|filter:notIn(configuration.zones)|filter:search" ng-click="configuration.zones.push(zone); notify('add', zone)">
-                            <a href="">{{ zone.replace('_', ' ')}}</a>
+                        <li class="list-group-item" ng-repeat="zone in availableZones|filter:notIn(configuration.zones)" ng-click="configuration.zones.push(zone); notify('add', zone)">
+                            <a href="">{{ zoneLabels[zone].replace('_', ' ')}}</a>
```
tomaskikutis, line 47: underscore

scripts/apps/dashboard/world-clock/timezones-all-labels.ts
tomaskikutis: yep, you need to make it a function like `getLabels = () => ({'Africa/Abidjan': gettext('Africa/Abidjan')})` which would only be called after the app has initialized.

If you check the console, you should see many errors that you can't call `gettext` before the translations are loaded.

scripts/core/datetime/datetime.ts
```diff
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import {gettext} from 'core/utils';
 import moment from 'moment-timezone';
 import {appConfig} from 'appConfig';
+import {labels as timezoneLabels} from 'apps/dashboard/world-clock/timezones-all-labels';
```
tomaskikutis, line 5: why not export it as `timezoneLabels`? It's easier to find usages when the name is the same as in the exporting file.

scripts/apps/dashboard/world-clock/timezones-all-labels.ts
```diff
@@ -1,6 +1,16 @@
...
 
-export const getTimezoneLabels = () => ({
+export const getTimezoneLabel = (tz) => {
+    const labels = getTimezoneLabels();
+
+    if (labels.hasOwnProperty(tz)) {
+        return labels[tz];
+    }
+
+    return tz;
```
tomaskikutis, line 10: It would be good to log a warning to sentry if it isn't found. Why wouldn't it be found? Is the list incomplete or what?

pablopunk: This way is more useful for example to make tests with non-existing timezones and not having to mockup all the labels

tomaskikutis: Why not import real labels and use them in tests?

pablopunk: I meant when you create dummy names like [here](https://github.com/superdesk/superdesk-client-core/blob/d881ecd0bd1186c5ee6d6f34ae965ef4fbe4698e/scripts/apps/dashboard/world-clock/tests/tzdata.spec.ts#L66)

tomaskikutis: Can't real values be used instead of those dummy names?

pablopunk: But is there any issues with the current implementation?

tomaskikutis: Code meant only for testing **inside implementation** is the issue.

We can restructure code by splitting blocks into functions so it's easier to test, but we shouldn't change the logic for the sake of testing.

Your version doesn't look as bad as the snippet below, but it is the same in principle.

```
export const getTimezoneLabel = (tz) => {
    if (isBeingTested) {
        return 'test label';
    }

    return getTimezoneLabels()[tz];
}
```

In your case, you are adding a conditional branch too, which is only meant for testing, only that it's not obvious that it's only for testing.

The best solution is simply passing valid data, so no exceptions are required as to how it would run in production.

If that's not possible or hard, you can mock the entire function:

```
// implementation code

const getLabel = (tz) => getTimezoneLabels()[tz];

export const timezone = {
    getLabel: getLabel,
};


// test code

timezone.getLabel = () => 'test label';

// ...
```

pablopunk: But why is my first implementation `getTimezoneLabels()[tz]` better than the current one `getTimezoneLabel(tz)`?

Also this one is more flexible because using a function could allow passing arguments like `(noWhitespace = true)` or something like that (not that we need it, but I don't see any drawback yet). The fact that is easier to write tests for it is just a side effect, not the only purpose of it.

tomaskikutis: I'm not suggesting to not use a function - only that the function shouldn't contain test-specific logic. In my example bellow `// implementation code` comment, I'm using a function too.

> The fact that is easier to write tests for it is just a side effect, not the only purpose of it.

When `labels.hasOwnProperty(tz)` evaluates to `false` except for tests? If the answer is never, then it means that piece of code is there only for testing.

pablopunk: Right now it only triggers for tests but I can totally imagine a feature where users create their own Timezones to display

tomaskikutis: Code should represent how an application works currently, not potentially. If we regularly add logic "just in case we need it someday", we'd not be able to differentiate between production code and other pieces that never run and are there only "for future use".

It'd make it harder to understand how the application works and when refactoring we would have to spend time maintaining that future code.

This is not a big issue, but a code smell that doesn't seem hard to fix. Can you do it?

pablopunk: fixed

tomaskikutis: Great, thanks.

pablopunk:
@petrjasek thanks, I think it's done now

PR #3506, opened 2020-06-16
Fix autosave getting called after a correction is published

tomaskikutis:
SDESK-5328


petrjasek:
I'm not sure if the onChangeMiddleware should be part of autosave, imo it should run on change even if the save is triggered before the autosave timeout

tomaskikutis:
Makes sense, I'll use it in both places then.

PR #3508, opened 2020-06-17
FIX: problem with locking when creating a related item from within parenting authoring item. [SDESK-5312]

GyanP:


scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -67,9 +71,21 @@ export function RelatedItemsDirective(
             scope.onCreated = (items: Array<IArticle>) => {
                 items.forEach((item) => {
                     scope.addRelatedItem(item);
+                    storage.setItem(item._id, scope.item._id);
                 });
             };
 
+            scope.$on('item:close', (evt, mainArticleId) => {
+                let parentItem = storage.getItem(mainArticleId) ? {_id: storage.getItem(mainArticleId)} : null;
```
tomaskikutis, line 79: It'd be good not to call `getItem` twice and save a result from the first call into a variable.

It'd also avoid calling it `parentItem`, since it's only an ID. I'd write it like this:

```
scope.$on('item:close', (evt, mainArticleId) => {
    const itemId = storage.getItem(mainArticleId);

    if (itemId != null) {
        autosave.get({_id: itemId}).then((result) => {
...
```

scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -67,9 +71,21 @@ export function RelatedItemsDirective(
             scope.onCreated = (items: Array<IArticle>) => {
                 items.forEach((item) => {
                     scope.addRelatedItem(item);
+                    storage.setItem(item._id, scope.item._id);
```
tomaskikutis, line 74: It's not good to use item ID as a storage key because it may be used for different purposes from other places in the app. It's not readable too if I saw it in the browser, I wouldn't be able to tell where is it getting set/used in the code.

It's better to use something like `open-item-after-related-closed--${id}`

PR #3513, opened 2020-06-21
feat(ui-framework): Updated version of ui framework

pavlovicnemanja:


tomaskikutis:
Don't forget to update it for extensions. For now, only datetime extension is using it. You can search for `"superdesk-ui-framework":` to find the places the version needs to be updated.

PR #3515, opened 2020-06-22
fix unstable tests - closes preview when editing is opened

tomaskikutis:
SDESK-5212

### DIFF
--- a/scripts/apps/archive/views/preview.html
+++ b/scripts/apps/archive/views/preview.html
@@ -73,8 +73,11 @@
         <div class="composite-nav" ng-if="selected.preview.type === 'composite'">
           <i class="filetype-icon-large-composite"></i>
           <p class="title preview-headline"
-             ng-if="selected.preview.headline"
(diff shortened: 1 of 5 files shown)


scripts/apps/archive/views/preview.html
```diff
@@ -86,9 +89,12 @@
         <div class="rewrite" ng-if="selected.preview.rewrite_id">{{ :: 'This story has been rewritten by: ' | translate }}<a ng-click="previewRewriteStory()"
             class="open-item"><i class="icon-external"></i></a></div>
 
-        <p class="title preview-headline"
-           ng-if="selected.preview.headline && selected.preview.type !== 'composite'"
-           ng-class="{condensed: !lock}" sd-html-preview="selected.preview.headline"></p>
+        <p class="title z preview-headline"
```
pablopunk, line 92: why `z`?

tomaskikutis: :upside_down_face:  Forgot to get rid of that. There were 2 elements with the same class, I was testing which is getting rendered

PR #3521, opened 2020-06-24
avoid error reporing on travis

petrjasek:
it's spamming sentry


petrjasek:
it's not error, but it's using error log level to get some attention

tomaskikutis:
What about filtering all "warn" level logs when running tests instead of targeting this one specific error?

petrjasek:
this one in particular is not a warning, but an error. and it's not always an issue, that depends on a client, that's why we have that config.

tomaskikutis:
> this one in particular is not a warning, but an error
> it's not error, but it's using error log level to get some attention

It might be a good idea to use another log level then to separate between not important logs like deprecation info VS important warnings related to business logic. Adding a config for an individual error is not a clean solution.

petrjasek:
might not need this after all

PR #3524, opened 2020-06-26
fix custom fields check when doing upload media validation

petrjasek:
SDANSA-420

### DIFF
--- a/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
+++ b/scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
@@ -13,23 +13,49 @@ import {appConfig} from 'appConfig';
  * @requires lodash
  * @requires api
  * @requires $rootScope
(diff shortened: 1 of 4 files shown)


scripts/apps/authoring/authoring/controllers/ChangeImageController.ts
```diff
@@ -13,23 +13,49 @@ import {appConfig} from 'appConfig';
...
-        const value = metadata[key];
+export function validateMediaFieldsThrows(validator, metadata, schema) {
+    const ERROR = gettext('Required field(s) missing');
+
+    Object.keys(validator).forEach((key) => {
+        if (!validator[key].required) {
+            return;
+        }
+
+        const specs = schema[key] || null;
```
tomaskikutis, line 29: What does `specs` stand for? Maybe `fieldOptions` or `schemaItem` would be a better name?

PR #3527, opened 2020-07-06
Add Save event for superdesk api

idsvn:
This event is similar to click Save button on the UI by calling `save` function in `AuthoringEditDirective` (as requested in https://github.com/superdesk/superdesk-belga/pull/228)

- Move `saveAuthoring` to separate function, otherwise, calling `save` will cause infinite loop because: `runBeforeUpdateMiddlware` => Run Extension => Trigger Save => Run `runBeforeUpdateMiddlware`
- If multiple extensions trigger `save` event, it will run `save` only once (`saveAuthoring` already do the diff comparison)

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
@@ -7,7 +7,6 @@ import {isPublished} from 'apps/archive/utils';
 import {resetFieldMetadata} from 'core/editor3/helpers/fieldsMeta';
 import {appConfig} from 'appConfig';
 import {gettext} from 'core/utils';
-import {addInternalEventListener} from 'core/internal-events';
 
 interface IScope extends ng.IScope {
     readOnlyLabel: string;
@@ -88,8 +87,6 @@ ArticleEditDirective.$inject = [
     '$interpolate',
     'suggest',
     'renditions',
-    'authoring',
-    'notify',
 ];
 export function ArticleEditDirective(
     autosave,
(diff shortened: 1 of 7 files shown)


idsvn:
@tomaskikutis It does not work. Not sure if there is problem in our autosave Coming Up implementation but it still causes infinite loop, we have checked and the [diff](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/authoring/authoring/services/AuthoringService.ts#L489) is indeed bigger than 0 so the next value should be different from the current one.
Strange thing is that if we use log point inside `api.save` callback, it does nothing.

tomaskikutis:
@idsvn The problem in your case was that you were calling `superdesk.ui.article.save();` before the updates were applied, so there were no changes to save. I have added `onUpdateAfter` to extensions API, try using that.

idsvn:
@tomaskikutis It's working now, thank you!
Should we keep `onUpdate` to avoid breaking changes? We could update `updateArticleOnProfileChange` to `onUpdateBefore` in `belga` repo, but not sure if there is any other extension use `onUpdate` hook?

tomaskikutis:
I have renamed `onUpdate` to `onUpdateBefore`. I'm not sure if it's used anywhere, but I expect it will be useful in the future, so I'm not removing it. The rename won't be an issue, because in case it was used somewhere, it will trigger a compile error and can be fixed in 5 minutes by using the new name.

idsvn:
OK, Only `updateArticleOnProfileChange` extension uses them in `Belga` repo so we will rename it in our PR.

PR #3529, opened 2020-07-13
fix user editing not working if username doesn't match

petrjasek:
even if it's readonly

SDCP-235

### DIFF
--- a/scripts/apps/users/directives/UserEditDirective.ts
+++ b/scripts/apps/users/directives/UserEditDirective.ts
@@ -35,7 +35,8 @@ export function UserEditDirective(api, notify, usersService, userList, session, 
             });
             scope.privileges = privileges.privileges;
             scope.features = features;
-            scope.usernamePattern = usersService.usernamePattern;
+            scope.usernamePattern = appConfig.user.username_pattern != null ?
+                new RegExp(appConfig.user.username_pattern) : usersService.usernamePattern;
             scope.twitterPattern = usersService.twitterPattern;
             scope.phonePattern = usersService.phonePattern;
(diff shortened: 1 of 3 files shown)


tomaskikutis:
How did invalid usernames get into the database? Why not fix the data instead of disabling validation?

petrjasek:
it's external data from azure

tomaskikutis:
In general, we should make data from external integrations to conform to Superdesk restrictions, but I guess it's an arbitrary rule which doesn't really matter except for usernames being easier to read without special chars. Or does it have a functional purpose?

If we want to turn off the username validation when azure integration is enabled, we should be doing it explicitly. It's easy to do by moving a pattern to server config and setting an empty string only if azure integration is enabled.

It'd be even better to move validation entirely to the server and only display server returned error messages on the client.

petrjasek:
in this very case that field is readonly, but due to that validation we have set when you are creating users it's blocking other updates, so here the fix is not about the pattern really, but about not doing validation on readonly field, which actually could be just text in this case and not an input

tomaskikutis:
What I don't like the most in the current solution is the disconnect between the code and the intention.

Not only it's not obvious that there is any connection between that field and azure integration, but there is also no code trail to follow where that connection would be discovered.

It would be the case if the pattern was taken from config and config would depend on azure integration explicitly.

Imagine you needed to make the field editable. The current solution would no longer work while the one based on the config would. It's not that we'll need it, but it shows that it's a more stable solution if it works in more situations.

petrjasek:
will check that failing e2e later, need to fix uat asap

PR #3530, opened 2020-07-13
FIX: Ninjs formatter - DateCreated field is not populated with data. [SDANSA-423]

GyanP:
Uploading multiple images will have different values for date created and we don't want to replace those values with ' ' empty strings.

### DIFF
--- a/scripts/apps/search/MultiImageEdit.ts
+++ b/scripts/apps/search/MultiImageEdit.ts
@@ -120,7 +120,15 @@ export function MultiImageEditController(
 
         $scope.images.forEach((item) => {
             if (item.selected) {
-                item[field] = $scope.metadata[field] || '';
+                if (field === 'extra') {
+                    Object.keys(item[field]).forEach((key) => {
+                        if (key !== 'DateCreated') {
+                            item[field][key] = $scope.metadata[field][key];
+                        }
+                    });
+                } else {
+                    item[field] = $scope.metadata[field] || '';
+                }
             }
         });
     };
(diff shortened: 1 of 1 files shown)


tomaskikutis:
Can this be solved without hardcoding field names? If I understand the ticket correctly, they don't like that `DateCreated` is present, but is an empty string. What about not taking any fields from image metadata if they are empty?

GyanP:
> What about not taking any fields from image metadata if they are empty?

Ya, I think you are right we should remove all the fields which are empty.

scripts/apps/search/MultiImageEdit.ts
```diff
@@ -120,7 +120,19 @@ export function MultiImageEditController(
 
         $scope.images.forEach((item) => {
             if (item.selected) {
-                item[field] = $scope.metadata[field] || '';
+                if (field === 'extra') {
+                    Object.keys(item[field]).forEach((key) => {
+                        if ($scope.metadata[field][key]) {
```
tomaskikutis, line 125: If we are writing this code to handle empty strings, could you make the condition `if (typeof x === 'string' && x.trim().length < 1)` so the purpose of the code is clear in the future? Trimming would also be useful in case the string only has whitespace.

scripts/apps/search/MultiImageEdit.ts
```diff
@@ -120,7 +120,19 @@ export function MultiImageEditController(
...
+                            item[field][key] = $scope.metadata[field][key];
+                        } else {
+                            delete item[field][key];
```
tomaskikutis, line 128: Instead of deleting in here, could it instead be prevented from being added in the first place?

PR #3538, opened 2020-07-20
[INGEST] [SETTINGS] implemented new `url_rquest` field type

jerome-poisson:
SDESK-5342

### DIFF
--- a/scripts/apps/ingest/directives/IngestSourcesContent.ts
+++ b/scripts/apps/ingest/directives/IngestSourcesContent.ts
@@ -3,6 +3,45 @@ import {cloneDeep} from 'lodash';
 import {gettext} from 'core/utils';
 import {appConfig} from 'appConfig';
 
+interface IFeedingServiceField {
+    id?: string;
+    type: string;
+    label: string;
+    readonly?: boolean;
+    placeholder?: string;
+    default_value?: string;
+    errors?: { [key: number]: any };
+    url?: string;
+}
+
+interface IProvider {
+    name: string;
+    source: string;
+    feeding_service: string;
+    feed_parser?: string;
+    content_types?: Array<any>;
+    allow_remove_ingested?: boolean;
+    content_expiry?: number;
+    config?: {};
+    private?: {};
+    ingested_count?: number;
+    accepted_count?: number;
+    token?: {};
+    is_closed?: boolean;
+    update_schedule?: {};
+    idle_time?: {};
+    last_updated?: string;
+    last_ingested_id?: string;
+    last_item_update?: string;
+    rule_set?: any;
+    routing_scheme?: any;
+    notifications?: {};
+    last_closed?: {};
+    last_opened?: {};
+    critical_errors?: {};
+    skip_config_test?: boolean;
+}
+
 IngestSourcesContent.$inject = ['ingestSources', 'notify', 'api', '$location',
     'modal', '$filter', 'privileges'];
 
@@ -578,6 +617,17 @@ export function IngestSourcesContent(ingestSources, notify, api, $location,
                     var feedingService: any = getCurrentService();
 
                     return feedingService ? feedingService.templateUrl : '';
+                };
+
+                /**
+                 * Do URL request specified in url_request field
+                 * @param provider ingest provider metadata
+                 * @param field url_request field metadata
+                 */
+                $scope.doUrlRequest = (provider: IProvider, field: IFeedingServiceField): void => {
+                    let provider_name = encodeURIComponent(provider.name);
+
+                    window.open(field.url.replace('{PROVIDER_NAME}', provider_name));
                 };
 
                 function getCurrentService() {
--- a/scripts/apps/ingest/views/settings/service-config.html
+++ b/scripts/apps/ingest/views/settings/service-config.html
@@ -22,6 +22,16 @@
         <div sd-ingest-config-errors></div>
     </div>
 
+    <div class="form__row" ng-if="field.type === 'url_request'" ng-show="isConfigFieldVisible(field)">
+        <button
+            type="button"
+            class="btn btn--primary"
+            ng-click="doUrlRequest(provider, field)">
+            {{field.label}}
+        </button>
+    </div>
+
+
     <div class="sd-line-input sd-line-input--is-select" ng-if="field.type === 'choices'" ng-show="isConfigFieldVisible(field)">
         <label class="sd-line-input__label" for="{{getConfigFieldId(field.id)}}">{{field.label | translate}}</label>
         <select id="{{getConfigFieldId(field.id)}}"


scripts/apps/ingest/views/settings/service-config.html
```diff
@@ -22,6 +22,16 @@
         <div sd-ingest-config-errors></div>
     </div>
 
+    <div class="form__row" ng-if="field.type === 'url_request'" ng-show="isConfigFieldVisible(field)">
+        <button
+            type="button"
+            class="btn btn--primary"
+            ng-click="doUrlRequest(field)">
+            {{field.label | translate}}
```
tomaskikutis, line 30: `| translate` should only be applied to static strings. Otherwise, the string can't be statically extracted for translation. You should remove the `translate` filter here, and translate the static string before it's assigned to a variable.

scripts/apps/ingest/directives/IngestSourcesContent.ts
```diff
@@ -580,6 +581,14 @@ export function IngestSourcesContent(ingestSources, notify, api, $location,
                     return feedingService ? feedingService.templateUrl : '';
                 };
 
+                /**
+                 * Do URL request specified in url_request field
+                 * @param {Object} field
+                 */
+                $scope.doUrlRequest = (field) => {
+                    $window.open(field.url);
```
tomaskikutis, line 589: why not use the native `window`? Is the angular version doing anything special?

scripts/apps/ingest/directives/IngestSourcesContent.ts
```diff
@@ -580,6 +581,14 @@ export function IngestSourcesContent(ingestSources, notify, api, $location,
                     return feedingService ? feedingService.templateUrl : '';
                 };
 
+                /**
+                 * Do URL request specified in url_request field
+                 * @param {Object} field
```
tomaskikutis, line 586: Use TypeScript to specify the types of parameters. It would also be good to specify an exact type, not only that it's an object.

scripts/apps/ingest/views/settings/service-config.html
jerome-poisson: Hey @tomaskikutis I was surprised too, but it's used like this all over the place in this file. If the word is translated it should work, the only thing which can't work is the marking.

scripts/apps/ingest/directives/IngestSourcesContent.ts
jerome-poisson: I guess it should work too, I've just checked how people are usually doing with Angular.

scripts/apps/ingest/directives/IngestSourcesContent.ts
jerome-poisson: alright, I was just doing the same way as for other functions in this file, but it's definitely better to use typescript, I'll update it.

scripts/apps/ingest/views/settings/service-config.html
tomaskikutis: Yep, it's used everywhere in this file, but it shouldn't be. Since you are editing the file already, it would be good if you could fix all instances.

It's an unstable solution to have to rely on other strings to be translated in order for this to work. If this label got renamed, for example, it would break, and the new string would not get extracted. It's quite easy to fix too.

jerome-poisson: _posting back this comment which was lost in some weird parallel universe_:

it should be marked in the backed though, but the issue is that I don't think we have a mechanism to match backend translation language to user preference in client.

tomaskikutis: > I don't think we have a mechanism to match backend translation language to user preference in client.

That's a good opportunity to fix back-end translations then. User language can be taken from `user.language` property.

jerome-poisson: well, that's definitely not something I'll do with this issue, it's a whole new thing.

jerome-poisson: as an update, after talking with Petr it seems that we have a solution for that, so I've translated the labels in the backend, and I will update this patch soon.

Not sure about the other labels of the files, I suspect that I would need to check all services, including customer ones, so that's probably work for a separate ticket. I'll have a look though.

jerome-poisson: fixed in this PR, I've created https://dev.sourcefabric.org/browse/SDESK-5431 for other labels.

jerome-poisson:
note: I've added a placeholder, I was needing it to associate the button with the provider in the backend.

scripts/apps/ingest/directives/IngestSourcesContent.ts
```diff
@@ -580,6 +580,16 @@ export function IngestSourcesContent(ingestSources, notify, api, $location,
                     return feedingService ? feedingService.templateUrl : '';
                 };
 
+                /**
+                 * Do URL request specified in url_request field
+                 * @param provider ingest provider metadata
+                 * @param field url_request field metadata
+                 */
+                $scope.doUrlRequest = (provider: Object, field: Object): void => {
```
tomaskikutis, line 588: Can you use the exact types for provider/field? Typing it as `Object` isn't really more useful than not specifying a type at all.

I would also drop the comments on parameters since they don't really tell much more than "this is a variable and it contains data".

jerome-poisson: Did follow what is done in the rest of the file for consistency, but sure I can remove.
About `field` there is no type for that, it's just description dict from the backend, that why it's not specified more than that.

tomaskikutis: > About `field` there is no type for that.

You are reading a `url` property from that variable and performing a `replace` operation which means it's a string. You could specify then `field: {url: string}`.

It doesn't matter how is it described in the back-end or even doesn't have types defined there - we can fire up the debugger and inspect which fields does the object contain - even if they are optional and we can type them as such - `{optionalProperty?: any}`

scripts/core/superdesk-api.d.ts
```diff
@@ -559,6 +559,17 @@ declare module 'superdesk-api' {
         preview?: boolean;
     }
 
+    export interface IFeedingServiceField extends IBaseRestApiResponse {
+        id?: string;
```
tomaskikutis, line 563: It shouldn't extend `IBaseRestApiResponse` - this type is only API response for a single resource like `users/x`. These responses have `_etag`, `_created`, `_updated`, `_id` fields.

Can you also put this interface in the file where you are using it instead of here? Here we put only interfaces that need to be exposed to extensions.

tomaskikutis (approved):

PR #3542, opened 2020-07-22
Users that are disabled or deactivated should not appear in 'mark for user' list. [SDBELGA-404]

GyanP:


tomaskikutis:
Have you seen my comment on JIRA?

> can you change the `users` endpoint so it doesn't return disabled or deactivated users by default and does it when such filters are passed using the [pyeve filtering format](https://docs.python-eve.org/en/stable/features.html?highlight=filter#filtering)?

`/users` should return all, except disabled or deactivated users
`/users?where={"include_deactivated": true, "include_disabled": true}` should return all including disabled,deactivated

In such case, `SelectUser.tsx` file shouldn't be modified at all since it would work with the defaults. We would only need to add special parameters to user management - the only place where disabled/deactivated users should be listed.

One thing I'm not sure about is query format. I would like to exclude deactivated/disabled users by default, but `where={disabled: true}` should only return disabled users as expected. I suggest using a separate parameter, we can call it "include" or "ignore_defaults" for example `/users?include=["is_active", "is_enabled"]`

@petrjasek what do you think?

petrjasek:
I would also go for a new param and keep the where filter as it is,
either using that include list or like `/users/?show_inactive=1&show_disabled=1`

tomaskikutis:
@GyanP when server and client PRs depend on each other, especially in a backward-incompatible way, always merge them together.

Since you have merged the server PR, but not this one, e2e tests are now failing for unrelated PRs. I'll merge this PR and it will get fixed so you don't need to do anything, just try to merge them together in the future.

Also, you can safely ignore failing travis tests if all tests are passing on fireq.

PR #3545, opened 2020-07-24
set chromedriver version to 84

petrjasek:


tomaskikutis:
I think you also need to update the webdriver on fireq

petrjasek:
yep I've updated fireq now to use the same package.json script as travis

PR #3547, opened 2020-07-27
show spiked items in personal view sent items tab

petrjasek:
SDANSA-428

### DIFF
--- a/scripts/apps/monitoring/services/CardsService.ts
+++ b/scripts/apps/monitoring/services/CardsService.ts
@@ -65,6 +65,8 @@ export function CardsService(search, session, desks, $location) {
 
         if (card.type === 'spike' || card.type === 'spike-personal') {
             params.spike = 'only';
+        } else if (card.type === 'personal' && card.sent) {
+            params.spike = 'include';
         }
 
         return params;
--- a/scripts/apps/monitoring/services/CardsService.ts
+++ b/scripts/apps/monitoring/services/CardsService.ts
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import {setFilters} from 'apps/search/services/SearchService';
+import {setFilters, IQueryParams} from 'apps/search/services/SearchService';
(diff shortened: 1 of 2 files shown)


scripts/apps/monitoring/services/CardsService.ts
```diff
@@ -65,6 +65,8 @@ export function CardsService(search, session, desks, $location) {
 
         if (card.type === 'spike' || card.type === 'spike-personal') {
             params.spike = 'only';
+        } else if (card.type === 'personal' && card.sent) {
+            params.spike = 'include';
```
tomaskikutis, line 69: I'd call it `include-sent` to be more precise

petrjasek: this param just controls if spiked items will be included or excluded, it doesn't handle sent filtering

tomaskikutis: it's good to have clear names even for local variables. In this case it's not clear what will be included when that parameter is set.

petrjasek: not sure if I follow here, if you set `spike` to `include` it includes spiked items. if it's `spike = 'include-spiked';` that's just duplication, isn't it?

tomaskikutis: You're right. It'd be helpful though to document possible values to `spike` option and their meanings.

scripts/apps/search/services/SearchService.ts
```diff
@@ -11,6 +11,7 @@ import {getDateFilters, getDateRangesByKey} from '../directives/DateFilters';
 import {gettext} from 'core/utils';
 import {KILLED_STATES} from 'apps/archive/constants';
 import {appConfig} from 'appConfig';
+import {string} from 'prop-types';
```
tomaskikutis, line 14: This is probably a mistake, you don't need this import.

PR #3550, opened 2020-07-30
Code improvements for list view

tomaskikutis:
SDESK-5419

* Import angular services when needed instead of passing them 10 levels deep
* Don't pass the entire `scope` to React components. Pass only required data.


scripts/apps/search/helpers.tsx
```diff
@@ -133,9 +133,21 @@ export function positionPopup(target, zIndex = 1000) {
...
-export function renderArea(area: string, itemProps, props?: { className?: string }, customRender: any = {}) {
+interface IItemProps {
+    item: any;
+    listConfig?: any;
+    singleLine?: any;
+    narrow?: any;
+}
+
+export function renderArea(
+    area: string,
```
pablopunk, line 144: is this only 'firstLine' and 'secondLine'? If so we could change the type of it so it doesn't allow any string

tomaskikutis: Since I already merged this, I'll update the type in some other unrelated PR ;D

PR #3552, opened 2020-08-04
[SDESK-5272] New SAMS extension with Set management

MarkLark86:


MarkLark86:
@pavlovicnemanja I have created some generic typescript components here that should satisfy for react components inside the ui-framework (https://github.com/superdesk/superdesk-client-core/pull/3552/commits/1c2b0c7b8c963c40378a35a618730807391d79d0). Could you have a look at those for me a advise if any changes are required please?

scripts/extensions/sams/src/components/sets/setContentPanel.tsx
```diff
@@ -0,0 +1,91 @@
...
+                );
+            }
+            return null;
+        }
+
+        render() {
+            return (
+                <RightPanel open={this.props.state !== CONTENT_PANEL_STATE.CLOSED}>
+                    <Panel side="right">
+                        {this.renderPanel()}
```
tomaskikutis, line 85: It's easier to read/understand code when all rendering logic is contained in the `render` method. If something is too big, it's better to make a separate component. If it's a simple `if` statement, ternaries or inline [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)s can be used.

scripts/extensions/sams/src/components/sets/setEditorPanel.tsx
```diff
@@ -0,0 +1,197 @@
...
+        }
+
+        onChange = {
+            name: (value: string) => this.onFieldChange('name', value),
+            description: (value: string) => this.onFieldChange('description', value),
+            destination_name: (value: string) => this.onFieldChange('destination_name', value),
+        };
+
+        onFieldChange(field: string, value: string): void {
+            const dirtyFields = ['state', 'name', 'description', 'destination_name'];
```
tomaskikutis, line 67: diffs can be computed dynamically in a generic way without knowing the field names. Why hardcode them?

There's a library we are using for computing diffs https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/helpers/CrudManager.tsx#L105

scripts/extensions/sams/src/components/sets/setListGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export class SetListGroup extends React.PureComponent<IProps> {
+    render() {
+        return (
+            <React.Fragment>
+                <ListHeader title={this.props.title} marginTop={this.props.marginTop} />
+                <ListItemGroup spaceBetweenItems={true}>
+                    {this.props.sets?.length === 0 ? (
+                        <ListItem inactive={true} shadow={2} noHover={true}>
```
tomaskikutis, line 33: There's a shorthand where you don't need to specify the value of a prop if it is `true`.

```<ListItem inactive noHover shadow={2}>```

File: scripts/extensions/sams/src/ui/list/index.ts [hunk 4]
```diff
@@ -0,0 +1,8 @@
+export {ListHeader} from './ListHeader';
```
tomaskikutis, line 1: Is this proxy file useful for anything? When I want to import something, I write the name and the editor autocompletes and auto imports it no matter where it is.

And if I didn't have auto-import it'd be annoying to navigate such empty proxy files just to get to the definition

scripts/extensions/sams/src/components/samsWorkspace.tsx
```diff
@@ -0,0 +1,92 @@
...
+import {ISamsAPI, MODAL_TYPES} from '../interfaces';
+
+import {ButtonGroup, Dropdown, SubNav} from 'superdesk-ui-framework/react';
+import {IMenuGroup} from 'superdesk-ui-framework/react/components/Dropdown';
+import {HeaderPanel, LayoutContainer, MainPanel} from '../ui';
+
+import {getManageSetsModalComponent} from './sets/manageSetsModal';
+
+interface IState {
+    currentModal: MODAL_TYPES;
```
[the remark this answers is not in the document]
MarkLark86: There will be more modals added to this workspace in the future

scripts/extensions/sams/src/components/sets/setListItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+        if (this.props.onEdit) {
+            e.preventDefault();
+            e.stopPropagation();
+            this.props.onEdit(this.props.set);
+        }
+    }
+
+    onDelete(e: React.MouseEvent<HTMLDivElement>) {
+        if (this.props.onDelete) {
+            e.preventDefault();
```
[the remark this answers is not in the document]
MarkLark86: This component encapsulates the entire list item, therefor it is better for this component to not propagate events when a `Delete` or `Edit` action has been requested by a user click.

scripts/extensions/sams/src/api/sets.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export function createSet(superdesk: ISuperdesk, item: ISet): Promise<ISetItem> {
+    const {gettext} = superdesk.localization;
+    const {notify} = superdesk.ui;
+
+    return superdesk.dataApi.create<ISetItem>(RESOURCE, item as ISetItem)
+        .then((set: ISetItem) => {
+            notify.success(gettext('Set created successfully'));
+
+            window.dispatchEvent(
```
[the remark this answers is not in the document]
MarkLark86: The original problem has been solved by using redux. As for multi-user scenarios, this will be covered at a later date

scripts/extensions/sams/src/components/sets/setListGroup.tsx
MarkLark86: Either way is an acceptable definition of an attribute. I prefer to clearly define the value of the attribute

scripts/extensions/sams/package.json
```diff
@@ -0,0 +1,26 @@
+{
```
tomaskikutis, line 1: Could you drop redux/reselect? Redux might be helpful for managing the global app state, but this is a single CRUD view. Adding redux/reselect only creates a layer of indirection and a pile of boilerplate to manage without any benefits I can see.

You could use a simple react component for separating data management:

```
interface ISet {
    _id: string;
    // ...
}

interface IProps {

}

interface IState {
    sets: Array<ISet>;
}

class SetsDataLayerComponent extends React.Component<IProps, IState> {
    constructor(props: IProps) {
        super(props);

        this.state = {
            sets: [],
        };

        this.addSet = this.addSet.bind(this);
        this.removeSet = this.removeSet.bind(this);
    }

    addSet() {
        this.setState({sets: this.state.sets.concat({_id: 'newid'})});
    }

    removeSet(id: string) {
        this.setState({sets: this.state.sets.filter(({_id}) => _id !== id)});
    }

    render() {
        return (
            <SetsManagementUi
                addSet={this.addSet}
                removeSet={this.removeSet}
            />
        );
    }
}
```

scripts/extensions/sams/src/components/samsWorkspace.tsx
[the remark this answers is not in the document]
tomaskikutis: In the case of more than one modal it might not be possible to keep the function stateless, but still, it could be a simple function call without a need to keep selected modal in the component state. Think notifications, for example, you call the function and a notification gets displayed. No need to store in the component state which one is being displayed at the moment.

scripts/extensions/sams/package.json
MarkLark86: This is the first of many PRs for SAMS integration into Superdesk. This workspace, for example, is going to have a lot more functionality. Loading Sets/StorageDestination into the store and using selectors can help reduce boilerplate in future sections of this UI, and also separate the state and view. This method was used heavily in Planning and proved very useful. It also helps keeps the view in sync with the data, i.e. update a set in the store and all views that use that set are automatically updated

tomaskikutis: > Loading Sets/StorageDestination into the store and using selectors can help reduce boilerplate in future sections of this UI, and also separate the state and view.

Regarding separation between state and view, the 50 LOC data layer component I posted already does it without requiring 10 files of boilerplate or external libraries. The same data layer component can be used for multiple UIs. Data can be synced by updating the data layer component state on WebSocket events. If there's a need for common code we can extract code to functions/components and reuse them this way.

I didn't work on planning much, but when I did I found it annoying and a waste of time to have to go through all the levels of abstraction only to discover a straight-forward setting of a value which could have been a one or two calls instead of 7.

Do you have an example from planning where by using redux/reselect you have been able to do something which wouldn't be possible using a simple wrapper component?

I'd say let's start with a minimalistic approach using the data component and introduce something more complex only if we hit into obstacles.

scripts/extensions/sams/src/api/sets.ts
```diff
@@ -0,0 +1,104 @@
...
+                    {detail: set},
+                ),
+            );
+
+            return set;
+        })
+        .catch((error: any) => {
+            notify.error(gettext('Failed to create the Set'));
+
+            throw new Error(error);
```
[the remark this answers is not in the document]
MarkLark86: The error response from the API are not currently using a well defined schema. Once this is done we can improve the user experience when it comes to errors or failures of the system (these errors could be more than validation failures)

scripts/extensions/sams/src/ui/list/index.ts
MarkLark86: It allows the developer to write
```js
import {ListItem, ListItemActionMenu, ListItemBorder, ListItemColumn, ListItemRow} from '../../ui/list';
```
instead of
```js
import {ListItem} from '../../ui/list/ListItem';
import {ListItemActionMenu} from '../../ui/list/ListItemActionMenu';
import {ListItemBorder} from '../../ui/list/ListItemBorder';
import {ListItemColumn} from '../../ui/list/ListItemColumn';
import {ListItemRow} from '../../ui/list/ListItemRow';
```

@fritzSF / @pavlovicnemanja As this part of the code will eventually be destined for the `ui-framework` library, which would you prefer? Or any other suggestions

scripts/extensions/sams/src/components/sets/setEditorPanel.tsx
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        onFieldChange(field: keyof ISetItem, value: any): void {
+            const updates = this.state.updates;
+            let dirty = true;
+
+            updates[field] = value;
+
+            if (this.props.original != null) {
+                dirty = hasItemChanged<ISetItem>(this.props.original, this.state.updates);
```
tomaskikutis, line 103: It's not wrong, but type argument doesn't provide any additional constraints if types of arguments are specified

scripts/extensions/sams/src/utils/api.ts
```diff
@@ -0,0 +1,32 @@
+// External Modules
+import {generate} from 'json-merge-patch';
+
+// Types
+import {IBaseRestApiResponse} from 'superdesk-api';
+
+export function generatePatch<T extends IBaseRestApiResponse>(original: T, updates: Partial<T>): Partial<T> {
```
tomaskikutis, line 7: I added this function because the code in the core was messy and not fixable in a reasonable time. I don't see a reason to carry this over to the new code. It should be fine for you to use `generate` directly. The only issue is that the library has poor types so maybe it might be worth creating a tiny wrapper just for types
```
export function generatePatch<T>(obj1: Partial<T>, obj2: Partial<T>): Partial<T> {
    return (generate(obj1, obj2) ?? {}) as Partial<T>
}
```

scripts/extensions/sams/src/api/sets.ts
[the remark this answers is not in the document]
MarkLark86: The problem is that there are multiple different ways the libraries, Flask/Eve/Cerberus, returns errors. So we decided to tackle this problem in a separate ticket, rather than write some code now and change that code again in the future

scripts/extensions/sams/src/utils/api.ts
```diff
@@ -1,32 +1,10 @@
...
-    delete patch['_etag'];
-    delete patch['_links'];
-
-    return patch;
+export function generatePatch<T>(original: Partial<T>, updates: Partial<T>): Partial<T> {
+    return (generate(original, updates) ?? {}) as Partial<T>;
 }
 
-export function hasItemChanged<T extends IBaseRestApiResponse>(original: T, updates: Partial<T>): boolean {
+export function hasItemChanged(original: any, updates: any): boolean {
```
tomaskikutis, line 8: Avoid `any` when possible. Wouldn't `hasItemChanged<T>(original: T, updates: Partial<T>)` work?

scripts/extensions/sams/src/api/sets.ts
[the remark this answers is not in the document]
tomaskikutis: Which ticket is that?

scripts/extensions/sams/src/utils/api.ts
MarkLark86: I'll update that tomorrow

MarkLark86:
> * Try not to use `as` TypeScript operator since it hides compilation errors. Recently there were improvements made to the compiler, but before that, it was possible to do `var a = {a:1} as {b: number};` and it would compile. Most of the time `as` can be avoided by fixing the types. I think I only used it once due to issues with a 3rd party library.

Done

> * We have a `GenericListPageComponent` which is meant for generic CRUD views(it's used for annotations library and internal destinations list). It has validation, filtering, preview and most of what you implemented in this GUI. It might require tweaking in order to be able to display items in multiple groups, but other than that it might be worth using it instead of implementing the list from scratch.

One of the advantages of this new extension is that we're able to utilise the UI-Framework to it's fullest extend, including the usage of the new React components. Through iterative development we can test this full usage in this sandbox extension, and work with the UI team to make the required changes (if any) without breaking any other part of Superdesk.

Through looking at the code for the `GenericListPageComponent` I can see that there are custom `style` properties being applied as well as no usage of the new React components.

scripts/extensions/sams/src/utils/redux.ts
```diff
@@ -0,0 +1,55 @@
+// External Modules
```
tomaskikutis, line 1: I appreciate your effort in converting this, but I don't think it's simpler than the previous version. It does remove some boilerplate but adds complexity, type safety isn't as strong and interfaces look a bit weird, at least to me since I don't understand them well.

I wasn't suggesting to merge actions and reducers together, just get rid of one-liner action creator functions and simply inline the actions like:


```
interface IIncrementAction {
    type: 'INCREMENT';
    amount: number;
}

interface IDecrementAction {
    type: 'DECREMENT';
    amount: number;
}

interface IResetAction {
    type: 'RESET';
}

type IAction = IIncrementAction | IDecrementAction | IResetAction;

function counter(state, action: IAction) {
    if (typeof state === 'undefined') {
        return 0;
    }

    switch (action.type) {
        case 'INCREMENT':
            return state + action.amount;
        case 'DECREMENT':
            return state - action.amount;
        case 'RESET':
            return 0;
        default:
            return assertNever(action.type);
    }
}

var store = Redux.createStore(counter);

store.dispatch({type: 'DECREMENT', amount: 1});

```

Could you revert this action-reducer merge and simply use inline actions in simple cases?

MarkLark86: I have reverted back to my original Redux implementation, as that implementation is based on https://redux.js.org/recipes/usage-with-typescript (a recipe for using Redux with Typescript by the Redux team). 

If any developer wishes to learn how state management is used in this extension, we can point that developer to the Redux website and the above recipe provided.

scripts/extensions/sams/src/store/sets/actions.ts
```diff
@@ -0,0 +1,89 @@
...
+            .then((response: boolean) => {
+                el.remove();
+                return response;
+            });
+    };
+}
+
+export function confirmBeforeDeletingSet(set: ISetItem): IThunkAction<void> {
+    return (dispatch, getState, {api}) => {
+        return dispatch(openDeleteConfirmationModal(set))
```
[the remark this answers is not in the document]
MarkLark86: The ThunkDispatch middleware is initialised with the Superdesk and SAMS API implementations: https://github.com/superdesk/superdesk-client-core/blob/051e2f51028443cc722e824384bfe85791dbe649/scripts/extensions/sams/src/apps/samsApp.tsx#L35

This allows any thunk action access to these APIs without the need of a HOC to pass through these API implementations. So the `openDeleteConfirmationModal` can easily have access to `superdesk.localization.gettext` and `superdesk.ui.confirm` methods: https://github.com/superdesk/superdesk-client-core/blob/051e2f51028443cc722e824384bfe85791dbe649/scripts/extensions/sams/src/store/sets/actions.ts#L79

scripts/extensions/sams/src/apps/samsApp.tsx
```diff
@@ -0,0 +1,61 @@
...
+                extension.exposes.store.dispatch<any>(loadStorageDestinations()),
+                extension.exposes.store.dispatch<any>(loadSets()),
+            ])
+                .then(() => {
+                    this.setState({ready: true});
+                });
+        }
+
+        componentWillUnmount() {
+            extension.exposes.store = undefined;
```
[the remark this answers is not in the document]
MarkLark86: I have removed this from `extension.exposes`.

To be fair, the usage of `exposes` has not been documented anywhere. In my opinion the core should never be accessing anything inside an extension. Other extensions may access `exposes` of other extensions, making it a direct dependency, but the core should not depend on any extension. That is where hooks in the core should expose ways for extensions to extend or enhance the core functionality

scripts/extensions/sams/src/store/sets/actions.ts
```diff
@@ -83,18 +75,22 @@ export function loadSets(): IThunkAction<Array<ISetItem>> {
...
 export function openDeleteConfirmationModal(set: ISetItem): IThunkAction<boolean> {
-    return (dispatch, _getState, {superdesk}) => {
+    return (_dispatch, _getState, {superdesk}) => {
         const {gettext} = superdesk.localization;
         const {confirm} = superdesk.ui;
 
-        dispatch(setDeleteConfirmationModalOpen(true));
+        const el = document.createElement('div');
+
+        el.classList.add('modal__backdrop', 'fade', 'in');
```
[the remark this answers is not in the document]
MarkLark86: The comments outline an issue between ui-framework and pure React modals. To fix this I would have to implement a custom React component just for a confirmation dialog. So for now this will do, but the underlying problem will need to be fixed in general

MarkLark86:
@petrjasek This PR is ready for review. The builds are failing, awaiting a prerelease of the ui-framework. I will modify package.json once this prerelease branch is ready

scripts/extensions/sams/src/apps/samsApp.tsx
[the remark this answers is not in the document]
tomaskikutis: I totally agree about core not accessing anything inside extensions. `exposes` is only used now by a "mark for user" extension and user activity widget which should be an extension on its own, but hasn't been converted to one yet.

scripts/extensions/sams/src/store/sets/actions.ts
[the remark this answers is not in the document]
tomaskikutis: It's hacky to do so. The purpose of thunk actions is to handle async data and it's confusing when it's used without any data and without dispatching any actions. Could you avoid dispatching fake actions and pass the `api` directly to that function as an argument?

scripts/extensions/sams/src/store/sets/types.ts
```diff
@@ -0,0 +1,56 @@
+// Types
+import {CONTENT_PANEL_STATE, ISetItem} from '../../interfaces';
+
+export const RECEIVE_SETS = 'sets__receive';
```
tomaskikutis, line 4: Is there a reason to export these constants instead of inlining them in the interface?

scripts/extensions/sams/src/components/sets/setListPanel.tsx
```diff
@@ -0,0 +1,98 @@
...
+export function getSetListPanel(superdesk: ISuperdesk) {
+    const {gettext} = superdesk.localization;
+
+    const mapStateToProps = (state: IApplicationState) => ({
+        sets: getSetsGroupedByState(state),
+        storageDestinations: getStorageDestinationsById(state),
+        currentSetId: getSelectedSetId(state),
+    });
+
+    const mapDispatchToProps = (dispatch: Dispatch) => ({
```
tomaskikutis, line 45: Can you remove instances of `mapDispatchToProps`?

scripts/extensions/sams/src/components/sets/setEditorPanel.tsx
```diff
@@ -0,0 +1,245 @@
...
+                name: (value: string) => this.onFieldChange('name', value),
+                description: (value: string) => this.onFieldChange('description', value),
+                destination_name: (value: string) => this.onFieldChange('destination_name', value),
+                state: (value: boolean) => this.onStateChange(value),
+            };
+        }
+
+        onFieldChange(field: keyof ISetItem, value: any): void {
+            const updates = this.state.updates;
+            let dirty = true;
```
tomaskikutis, line 98: What about computing dirtiness in the render method? It would allow dropping `this.onChange` and `onFieldChange` which don't seem to be doing anything, but passing values around.

scripts/extensions/sams/src/store/sets/types.ts
MarkLark86: This has already been covered in the link I provided: https://redux.js.org/recipes/usage-with-typescript#type-checking-actions--action-creators

scripts/extensions/sams/src/components/sets/setListPanel.tsx
MarkLark86: Using `mapDispatchToProps` makes the view component itself more declarative as it does not need to know about `dispatch` and the usage thereof, instead it only has to worry about what prop to execute. https://react-redux.js.org/using-react-redux/connect-mapdispatch#more-declarative

scripts/extensions/sams/src/components/sets/setEditorPanel.tsx
MarkLark86: 1) It is better to calculate the `dirty` state only when the attributes change, not on every single render
2) The updates are stored in the local state, and is used when creating or updating an item. The untouched original item is passed in through `props.original` and the updates are stored in `state.updates`
https://github.com/superdesk/superdesk-client-core/blob/8bee669e05cadd9db6c1490ec4a82086d71c328a/scripts/extensions/sams/src/components/sets/setEditorPanel.tsx#L130
https://github.com/superdesk/superdesk-client-core/blob/8bee669e05cadd9db6c1490ec4a82086d71c328a/scripts/extensions/sams/src/components/sets/setEditorPanel.tsx#L132

tomaskikutis: My bad, I thought dirtiness was checked via reference equality. Having `onChange` there still doesn't seem useful compared to calling `onFieldChange`/`onStateChange` directly when handling input changes.

scripts/extensions/sams/src/store/sets/actions.ts
```diff
@@ -0,0 +1,145 @@
...
+            });
+    };
+}
+
+export function createSet(item: Partial<ISetItem>): IThunkAction<ISetItem> {
+    return (dispatch, _getState, {api}) => {
+        return api.sets.create(item)
+            .then((newSet: ISetItem) => {
+                dispatch(loadSets());
+                dispatch(previewSet(newSet._id));
```
tomaskikutis, line 140: Shouldn't preview action be dispatched only after `loadSets` promise is resolved?

scripts/extensions/sams/src/components/sets/setListPanel.tsx
MarkLark86: The usage of `mapDispatchToProps` is the convention being used in this extension

scripts/extensions/sams/src/store/sets/actions.ts
MarkLark86: Probably, but through the use of Redux the preview get's rendered properly anyway. The code works fine

scripts/extensions/sams/src/components/sets/setEditorPanel.tsx
MarkLark86: By convention I do not use anonymous functions in the render function. Using the `onChange` object provides the same reference to the callback on every render

scripts/extensions/sams/src/store/sets/types.ts
MarkLark86: As I have stated earlier:
> If any developer wishes to learn how state management is used in this extension, we can point that developer to the Redux website and the above recipe provided.

The code works perfectly fine this way, I do not see the problem

scripts/extensions/sams/src/components/sets/setListPanel.tsx
tomaskikutis: I thought we agreed on a compromise that if we are to use redux, we would minimize the boilerplate, no?

Convention or not, we should be able to justify the usage of any pattern and drop everything that doesn't provide significant value.

I think I refuted the argument about it being more declarative. Are there other reasons you think it's worth using? As I mentioned before, the reason I don't want it used, because it's an abstraction which doesn't help in any way, unless you want to connect the same component twice, which we are not doing.

I don't think we should have any different conventions in core/extensions since it's the same team maintaining both.

scripts/extensions/sams/src/store/sets/actions.ts
tomaskikutis: It will work fine in most cases when the internet connection is stable. If it's not and HTTP request to fetch sets takes longer, the preview will get opened before the sets are loaded.

We should ensure that there aren't crashes or bugs because of a slower connection. Can you fix this and similar situations and not dispatch actions synchronously that depend on an asynchronous result?

scripts/extensions/sams/src/store/sets/types.ts
tomaskikutis: If you don't see the problem, but I see it, why not fix it? Or do you see a problem with not using constants?

The problem that I see is that we are adding additional lines of code that aren't needed. If the problem can be solved in 50 lines of code, why write 500 lines to solve it? Why create variables and manage imports when it can be avoided?

scripts/extensions/sams/src/store/sets/actions.ts
MarkLark86: https://github.com/superdesk/superdesk-client-core/blob/74e4f93e37ff21e78b6624b4e0e98626fa26958f/scripts/extensions/sams/src/components/sets/setPreviewPanel.tsx#L46

tomaskikutis: It might not crash in this scenario, but it's always better to handle things correctly as close to the top level as possible. For example, if the preview component can't be rendered if the set is null, why even allow nullable set prop instead of ensuring that it's not null before the component is rendered?

scripts/extensions/sams/src/components/sets/setEditorPanel.tsx
tomaskikutis: > onChange: Dictionary<string, (value: any) => void>;

Yep, I agree that having the same references is useful for checking reference equality. Could you declare it as separate functions so we don't lose type safety due to `any` type?

scripts/extensions/sams/src/store/sets/selectors.ts
```diff
@@ -0,0 +1,101 @@
...
+        sets.forEach((set: ISetItem) => {
+            switch (set.state) {
+            case SET_STATE.USABLE:
+                groupedSets.usable.push(set);
+                break;
+            case SET_STATE.DISABLED:
+                groupedSets.disabled.push(set);
+                break;
+            case SET_STATE.DRAFT:
+            default:
```
pablopunk, line 93: could the state be any value other than these 3? otherwise I would use `default: assertNever(set.state)` [like here](https://github.com/superdesk/superdesk-client-core/blob/5cf368ccfd5fb8b060738019994828cab82166c8/scripts/core/ui/components/alert.tsx#L14-L15)

scripts/extensions/sams/src/components/sets/setListItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+            event.stopPropagation();
+            this.props.deleteSet(this.props.set);
+        }
+    }
+
+    render() {
+        const {set, storageDestination} = this.props;
+        const storageDestinationText = storageDestination == null ?
+            '' :
+            `${storageDestination._id} / ${storageDestination.provider}`;
```
pablopunk, line 54: I wonder if we will need translations here in the future. For example I can see now `Default` as the id and that's probably something we want to translate.

MarkLark86: This is actually defined in the `settings.py` from the SAMS API, so this will be instance specific.

scripts/extensions/sams/src/store/sets/actions.ts
MarkLark86: Updated in https://github.com/superdesk/superdesk-client-core/pull/3552/commits/61cafacbf209090496b108c5132368421d774469

scripts/extensions/sams/src/store/index.ts
```diff
@@ -0,0 +1,76 @@
...
+                throw error;
+            }
+        };
+    };
+}
+
+let store: Store | undefined;
+
+export function createReduxStore(
+    extraArguments: any,
```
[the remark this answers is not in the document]
MarkLark86: I have added the `IExtraArguments` to the signature of `createReduxStore`.
As for the think dispatches, if you use `IThunkAction` type from https://github.com/superdesk/superdesk-client-core/blob/61cafacbf209090496b108c5132368421d774469/scripts/extensions/sams/src/store/types.ts the IDE will recognise the types in the `extraArgument` param.
https://github.com/superdesk/superdesk-client-core/blob/61cafacbf209090496b108c5132368421d774469/scripts/extensions/sams/src/store/sets/actions.ts#L115

scripts/extensions/sams/src/store/sets/types.ts
MarkLark86: You're over-exaggerating the boilerplate added by following the examples provided by the Redux team

scripts/extensions/sams/src/components/sets/setListPanel.tsx
MarkLark86: I spent an entire day rewriting the redux integration to cater for your needs to reduce boilerplate, which was ultimately rejected.

You have the opinion that you don't want `mapDispatchToProps`, I have the opinion (and so does the Redux team) that it is more declarative if the `dispatch` function is not used inside the React component. I believe that any usage of Redux directly should not be used inside any React Component, that is what `mapStateToProps` and `mapDispatchToProps` is for. Using `mapDispatchToProps` does not add much boilerplate at all (in this example it is only 6 lines).

tomaskikutis: > I spent an entire day rewriting the redux integration to cater for your needs to reduce boilerplate, which was ultimately rejected.

Thank you for doing that and am sorry if I wasn't clear enough. I only asked to drop action creator functions and inline the actions, not merge actions and reducers into one system.

Regarding `mapDispatchToProps`, it's not only about lines, but also about call stack and having to jump around the code more to discover what the code does.

For example, on button click a prop is called, I then have to jump to that prop to see what it does - I discover that it does nothing, but is just a proxy for dispatching an action. I then go to action definition to see its name and search for a reducer to finally find the actual state changes done by clicking that button. If I can skip one layer and see the actual state change quicker that's already more convenient.

> I believe that any usage of Redux directly should not be used inside any React Component

And I believe that everyone should be able to justify technical decisions rationally and explain how it helps or harms code quality.

Anyway, let's not spend too much time on this, keep the mapping and I will do an extra jump in the call stack :)

scripts/extensions/sams/src/store/sets/types.ts
tomaskikutis: I am over-exaggerating, but that was just an example to illustrate the point. If there are lines that aren't really useful for anything and can be dropped without a negative impact on code quality, why not do so? Are we obligated to repeat patterns from documentation pages even if they aren't useful?

Let me know if I'm wrong and there is any benefit to having those constants.

MarkLark86: These constants are used in multiple places:
https://github.com/superdesk/superdesk-client-core/blob/61cafacbf209090496b108c5132368421d774469/scripts/extensions/sams/src/store/sets/types.ts#L6
https://github.com/superdesk/superdesk-client-core/blob/61cafacbf209090496b108c5132368421d774469/scripts/extensions/sams/src/store/sets/actions.ts#L21
https://github.com/superdesk/superdesk-client-core/blob/61cafacbf209090496b108c5132368421d774469/scripts/extensions/sams/src/store/sets/reducers.ts#L25

This is quite common practice in any programming language. If you think this code will cause bugs or will not work in any way, I will change it.

tomaskikutis: Rain dances were also a common practice. We shouldn't be copying things without understanding them just because others do it. That's the main issue I'm concerned about much more than the constants themselves.

I'll help you out and link the reasons why people use constants for action names: https://stackoverflow.com/a/37054518/1175593

These are good points, but they are irrelevant when TypeScript is used because there already is a union of all possible actions in a single place that satisfies all listed goals.

scripts/extensions/sams/src/api/sets.ts
```diff
@@ -0,0 +1,74 @@
...
+    return superdesk.dataApi.patch<ISetItem>(RESOURCE, original, updates)
+        .then((set: ISetItem) => {
+            notify.success(gettext('Set updated successfully'));
+
+            return set;
+        })
+        .catch((error: any) => {
+            notify.error(gettext('Failed to update the Set'));
+
+            return Promise.reject(error);
```
pablopunk, line 57: is this different than just `throw error`?

MarkLark86: I originally had this throw an error, but changed it to use `Promise.reject` after a peer review

tomaskikutis: @pablopunk errors are for crash-like conditions which signify that something is wrong with the code and it should be logged. Validation issues are expected application behaviour.

pablopunk: I was just asking, but now that you mention it, you're right on that but that doesn't have to do with `throw` vs `reject`. Also in this case the error could be something else other than a validation error, like a network error, and it will be catched by this block too

tomaskikutis: Validation issues were being logged as errors in the console, that's why I asked Mark not to throw it. I agree that we could put a conditional here and throw errors for failures which are not because of validation.

MarkLark86: There are tickets in the pipeline to improve the validation vs errors, from the SAMS API right through to this front end. We can improve this handling with those tickets

tomaskikutis (approved):

pablopunk (approved):

PR #3554, opened 2020-08-04
fix page retrieval

tomaskikutis:


petrjasek:
does it affect 2.0?

tomaskikutis:
Probably not, but I'd port cherry-pick it anyway in case a new extension or method gets added and it could break then.

PR #3557, opened 2020-08-04
handle missing related items

petrjasek:
when related item is not available anymore it was
showing the loader indefinitely. now it will remove such
item from related and show a warning to user.

SDESK-5252

### DIFF
--- a/scripts/apps/relations/directives/RelatedItemsDirective.ts
+++ b/scripts/apps/relations/directives/RelatedItemsDirective.ts
@@ -17,7 +17,7 @@ interface IScope extends IDirectiveScope<void> {
     item: IArticle;
     loading: boolean;
     reorder: (start: {index: number}, end: {index: number}) => void;
-    relatedItems: Array<IArticle>;
+    relatedItems: {[key: string]: IArticle};
     onchange: () => void;
     addRelatedItem: (item: IArticle) => void;
     isEmptyRelatedItems: (fieldId: string) => void;
@@ -239,7 +239,16 @@ export function RelatedItemsDirective(
                 scope.loading = true;
                 relationsService.getRelatedItemsForField(scope.item, scope.field._id)
                     .then((items) => {
-                        scope.relatedItems = items;
+                        scope.relatedItems = {};
+                        Object.keys(items).forEach((key) => {
+                            if (items[key] != null) {
+                                scope.relatedItems[key] = items[key];
+                            } else {
+                                scope.removeRelatedItem(key);
+                                notify.warning(gettext('Related item is not available.'));
+                            }
+                        });
+                    }).finally(() => {
                         scope.loading = false;
                     });
             };
--- a/scripts/apps/relations/services/RelationsService.ts
+++ b/scripts/apps/relations/services/RelationsService.ts
@@ -21,7 +21,14 @@ export function RelationsService(api, $q) {
 
         return $q.all(relatedItemsKeys.map((key) => {
             if (isLink(associations[key])) {
-                return api.find('archive', associations[key]._id);
+                return api.find('archive', associations[key]._id)
+                    .catch((response) => {
+                        if (response?.status === 404) {
+                            return null;
+                        }
+
+                        return Promise.reject(response);
+                    });
             }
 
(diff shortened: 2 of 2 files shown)


scripts/apps/relations/services/RelationsService.ts
```diff
@@ -21,7 +21,14 @@ export function RelationsService(api, $q) {
 
         return $q.all(relatedItemsKeys.map((key) => {
             if (isLink(associations[key])) {
-                return api.find('archive', associations[key]._id);
+                return api.find('archive', associations[key]._id)
+                    .catch((response) => {
+                        if (response?.status === 404) {
+                            return null;
```
tomaskikutis, line 27: what's the difference between returning `null` and having it rejected? Does the promise become resolved if you return from `catch`?

scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -239,7 +239,16 @@ export function RelatedItemsDirective(
...
                 relationsService.getRelatedItemsForField(scope.item, scope.field._id)
                     .then((items) => {
-                        scope.relatedItems = items;
+                        scope.relatedItems = {};
+                        Object.keys(items).forEach((key) => {
+                            if (items[key] != null) {
+                                scope.relatedItems[key] = items[key];
+                            } else {
+                                scope.removeRelatedItem(key);
+                                notify.warning(gettext('Related item is not available.'));
```
tomaskikutis, line 248: Why is it not available? Is it due to spiking/expiry? I'd say we should handle it on the back-end so that when the item disappears, it should also disappear where it's linked.

scripts/apps/relations/services/RelationsService.ts
petrjasek: yes it's resolve

scripts/apps/relations/directives/RelatedItemsDirective.ts
petrjasek: I think it is spiking/expiry, but couldn't find the item to check :smile:
in general backend can handle it too, it might be a bit more complicated if that item is being edited/locked at that very same time when such related item is being removed and it wouldn't be visible to users.

tomaskikutis: It would be more work to do it on the back-end, but it'd be a proper solution. Non-GUI stuff shouldn't be done on the front-end. Say we start working on a mobile app, would we copy-paste such workarounds across both interfaces?

What about implementing a "force-change" feature, which would send a patch with a message over a websocket and the client would update the UI with new data?

tomaskikutis: I realized it would also generate incorrect item history by displaying that a user removed a related item when in fact the item disappeared at a different date and was caused by a different user or system action.

petrjasek: I would make such a complex solution when we refactor authoring, but imo at this point it's over engineering. we would do it on backend not to duplicate the hadling in future client app but we have to handle anyhow the changes backend will produce in both and that imo introduces another level of complexity. in the long run it makes sense, but that's another task/story imo

tomaskikutis: I agree that it may not be easy to accomplish this now due to our accumulated tech debt, but in general, it's not a complex feature. It would listen for an event, update data & re-render. Sounds like <100 lines. No?

petrjasek: I'm just not sure about cases when user is just editing such field and backend sends notification to update

tomaskikutis: In this case, we are talking about related items, so there would only be an ID in the article and nothing could be edited.

I'm not sure if there would be use cases to force changes for other fields of locked items. We could force-save before applying changes. That would enable users to see that overwritten data in history.

petrjasek: yep something like that, I will create a ticket to follow up on this in future

PR #3562, opened 2020-08-06
add duplicateToPersonal action

petrjasek:
it uses copy on the backend, but adding a new action to keep UI
consistent

uses https://github.com/superdesk/superdesk-core/pull/1956

SDANSA-417


petrjasek:
not really

petrjasek:
actually there is on the api side, for duplication you need to specify target desk while for copy only item id

PR #3564, opened 2020-08-13
Make monitoring group dropdown labels translatable

tomaskikutis:
SDCP-280


scripts/apps/workspace/content/constants.ts
```diff
@@ -116,7 +116,7 @@ function getLabelForStageName(stageName: IStage['name']): string | null {
     }
 }
 
-function getLabelForStageType(stageType: ICard['type']): string | null {
+export function getLabelForStageType(stageType: ICard['type']): string | null {
```
pablopunk, line 119: you exported this and forgot to use it somewhere else or is just a export because it should be there?

tomaskikutis: I was trying different ways to solve the problem and one of them required the export. Later I decided to do it differently and forgot to remove the export :)

pablopunk (approved):

PR #3566, opened 2020-08-18
Improve display of validation error messages. [SDESK-5444]

GyanP:


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -18,6 +18,17 @@ import {addInternalEventListener} from 'core/internal-events';
 import {validateMediaFieldsThrows} from '../controllers/ChangeImageController';
 import {getLabelNameResolver} from 'apps/workspace/helpers/getLabelForFieldId';
 
+export function getErrorMessage(message, notify) {
+    return getLabelNameResolver().then((getLabelForFieldId) => {
+        const field = message.split(' ')[0];
+        const fieldLabel = getLabelForFieldId(field.toLocaleLowerCase());
+        const text = message.substr(message.indexOf(' ') + 1);
+
+        notify.error(gettext(`{{field}} ${text}`, {field: fieldLabel.toLocaleUpperCase()}));
```
tomaskikutis, line 27: `gettext` should only be used on static strings. I see that you have added string extraction separately, but it's a hack since strings that are coming from the back-end should already come translated. Could this be done?

scripts/apps/authoring/macros/macros.ts
```diff
@@ -81,7 +81,11 @@ function MacrosService(api, notify) {
             commit: !!commit,
         }).then((res) => res, (err) => {
             if (angular.isDefined(err.data._message)) {
-                notify.error(gettext('Error: {{message}}', {message: err.data._message}));
+                const error_messages = JSON.parse(err.data._message.replace(/'/g, '"'));
```
tomaskikutis, line 84: Why is it needed to replace quotes? Does the error come as a invalid JSON string? Could it come as an object/array and not a string?

GyanP: yes, there was an error while doing JSON parse. But now I have handled it in the back end itself.

tomaskikutis (approved):

petrjasek:
seems like those e2e errors are consistent on fireq, can you check it @GyanP ?

petrjasek:
hm my bad, I can see those on other PRs too

PR #3574, opened 2020-08-24
Fix editor3 unstable test

pablopunk:
SDESK-5212

### DIFF
--- a/spec/helpers/monitoring.ts
+++ b/spec/helpers/monitoring.ts
@@ -331,11 +331,11 @@ class Monitoring {
         this.previewAction = function(groupIndex, itemIndex) {
             const item = this.getItem(groupIndex, itemIndex);
 
-            browser.wait(ECE.elementToBeClickable(item));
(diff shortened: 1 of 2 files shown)


spec/helpers/monitoring.ts
```diff
@@ -331,11 +331,11 @@ class Monitoring {
         this.previewAction = function(groupIndex, itemIndex) {
             const item = this.getItem(groupIndex, itemIndex);
 
-            browser.wait(ECE.elementToBeClickable(item));
+            browser.wait(ECE.elementToBeClickable(item), 5000);
```
tomaskikutis, line 334: I think that if you don't pass the number to `brower.wait` it waits for what is specified as default timeout which is a few minutes if I remember correctly, so I don't think limiting it down to 5 seconds would help here.

tomaskikutis (approved):

spec/helpers/monitoring.ts
pablopunk: you're right, it was the `waitFor` below the one causing the timeout

PR #3575, opened 2020-08-25
Auto tagging

tomaskikutis:
SDNTB-642


package.json
```diff
@@ -120,7 +121,7 @@
     "sass-loader": "6.0.6",
     "shortid": "2.2.8",
     "style-loader": "0.20.2",
-    "superdesk-ui-framework": "2.2.6",
+    "superdesk-ui-framework": "github:superdesk/superdesk-ui-framework#master",
```
petrjasek, line 124: 2.2.11 is out now

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -0,0 +1,507 @@
...
+                            guid,
+                            language,
+                            headline,
+                            body_html,
+                        },
+                    },
+                }).then((res) => {
+                    const resClient = toClientFormat(res.analysis);
+
+                    this.setState({
```
petrjasek, line 103: it should check if component is still mounted before setting the state

petrjasek (approved):

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -0,0 +1,516 @@
...
+                httpRequestJsonLocal<{analysis: IServerResponse}>({
+                    method: 'POST',
+                    path: '/ai/',
+                    payload: {
+                        service: 'imatrics',
+                        item: {
+                            guid,
+                            language,
+                            headline,
+                            body_html,
```
petrjasek, line 99: can you add abstract here as well? or maybe just send the whole item. it's needed for SDNTB-667

tomaskikutis: Will add after server changes in order not to break the testing instance.

petrjasek: it's merged on the backend now

PR #3584, opened 2020-08-27
don't call setState after the component is unmounted

tomaskikutis:


petrjasek:
is it related to any issue? or merging just to devel is ok

tomaskikutis:
There isn't an issue. I just found a console error and fixed it. Merging just to develop should be fine.

petrjasek (approved):

PR #3588, opened 2020-08-31
Translate 'main' package group

pablopunk:
SDCP-291

### DIFF
--- a/scripts/apps/packaging/directives/AddPackageDropdown.ts
+++ b/scripts/apps/packaging/directives/AddPackageDropdown.ts
@@ -1,7 +1,8 @@
 import {AuthoringWorkspaceService} from 'apps/authoring/authoring/services/AuthoringWorkspaceService';
+import {IPackagesService} from 'types/Services/Packages';
 
 AddPackageDropdown.$inject = ['$rootScope', 'api', 'packages', 'authoringWorkspace'];
-export function AddPackageDropdown($rootScope, api, packages, authoringWorkspace: AuthoringWorkspaceService) {
+export function AddPackageDropdown($rootScope, api, packages: IPackagesService, authoringWorkspace: AuthoringWorkspaceService) {
     return {
         templateUrl: 'scripts/apps/packaging/views/sd-add-package-dropdown.html',
         link: function(scope) {
@@ -17,7 +18,7 @@ export function AddPackageDropdown($rootScope, api, packages, authoringWorkspace
             scope.groupList = scope.groupList || packages.groupList;
 
             scope.select = function(group) {
-                packages.addPackageGroupItem(group, scope.item);
+                packages.addPackageGroupItem(group, scope.item, null);
             };
         },
     };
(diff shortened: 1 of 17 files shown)


scripts/apps/packaging/views/sd-package-items-edit.html
```diff
@@ -2,7 +2,7 @@
     <ul class="package-edit" sd-sort-package-items>
         <li ng-repeat="group in list">
             <div class="group-info">
-                <h6>{{ group.id }}</h6>
+                <h6>{{ getPackageLabel(group.id) }}</h6>
```
tomaskikutis, line 5: It'd be good to avoid the separation of groups and their translations if we can. Could you instead use
`<h6>{{group.label}}</h6>`

and initialize the translation at the same place where the group is initialized?

scripts/apps/packaging/services/PackagesService.ts
```diff
@@ -1,5 +1,13 @@
 import _ from 'lodash';
 import {AuthoringWorkspaceService} from 'apps/authoring/authoring/services/AuthoringWorkspaceService';
+import {gettext} from 'core/utils';
+
+function getRefLabel(ref) {
```
tomaskikutis, line 5:
```
gettext(ref);
        ^ only static strings should be passed to gettext
```


Aside from that, the group object and label are still separate. It could be united by defining the main group like this:

`const mainGroup = {idRef: 'main', label: gettext('Main')};`

This way, there's no `default` switch condition and we are can be certain that the string is translated and that no dead translations will remain if objects are removed from the code.

pablopunk: oops I meant to put `gettext('main')`, my bad.

tomaskikutis (approved):

scripts/apps/packaging/directives/AddPackageDropdown.ts
```diff
@@ -1,7 +1,8 @@
 import {AuthoringWorkspaceService} from 'apps/authoring/authoring/services/AuthoringWorkspaceService';
+import {IPackagesService} from 'types/Services/Packages';
 
 AddPackageDropdown.$inject = ['$rootScope', 'api', 'packages', 'authoringWorkspace'];
-export function AddPackageDropdown($rootScope, api, packages, authoringWorkspace: AuthoringWorkspaceService) {
+export function AddPackageDropdown($rootScope, api, packages: IPackagesService, authoringWorkspace: AuthoringWorkspaceService) {
```
petrjasek, line 5: I think this one is too long still

pablopunk: I'll fix it. How can the tests pass on my local copy and not here? 🤔

pablopunk:
`npm test` passed already and the other tests passed in previous commits so I'll merge this one

scripts/apps/packaging/directives/AddPackageDropdown.ts
tomaskikutis: I think test instances run tests, not against your branch, but against target branch(develop in this case) + your branch merged. Because of this they can pass locally, but still fail on the test server. I'm not sure if it works this way with unit tests, but it does with e2e tests.

PR #3591, opened 2020-09-01
FIX: problem with locking when creating a related item from within parenting authoring item. [SDESK-5312]

GyanP:


tomaskikutis (approved):

scripts/apps/relations/directives/RelatedItemsDirective.ts
```diff
@@ -361,9 +360,10 @@ export function RelatedItemsDirective(
                 'item:lock',
                 'item:unlock',
                 'content:update',
-            ].map((eventName) =>
-                $rootScope.$on(eventName, onItemEvent),
-            );
+                'item:close',
+            ].map((eventName) => {
+                $rootScope.$on(eventName, onItemEvent);
```
petrjasek, line 365: just wondering wouldn't `scope.$on` work here?

GyanP: I am not too sure but I think you need to use a `$rootScope` instead of `$scope`. That's because since there is no parent-child relationship, each controller/directive will have its own scope and there won't be any inherited scope.

PR #3593, opened 2020-09-01
wait for editors to initialize before performing actions on the store

tomaskikutis:
SDANSA-425


pablopunk:
there's a test failing here, probably related to these changes

scripts/apps/authoring/macros/macros.ts
```diff
@@ -352,7 +352,16 @@ function MacrosReplaceDirective(editorResolver) {
                 return scope.diff[from] || null;
             }
 
-            init(scope.diff);
+            // There may be multiple instances of editors. Try waiting for all.
+            const initializeMacros = debounce(once(() => {
+                init(scope.diff);
```
petrjasek, line 357: not sure this is the case, but you might need to call `scope.apply();` here to make it re-render anything angular:
```ts
scope.apply(() => init(scope.diff));
```

petrjasek (approved):

PR #3598, opened 2020-09-04
Add America/Edmonton timezone to json

pablopunk:
SDCP-279


scripts/apps/dashboard/world-clock/timezones-all.json
```diff
@@ -1 +1,1979 @@
...
+            "-5:50:36 - LMT 1883_10_18_12_9_24 -5:50:36",
+            "-6 US C%sT 1920 -6",
+            "-6 Chicago C%sT 1936_2_1_2 -6",
+            "-5 - EST 1936_10_15_2 -5",
+            "-6 Chicago C%sT 1942 -6",
+            "-6 US C%sT 1946 -6",
+            "-6 Chicago C%sT 1967 -6",
+            "-6 US C%sT"
+        ],
+        "America/Edmonton": [],
```
petrjasek, line 84: does it actually work without any specs?

pablopunk: it does, I tried it with specs, without them, and with wrong specs. I think they're just useful if the actual timezone id doesn't exist on moment.js, and this one does exist

petrjasek: I see, cool

petrjasek (approved):

PR #3601, opened 2020-09-08
SDESK:5477 Update: use asset count api to load assets count

dev-tsoni:


scripts/extensions/sams/src/interfaces.ts
```diff
@@ -27,6 +27,7 @@ export interface ISetItem extends IBaseRestApiResponse {
     destination_name?: string;
     destination_config?: Dictionary<string, any>;
     destination?: IStorageDestinationItem;
+    count?: number;
```
MarkLark86, line 30: It would make it easier for a developer to understand resources from the API if the `I*Item` interfaces use the same schema as defined in the server (this could also cause issues when attempting to save these items, i.e. which attributes are for the browser and which ones are for the API).

In this case it would be better if the Set Asset counts were to be stored in the `Set` branch of the Redux store.
In this case you would need to:
In `store/sets/types.ts`
* Add a `Dictionary<string, number>` to `ISetState`
* Add a new action type to store the set counts

In `store/sets/actions.ts`
* A new thunk action that will call the `api.assets.getCount()`

In `store/sets/reducers.ts`
* Add a new `case` for storing these set counts into the state

In `store/sets/selectors.ts`
* Add a selector for these set counts.

Once the above is done, you should then be able to use the selector in `SetListItem`

scripts/extensions/sams/src/store/sets/actions.ts
```diff
@@ -60,9 +60,19 @@ export function loadSets(): IThunkAction<Array<ISetItem>> {
     return (dispatch, _getState, {api}) => {
         return api.sets.getAll()
             .then((sets: Array<ISetItem>) => {
-                dispatch(receiveSets(sets));
-
-                return Promise.resolve(sets);
+                let set_ids: string[] = sets.map((set) => {
+                    return set._id
+                })
+                return api.assets.getCount(set_ids)
```
MarkLark86, line 66: And instead of enhancing the `IAssetItem` here you could call the new `ThunkAction` to load and store the set counts counts every time `loadSets` is called

MarkLark86 (changes requested):

scripts/extensions/sams/src/components/sets/setListItem.tsx
```diff
@@ -58,7 +59,7 @@ export class SetListItem extends React.PureComponent<IProps> {
                 <ListItemBorder />
                 {set.state === SET_STATE.DRAFT ? null : (
                     <ListItemColumn>
-                        <Badge text="0" type="success" />
+                        <Badge  text = {count.toString()} type="success" />
```
MarkLark86, line 62: I'm not sure why our lint rules didn't pick this up, could you remove the spaces around the `=` in this assignment along with the extra space before it (just to keep a standard with the rest of the codebase)

MarkLark86:
There has also been some changes to the lint rules. So you'll need to:

- rebase from the `sams` branch
- re-run `npm run lint` locally
- fix any lint issues

PR #3603, opened 2020-09-09
[SDESK-5260] SAMS: Asset list and filter views

MarkLark86:


scripts/core/get-superdesk-api-implementation.tsx
```diff
@@ -115,8 +123,12 @@ export function getSuperdeskApiImplementation(
     return {
         dataApi: dataApi,
         dataApiByEntity,
+        elasticsearch: elasticsearchApi,
```
tomaskikutis, line 126: It would be very useful to remove the knowledge about elastic from the client and instead have an API with a generic query format for all endpoints that could work with any back-end whether it is mongo, elastic or if we add anything else.

I've created SDESK-4872 to address this after a discussion with Petr last December #3243.

Most of the time our queries are quite straight-forward and contain mostly of `=`, `<`, `>` and `contains` operators so it shouldn't be too hard to implement this query format.

We have to start somewhere, so I suggest first doing it for SAMS and then we could add it to other endpoints. I'd use a subset of mongo query format starting with [comparison](https://docs.mongodb.com/manual/reference/operator/query-comparison/) + AND/OR operators. That should probably do it for most part.

What do you think @MarkLark86 @petrjasek?

MarkLark86: This will definitely be a good improvement to add to the SAMS API itself, but at a later date. I do see the benefits in having a storage agnostic search query language, but we have time constraints at the moment.

scripts/core/helpers/url.ts
```diff
@@ -0,0 +1,103 @@
...
+    const hash = window.location.hash;
+    const index = hash.indexOf('?');
+
+    return index < 0 ?
+        hash.slice(1) :
+        hash.slice(1, index);
+}
+
+export function setUrlPage(page: string): Promise<void> {
+    return ng.getService('$location').then(($location) => {
```
pablopunk, line 14: is there any difference between using this and [ng.get](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/services/ng.ts#L27)? Apart from being async

scripts/extensions/sams/src/ui/SearchBar.tsx
```diff
@@ -1,25 +1,97 @@
...
+            value: this.props.initialValue ?? '',
+        };
+
+        this.handleKeydown = this.handleKeydown.bind(this);
+        this.handleChange = this.handleChange.bind(this);
+        this.clearValue = this.clearValue.bind(this);
+        this.handleSubmit = this.handleSubmit.bind(this);
+    }
+
+    htmlId = nextId();
```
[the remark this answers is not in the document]
MarkLark86: I would have thought so too. Their docs state:
`NOTE: Don't initialize htmlId in React lifecycle methods like render(). htmlId should stay the same during component lifetime.` (https://www.npmjs.com/package/react-id-generator#nextid)

And their example shows placing this attribute in the class level (the ui-framework itself defines this variable in the same way).

scripts/core/helpers/url.ts
MarkLark86: `getService` waits until the `$injector != null` before returning the service instance, where as `get` will throw an error if `$injector` is not defined. Considering this is changing the page, I didn't see any need to have this code run synchronously.

tomaskikutis: Before initializing the app/extensions, we wait for angular to load, so it can be safely assumed that all code called from within react components can get services synchronously and will not throw an error.

pablopunk (approved):

scripts/extensions/sams/src/ui/SearchBar.tsx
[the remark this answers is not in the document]
tomaskikutis: A constructor of a class isn't a React lifecycle method, so you can definitely put it there. The way it's defined in that library example is using syntax from ecmascript proposal for [class fields](https://github.com/tc39/proposal-class-fields). It's syntactic sugar and works for the most part the same as it was defined in the constructor.

I would like to avoid this syntax though in order to minimize the number of different patterns used across the codebase.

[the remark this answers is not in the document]
MarkLark86: This code is within the `ui` folder. It would be beneficial if we use the same patterns in this folder that the `UI-Framework` applies, so there is less changes they need to apply when porting this code across.

pablopunk (approved):

dev-tsoni (approved):

PR #3609, opened 2020-09-14
prevent http request when template is not executed

tomaskikutis:
SDESK-5528

`http://localhost:9000/%7B%7Bmedia.renditions.viewImage.href%7D%7D 404 (Not Found)`

The error used to be displayed even when `ng-if` wasn't resolving to true. I suspect that's because angular first renders the template as is and only after that populates placeholders with values and does other modifications to templates.

It wasn't causing any real error in the app, just initializing a faulty request.


petrjasek (approved):

pablopunk:
yeah I fixed a similar issue in the past, `ng-src` is the way to go

PR #3613, opened 2020-09-15
Activate save button when content changes in multi edit mode

tomaskikutis:
SDCP-364

### DIFF
--- a/scripts/apps/authoring/multiedit/multiedit.ts
+++ b/scripts/apps/authoring/multiedit/multiedit.ts
@@ -12,6 +12,7 @@ import {gettext} from 'core/utils';
 import {isPublished} from 'apps/archive/utils';
 import _ from 'lodash';
 import {AuthoringWorkspaceService} from '../authoring/services/AuthoringWorkspaceService';
+import {generatePatchIArticle} from 'core/helpers/CrudManager';
 
 MultieditService.$inject = ['storage', 'superdesk', 'authoringWorkspace', 'referrer', '$location'];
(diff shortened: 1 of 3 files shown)


scripts/apps/authoring/multiedit/multiedit.ts
```diff
@@ -189,7 +190,7 @@ function MultieditArticleDirective(authoring, content, multiEdit, lock, $timeout
             function openItem() {
                 authoring.open(scope.article).then((item) => {
                     scope.origItem = item;
-                    scope.item = _.create(item);
+                    scope.item = JSON.parse(JSON.stringify(item));
```
pablopunk, line 193: wouldn't `{...item}` work here?

tomaskikutis: It may or may not :) `{...item}` produces a shallow copy, so if nested objects got modified in the original, it would also be modified in the copy automatically, because they'd have the same reference
screenshot-2020-09-15--14-42

pablopunk: yeah I know about that, I was just asking cause I thought `item` was not being modified at all but I guess it is

pablopunk (approved):

scripts/apps/authoring/multiedit/multiedit.ts
tomaskikutis: The same object is used in article edit directive so I wouldn't be surprised if it was modified. I'm saving time and fixing it up front :smile:

PR #3614, opened 2020-09-16
Allow print previewing images

tomaskikutis:
SDANSA-421

### DIFF
--- a/scripts/apps/authoring/views/full-preview-item.html
+++ b/scripts/apps/authoring/views/full-preview-item.html
@@ -147,7 +147,7 @@
         </div>
 
         <div ng-if="!hideMedia && item.type === 'picture' && item.renditions.original.href" style="margin-bottom: 20px">
-            <img ng-src="{{item.renditions.original.href}}" />
+            <img ng-src="{{item.renditions.baseImage.href}}" />
         </div>
 
         <div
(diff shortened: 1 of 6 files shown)


scripts/apps/authoring/views/full-preview-item.html
```diff
@@ -146,6 +146,10 @@
             <p><i class="icon-picture" ng-if="hideMedia"></i>{{item.associations.featuremedia.description_text}}</p>
         </div>
 
+        <div ng-if="!hideMedia && item.type === 'picture' && item.renditions.original.href" style="margin-bottom: 20px">
+            <img ng-src="{{item.renditions.original.href}}" />
```
petrjasek, line 150: not sure we should use the original rendition, it's huge file usually for ansa, baseImage might be enough (maybe we can have it configurable)

petrjasek (approved):

scripts/apps/authoring/views/full-preview-item.html
```diff
@@ -147,7 +147,7 @@
         </div>
 
         <div ng-if="!hideMedia && item.type === 'picture' && item.renditions.original.href" style="margin-bottom: 20px">
```
petrjasek, line 149: would be good to change it here as well

tomaskikutis: Here's fix PR https://github.com/superdesk/superdesk-client-core/pull/3627

PR #3618, opened 2020-09-18
update test server timeouts

petrjasek:
prepopulate can take close to 10s so add some buffer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ addons:
     chrome: stable
 
 before_script:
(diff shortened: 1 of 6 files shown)


tomaskikutis (approved):

petrjasek:
that's actually to make it fail faster

tomaskikutis:
> that's actually to make it fail faster

That's not a bad idea, but I would allow longer timeouts - 20s instead of 2s for example. I would also be good then to lower the default timeout.

petrjasek:
yep we can try to change that default, will do some testing what works there, atm we have it set to 200s which seems like way to much

PR #3619, opened 2020-09-21
News preview. [SDANSA-286]

GyanP:


petrjasek:
imo we should rather use order from content profile and not hardcode ansa field ids in core

tomaskikutis (commented):
I agree with Petr's suggestion.

tomaskikutis (approved):

PR #3623, opened 2020-09-22
wrong search filter label for Wire (category) CV. [SDCP-368]

GyanP:


petrjasek:
I think the issue is there also for other fields like genre, imo it should work the same for all of them

scripts/apps/search/views/search-filters.html
```diff
@@ -63,7 +63,7 @@
         </ul>
     </div>
 
-    <div sd-toggle-box data-style="circle" data-title="{{:: 'Services' | translate }}" ng-hide="isEmpty('category')" data-open="true">
+    <div sd-toggle-box data-style="circle"  data-title="{{:: category.display_name | translate }}" ng-hide="isEmpty('category')" data-open="true">
```
tomaskikutis, line 66: Don't use `translate` on a dynamic value. See where that display name is getting set and translate there.

scripts/apps/search/directives/SearchFilters.ts
```diff
@@ -40,7 +41,10 @@ class LinkFunction {
         // fetch available languages
         metadata.initialize()
             .then(() => {
-                scope.category = metadata?.cvs?.find((cv) => cv._id === 'categories');
+                this.scope.filterLabels = {};
+                metadata?.cvs?.forEach((cv) => {
+                    this.scope.filterLabels[cv._id] = gettext(cv.display_name);
```
tomaskikutis, line 46: `gettext` won't work on dynamic user generated and editable values like vocabulary names. Vocabulary names aren't translatable yet. You should add an additional object if you want to handle translations.

I think it'd work to add translations object at the root of the vocabulary which would support multiple fields:
screenshot-2020-09-23--13-18

What do you think @petrjasek ?

petrjasek: yep that would work

tomaskikutis (approved):

PR #3624, opened 2020-09-23
[SDESK-5270] SAMS Authoring Attachments Widget

MarkLark86:


scripts/core/superdesk-api.d.ts
```diff
@@ -105,6 +105,7 @@ declare module 'superdesk-api' {
             monitoring?: {
                 getFilteringButtons?(deskId: string): Promise<Array<IMonitoringFilter>>;
             };
+            configurableUiComponents?: IConfigurableUiComponents;
```
tomaskikutis, line 108: The idea behind configurable UI components is that only one of the same type should be allowed per instance. That's why I'm registering them in [startApp](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/index.ts#L112) and not allowing extensions to provide them.

I've already implemented an API for side widgets and I suggest you use that(and disable the original one if the one from extensions is present). It's not merged to develop yet, but you can merge it to `sams` branch since it won't change much, except for styles.

tomaskikutis: Here's the PR https://github.com/superdesk/superdesk-client-core/pull/3575/files#diff-77f7ccd12c27cd27aa6bdedde8c6be34R80

MarkLark86: I have reverted to use the intended method of ConfigurableUiComponents and removed the ability for extensions to provide these.
Attachments can be used in two places, the sidebar widget and as a field in the editor. Because of this, I will continue to use the ConfigurableUiComponents functionality and let the instance provide he config for the intended component to use.

I have also removed the usage of `redux` for attachments, as it was only really used to store the list of files. The new HOC now provides the API via props

tomaskikutis: > Attachments can be used in two places, the sidebar widget and as a field in the editor.

You could also expose the same component to multiple extension points. I've done that with "mark for user" widget for example. There's also an [extension point for article fields](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L72) in the API.

MarkLark86: This field/widget replaces an existing one though, not adding a new one. Displaying and providing very similar experience to the one that's being replaced. ConfigurableUiComponents fits that description quite well.

scripts/extensions/sams/src/apps/samsAttachmentsWidget.tsx
```diff
@@ -0,0 +1,255 @@
+/* eslint-disable react/no-multi-comp */
```
pablopunk, line 1: we should consider putting this on [superdesk-client-code-style](https://github.com/superdesk/superdesk-client-code-style) cause I've seen it in other components

scripts/extensions/sams/package.json
```diff
@@ -16,7 +16,7 @@
     "@types/react": "16.9.44",
     "@types/redux-logger": "^3.0.8",
     "superdesk-code-style": "1.3.0",
-    "typescript": "3.7.2"
+    "typescript": "3.9.7"
```
pablopunk, line 19: I think we already have typescript@4 in client-core, we could put it here too

scripts/core/notify/notify.tsx
```diff
@@ -132,3 +133,20 @@ export default angular.module('superdesk.core.notify', ['superdesk.core.translat
         // eslint-disable-next-line
         return ReactDOM.render(<NotifyComponent />, targetEl); // returns instance of NotifyComponent
     }]);
+
+export const notify = {
+    info: (text: string, displayDuration?: number, options?: any) => {
+        ng.get('notify').info(text, displayDuration, options);
+    },
+    success: (text: string, displayDuration?: number, options?: any) => {
+        // eslint-disable-next-line angular/no-http-callback
```
pablopunk, line 142: why does this rule need to be disabled on `success` and `error` but not on `warning` and `info`?

scripts/extensions/sams/src/containers/FileUploadModal.tsx
```diff
@@ -87,6 +93,26 @@ export class FileUploadModal<T> extends React.Component<IProps<T>, IState> {
         }
     }
 
+    getInitialItems(): Array<IUploadItem> {
+        const items: Array<IUploadItem> = [];
+
+        if (this.props.initialFiles != null && this.props.initialFiles?.length > 0) {
```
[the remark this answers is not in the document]
MarkLark86: Lint was failing without the null check

scripts/extensions/sams/package.json
MarkLark86: The `develop` branch is not currently set to 4. I will update the `sams` branch once `develop` updates

scripts/core/notify/notify.tsx
MarkLark86: I think this is a bug with the linting, for some reason it saw this object as an angular http callback....

scripts/extensions/sams/package.json
pablopunk: ah that's right, it's on a PR, nevermind

scripts/extensions/sams/src/containers/FileUploadModal.tsx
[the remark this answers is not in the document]
pablopunk: hmm weird, what was the error?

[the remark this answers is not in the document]
tomaskikutis: It's not an error. `this.props.initialFiles?.length` evaluates to `number | undefined` and it's not valid to compare
`undefined > 0`. It could be solved by using a default -  `if ( (this.props.initialFiles?.length ?? 0) > 0 ) {...}`,
but inside the `if` statement, `forEach` is called on `initialFiles` which still hasn't been checked if it's null or not.

At run-time it's impossible for it to be null, because the length is checked, but the compiler doesn't understand number comparisons and `5 > 4` evaluates to `boolean`, not `true`;

It could be written in a shorter way if no comparison to a number was made:
```
if (this.props.initialFiles?.length) {
    this.props.initialFiles.forEach(...); // compiler understands that this.props.initialFiles is not null
}
```

pablopunk (approved):

PR #3626, opened 2020-09-24
SDESK-5479 Update: admin delete disabled set with no asset

dev-tsoni:


scripts/extensions/sams/src/components/sets/setListItem.tsx
```diff
@@ -78,7 +78,7 @@ export class SetListItem extends React.PureComponent<IProps> {
                     </ListItemRow>
                 </ListItemColumn>
                 <ListItemActionMenu row={true}>
-                    {this.props.deleteSet == null ? null : (
+                    {this.props.deleteSet == null || set.state === SET_STATE.DISABLED && count ? null : (
```
MarkLark86, line 81: I'm not sure checking the state here is required. If the `count > 0` then we should never allow deleting the Set regardless of it's state

scripts/extensions/sams/src/components/sets/setPreviewPanel.tsx
```diff
@@ -59,7 +61,7 @@ export function getSetPreviewPanel(superdesk: ISuperdesk) {
                 ariaValue: 'close',
             }];
 
-            if (set.state === SET_STATE.DRAFT) {
+            if (set.state === SET_STATE.DRAFT || set.state === SET_STATE.DISABLED && !count) {
```
MarkLark86, line 64: Same here

scripts/extensions/sams/src/components/sets/setListItem.tsx
dev-tsoni: Resolved

scripts/extensions/sams/src/components/sets/setPreviewPanel.tsx
dev-tsoni: Resolved

MarkLark86 (approved):

PR #3628, opened 2020-09-25
After rearranging gallery carousel items the next or previous buttons do not function as expected. [SDESK-5540]

GyanP:
When we rearrange the images, the carousel is not updated because the scope.$watch('currentIndex') is not called.

### DIFF
--- a/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
@@ -331,6 +331,10 @@ export function ItemCarouselDirective(notify) {
                         updated = true;
                     },
                 });
+
+                if (scope.currentIndex != null) {
+                    carousel?.trigger('to.owl.carousel', [scope.currentIndex]);
+                }
             }
 
             const removeAddImageEventListener = addInternalEventListener('addImage', (event) => {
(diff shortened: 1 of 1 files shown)


GyanP:
When we rearrange the images, the carousel is not updated because the scope.$watch('currentIndex') is not called.

scripts/apps/authoring/authoring/directives/ItemCarouselDirective.ts
```diff
@@ -331,6 +331,10 @@ export function ItemCarouselDirective(notify) {
                         updated = true;
                     },
                 });
+
+                if (scope.currentIndex != null) {
+                    carousel?.trigger('to.owl.carousel', [scope.currentIndex]);
```
petrjasek, line 336: seems like `carousel` will be always set here, so no need for `?`

tomaskikutis: I originally added it because it wasn't set when running unit tests and it was a lot of trouble to fix it properly.

petrjasek: yep I guess in that watch it was useful, but here it shouldn't be needed

tomaskikutis (approved):
Please check if this PR doesn't cause a regression for SDESK-5513

PR #3630, opened 2020-09-28
SDESK-5266 Add: Asset Preview Panel

dev-tsoni:


scripts/extensions/sams/src/components/assets/assetListPanel.tsx
```diff
@@ -35,24 +70,64 @@ export function getAssetListPanel(superdesk: ISuperdesk) {
                 );
             } else if (this.props.listStyle === ASSET_LIST_STYLE.GRID) {
                 return (
-                    <GridList className="sd-margin--1">
-                        {this.props.assets.map((asset) => (
-                            <AssetGridItem key={asset._id} asset={asset} />
-                        ))}
-                    </GridList>
+                    <PageLayout
```
MarkLark86, line 73: The `<PageLayout>` component is designed to be used higher up in the hierarchy of components. In this use case it is defined in the `SamsWorkspaceComponent`. This will also enable this component to focus solely on listing of the Assets.

scripts/extensions/sams/src/store/assets/actions.ts
```diff
@@ -44,6 +46,19 @@ export function setAssetListStyle(style: ASSET_LIST_STYLE): IAssetActionTypes {
...
 }
 
+export function previewAsset(asset: string): IAssetActionTypes {
+    return {
+        type: MANAGE_ASSETS_PREVIEW,
+        payload: asset,
+    };
+}
+
+export function onAssetPreviewPanelClosed(): IAssetActionTypes {
```
MarkLark86, line 56: This action name could cause confusion, as it seems this action reacts to the panel closing when in reality this action closes the panel. Have a look at the `Sets` store for example names so that we keep a common format across store branches.
https://github.com/superdesk/superdesk-client-core/tree/sams/scripts/extensions/sams/src/store/sets

scripts/extensions/sams/src/components/assets/assetPreviewPanel.tsx
```diff
@@ -0,0 +1,62 @@
...
+                                <Text>{asset.name}</Text>
+
+                                <FormLabel text={gettext('Description')} style="light" />
+                                <Text>{asset.description}</Text>
+
+                                <FormLabel text={gettext('Filename')} style="light" />
+                                <Text>{asset.filename}</Text>
+
+                                <FormLabel text={gettext('FileLength')} style="light" />
+                                <Text>{asset.length}</Text>
```
MarkLark86, line 45: There is also a utility function to convert bytes to a more human readable format:
https://github.com/superdesk/superdesk-client-core/blob/sams/scripts/extensions/sams/src/utils/ui.tsx#L35

scripts/extensions/sams/src/components/assets/assetListPanel.tsx
dev-tsoni: Have moved assetPreviewPanel to SamsWorkspaceComponent.

scripts/extensions/sams/src/store/assets/actions.ts
dev-tsoni: Resolved.

scripts/extensions/sams/src/components/assets/assetPreviewPanel.tsx
dev-tsoni: Resolved.

MarkLark86 (approved):

PR #3616, opened 2020-09-17
Fix copy-paste from MS Word on macOS

pablopunk:
SDESK-5530

There's an issue on draft-js that I [explain here](https://github.com/facebook/draft-js/commit/5081c872304384d3a213159bd8cc16e164bdf85a#r42395763) where copy & pasting from certain documents in Word won't paste spaces after links.

The actual fix is here https://github.com/pablopunk/draft-js/commit/df890ab9570dcb20a6e287116abdd9c247e510c2

PR #3631, opened 2020-09-29
cherry-pick #3616 on master (1.33)

pablopunk:
SDESK-5530


petrjasek:
would be good to use rather commit id than a branch so it's not changing later

petrjasek (approved):

PR #3637, opened 2020-10-01
feat(editor3): Remove all formatting button

pablopunk:
SDANSA-422

Pressing the button will convert all the text of the editor in plain
text (even removing line breaks).

### DIFF
--- a/scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
+++ b/scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
@@ -93,6 +93,7 @@ const EDITOR3_RICH_FORMATTING_OPTIONS = [
     'table',
     'formatting marks',
     'remove format',
+    'remove all format',
     'annotation',
     'comments',
     'suggestions',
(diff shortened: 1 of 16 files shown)


scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
```diff
@@ -93,6 +91,7 @@ const EDITOR3_RICH_FORMATTING_OPTIONS = [
     'table',
     'formatting marks',
     'remove format',
+    'remove all format',
```
tomaskikutis, line 94: It should probably be called "remove formatting", not "remove format".

tomaskikutis (approved):

scripts/apps/workspace/content/directives/ContentProfileSchemaEditor.ts
pablopunk: I thought I read it this way on the requirements but I can't see it anymore so I'll change it cause it makes more sense

pablopunk: you mean "remove all formatting" or change the old one too?

tomaskikutis: It'd be good to rename both since it's the same term.

pablopunk: I think that would remove the option from current instances (as this is an id). Also, I think I'm gonna put all these into a type so we can keep track of them in other files

tomaskikutis: You can keep the ID and only change naming in function/variable names.

PR #3638, opened 2020-10-02
SAMS: Improvements to APIs, Apps and Item Actions

MarkLark86:


petrjasek (approved):
looks good, just going forward I would rather not do forms via markup but define it in json and have some core api to render it

MarkLark86:
> looks good, just going forward I would rather not do forms via markup but define it in json and have some core api to render it

Sounds like a good idea. Once we have proven the usage of React component from the UI-Framework we could implement the new system using these components

tomaskikutis:
> looks good, just going forward I would rather not do forms via markup but define it in json and have some core api to render it

We have a `GenericListPageComponent` for it. It's used in [internal destinations](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/internal-destinations/InternalDestinations.tsx#L148), annotations library, the refactored content profiles PR and I expect to use it in more places to replace angular CRUD views and use it for new ones.

PR #3641, opened 2020-10-05
Add send to personal space in action menu. [SDANSA-424]

GyanP:


scripts/apps/ingest/services/SendService.ts
```diff
@@ -205,7 +205,7 @@ export function SendService(
                     notify.error(gettext('Failed to get item.'));
                     return item;
                 });
-        } else if (data.desk === 'PERSONAL_SPACE') {
+        } else if (data.desk === 'PERSONAL_SPACE' || action === 'send_to_personal') {
```
tomaskikutis, line 208: Since you are now checking if ` action === 'send_to_personal'`, do you still need the desk check? When is it relevant?

GyanP: yup, we need it there because whenever we use **send to**, the action which is called is `send to` not `send to personal` and there it check for the desk, if the desk is personal or not.

tomaskikutis (approved):

PR #3642, opened 2020-10-06
Improve authors dropdown performance

pablopunk:
SDESK-5595

This solution is not perfect but the performance is a bit better. Old dropdowns still need to be replaced with the new ones to prevent issues like this one.

* Lazy load avatars inside scroll elements
* Removed vs-repeat from loop: Recalculating DOM styles takes a huge amount of time but this seems to help.

### Before

bad blurred gif

### After

good blurred gif

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
     "react-bootstrap": "0.32.1",
     "react-debounce-input": "3.2.0",
     "react-dom": "16.9.0",
+    "react-lazyload": "^3.0.0",
     "react-mentions": "1.2.2",
     "react-paginate": "6.3.0",
     "react-portal": "4.1.3",
--- a/package.json
+++ b/package.json
@@ -171,7 +171,7 @@
     "start-client-server": "http-server dist -p 9000 -s &",
     "stop-client-server": "fuser -k 9000/tcp",
     "protractor": "protractor protractor.conf.js",
-    "webdriver-manager": "webdriver-manager update --gecko false --standalone false --versions.chrome=84.0.4147.30",
+    "webdriver-manager": "webdriver-manager update --gecko false --standalone false --versions.chrome=86.0.4240.75",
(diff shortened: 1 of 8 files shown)


tomaskikutis:
What about limiting items in the list to 10 or 20 instead? It would make it more like autocomplete instead of browsing everything. How many items did they have?

I don't like the modification of the avatar component, because it's not really a problem until you don't load hundreds of them which is the main issue that should be addressed.

pablopunk:
@tomaskikutis I tried that first but I realized the search didn't work so I obviously thought they weren't using it at all. Now I took another look and it does work on their instance. Seems like in newer versions we don't allow to search anymore for parent items in the dropdown (users are parents of their roles in this dropdown tree) if the "disable category selection" is checked. I'm gonna try to fix this, but they still want to see all users at once in the dropdown (obviously it's gonna be slow until they get rid of those 2k users), but it'll be fine for all other instances.

About the avatars. I think it's a good idea to lazy load all images that could appear in scroll containers to avoid making multiple requests that could potentially be pointless. Also, it's optional in the component.

tomaskikutis:
> obviously it's gonna be slow until they get rid of those 2k users

Not necessarily. It would perform well with any amount of items if we used virtual scrolling/pagination. Lazy loading images helps, but it's more of a work-around and not a stable long-term solution. Imagine avatar component made HTTP requests in order to display online/offline status or to display a number of assigned items. We can't lazy load that - and shouldn't attempt to. Virtual scrolling/pagination removes the problem entirely.

pablopunk:
@tomaskikutis yeah I agree about the scrolling, that's why I said we need to refactor this with the new components, the code is a mess right now to include more features.

tomaskikutis (approved):

pablopunk:
There are 3 tests failing on develop so I'll merge this one

PR #3643, opened 2020-10-06
Use the refactored list in spike view

tomaskikutis:
SDESK-5237

REQUIRES: SDESK-5554


scripts/apps/monitoring/directives/ArticlesListReact.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {openArticle} from 'core/get-superdesk-api-implementation';
+import {ISuperdeskQuery} from 'core/query-formatting';
+import {dataApi} from 'core/helpers/CrudManager';
+import ng from 'core/services/ng';
+import {IArticle, IDesk, IUser} from 'superdesk-api';
+import {ArticlesListByQueryWithFilters} from 'core/ArticlesListByQueryWithFilters';
+
+interface IProps {
+    heading: string;
+    getQuery(userId: string, deskId: string): ISuperdeskQuery;
```
petrjasek, line 13: not sure about those params here, imo the component shouldn't be passing anything to `getQuery` method but rather the parent component which knows what is needed there.

scripts/core/ArticlesListByQuery.tsx
```diff
@@ -0,0 +1,205 @@
...
+                            return _items;
+                        })}
+                        shouldReloadTheList={(changedFields) => {
+                            /** TODO: Have websockets transmit the diff.
+                             * The component should not update when field value changes do not affect the query -
+                             * for example, if the query is {desk: 'X'} and an update is about an item moved
+                             * from desk Y to Z.
+                             */
+
+                            const queryFields = getQueryFieldsRecursive(this.props.query.filter);
```
petrjasek, line 119: this doesn't handle full text search

scripts/core/ArticlesListV2.tsx
```diff
@@ -0,0 +1,164 @@
...
+                                if (
+                                    extra.resource === 'archive'
+                                    || extra.resource === 'archive_unspike'
+                                ) {
+                                    const reloadTheList = this.props?.shouldReloadTheList(
+                                        new Set(Array.from(Object.keys(extra.fields))),
+                                    ) ?? false;
+
+                                    if (reloadTheList) {
+                                        this.lazyLoaderRef.reset();
```
petrjasek, line 109: would be good to do some throttling, there are some multi actions updating many items at the same time

scripts/core/ArticlesListByQuery.tsx
```diff
@@ -0,0 +1,205 @@
...
+                this.setState({itemCount: res._meta.total}, () => {
+                    resolve(res);
+                });
+            });
+        });
+    }
+    componentDidMount() {
+        this.loadItems(0, 1).then((res) => {
+            this.setState({
+                itemCount: res._meta.total,
```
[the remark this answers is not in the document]
tomaskikutis: It's only about getting the count. It was required for a virtual scrolling library so it can compute the correct height of the scroll-box. I ended up not using that library due to performance issues of our list item component, but I'll be doing more work on it so I'll keep that logic for now.

I think I've tried to call it with `0, 0` but server threw an error.

scripts/core/ArticlesListV2.tsx
tomaskikutis: Good point. I've added it.

petrjasek (approved):

PR #3648, opened 2020-10-13
Sdesk 5267 dowload one or more assets

dev-tsoni:


scripts/extensions/sams/src/components/assets/assetListItem.tsx
```diff
@@ -46,16 +49,26 @@ export class AssetListItem extends React.PureComponent<IProps> {
...
         const {gettext, longFormatDateTime} = superdeskApi.localization;
         const actions = getDropdownItemsForActions(this.props.asset, this.props.actions);
 
         return (
-            <ListItem onClick={this.onItemClick} selected={this.props.selected} shadow={1}>
+            <ListItem onClick={this.onItemClick} selected={this.props.selected || this.props.itemSelected} shadow={1}>
                 <ListItemBorder />
                 <ListItemColumn>
                     <Icon name={getIconTypeFromMimetype(this.props.asset.mimetype)} />
+                    <div className="sd-grid-item__checkbox" onClick={this.onCheckboxClick}>
```
MarkLark86, line 66: There is a specific set of classes to use with the `ListItem` checkboxes. See the Planning playground for examples
https://ui-framework.superdesk.org/#/playgrounds/planning

scripts/extensions/sams/src/components/assets/assetPreviewPanel.tsx
```diff
@@ -33,6 +34,34 @@ export class AssetPreviewPanel extends React.PureComponent<IProps> {
...
+                    <div className="side-panel__content-block side-panel__content-block--flex">
+                        <div className="side-panel__content-block-inner side-panel__content-block-inner--grow" />
+                        <div className="side-panel__content-block-inner side-panel__content-block-inner--right">
+                            <Dropdown
+                                align = "right"
+                                append = {true}
+                                items={[
+                                    {
+                                        type: 'group',
+                                        label: 'Actions',
```
MarkLark86, line 46: Use `gettext` to flag translation

scripts/extensions/sams/src/components/assets/assetPreviewPanel.tsx
```diff
@@ -33,6 +34,34 @@ export class AssetPreviewPanel extends React.PureComponent<IProps> {
...
+                                        items: [
+                                            'divider',
+                                            {
+                                                label: 'Download', icon: 'download',
```
MarkLark86, line 50: Use `gettext` to flag translation on the label

scripts/extensions/sams/src/components/workspaceSubnav.tsx
```diff
@@ -179,86 +189,147 @@ export class WorkspaceSubnav extends React.PureComponent<IProps> {
...
-                        <Dropdown items={items}>
+        if (this.selectedAssets()) {
+            return (
+                <React.Fragment>
+                    <SubNav zIndex={2}>
+                        <div className="multi-action-bar">
+                            <button className="toggle" onClick={this.props.closeMultiActionBar}>
+                                <i className="icon-chevron-up-thin" />
+                            </button>
+                            <button className="btn" onClick={this.props.closeMultiActionBar}>cancel</button>
```
MarkLark86, line 200: Use `gettext` to flag translation

scripts/extensions/sams/src/components/workspaceSubnav.tsx
```diff
@@ -179,86 +189,147 @@ export class WorkspaceSubnav extends React.PureComponent<IProps> {
...
+                            <span id="multi-select-count">{this.props.selectedAssetIds?.length} Item selected</span>
```
MarkLark86, line 201: Use `gettextPlural` to flag translation. Example `gettextPlural` usage:
https://github.com/superdesk/superdesk-client-core/blob/sams/scripts/extensions/sams/src/containers/FileUploadModal.tsx#L202

MarkLark86 (changes requested):

scripts/extensions/sams/src/components/assets/assetGridItem.tsx
```diff
@@ -66,13 +76,19 @@ export class AssetGridItem extends React.PureComponent<IProps> {
...
                 onClick={this.onItemClick}
-                selected={this.props.selected}
+                selected={this.props.selected || this.props.itemSelected}
             >
                 <GridItemThumb
                     uploading={true}
                     remove={this.props.remove && this.onRemove}
                     icon={typeIcon}
                 >
+                    <div className="sd-grid-item__checkbox" onClick={this.onCheckboxClick}>
```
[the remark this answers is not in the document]
dev-tsoni: I am leaving the checkbox for now, when your attachment PR will get merged, i'll update my PR.

scripts/extensions/sams/src/components/assets/assetListItem.tsx
dev-tsoni: I used the first example in Planning playground.

MarkLark86: By design the UI Framework uses BEM for naming convention and to enforce usage. In this case, classes starting with `sd-grid-item__` should only be used with a parent of `sd-grid-item`.
https://ui-framework.superdesk.org/#/components/list-item shows how to use `sd-list-item` and at the bottom of that page you will see an example of the checkbox on a list item

scripts/core/superdesk-api.d.ts
```diff
@@ -963,6 +963,8 @@ declare module 'superdesk-api' {
             formatFiltersForServer?: (filters: ICrudManagerFilters) => ICrudManagerFilters,
         ): Promise<IRestApiResponse<T>>;
         queryRaw<T>(endpoint, params?: Dictionary<string, any>): Promise<T>;
+        queryZip<T>(endpoint, params?: Dictionary<string, any>): Promise<void>;
+        queryAssetBinary<T>(endpoint): any;
```
MarkLark86, line 967: This function doesn't return anything, so `void` would be better than `any`

tomaskikutis: What do these new methods do?

dev-tsoni: queryZip is used to request compressed binary zip for multiple assets and download the returned compressed binary as zip. 
And queryAssetBinary is used to open a single asset into a new window.
Can not use queryRaw as response from server is Binary and not Json.

tomaskikutis: I see. It should not be in this part of the API though. If I'm not mistaken `queryZip` does 2 things - queries using `httpRequestBase` to receive a blob and downloads that blob to a browser.

I suggest to rename `queryRaw` to `queryRawJson` and create `queryRaw` that would be calling `httpRequestBase` in order to return the response that hasn't been converted to JSON.

After that, you could add a method to helpers or utilities selection `download(blob, applicationType): void` that would do the downloading separately.

`queryAssetBinary` doesn't query anything and doesn't even initiate an AJAX request. I would drop it entirely since it's basically a wrapper for `window.open`. You can get base server url any time from config that is [exposed to extensions](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L996).

dev-tsoni: I agree with you. I'll have a discussion with Mark about the changes tomorrow and update you accordingly.

MarkLark86 (approved):

PR #3653, opened 2020-10-14
Handle double-clicks or repeated clicking - avoid performing the action multiple times [SDESK-5597]

GyanP:


petrjasek:
would be good to do the same for send&publish

images/three-dots-white.svg
```diff
@@ -0,0 +1,32 @@
+<svg width="120" height="30" viewBox="0 0 120 30" xmlns="http://www.w3.org/2000/svg" fill="#FFFFFF">
```
tomaskikutis, line 1: Where did you get this SVG from? I don't think you need it, there's a `sd-loader` CSS class available.

styles/sass/layouts.scss
```diff
@@ -1548,4 +1548,15 @@ body.prepare-print {
             border-bottom: 0;
         }
     }
-}
\ No newline at end of file
+}
+
+.loading-overlay-for-publish {
```
tomaskikutis, line 1553: Try making things reusable and avoid unnecessary coupling. In this case, it's just a button that displays a spinner when in loading state. It isn't related to publishing except that you are using it there. Including publishing in the name makes it seem that it only works there, which shouldn't be the case.

images/three-dots-white.svg
GyanP: > Where did you get this SVG from?

I got it from [here](https://github.com/superdesk/superdesk-client-core/blob/develop/images/three-dots.svg). I just change its color to white. 

ezgif com-gif-maker
There are some issues because of which I didn't use the sd-loader class.

-  It shows the background color while hovering on it.
- On Hovering the width of the background color is overflowed.

So If I made any changes in the properties of sd-loader, It will reflect everywhere that's why I made these changes.

tomaskikutis (approved):

PR #3654, opened 2020-10-16
SDBELGA-423 Store search panel show/hide state in the user preferences.

ride90:
related to https://github.com/superdesk/superdesk-core/pull/2019

### DIFF
--- a/scripts/apps/search/directives/SearchContainer.ts
+++ b/scripts/apps/search/directives/SearchContainer.ts
@@ -1,15 +1,14 @@
 import {isEmpty, omit} from 'lodash';
 import {gettext} from 'core/utils';
 
-
 export function SearchContainer() {
     const FILTERS_PANEL_PREFERENCES_KEY = 'search:filters_panel_open';
     var filtersPanelPreferences = {
         [FILTERS_PANEL_PREFERENCES_KEY]: {
             'type': 'bool',
-            'default': true
-        }
-    }
+            'default': true,
+        },
+    };
 
     return {
         controller: ['$scope', '$location', 'pageTitle', 'preferencesService',
--- a/scripts/apps/search/directives/SearchContainer.ts
+++ b/scripts/apps/search/directives/SearchContainer.ts
@@ -1,16 +1,24 @@
 import {isEmpty, omit} from 'lodash';
 import {gettext} from 'core/utils';
 
+
 export function SearchContainer() {
+    const FILTERS_PANEL_PREFERENCES_KEY = 'search:filters_panel';
+    var filtersPanelPreferences = {
+        [FILTERS_PANEL_PREFERENCES_KEY]: {
+            'type': 'bool',
+            'default': true
+        }
+    }
+
     return {
-        controller: ['$scope', '$location', 'pageTitle',
-            function SearchContainerController($scope, $location, pageTitle) {
+        controller: ['$scope', '$location', 'pageTitle', 'preferencesService',
+            function SearchContainerController($scope, $location, pageTitle, preferencesService) {
                 const query = omit($location.search(), '_id', 'repo');
 
                 this.flags = $scope.flags || {};
                 this.flags.facets = !isEmpty(query);
                 pageTitle.setUrl(gettext('Search'));
-
                 $scope.$watch('search.activeProvider', (activeProvider) => {
                     if (activeProvider) {
                         if (activeProvider.advanced_search !== undefined) {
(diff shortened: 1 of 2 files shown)


scripts/apps/search/directives/SearchContainer.ts
```diff
@@ -28,6 +36,18 @@ export function SearchContainer() {
                         }
                     }
                 });
+                preferencesService.get().then((result) => {
+                    if (result !== undefined && FILTERS_PANEL_PREFERENCES_KEY in result) {
```
pablopunk, line 40: if result is null then `null !== undefined` will be `true`. We use this convention to check bot null and undefined: `(result != null)`

scripts/apps/search/directives/SearchContainer.ts
```diff
@@ -1,16 +1,24 @@
 import {isEmpty, omit} from 'lodash';
 import {gettext} from 'core/utils';
 
+
 export function SearchContainer() {
+    const FILTERS_PANEL_PREFERENCES_KEY = 'search:filters_panel';
```
tomaskikutis, line 6: Make preference name more exact - `search:filters_panel_open`

scripts/apps/search/directives/SearchContainer.ts
```diff
@@ -28,6 +36,18 @@ export function SearchContainer() {
...
+                    if (result !== undefined && FILTERS_PANEL_PREFERENCES_KEY in result) {
+                        this.flags.facets = result[FILTERS_PANEL_PREFERENCES_KEY]['enabled'];
+                    } else {
+                        this.flags.facets = filtersPanelPreferences[FILTERS_PANEL_PREFERENCES_KEY]['default'];
+                    }
+                });
+                $scope.toggleFiltersPane = () => {
+                    filtersPanelPreferences[FILTERS_PANEL_PREFERENCES_KEY]['enabled'] = !$scope.flags.facets;
+                    preferencesService.update(filtersPanelPreferences);
+                    $scope.flags.facets = !$scope.flags.facets;
```
pablopunk, line 49: it looks like we're saving the boolean value in two places

1. `filtersPanelPreferences[FILTERS_PANEL_PREFERENCES_KEY]['enabled']`
2. `$scope.flags.facets`

is this necessary?

pablopunk (changes requested):

scripts/apps/search/directives/SearchContainer.ts
```diff
@@ -28,6 +36,18 @@ export function SearchContainer() {
                         }
                     }
                 });
+                preferencesService.get().then((result) => {
+                    if (result !== undefined && FILTERS_PANEL_PREFERENCES_KEY in result) {
+                        this.flags.facets = result[FILTERS_PANEL_PREFERENCES_KEY]['enabled'];
+                    } else {
+                        this.flags.facets = filtersPanelPreferences[FILTERS_PANEL_PREFERENCES_KEY]['default'];
+                    }
```
tomaskikutis, line 44: I think it'd be simpler if the preference was a boolean instead of an object.

`this.flags.facets = result[FILTERS_PANEL_PREFERENCES_KEY] ?? true`

scripts/apps/search/directives/SearchContainer.ts
ride90: done

scripts/apps/search/directives/SearchContainer.ts
ride90: yes, one is for scope another is for "payload". I've upadted that part a little bit.

scripts/apps/search/directives/SearchContainer.ts
ride90: done

scripts/apps/search/directives/SearchContainer.ts
ride90: In this case, we can not a default value, but the main reason is to keep it consistent with other boolean preferences definitions e.q
```
superdesk.register_default_user_preference('spellchecker:status', {
        'type': 'bool',
        'enabled': True,
        'default': True
    })
```

ride90: I think we can change it to be just a boolean, we don't really use `default` value defined on the server side, and I believe that `'type': 'bool'` is only for UI, but this setting is hidden from UI, so we are fine.

ride90: I've tried, but server-side does not support a privilege to be just a bool.

tomaskikutis: You could register a general preference for `global_search`
`superdesk.register_default_user_preference('global_search', {})`
and then set `filtering_open: true` inside that object.

ride90: I've got your point, that makes sense but currently, it's designed in a different way.
Instead of having `global_search. filtering_open` object we have `search:filters_panel_open` preference key. 
With `superdesk.register_default_user_preference('global_search', {})` approach, for example, would be difficult to hide some `global_search` configs from the UI and display the rest. Also, I'm not sure how to extend it in the future.

With current notation, you can also define `category`, `privileges`, and `label`  which are used in the UI.
For example, we can have the next preferences:

```
superdesk.register_default_user_preference('search:filters_panel_open', {
        'type': 'bool',
        'enabled': True,
        'default': True
    })
superdesk.register_default_user_preference('search:some_ui_config', {
        'type': 'bool',
        'enabled': False,
        'default': False,
        'label': _('Enable some feature'),
        'category': _('Search'),
        'privileges': ['some_privelege']
    })
superdesk.register_default_user_preference('search:another_ui_config', {
        'type': 'string',
        ..
        'label': _('Another search feature'),
        'category': _('Search')
    })
```
in this case, the second and third preferences are grouped (category) and displayed in UI, first one is hidden.

tomaskikutis: Ok, let's keep it then

tomaskikutis (approved):

pablopunk (approved):

PR #3656, opened 2020-10-20
Field is not present inside vocabularies.

GyanP:
The field is not present inside vocabularies but the ingest repo uses the default scheme (because before the item is fetched, it doesn't know what profile to use) and since the moment, the item has the default field although it may not be visible to the users in the UI.
Superdesk-Search-Tab

### DIFF
--- a/scripts/apps/search/directives/SearchFilters.ts
+++ b/scripts/apps/search/directives/SearchFilters.ts
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import moment from 'moment';
 import {getDateFilters} from './DateFilters';
+import {gettext} from 'core/utils';
 
 class LinkFunction {
     scope: any;
@@ -10,6 +11,7 @@ class LinkFunction {
     desks: any;
     aggregationsMapper: any;
     session: any;
+    defaultFilterLabels: any;
 
     constructor(desks, tags, $location, scope, elem, metadata, session) {
         this.scope = scope;
--- a/scripts/apps/search/directives/SearchFilters.ts
+++ b/scripts/apps/search/directives/SearchFilters.ts
@@ -38,6 +38,12 @@ class LinkFunction {
         };
         this.scope.dateFilters = getDateFilters()
             .filter((dateFilter) => metadata.search_config?.[dateFilter.fieldname] == null);
+        this.defaultFilterLabels = {
+            categories: gettext('CATEGORY'),
+            genre: gettext('GENRE'),
+            priority: gettext('PRIORITY'),
+            urgency: gettext('URGENCY'),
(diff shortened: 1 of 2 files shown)


scripts/apps/search/views/search-filters.html
```diff
@@ -63,7 +63,7 @@
         </ul>
     </div>
 
-    <div sd-toggle-box data-style="circle"  data-title="{{:: filterLabels.categories }}" ng-hide="isEmpty('category')" data-open="true">
+    <div sd-toggle-box data-style="circle"  data-title="{{:: filterLabels.categories || 'CATEGORIES' }}" ng-hide="isEmpty('category')" data-open="true">
```
tomaskikutis, line 66: Can you make an if/else branch in [SearchFilters.ts](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/search/directives/SearchFilters.ts#L56) add create defaults there? Don't forget the translations.

tomaskikutis:
> The field is not present inside vocabularies but the ingest repo uses the default scheme

What field are you referring to? Did you mean content profile instead of default scheme?

scripts/apps/search/directives/SearchFilters.ts
```diff
@@ -38,6 +38,12 @@ class LinkFunction {
         };
         this.scope.dateFilters = getDateFilters()
             .filter((dateFilter) => metadata.search_config?.[dateFilter.fieldname] == null);
+        this.defaultFilterLabels = {
+            categories: gettext('CATEGORY'),
```
tomaskikutis, line 42: Can you pass `Category`(and other labels) sentence-cased to `gettext` and uppercase using CSS? By uppercasing using CSS we can reduce the number duplicate of translations in transifex.

scripts/apps/search/directives/SearchFilters.ts
```diff
@@ -56,6 +62,13 @@ class LinkFunction {
                 metadata?.cvs?.forEach((cv) => {
                     this.scope.filterLabels[cv._id] = getLocaleName(cv, this.session.identity.language);
                 });
+                Object.keys(this.defaultFilterLabels).forEach((key) => {
```
tomaskikutis, line 65: Adding a variable would make it more readable

```
Object.keys(this.defaultFilterLabels).forEach((key) => {
    const hasKey = Object.keys(this.scope.filterLabels).includes(key);

    if (!hasKey || (hasKey && this.scope.filterLabels[key] == null)) {
        this.scope.filterLabels[key] = this.defaultFilterLabels[key];
    }
});
```

tomaskikutis (approved):

PR #3658, opened 2020-10-20
Make it easier to run end-to-end tests.

tomaskikutis:
* Browser drivers are installed automatically if needed when running protractor.
* Browser version is used in order to download the correct drivers.

### DIFF
--- a/run-end-to-end-tests.js
+++ b/run-end-to-end-tests.js
@@ -1,0 +1,63 @@
+const execSync = require('child_process').execSync;
+
+function ensurePackageInstalled() {
+    return new Promise((resolve, reject) => {
+        try {
+            require.resolve('webdriver-manager');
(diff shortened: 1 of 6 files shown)


pablopunk (approved):

run-end-to-end-tests.js
```diff
@@ -54,7 +50,10 @@ function installWebdriverDriver() {
 ensurePackageInstalled()
     .then(installWebdriverDriver)
     .then(() => {
-        const argumentsToForward = process.argv.reduce((acc, item, index) => index < 2 ? acc : acc.concat(item), []).join(' ');
+        const argumentsToForward = process.argv.reduce(
```
petrjasek, line 53: imo it would be more readable as `process.argv.slice(2).join(' ');`

PR #3660, opened 2020-10-21
Show countries dropdown as multi selection dropdown

pablopunk:
SDBELGA-422

Multi selection dropdown allows searching through terms, which is also
useful for single selection fields like this one.


tomaskikutis (approved):
It's not the nicest solution, but will do for now since we'll be deleting this code anyway when moving to react.

pablopunk:
@tomaskikutis yep, that's exactly what I told Karel... These components are probably on the top priorities to refactor, they're used a lot and they're old and big/complex

PR #3662, opened 2020-10-22
Send params for publish from personal

GyanP:


scripts/apps/authoring/authoring/services/AuthoringService.ts
```diff
@@ -497,7 +497,8 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
             helpers.filterDefaultValues(diff, origItem);
 
             if (_.size(diff) > 0) {
-                return api.save('archive', origItem, diff).then((__item) => {
+                return api.save('archive', origItem, diff, {},
+                    {publish_from_personal: appConfig?.features?.publishFromPersonal}).then((__item) => {
```
tomaskikutis, line 501: We should avoid having undefined keys in object. Either add a default using `appConfig?.features?.publishFromPersonal ?? false` or only add the key if it's true.

```
if (appConfig?.features?.publishFromPersonal === true) {
    params['publish_from_personal'] = true;
}
```

tomaskikutis (approved):

scripts/apps/authoring/authoring/services/AuthoringService.ts
petrjasek, line 501: wouldn't this actually work even if it's set to `false` ?

GyanP: I think It would work.

tomaskikutis (approved):

PR #3669, opened 2020-10-28
feat(editor3): UI options for characters limit

pablopunk:
SDANSA-447

Needs https://github.com/superdesk/superdesk-core/pull/2034


scripts/apps/authoring/authoring/components/CharacterCountConfigButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+                }}
+            >
+                <i className="icon-settings" />
+            </button>
+        );
+    }
+}
+
+function CharacterCountConfigModal({closeModal, onChange, value}) {
+    const [radioValue, radioValueSet] = React.useState<
```
tomaskikutis, line 102: Still trying to sneak hooks in ;D My bad, I didn't follow through with a lint rule :smile:

scripts/apps/authoring/authoring/components/CharacterCount.tsx
```diff
@@ -0,0 +1,42 @@
...
+    input = html ? cleanHtml(input) : input;
+    input = input.replace(/\r?\n|\r/g, '');
+
+    return input.length;
+}
+
+export function CharacterCount(props: IProps) {
+    const {limit} = props;
+    const numChars = getEditorFieldCharactersCount(props.item, props.html);
+    const highlightLimit = limit && numChars >= limit;
```
tomaskikutis, line 24: Since you are checking if limit is nullish, it should probably be nullable in the interface?

scripts/apps/authoring/authoring/components/CharacterCount.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import classNames from 'classnames';
+import {gettext} from 'core/utils';
+import {cleanHtml} from '../helpers';
+
+interface IProps {
+    limit: number;
+    html: boolean;
+    item: string;
```
tomaskikutis, line 9: I assume `item` is plain text version of `html` and only one can be passed? If so, it'd be better to make it more explicit by defining a more precise type.

```
interface IProps {
    limit: number;
    value: {text: string} | {html: string};
}
```

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -428,6 +433,12 @@ export class Editor3Component extends React.Component<IProps, IState> {
             return 'handled';
         }
 
+        if (this.props.limit && this.props.limitBehavior === 'limit') {
+            if (preventInputWhenLimitIsPassed(editorState, chars, this.props.limit)) {
```
tomaskikutis, line 437: The limit should apply to all types changes to editor state, not only to keystrokes. For example, it doesn't apply the limit when pasting.

scripts/core/editor3/helpers/characters-limit.ts
```diff
@@ -0,0 +1,178 @@
...
+    const endSelectedTextLength = currentSelection.getEndOffset();
+    const keyAfterEnd = currentContent.getKeyAfter(endKey);
+
+    if (blocksAreTheSame) {
+        length +=
+            currentSelection.getEndOffset() - currentSelection.getStartOffset();
+    } else {
+        let currentKey = startKey;
+
+        while (currentKey && currentKey !== keyAfterEnd) {
```
tomaskikutis, line 52: It'd be good to use a for loop instead to avoid infinite loop & browser freeze.

scripts/core/editor3/store/index.ts
```diff
@@ -242,10 +298,14 @@ export function getInitialContent(props) {
...
     if (props.editorState != null) {
         var contentState = convertFromRaw(
-            (props.editorState instanceof Array) ? props.editorState[0] : props.editorState,
+            props.editorState instanceof Array
+                ? props.editorState[0]
+                : props.editorState,
         );
 
-        return initializeHighlights(EditorState.createWithContent(contentState)).getCurrentContent();
+        return initializeHighlights(
```
tomaskikutis, line 306: It'd be good not to reformat the code that doesn't violate our code style configuration.

scripts/apps/authoring/authoring/components/CharacterCountConfigButton.tsx
```diff
@@ -0,0 +1,150 @@
...
+            <ModalHeader>{gettext('Character limit settings')}</ModalHeader>
+            <ModalBody>
+                <p>
+                    {gettext(
+                        'You can either completely block further writing after the character' +
+                            'limit is reached or highlight exceeding characters in red.',
+                    )}
+                </p>
+                <CheckGroup orientation="vertical">
+                    <Radio<CharacterLimitUiBehavior>
```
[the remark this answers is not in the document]
pablopunk: If you can't type more characters after the limit, then you wouldn't have anything to highlight in red 🤔 

About the syntax, it does look a bit weird but it's valid. It's just declaring the type of the value; if you see the next lines they're not children, but attributes. It's like doing this:

```jsx
const RadioComponent = Radio<CharacterLimitUiBehavior>

return <RadioComponent value={value} ... />
```

I can do something similar if you think it's more readable

scripts/apps/authoring/authoring/components/CharacterCountConfigButton.tsx
pablopunk: hahaha I swear I didn't mean to sneak it, I'm just used to it. Anyway they're pretty straightforward and they're becoming the standard in React, we can talk about it with other devs if you want

scripts/core/editor3/components/Editor3Component.tsx
pablopunk: True, right now if you paste some text that overflows the field it does highlight it in red (instead of chopping it). I thought this was more user friendly, otherwise you would lose the rest of the text when pasting, not sure if that's a good UX

scripts/core/editor3/helpers/characters-limit.ts
pablopunk: Agree, I'm gonna commit a simpler way I found to count selection characters

scripts/core/editor3/store/index.ts
pablopunk: yeah I don't know why I pushed this, I used an auto-formatter for my new code and I probably included this in

scripts/apps/authoring/authoring/components/CharacterCountConfigButton.tsx
tomaskikutis: I don't think there is a standard here, just two APIs - hooks or `setState` - both work and have their own pros/cons.

All the components in this project use `setState` / redux so I wouldn't like to bring in another way to manage state if it doesn't provide significant benefits. This component is very basic and doesn't do anything so I'd keep it as a `PureComponent`.

scripts/core/editor3/components/Editor3Component.tsx
tomaskikutis: > otherwise you would lose the rest of the text when pasting, not sure if that's a good UX

That's how `maxlength` works with native input/textarea so I guess it'd be good enough here as well. If requirements allow it, I guess we could keep highlighting too.

scripts/apps/authoring/authoring/components/CharacterCountConfigButton.tsx
pablopunk: Hooks are meant to replace class-based state. In React docs they say that it's the recommended way of creating new components. I never meant to replace the current classes we have, unless we create custom hooks, which in the future could be a very powerful way of handling CRUD operations, like on this example:

```tsx
const UserComponent = ({ id }) => {
  const [user, editUser, removeUser] = userApi.useUser(id) // custom hook that connects an API with a UI component
  return <div>{user?.name}</div>
}
```

I don't mean that it cannot be made with class-logic but IMO it's a cleaner way of connecting the data to the UI, smaller and more intuitive to read.

Also, setState is for component state and redux for app state, that we only use on editor3 for now, so I don't see an issue with hooks for component state.

scripts/apps/authoring/authoring/components/CharacterCountConfigButton.tsx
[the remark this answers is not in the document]
pablopunk: >does it cut it then?

no, it doesn't cut the text if you change the config. I guess it's the same thing when pasting, it should work the same way IMO

>version is a bit confusing when seeing for the first time

yeah I agree, I did it like this cause it's the way it's done in the ui-framework (it doesn't infer the type so it throws an error).

scripts/apps/authoring/authoring/components/CharacterCountConfigButton.tsx
tomaskikutis: > Hooks are meant to replace class-based state. In React docs they say that it's the recommended way of creating new components.

Class components aren't deprecated and there are no plans to do so. Given it's a core API I doubt it will ever be removed.

I don't mind hooks in principle. If all our code was in hooks and you wanted to add a class component, I would ask you not to. I want the project to use as little different patterns as possible so there is less to learn in order to maintain it. Hooks aren't not just an addition, as a library would be, but a totally different paradigm. Everyone would need to learn how all life-cycle methods are implemented with hooks and IMO we could spend that time more effectively refactoring business logic.

pablopunk: You're kinda right in the sense that it's better to keep it uniform but getting stuck in a way a library works instead of exploring the future of it (classes won't be deprecated soon, but all the improvements will come from hooks logic) is not great either. If we work with react it's reasonable to keep learning it and keep moving with the library and the community.

scripts/core/editor3/components/Editor3Component.tsx
pablopunk: Ok, I talked with Ana and they want the behavior we currently have. Allow pasting (highlighting overflow) and not chopping the overflow when switching options

scripts/apps/authoring/authoring/components/CharacterCountConfigButton.tsx
tomaskikutis: I get your point. How about we keep class components in the core and build a next few extensions using hooks to try it out?

pablopunk: hmmm you mean moving component to a new extension?

tomaskikutis: No :) Just using a class component here and when we have to make a new extension build it all with hooks.

tomaskikutis (approved):

PR #3671, opened 2020-10-28
Sdesk 5265 edit asset metadata

dev-tsoni:


scripts/extensions/sams/src/components/assets/assetEditorPanel.tsx
```diff
@@ -55,31 +140,48 @@ class AssetEditorPanelComponent extends React.PureComponent<IProps> {
...
+                            <Button
+                                text={gettext('Save')}
+                                type="primary"
+                                disabled={!this.state.isDirty || this.state.submitting}
+                                onClick={this.onSave}
+                            />
+                        </ButtonGroup>
+                    </PanelHeaderSlidingToolbar>
+                </PanelHeader>
                 <FormGroup>
```
MarkLark86, line 161: These `FormGroup`'s are render-able with the `AssetEditor`. Instead of duplicating these render items, it might be better to render `AssetEditor` here instead

scripts/extensions/sams/src/ui/grid/GridItem.tsx
```diff
@@ -18,7 +19,7 @@ export class GridItem extends React.PureComponent<IProps> {
         );
 
         return (
-            <div className={classes} onClick={this.props.onClick}>
+            <div className={classes} onClick={this.props.onClick} onDoubleClick={this.props.onDoubleClick}>
```
MarkLark86, line 22: When handling double clicks, it makes for a smoother user experience if we place a `window.setTimeout` over the single click prop (otherwise when double clicking, the single click event will still be fired), and clear this timeout on double click.

Example usage (see`select` and `dbClick` functions):
https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/search/components/Item.tsx#L166

scripts/extensions/sams/src/apps/samsWorkspace.tsx
```diff
@@ -208,15 +231,24 @@ export class SamsWorkspaceComponent extends React.Component<IProps, IState> {
...
-                            setName={this.props.setName}
-                            onPanelClosed={this.props.onPanelClosed}
-                            downloadAsset={this.onDownloadSingleAssetCompressedBinary}
-                            deletAsset={this.onDeleteAsset}
-                        />
+                    rightPanelOpen={ContentPanel != null}
+                    rightPanel={ContentPanel == null ? (
+                        <div />
+                    ) : (
+                        <PanelContent>
```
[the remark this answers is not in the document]
dev-tsoni: Resolved.

scripts/extensions/sams/src/components/assets/assetEditorPanel.tsx
dev-tsoni: Resolved.

MarkLark86 (approved):

PR #3675, opened 2020-10-31
Add missing preview fields

tomaskikutis:
SDESK-5624

Preview fields like fields in authoring were hardcoded and it was hard to follow for which fields preview was implemented. I went over all hardcoded and custom fields and have built a mapper to field types which at the moment we have 9. I have decoupled the preview logic from field ids so all preview components now only rely on field type.


scripts/apps/authoring/preview/field-types/urls.tsx
```diff
@@ -0,0 +1,19 @@
...
+    urls: Array<{url: string; description: string}>;
+}
+
+export class UrlsPreview extends React.Component<IProps> {
+    render() {
+        return (
+            <div>
+                {
+                    this.props.urls.map(({url, description}) => (
+                        <div key={url}>{description}({url})</div>
```
pablopunk, line 13: should urls be inside an `a` tag so the user can just click on it?

scripts/apps/authoring/preview/getAuthoringField.ts
```diff
@@ -0,0 +1,161 @@
...
+import {getRelatedArticles, getRelatedMedia} from '../authoring/controllers/AssociationController';
+
+export function getAuthoringField(
+    fieldId: string,
+    item: IArticle,
+    customFieldsVocabularies: Array<IVocabulary>,
+): IAuthoringField {
+    const customField = customFieldsVocabularies.find(({_id}) => _id === fieldId);
+
+    if (customField?.field_type === 'text') {
```
pablopunk, line 13: This function looks more like a `switch` instead of so many `else if`

pablopunk (approved):

scripts/apps/authoring/preview/field-types/urls.tsx
tomaskikutis: It's meant to be printed so the link has to be inline. I have added a fix to make it clickable.

pablopunk (approved):

PR #3678, opened 2020-11-02
User without Templates permission in authoring cannot see the option 3dots - 'Save as template'. [SDANSA-374]

GyanP:


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -161,6 +161,16 @@ export function AuthoringDirective(
                 }
             }
 
+            /**
+             * Get the Current Template for the item.
+            */
+            function getCurrentTemplate() {
+                api('content_templates').getById($scope.item?.template)
```
tomaskikutis, line 168: You shouldn't be using optional chaining here. If `item` is null, you won't get the data anyway even if you avoid null reference exception here.

It should be checked before doing the request:

```
if (typeof $scope.item?.template !== 'string') {
    logger.error('template must be present');
    return;
}

api('content_templates').getById($scope.item.template)
```

tomaskikutis (approved):

PR #3683, opened 2020-11-05
Default content profile and template should be required in Desk configuration and fix the e2e test cases. [SDESK-5616]

GyanP:


scripts/apps/desks/views/desk-config-modal.html
```diff
@@ -151,25 +151,50 @@ <h3 class="modal__heading" ng-show="desk.edit._id" translate>Edit "{{desk.edit.n
...
+                            <div class="sd-line-input sd-line-input--is-select">
+                                <select
+                                    ng-model="desk.edit.default_content_template"
+                                    class="sd-line-input__select"
+                                    required
+                                    data-test-id="field--default-content-template"
+                                >
                                 <option value=""></option>
-                            </select>
+                                <option ng-repeat="template in templates" value="{{template._id}}" data-test-id="template {{template.template_name}}" ng-selected="desk.edit.default_content_template === template._id">
```
tomaskikutis, line 170: You don't need `data-test-id` here. It's value is template name with a prefix, so we can instead add `data-test-id` on a `select` itself and then select by option value without having to add ids for every option.

`el(['field--default-content-template'], by.cssContainingText('option', 'template 5')).click();`

scripts/apps/desks/views/desk-config-modal.html
```diff
@@ -151,25 +151,50 @@ <h3 class="modal__heading" ng-show="desk.edit._id" translate>Edit "{{desk.edit.n
...
+                                    Select a default content profile which is used for articles fetched from ingest. If you don't see any, create one
+                                   <a href="#/settings/content-profiles">here</a>
+                                </span>
+                            </div>
+                            <div  class="sd-line-input sd-line-input--is-select">
+                                <select ng-model="desk.edit.default_content_profile"
+                                        class="sd-line-input__select"
+                                        required>
+                                    <option value=""></option>
+                                    <option ng-repeat="profile in profiles" value="{{profile._id}}" data-test-id="profile {{profile.label}}" ng-selected="desk.edit.default_content_profile === profile._id">
```
tomaskikutis, line 193: Add `data-test-id` on a select instead of options. Also try avoiding dynamic test IDs where possible as they are harder to find.

spec/helpers/desks.ts
```diff
@@ -267,6 +269,24 @@ class Desks {
             element(by.model('desk.edit.desk_type')).$('[value="' + deskType + '"]').click();
         };
 
+        /**
+         * Set desk default content template
+         * @param {string} template name
+         **/
+        this.setDeskDefaultContentTemplate = function(template) {
+            element(by.model('desk.edit.default_content_template'))
```
tomaskikutis, line 277: Don't use `by.model` for new code. It relies on angular and we'll have to rewrite it when moving to react. It's best to use `data-test-id` to query things since it's not coupled to any framework.

scripts/apps/desks/views/desk-config-modal.html
```diff
@@ -161,15 +161,13 @@ <h3 class="modal__heading" ng-show="desk.edit._id" translate>Edit "{{desk.edit.n
                             </div>
                             <div class="sd-line-input sd-line-input--is-select">
                                 <select
+                                    ng-options="template._id as template.template_name for template in templates"
```
tomaskikutis, line 164: You could have left `<option ng-repeat="template in templates" ...` I was only talking about `data-test-id` attribute. This is fine too though.

tomaskikutis (approved):

PR #3684, opened 2020-11-05
Sdesk 5580 set maximum size of asset on a per set basis

dev-tsoni:


scripts/extensions/sams/src/utils/ui.tsx
```diff
@@ -53,6 +53,18 @@ export function getIconTypeFromMimetype(mimetype: string) {
     }
 }
 
+export function getFileSizeFromHumanReadable(fileSize: number, dataUnit: string) {
+    if (dataUnit.startsWith('Bytes')) {
```
MarkLark86, line 57: It might be more precise if we used an enum for `dataUnit` rather then using strings, similar to how we use enums for `SET_STATE` | `ASSET_STATE` etc.

Then use these values as the `value` attribute to the options in the select field.

scripts/extensions/sams/src/components/sets/setEditorPanel.tsx
```diff
@@ -251,13 +251,13 @@ export class SetEditorPanelComponent extends React.Component<IProps, IState> {
...
                                     <Select
-                                        label={gettext('Stoeage Unit')}
+                                        label={gettext('Storage Unit')}
                                         onChange={this.onChange.storage_unit}
                                     >
-                                        <Option>{gettext('Bytes')}</Option>
-                                        <Option>{gettext('KB')}</Option>
-                                        <Option>{gettext('MB')}</Option>
-                                        <Option>{gettext('GB')}</Option>
+                                        <Option>{DATA_UNIT.BYTES}</Option>
```
MarkLark86, line 257: To provide translations here, the `DATA_UNIT.*` values will need to be applied as the `value` attribute of the tag, and the innerHTML use `gettext`. i.e.
`<Option value={DATA_UNIT.BYTES}>{gettext('Bytes')}</Option>`

dev-tsoni: Okay.

MarkLark86 (approved):

PR #3687, opened 2020-11-06
[SDESK-5636] Pass locale settings when using DatePicker

MarkLark86:


scripts/core/helpers/locale.ts
```diff
@@ -23,3 +25,30 @@ export function getMonthNames(targetLocale: string): Array<string> {
 
     return monthNames;
 }
+
+export function getLocaleForDatePicker(targetLocale?: string): DatePickerLocaleSettings {
+    function getLocale() {
+        return {
+            firstDayOfWeek: appConfig.startingDay,
```
tomaskikutis, line 32: firstDayOfWeek is already exposed via `appConfig` as `startingDay`. I think we shouldn't duplicate it here.

scripts/core/superdesk-api.d.ts
```diff
@@ -1362,6 +1372,7 @@ declare module 'superdesk-api' {
             formatDate(date: Date): string;
             formatDateTime(date: Date): string;
             longFormatDateTime(date: Date | string): string;
+            localeForDatePicker(targetLocale?: string): DatePickerLocaleSettings;
```
tomaskikutis, line 1375: It may be used for other things than date picker, for example to print dates in different components, so I'd make the naming more generic like `getDateTimePeriodNames`

scripts/core/helpers/locale.ts
MarkLark86: The `DatePicker` this is referring to is the UI Framework's date picker, and the return type is the same type of the property. This method exists so the developer can use this function to supply the property to that component.

I could move this to a new helper file, i.e. `scripts/core/helpers/ui-framework.ts` or something along those lines?

scripts/core/superdesk-api.d.ts
MarkLark86: See my comment above as to the usage of this helper function

scripts/core/helpers/locale.ts
tomaskikutis: Sorry, I didn't notice your reply.

I agree that it would work to move it to `helpers/ui-framework.ts`, but I wouldn't like to expose any UI framework related things in Superdesk API because every extension already depends on ui-framework directly.

I think a better solution would be to add a method to ui framework to set locale once so it could reuse it for different components without having to pass it directly - similarly how we don't pass a language every time we call `gettext`.

MarkLark86: The locale here could change depending on it's usage.

For example, when using a DatePicker in a text item this field should use the locale of that item, similar to how we change the locale for certain CVs. I have added this ability into the `datetimeField` extension https://github.com/superdesk/superdesk-client-core/blob/17541650ba8e9b4fcd91aa7c7a7473b6082d7658/scripts/extensions/datetimeField/src/getDateTimeField.tsx#L47-L48
https://github.com/superdesk/superdesk-client-core/blob/17541650ba8e9b4fcd91aa7c7a7473b6082d7658/scripts/extensions/datetimeField/src/getDateTimeField.tsx#L55-L57

tomaskikutis: I see. I still don't like putting ui-framework related stuff into the superdesk-api. What about modifying the date picker in the UI-framework so it would be enough to pass a string based locale like `fr-CA`? That way that helper function could live in ui-framework itself. The `startingDay` config option probably always matches the locale string so we could probably drop it someday.

MarkLark86: Nemanja wanted Superdesk to provide the translations. Ideally this should use `gettext` not `moment`, as all other translations are controlled via this mechanism

MarkLark86: I also understand your concerns, we should try to minimise UI specifics in the API.

But at the same time, all extension's have access to both the UI framework and the Superdesk API. This is a small utility function to aid developer experience with translations and the DatePicker

tomaskikutis: Ok, let's keep it like that then.

tomaskikutis (approved):

PR #3688, opened 2020-11-06
SDESK-5639 Multiple selection bar is not erased after deletion.

dev-tsoni:


MarkLark86:
I'm not sure why, but on my local this bug (and SDESK-5640 - Preview remains open) are still occurring.

It might make it easier to control these side effects if we had 2 new redux actions, `deleteAsset` and `deleteMultipleAssets`. These side effects can:
* Remove the Asset from the list of selected Assets
* If the Asset is currently being previewed, then close it.

Later when more functionality is added, i.e. multi-user scenarios etc, we can manage these side effects in these actions rather than in UI components themselves

MarkLark86 (approved):

PR #3691, opened 2020-11-09
Disallow editing metadata of an item that is already in edit mode in order to avoid etag conflicts

tomaskikutis:
SDESK-5647

### DIFF
--- a/scripts/apps/authoring/authoring/index.ts
+++ b/scripts/apps/authoring/authoring/index.ts
@@ -29,7 +29,7 @@ import {FullPreviewItemDirective} from './directives/FullPreviewItemDirective';
 import {AuthoringTopbar2React} from './authoring-topbar2-react';
 import {appConfig} from 'appConfig';
 import {FullPreview} from '../preview/fullPreview';
-import {isLocked} from 'core/get-superdesk-api-implementation';
+import {sdApi} from 'api';
 
 export interface IOnChangeParams {
     item: IArticle;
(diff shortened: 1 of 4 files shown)


scripts/apps/authoring/authoring/index.ts
```diff
@@ -29,6 +29,7 @@ import {FullPreviewItemDirective} from './directives/FullPreviewItemDirective';
 import {AuthoringTopbar2React} from './authoring-topbar2-react';
 import {appConfig} from 'appConfig';
 import {FullPreview} from '../preview/fullPreview';
+import {isLocked} from 'core/get-superdesk-api-implementation';
```
petrjasek, line 32: would rather have this in some `core/article` module

tomaskikutis: Me too. I wasn't sure where to put it. It seems most things in "extensions API" would work as "core API" too. The problem is that some things in extensions API depend on angular services which makes them async when they shouldn't be. I didn't want to rush with core API so I don't have to add messy code from the start.

One of the significant decisions we have to make is how do we handle relations. If an article has an id of a desk or an id of a user that's assigned to that article, and we wan't to display desk/user name in the list view, how do we get the full object? It will not perform well if we sent a HTTP request for each article in the list to get that related entity object.

We could get all entities on app start and cache them, but it wouldn't work when there's a lot of data. Would it work to have HTTP API optionally return related entities instead of IDs only?

petrjasek: I think we need some datalayer on the client, which would be able to fetch related entities or return them from cache if it already has them. And in general it should avoid having multiple versions of the same entity.

petrjasek (approved):

scripts/apps/authoring/authoring/index.ts
tomaskikutis: I like the data layer approach because it reduces traffic & duplicate entities being sent, but without optimizing it can make UI slower. If we have 50 items and 3 types of related entities, say content profile, user avatar, desk name, it's already 150 related entities and 3 requests - and we are ready to render only after the slowest of them arrives.

On the other hand, if we cached a few hundred of most frequently accessed entities of each type, it would probably be quick for most cases.

petrjasek: yep if we handle that in the layer we just need to tell it what entities and what fields will be needed, then it can either make specific request or multiple requests to fetch it. client will render only once it has all the data anyhow.

PR #3694, opened 2020-11-10
Fix overflow highlights when CP changes

pablopunk:
SDANSA-446
SDANSA-447

### DIFF
--- a/scripts/core/editor3/actions/editor3.tsx
+++ b/scripts/core/editor3/actions/editor3.tsx
@@ -4,7 +4,10 @@ import {logger} from 'core/services/logger';
 import {SelectionState, convertFromRaw, EditorState} from 'draft-js';
 import {IArticle} from 'superdesk-api';
 import {getFieldMetadata, fieldsMetaKeys} from '../helpers/fieldsMeta';
-import {CharacterLimitUiBehavior} from 'apps/authoring/authoring/components/CharacterCountConfigButton';
+import {
+    CharacterLimitUiBehavior,
+    DEFAULT_UI_FOR_EDITOR_LIMIT,
+} from 'apps/authoring/authoring/components/CharacterCountConfigButton';
 
 /**
  * @ngdoc method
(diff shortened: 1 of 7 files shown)


tomaskikutis:
Are you cherry picking code from develop branch or are these new changes that would have to be merged to develop?

pablopunk:
new changes that should be merged to develop, the same as https://github.com/superdesk/superdesk-client-core/pull/3690

scripts/core/editor3/actions/editor3.tsx
```diff
@@ -236,8 +239,10 @@ export function changeCase(changeTo: ITextCase, selection: SelectionState) {
 
 export type EditorLimit = { ui: CharacterLimitUiBehavior, chars: number };
 export function changeLimitConfig(payload: EditorLimit) {
+    const config = payload.ui ? payload : {...payload, ui: DEFAULT_UI_FOR_EDITOR_LIMIT};
```
tomaskikutis, line 242: when writing ifs/ternaries against non-boolean values it'd be good to do `x != null` check instead of simply `if(x)`so it's more clear. For example an empty string is not null, but would evaluate to false in if/ternary.

tomaskikutis (approved):

scripts/core/editor3/actions/editor3.tsx
pablopunk: agree, but this should still evaluate to `false` if the string is empty. Although, that shouldn't happen unless some data gets corrupted

tomaskikutis: string example wasn't supposed to apply here directly since `payload.ui` is an object. I just wanted to illustrate that `if (x)` is not the same as `if (x != null)`

pablopunk: it is a string (`type CharacterLimitUiBehavior = 'highlight' | 'limit'`), that's why I did took it as an example

PR #3695, opened 2020-11-10
Fix eslint warnings

tomaskikutis:
There were more than a 100 warnings in eslint output which has been annoying for a long time and became even more annoying when github started displaying them inline when previewing diffs.

I removed tests which were disabled for 3-5 years, improved code for some tests and in some cases simply disabled warnings for specific lines so even if we don't fix it right away, we would be able to see warnings for new code.

I also disabled some rules here https://github.com/superdesk/superdesk-client-code-style/pull/7


scripts/apps/authoring/tests/authoring.spec.ts
```diff
@@ -90,22 +92,6 @@ describe('authoring', () => {
             expect(ITEM._locked).toBe(true);
         }));
 
-    it('unlocks a locked item and locks by current user',
```
petrjasek, line 93: why removing this one?

scripts/apps/workspace/tests/workspace.spec.ts
```diff
@@ -160,7 +160,7 @@ describe('workspace', () => {
         });
 
         describe('selectWorkspace() scope method', () => {
-            it('can set workspace type', inject(() => {
+            it('selectWorkspace()can set workspace type', inject(() => {
```
pablopunk, line 163: it's missing a space after `()`

scripts/apps/authoring/tests/authoring.spec.ts
tomaskikutis: `expect('a', 'b')` does not check anything - it should be `expect('a').toBe('b')`. When I corrected the syntax it failed since (`$location.path()` would evaluate to an empty string) so I removed it together with all other tests which haven't worked for years.

petrjasek (approved):

PR #3698, opened 2020-11-12
Add the new badge for 'Correction Workflow'. [SDCP-391]

GyanP:


tomaskikutis (approved):

scripts/apps/archive/constants.ts
```diff
@@ -59,6 +59,16 @@ export enum ITEM_STATE {
      */
     RECALLED = 'recalled',
 
+    /**
+     * Correction, If Correction workflow is true, Correction is send on a desk.
+     */
+    CORRECTED = 'corrected',
+
```
tomaskikutis, line 66: Can you rephrase comments so they are more clear? Now I don't understand any additional information from them that is not already in the enum name.

I assume if an item is `CORRECTED` it will always published and items that are `BEING_CORRECTED`, will always be not published, but that were published in the past?

scripts/core/superdesk-api.d.ts
```diff
@@ -319,6 +329,8 @@ declare module 'superdesk-api' {
         lock_session: any;
         rewritten_by?: IArticle['_id'];
         rewrite_of?: IArticle['_id'];
+        correction_by?: IArticle['_id'];
+        corrected_of?: IArticle['_id'];
```
tomaskikutis, line 333: it would be more correct to name it `correction_by`/`correction_of`

GyanP: Petr and I are discussing whether we need these fields or not, if yes then I'll for sure rename them.

scripts/apps/archive/constants.ts
```diff
@@ -60,13 +60,13 @@ export enum ITEM_STATE {
     RECALLED = 'recalled',
 
     /**
-     * Correction, If Correction workflow is true, Correction is send on a desk.
+     * Correction, If Correction workflow is true, correction, copy of published article which we can edit.
```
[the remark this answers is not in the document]
GyanP: No.

PR #3702, opened 2020-11-17
SDESK-5474-provide-human-friendly-error-messages-sams

dev-tsoni:


MarkLark86 (commented):
Missing implementation in the following functions:

- https://github.com/superdesk/superdesk-client-core/blob/b22cc2a9892d2bdaaa3915b80a682c32f4813dfd/scripts/extensions/sams/src/api/assets.ts#L38 SAMS API response can be processed from the `superdeskApi.dataApi.uploadFileWithProgress` call

- https://github.com/superdesk/superdesk-client-core/blob/b22cc2a9892d2bdaaa3915b80a682c32f4813dfd/scripts/extensions/sams/src/api/assets.ts#L357 SAMS API response can be processed from the `superdeskApi.dataApi.patch` call

MarkLark86 (approved):

PR #3703, opened 2020-11-17
The publish button should be present on top of the editor also in personal space if the user has the publish privilege. [SDANSA-434]

GyanP:


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -713,7 +714,9 @@ export function AuthoringDirective(
             }
 
             $scope.showCustomButtons = function(item) {
-                if ($location.path() === '/workspace/personal') {
+                if ($scope.isPersonalSpace && appConfig?.features?.publishFromPersonal) {
```
tomaskikutis, line 717: I was trying to see what custom buttons are. Looks like there are 4:

* To Desk
* Close & Continue
* Publish
* Publish & Continue

It would be good to clean up the logic of `showCustomButtons` to be something like

```
$scope.showCustomButtons = () => toDeskAvailable && closeAndContinueAvailable && publishAvailable;
```

It would also need some comments/improvement of variable names - it is not clear at all what "to desk" or "close and continue" means.

scripts/apps/authoring/authoring/services/AuthoringService.ts
```diff
@@ -368,6 +368,15 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         helpers.filterDefaultValues(extDiff, orig);
         var endpoint = 'archive_' + action;
 
+        if (appConfig.features.publishFromPersonal && !orig?.task?.desk && !extDiff?.task?.desk) {
+            var currentDeskId = session.identity.desk || desks.getCurrentDeskId();
```
tomaskikutis, line 372: It's not very good that you need a fallback here. It would be good to fix one of the methods to it works reliably.

scripts/apps/authoring/views/authoring-topbar.html
```diff
@@ -77,7 +78,8 @@
                 <span class="btn__text--short" translate>P</span>
         </button>
         <button
-                ng-if="$root.config.features.customAuthoringTopbar.publishAndContinue === true && canPublishOnDesk()"
+                ng-if="$root.config.features.customAuthoringTopbar.publishAndContinue === true
+                && canPublishOnDesk() && publishAndContinueAvailable"
```
petrjasek, line 82: would be good to have that logic in one place and here just check `ng-if="publishAndContinueEnabled"` otherwise you have to check both places to figure out what is going on

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -712,11 +727,25 @@ export function AuthoringDirective(
...
-                return item.task && item.task.desk && item.state !== 'draft' || $scope.dirty;
+            // ToDesk -- Send an Item to a desk
+
+            // closeAndContinue -- Create an update of an item and Close the item.
+
+            // publish -- publish an item
+
+            // publishAndContinue -- Publish an item and Create an update.
+
+            $scope.showCustomButtons = () => {
```
petrjasek, line 738: I don't like this function which seems to be a check if any button is visible also setting those other variables, shouldn't that be rather done when item is updated and this would only return the status/

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -138,6 +138,35 @@ export function AuthoringDirective(
...
+                return $scope.item.task && $scope.item.task.desk && $scope.item.state !== 'draft' || $scope.dirty;
+            }
+
+            // ToDesk -- Send an Item to a desk
+
+            // closeAndContinue -- Create an update of an item and Close the item.
+
+            // publish -- publish an item
+
+            // publishAndContinue -- Publish an item and Create an update.
```
petrjasek, line 154: looks good, maybe those comments could be where you define those variables like:
```ts
$scope.toDeskEnabled = false; // Send an item to desk
```
otherwise this might later show up in some diffs and you wouldn't see the relation

petrjasek (approved):

PR #3704, opened 2020-11-20
Add desk in item associations [SDANSA-434]

GyanP:


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -480,6 +480,15 @@ export function AuthoringDirective(
                         ...(item.task ?? {}),
                         desk: currentDeskId,
                     };
+
+                    if (item.associations) {
+                        Object.keys(item.associations).forEach((key) => {
+                            item.associations[key].task = {
```
petrjasek, line 486: it should also check if `item.associations[key]` is not `null`

petrjasek (approved):

PR #3709, opened 2020-11-23
FIX: Unable to locate the authoring item in the personal space [SDANSA-434]

GyanP:


tomaskikutis:
Which use case should these changes solve?

GyanP:
> Which use case should these changes solve?

Whenever we publish an item from personal space without adding the required fields, it will show the required fields error so now when we add the required fields and save the item, it will remove the item from personal space.

Reason: Item which does not contain the desk are the personal space articles. so when we publish an item we add the desk but if it didn't publish, it still contains the desk in the item, so that's why when we perform any operation it will remove the item from personal space.

authoring_disappears_when_empty_metadata_media_associated1

tomaskikutis:
I see. I don't like that changes are applied once in order to make it publishable and then applied again on failure to simulate a reset. It's always good to prevent this. For example these changes might appear in the history. If I try to perform an action and it fails, I expect that nothing would be modified.

It would be much cleaner to have a separate endpoint or query param like `/publish?allowFromPersonal=true` where an item without a desk would be sent, the server would try to publish it, and add a desk only at the last step after it has been checked that there are no errors. Maybe we don't even need a query param. If there's a config enabled that allows publishing from personal space, server should accept an item without a desk by default.

scripts/apps/authoring/authoring/services/AuthoringService.ts
```diff
@@ -368,7 +368,14 @@ export function AuthoringService($q, $location, api, lock, autosave, confirm, pr
         helpers.filterDefaultValues(extDiff, orig);
         var endpoint = 'archive_' + action;
 
-        return api.update(endpoint, orig, extDiff, {publishing_warnings_confirmed: publishingWarningsConfirmed})
+        var params = {
+            publishing_warnings_confirmed: publishingWarningsConfirmed,
+            desk_id: appConfig.features.publishFromPersonal && !orig?.task?.desk && !extDiff?.task?.desk
```
tomaskikutis, line 373: If `desk_id` is only needed when publishing from personal, can you only add it to params only in that case?

```
if (publishFromPersonal) {
    params.desk_id = ...;
}
```

tomaskikutis (approved):
Looks much better now :tada:

PR #3713, opened 2020-11-24
[INGEST] [SETTINGS] Use Object ID instead of provider name to identit…

jerome-poisson:
…y provider

Provider name was used to identify the requesting provider when doing
URL request, but the provider name may not be set yet at this point.
This has been replaced by a generated ID which is compatible with
MongoDB. This way it can be used in the backend in the `_id` field.

SDESK-5342

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "ts-loader": "3.5.0",
     "tslint": "5.11.0",
     "typescript": "3.9.7",
-    "uuid": "<9.0.0",
+    "uuid": "8.3.1",
     "webdriver-manager": "12.1.7",
     "webpack": "3.11.0",
     "webpack-dev-server": "2.11.1"
--- a/package.json
+++ b/package.json
@@ -124,6 +124,7 @@
     "ts-loader": "3.5.0",
     "tslint": "5.11.0",
     "typescript": "3.9.7",
+    "uuid": "<9.0.0",
     "webdriver-manager": "12.1.7",
     "webpack": "3.11.0",
     "webpack-dev-server": "2.11.1"
--- a/scripts/apps/ingest/directives/IngestSourcesContent.ts
+++ b/scripts/apps/ingest/directives/IngestSourcesContent.ts
@@ -638,7 +638,6 @@ export function IngestSourcesContent(ingestSources, notify, api, $location,
                  * @param field url_request field metadata
                  */
                 $scope.doUrlRequest = (provider: IProvider, field: IFeedingServiceField): void => {
-                    let provider_name = encodeURIComponent(provider.name);
                     provider.url_id = mongoDBObjectId();
 
                     window.open(field.url.replace('{OID}', provider.url_id));
--- a/scripts/apps/ingest/directives/IngestSourcesContent.ts
+++ b/scripts/apps/ingest/directives/IngestSourcesContent.ts
@@ -40,6 +40,7 @@ interface IProvider {
     last_opened?: {};
     critical_errors?: {};
     skip_config_test?: boolean;
+    url_id?: string;
 }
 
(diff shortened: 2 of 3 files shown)


scripts/apps/ingest/directives/IngestSourcesContent.ts
```diff
@@ -619,15 +620,28 @@ export function IngestSourcesContent(ingestSources, notify, api, $location,
...
+
+
                 /**
                  * Do URL request specified in url_request field
                  * @param provider ingest provider metadata
                  * @param field url_request field metadata
                  */
                 $scope.doUrlRequest = (provider: IProvider, field: IFeedingServiceField): void => {
                     let provider_name = encodeURIComponent(provider.name);
+                    provider.url_id = mongoDBObjectId();
```
tomaskikutis, line 642: Why is new `url_id` generated every time and on the client? Couldn't the provider object coming from the server already have that ID present?

tomaskikutis: It's also not good to couple database specific things like mongo ID format to UI code. Client should talk to API and not be coupled to implementation specific things like what kind of database is used.

jerome-poisson: Hi Tomas,

I agree that it's not ideal, but the issue is that at this point we have nothing from backend, the `provider` object is AFAIK created by the client (the `save` button may have not been pressed at all), and we need to identify the HTTP request origin to link it with the provider.

The use of Mongo DB Object ID instead of something like e.g. UUID, is because we can use it directly as an ID in backend, which is indexed. Using something else would involve the creation of a new field, and it was requested in review of the [backend PR review](https://github.com/superdesk/superdesk-core/pull/1941#discussion_r484360976) (cc @petrjasek) to use `_id`.

To make it cleaner, we would need that the client requests a new provider object when `add new` is pressed in ingest settings, and that would involve more work (probably a new endpoint in backend + request by client).

petrjasek: `_id` in mongo doesn't have to be ObjectId, as long as it's unique it should be fine

jerome-poisson: @petrjasek really? I was pretty sure it had to be. In this case if I use a uuid here, is it fine?

tomaskikutis: > I agree that it's not ideal, but the issue is that at this point we have nothing from backend, the provider object is AFAIK created by the client (the save button may have not been pressed at all), and we need to identify the HTTP request origin to link it with the provider.

I checked `ingest_providers` collection in the database and all items have `_id`s. Do you want to do authentication when creating a new provider before it has been saved?

jerome-poisson: @tomaskikutis yes they have `_id` once in the DB, but the authentication is done before the provider is saved. When you click on the save button, a test is done to check the settings, which uses the authentication (it basically tries to connect to the mailbox), and if it fails, the save is denied and the provider is not created in backend, thus the URL request is done before anything is created in backend.

jerome-poisson: @tomaskikutis @petrjasek indeed `_id` is working with any string, I've updated the PR to use a UUID so no more implementation specific ID.

package.json
```diff
@@ -124,6 +124,7 @@
     "ts-loader": "3.5.0",
     "tslint": "5.11.0",
     "typescript": "3.9.7",
+    "uuid": "<9.0.0",
```
tomaskikutis, line 127: Can you remove `<` so it's fixed to this exact version?

tomaskikutis (approved):

PR #3722, opened 2020-12-02
Use new list view for highlights

tomaskikutis:
SDESK-5237


tomaskikutis:
This branch was forked from https://github.com/superdesk/superdesk-client-core/pull/3643 and commits before ae62e47 are from that branch.

scripts/core/MultiSelectHoc.tsx
```diff
@@ -12,13 +20,24 @@ export interface IMultiSelectOptions {
...
-interface Istate {
+interface IState {
     selected: Map<string, IArticle>;
 }
 
-export class MultiSelectHoc extends React.PureComponent<IProps, Istate> {
+export class MultiSelectHoc extends React.PureComponent<IProps, IState> {
+    private removeContentUpdateListener: () => void;
+    private removeResourceDeletedListener: () => void;
+    private maybeUnselectItems: (ids: globalThis.Set<string>) => void;
```
pablopunk, line 39: doesn't it need the underscore in the name? [this one](https://github.com/superdesk/superdesk-client-core/pull/3722/files?diff=unified&w=1#diff-e7af39c7e776bf103574695df94d8489954718dd1f293fb8eca59173b6770426R47) has it

scripts/core/ArticlesListByQueryWithFilters.tsx
```diff
@@ -195,135 +201,212 @@ export class ArticlesListByQueryWithFilters extends React.PureComponent<IProps,
...
+                            $and: [
+                                query.filter,
+                                {'_id': {$in: ids.toJS()}},
+                            ],
+                        },
+                    };
+
+                    const elasticQuery = toElasticQuery(queryForSpecificItems);
+
+                    return httpRequestJsonLocal<IRestApiResponse<{_id: string}>>({ // projections only return _id
```
pablopunk, line 327: don't we have some functions already to perform a search instead of using http manually?

scripts/core/MultiSelectHoc.tsx
tomaskikutis: We don't have a convention on this. In a given react component probably 9 of 10 methods and variables would be private so I wouldn't put underscores everywhere, especially when there is a `private` keyword. Regarding `_mounted`, I don't remember why did I add the underscore. Probably because it's a property that doesn't have to do anything with business logic and is just boring and unimportant implementation detail. I would rather drop it there than add underscores to all private members.

spec/monitoring_spec.ts
```diff
@@ -751,6 +761,7 @@ describe('monitoring', () => {
         monitoring.saveSettings();
 
         monitoring.openMonitoring();
+        browser.sleep(3000); // wait for monitoring groups to load
```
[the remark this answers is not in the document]
tomaskikutis: There is a function to wait for one element, but here, there are many groups which finish loading independently at different times so there's no one element to wait for.

I have addressed it in https://github.com/superdesk/superdesk-client-core/pull/3734 by adding loading state to each group and then waiting until there are no loading elements left. I will drop the sleep statement.

Regarding the test being disabled, it wasn't disabled in this PR. I temporarily disabled it on develop so all other tests are passing and we can continue the work on other PRs. Meanwhile Petr has found the issue and merged the fix https://github.com/superdesk/superdesk-client-core/pull/3739.

scripts/apps/monitoring/directives/MonitoringView.ts
```diff
@@ -134,6 +145,41 @@ export function MonitoringView(
...
+                                        {'original_creator': {$eq: user._id}},
+                                    ]},
+                                    {'state': {$ne: 'draft'}},
+                                ]},
+                                {'package_type': {$ne: 'takes'}},
+                                {'highlights': {$eq: $location.search().highlight}},
+                            ]},
+                            sort: [{'versioncreated': 'desc'}],
+                            page: 1,
+                            max_results: 20,
```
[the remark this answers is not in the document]
tomaskikutis: It's lazy loaded and 20 is the page size. It's a bit low, I'll raise it to at least 50.

scripts/core/ArticlesListByQueryWithFilters.tsx
tomaskikutis: We do and this is one of those functions :smile:

There's also [dataApi](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/helpers/CrudManager.tsx#L135), but it uses mongo format for filtering. This one is more generic.

scripts/apps/monitoring/directives/MonitoringView.ts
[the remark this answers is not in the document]
pablopunk: I was gonna suggest making it a constant cause it's used several times on this file

scripts/core/ArticlesListByQueryWithFilters.tsx
pablopunk: I meant this is inside a render function inside a component, not inside a helper function to create requests

tomaskikutis: This `httpRequestJsonLocal` call is <10 LOC. I don't think the helper that would be created out of this would be much simpler. For example, in this case only `_id` of items are returned which wouldn't be useful in other places.

pablopunk (approved):

PR #3723, opened 2020-12-03
Display three dot menu in monitoring when items are selected

tomaskikutis:


petrjasek (approved):

tomaskikutis:
I have reverted some e2e fixes because they were cherry picked from [highlights refactoring branch](https://github.com/superdesk/superdesk-client-core/pull/3722) and were specific to changes to highlights view.

PR #3726, opened 2020-12-08
handle privileges for search providers

petrjasek:
SDBELGA-435

### DIFF
--- a/scripts/apps/search/controllers/SearchMenuController.ts
+++ b/scripts/apps/search/controllers/SearchMenuController.ts
@@ -13,9 +13,11 @@ const isSameRepo = (shortcut, provider) => {
 
 SearchMenuController.$inject = [
     '$rootScope', '$scope', '$filter', '$location', '$route', 'searchProviderService', 'api', 'savedSearch',
+    'privileges',
 ];
 export default function SearchMenuController(
     $rootScope, $scope, $filter, $location, $route, searchProviderService, api, savedSearch,
+    privileges,
 ) {
     let providerLabels = {};
 
@@ -100,7 +102,7 @@ export default function SearchMenuController(
 
     // init search providers
     if (get($rootScope.config, 'features.searchShortcut')) {
-        api.search_providers.query({max_results: 200, where: {is_closed: {$ne: true}}})
+        api.search_providers.query({max_results: 200, where: {is_closed: {$ne: true}}, with_privileges: true})
             .then((result) => {
                 this.providers = $filter('sortByName')(result._items, 'search_provider');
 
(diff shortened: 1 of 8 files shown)


petrjasek:
uses https://github.com/superdesk/superdesk-core/pull/2070

scripts/apps/search/controllers/SearchMenuController.ts
```diff
@@ -100,17 +102,19 @@ export default function SearchMenuController(
 
     // init search providers
     if (get($rootScope.config, 'features.searchShortcut')) {
-        api.search_providers.query({max_results: 200, where: {is_closed: {$ne: true}}})
+        api.search_providers.query({max_results: 200, where: {is_closed: {$ne: true}}, with_privileges: true})
```
tomaskikutis, line 105: Why `with_privileges` parameter needed?

petrjasek: that's to only return providers which user can actually use, admins will have privileges for managing all providers but not necessarily be able to use them for fetching content

tomaskikutis: Wouldn't it work to use the session on the back-end to check that?

petrjasek: it wouldn't know what the user is doing, same user should see all in one case but not all in the other, so that param tell the backend what is the usecase atm

tomaskikutis: > same user should see all in one case but not all in the other

Is that about viewing/managing? Wouldn't it be better then to have 2 privileges?

petrjasek: there is a privilege for managing, for usage it's allowed per provider

tomaskikutis: So what is that use case where the same user should see all in one case but not all in the other?

petrjasek: when managing user should see all, that's based on privilege. when searching user should see it based on which providers are allowed for his user role.

tomaskikutis: I see. I have done a similar thing with templates. I think it would be easier to understand if the query param was `manage=true`. `with_privileges` is seems vague to me.

test-server/requirements.txt
```diff
@@ -1,3 +1,3 @@
 gunicorn==20.0.4
 honcho==1.0.1
-git+git://github.com/superdesk/superdesk-core.git@develop#egg=Superdesk-Core
+git+git://github.com/petrjasek/superdesk-core.git@feat-search-providers-privileges#egg=Superdesk-Core
```
tomaskikutis, line 3: don't forget to set it back to superdesk before merging

tomaskikutis (approved):

PR #3729, opened 2020-12-10
SDESK-5671 ADD: lock asset for editing

dev-tsoni:


scripts/extensions/sams/src/api/assets.ts
```diff
@@ -542,3 +544,45 @@ export function updateAsset(original: IAssetItem, updates: Partial<IAssetItem>):
...
         });
 }
+
+export function lockAsset(original: IAssetItem, updates: Dictionary<string, any>): Promise<Partial<IAssetItem>> {
+    const {gettext} = superdeskApi.localization;
+    const {notify} = superdeskApi.ui;
+
+    return superdeskApi.dataApi.patch<IAssetItem>(LOCK_ASSET, original, updates)
+        .then((asset: IAssetItem) => {
+            notify.success(gettext('Asset locked successfully'));
```
MarkLark86, line 554: Lock/Unlock notifications on success are not necessary

scripts/extensions/sams/src/ui/grid/GridItem.tsx
```diff
@@ -37,7 +38,8 @@ export class GridItem extends React.PureComponent<IProps> {
             'sd-grid-item',
             {
                 'sd-grid-item--with-click': this.props.onClick != null,
-                'sd-grid-item--selected': this.props.selected,
+                'sd-grid-item--with-click locked': this.props.doubleClickSelected,
```
MarkLark86, line 41: It might be more meaningful to have this property directly related to an item being locked. i.e.

`'locked: this.props.locked,`

scripts/extensions/sams/src/apps/samsWorkspace.tsx
```diff
@@ -156,12 +156,22 @@ export class SamsWorkspaceComponent extends React.Component<IProps, IState> {
...
         this.onDeleteAsset = this.onDeleteAsset.bind(this);
+        this.onEditAsset = this.onEditAsset.bind(this);
     }
 
     onDeleteAsset(asset: IAssetItem): void {
         this.props.deleteAsset(asset);
     }
 
+    onEditAsset(asset: IAssetItem): void {
+        const updates: Dictionary<string, any> = {'lock_action': 'edit'};
```
MarkLark86, line 167: This might be better placed in a set of actions/reducers, with another action/reducer for each to update the Asset in the store.

This way after successfully locking/unlocking an item, the Asset in the redux store is also updated. This will then automatically update list views to pass the new lock status to item list/grid item component (as well as updating the `_etag` attribute)

scripts/extensions/sams/src/components/assets/assetListItem.tsx
```diff
@@ -72,7 +72,9 @@ export class AssetListItem extends React.PureComponent<IProps> {
             <ListItem
                 onClick={this.onItemClick}
                 onDoubleClick={this.onItemDoubleClick}
-                selected={this.props.selected || this.props.itemSelected} shadow={1}>
+                selected={this.props.selected || this.props.itemSelected}
+                shadow={1}
+            >
                 <ListItemBorder />
```
MarkLark86, line 78: This component has a `state` property that we can set to `locked` if this Asset is locked (i.e. has `lock_user`, `lock_session`, set). This might be handy to have `isItemLocked` as a utility function so we can use it elsewhere.
https://github.com/superdesk/superdesk-client-core/blob/sams/scripts/extensions/sams/src/ui/list/ListItemBorder.tsx#L5

scripts/extensions/sams/src/components/assets/assetGridItem.tsx
```diff
@@ -94,7 +95,8 @@ export class AssetGridItem extends React.PureComponent<IProps> {
             <GridItem
                 onClick={this.onItemClick}
                 onDoubleClick={this.onItemDoubleClick}
-                selected={this.props.selected || this.props.itemSelected}
+                singleClickSelected={this.props.selected || this.props.itemSelected}
+                doubleClickSelected={this.props.selected && this.props.itemSelectedForEdit}
```
MarkLark86, line 99: Again, using a utility function to set `locked` property here if the Asset is locked

MarkLark86 (approved):

scripts/extensions/sams/src/components/assets/assetGridItem.tsx
```diff
@@ -94,7 +95,8 @@ export class AssetGridItem extends React.PureComponent<IProps> {
             <GridItem
                 onClick={this.onItemClick}
                 onDoubleClick={this.onItemDoubleClick}
-                selected={this.props.selected || this.props.itemSelected}
+                singleClickSelected={this.props.selected || this.props.itemSelected}
+                locked={this.props.selected && this.props.itemSelectedLocked}
```
MarkLark86, line 99: The red border should always show if the Asset is locked, regardless if the item is selected or not (this goes for both Grid and List item views)

scripts/extensions/sams/src/ui/grid/GridItem.tsx
```diff
@@ -5,7 +5,8 @@ interface IProps {
     children: React.ReactNode;
     onClick?(): void;
     onDoubleClick?(): void;
-    selected?: boolean;
+    singleClickSelected?: boolean;
```
MarkLark86, line 8: Now that we have a `locked` property, this could be renamed back to `selected`

scripts/extensions/sams/src/utils/assets.ts
```diff
@@ -93,3 +93,16 @@ export function getMimetypeHumanReadable(mimetype?: string): string {
         return gettext('Other');
     }
 }
+
+export function verifyAssetBeforeLocking(asset: Partial<IAssetItem>, lock_action?: string): boolean {
+    const session_id = superdeskApi.session.getSessionId();
+    const user_id_verified = superdeskApi.session.getCurrentUser().then((user) => {
```
MarkLark86, line 99: Because `session.getCurrentUser` returns a promise, this promise will not be resolved by the time it is used in the if condition below.

Maybe for now we can have a synchronous function `session.getCurrentUserId()` which simply returns `session.identity._id` (as `session.identity` is actually set upon login)

scripts/extensions/sams/src/components/assets/assetGridItem.tsx
MarkLark86: Also we could identify the lock state inside `assetGridItem` and `assetListItem` components directly by checking `lock_session` has a value. So we wouldn't need to send through `itemSelectedLocked`. As this visual lock indicates that the actual Asset in the SAMS API is locked

scripts/extensions/sams/src/store/assets/actions.ts
```diff
@@ -160,6 +163,37 @@ export function editAsset(assetId?: string): IAssetActionTypes {
     };
 }
 
+export function unlockAsset(asset: IAssetItem): IThunkAction<void> {
+    return (dispatch) => {
+        return samsApi.assets.unlockAsset(asset, {})
+            .then(() => {
+                return dispatch(queryAssetsFromCurrentSearch(LIST_ACTION.REPLACE));
```
MarkLark86, line 170: Reloading the lists would be overkill here. It would be more efficient to have a separate action/reducer to update a single Asset in the redux store (using the updated attributes from the response to the `lock` or `unlock` api call)

MarkLark86 (approved):

PR #3735, opened 2020-12-14
init `desk_metadata` before using it

petrjasek:
SDESK-5683

### DIFF
--- a/scripts/apps/desks/controllers/DeskConfigController.ts
+++ b/scripts/apps/desks/controllers/DeskConfigController.ts
@@ -27,6 +27,7 @@ export function DeskConfigController($scope, $controller, notify, desks, modal) 
     const initializeDesk = function(desk, step) {
         $scope.modalActive = true;
         $scope.step.current = step;
+        desk.desk_metadata = desk.desk_metadata ?? {};
         $scope.desk.edit = _.cloneDeep(desk);
         $scope.desk.orig = _.cloneDeep(desk);
     };
--- a/scripts/apps/desks/controllers/DeskConfigController.ts
+++ b/scripts/apps/desks/controllers/DeskConfigController.ts
@@ -27,7 +27,6 @@ export function DeskConfigController($scope, $controller, notify, desks, modal) 
     const initializeDesk = function(desk, step) {
         $scope.modalActive = true;
         $scope.step.current = step;
-        desk.desk_metadata = desk.desk_metadata ?? {};
         $scope.desk.edit = _.cloneDeep(desk);
         $scope.desk.orig = _.cloneDeep(desk);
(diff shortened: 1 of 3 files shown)


tomaskikutis (approved):

scripts/apps/desks/controllers/DeskConfigController.ts
```diff
@@ -27,6 +27,7 @@ export function DeskConfigController($scope, $controller, notify, desks, modal)
     const initializeDesk = function(desk, step) {
         $scope.modalActive = true;
         $scope.step.current = step;
+        desk.desk_metadata = desk.desk_metadata ?? {};
```
tomaskikutis, line 30: Why not initialize it in desk service instead?

petrjasek: yep that works too, makes more sense, will update

scripts/apps/desks/services/DesksFactory.ts
```diff
@@ -284,6 +284,7 @@ export function DesksFactory($q, api, preferencesService, userList, notify,
         },
         fetchDeskById: function(Id) {
             return api.desks.getById(Id).then((_desk) => {
+                _desk.desk_metadata = _desk.desk_metadata ?? {};
```
tomaskikutis, line 287: On the other hand, if the property is optional, we could mark it so in the types and access using `desk?.desk_metadata` and if it's not optional, server could always return an empty object so we don't need to override it here. Otherwise, desks will also be accessible directly via the endpoint in any other place in the code and will not have that property,

petrjasek: it's optional, we only need it really set on that ui component in desk settings, but there it's accessed from directive which is expecting it to be set (this component is used for different types and is not desk specific).

tomaskikutis: > we only need it really set for ui component in desk settings

Is it a real property that desks really have and store it in the database or is it front-end-only pseudo-property like `IArticle['selected']` that should not be saved to the database?

If so, I would then keep it in that UI component and would not spread this bad practice to other parts of the code.

petrjasek: it's real property and gets stored

tomaskikutis: I would then set that as optional in the interface and use optional chaining for accessing data. For angular templates optional chaining will probably not work, so we could initialize a default.

The current solution is fine, but I would add the property as optional to the interface so developers know that optional chaining has to be used in other places.

tomaskikutis (approved):

PR #3744, opened 2020-12-21
Wrong label in preview: Instead of 'Wire' (Category) we show 'Service'. [SDCP-418]

GyanP:


scripts/apps/archive/views/metadata-view.html
```diff
@@ -38,43 +38,43 @@
...
             <dd>{{ item.dateline.text }}</dd>
         </dl>
         <dl>
-            <dt translate>Byline</dt>
+            <dt translate>{{label('byline')}}</dt>
             <dd>{{ item.byline }}</dd>
         </dl>
         <dl>
-            <dt translate>Sign-Off</dt>
+            <dt translate>{{label('sign_off')}}</dt>
```
petrjasek, line 77: I suppose we don't need that translate here anymore, do we?

tomaskikutis: Yep, all translate attributes need to be removed where `label` function is used.

scripts/apps/archive/directives/MediaMetadata.ts
```diff
@@ -13,6 +14,10 @@ export function MediaMetadata(userList, archiveService, metadata) {
             scope.$watch('item', reloadData);
             scope.isCorrectionWorkflowEnabled = appConfig?.corrections_workflow;
 
+            vocabularies.getVocabularies().then((vocabulariesCollection) => {
```
tomaskikutis, line 17: It'd be better to use `getLabelNameResolver`

tomaskikutis (approved):

PR #3745, opened 2020-12-22
SDESK-5676 ADD: selector to get disabledSetIds and disable child asse…

dev-tsoni:
Have added a selector to get disable set Ids, and then disabled the editing of assets belonging to these sets.


MarkLark86 (commented):
Does this update the UI, by removing the edit action from list/preview 3-dot menu's?

MarkLark86 (approved):

PR #3746, opened 2020-12-23
FIX: For the first time, the preview shows ANPA_CATEGORY. [SDCP-418]

GyanP:
For the first time, the label function is not called, it just resolves the getLabelNameResolver, that's why I have used this approach.

### DIFF
--- a/scripts/apps/archive/directives/MediaMetadata.ts
+++ b/scripts/apps/archive/directives/MediaMetadata.ts
@@ -1,10 +1,10 @@
 import _ from 'lodash';
 import {appConfig} from 'appConfig';
-import {getLabelForFieldId} from 'apps/workspace/helpers/getLabelForFieldId';
+import {getLabelNameResolver} from 'apps/workspace/helpers/getLabelForFieldId';
 
-MediaMetadata.$inject = ['userList', 'archiveService', 'metadata', 'vocabularies'];
+MediaMetadata.$inject = ['userList', 'archiveService', 'metadata', '$timeout'];
 
-export function MediaMetadata(userList, archiveService, metadata, vocabularies) {
+export function MediaMetadata(userList, archiveService, metadata, $timeout) {
     return {
         scope: {
             item: '=',
@@ -14,8 +14,12 @@ export function MediaMetadata(userList, archiveService, metadata, vocabularies) 
             scope.$watch('item', reloadData);
             scope.isCorrectionWorkflowEnabled = appConfig?.corrections_workflow;
 
-            vocabularies.getVocabularies().then((vocabulariesCollection) => {
-                scope.label = (id) => getLabelForFieldId(id, vocabulariesCollection);
+            scope.label = (id) => id;
+
+            getLabelNameResolver().then((_getLabelForFieldId) => {
+                $timeout(() => {
+                    scope.label = _getLabelForFieldId;
+                });
             });
 
             function reloadData() {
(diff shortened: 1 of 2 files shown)


tomaskikutis:
Couldn't the issue be fixed inside `getLabelNameResolver`?

GyanP:
> Couldn't the issue be fixed inside `getLabelNameResolver`?

Nope, that's why I used this approach.

tomaskikutis:
And what exactly was the issue? Looks like the only difference is the list of vocabularies.

GyanP:
> And what exactly was the issue? Looks like the only difference is the list of vocabularies.

Here's a small video explaining the exact issue:

https://user-images.githubusercontent.com/10600741/103073028-d26eb080-45ec-11eb-81c7-748943af0080.mp4

tomaskikutis:
I was asking more about the issue in the code. You are replacing `getLabelNameResolver` with almost the same code as inside that function, except passing a different list of vocabularies. I'm guessing that a vocabulary that is needed to get these labels does not get loaded inside `getLabelNameResolver`. I would like you to check the reason and try to fix it inside that function.

scripts/apps/archive/directives/MediaMetadata.ts
```diff
@@ -14,8 +14,12 @@ export function MediaMetadata(userList, archiveService, metadata, vocabularies)
             scope.$watch('item', reloadData);
             scope.isCorrectionWorkflowEnabled = appConfig?.corrections_workflow;
 
-            vocabularies.getVocabularies().then((vocabulariesCollection) => {
-                scope.label = (id) => getLabelForFieldId(id, vocabulariesCollection);
+            scope.label = (id) => id;
```
tomaskikutis, line 17: It would be better to add a "loading" property which would be `true` until the real `scope.label` is set and to add `ng-if` in the template so `scope.label` would only be called after it was set. By doing that we would also avoid this dummy `scope.label` function which doesn't really resolve the label, but only returns the ID.

scripts/apps/archive/directives/MediaMetadata.ts
```diff
@@ -17,7 +17,9 @@ export function MediaMetadata(userList, archiveService, metadata) {
             scope.label = (id) => id;
 
             getLabelNameResolver().then((_getLabelForFieldId) => {
-                scope.label = _getLabelForFieldId;
+                $timeout(() => {
```
petrjasek, line 20: it would be probably better to use `scope.$applyAsync` to avoid triggering digest on the whole app

tomaskikutis: @petrjasek could you elaborate on `$apply` vs `$applyAsync`? If I understand correctly, `applyAsync` "debounces" the function call, but works the same, no? Can you choose to trigger digest to the whole app vs a section?

petrjasek: `$apply` only works if you trigger it outside angular digest, like in some jquery on click handler. `$applyAsync` will work also if it's called during digest, scheduling the call to next iteration. both can run on the directive scope only, so both are fine. problem is `$timeout` which triggers global digest

tomaskikutis: I see. I didn't know that `$timeout` triggers a global digest. I was chatting with @GyanP and he said it worked to call `scope.$apply` after setting `loading` to false. Timeout won't be needed then.

scripts/apps/archive/directives/MediaMetadata.ts
```diff
@@ -64,6 +63,12 @@ export function MediaMetadata(userList, archiveService, metadata, $timeout) {
 
                 return term.name;
             };
+
+            scope.$watch('label', (newValue) => {
+                if (newValue) {
+                    scope.loading = false;
```
tomaskikutis, line 69: Shouldn't this be set at the same place where `scope.label = _getLabelForFieldId;`? `$watch` wouldn't be needed then which would make the code simpler.

tomaskikutis (approved):

PR #3747, opened 2020-12-23
update selectUser component

tomaskikutis:
SDESK-5692

REQUIRES: https://github.com/superdesk/superdesk-ui-framework/pull/531


scripts/core/ui/components/SelectUser.tsx
```diff
@@ -1,120 +1,139 @@
...
+        this.abortController = null;
     }
 
     componentDidMount() {
         this._mounted = true;
 
-        this.queryUsers();
+        if (this.props.selectedUserId != null) {
+            dataApi.findOne<IUser>('users', this.props.selectedUserId).then((selectedUser) => {
+                this.setState({selectedUser});
```
petrjasek, line 34: in theory this could get unmounted meanwhile

scripts/core/ui/components/SelectUser.tsx
```diff
@@ -1,120 +1,139 @@
...
+                            }
+                            : {},
+                    );
+
+                    return new Promise((resolve) => {
+                        fetch(`${appConfig.server.url}/users?where=${query}&max_results=50`, {
+                            signal: this.abortController.signal,
+                            method: 'GET',
+                            headers: {
+                                'Authorization': ng.get('session').token,
```
petrjasek, line 88: I would expect to have some helper for this, I mean for the whole fetch part

scripts/core/ui/components/SelectUser.tsx
```diff
@@ -1,120 +1,139 @@
...
+                                <div>
+                                    {gettext('Select a user')}
+                                </div>
+                            )
+                            : (
+                                <div style={{display: 'flex', alignItems: 'center'}}>
+                                    <UserAvatar user={user} displayStatus={true} />
+                                    <div style={{marginLeft: 14, padding: '4px 0'}}>
+                                        <div>{user.display_name}</div>
+                                        <div style={{fontSize: 12}}>@{user.username}</div>
```
petrjasek, line 123: not sure why hardcoding font size? would that be accesible?

scripts/core/ui/components/SelectUser.tsx
```diff
@@ -1,120 +1,139 @@
...
+                                        },
+                                    },
+                                    {
+                                        username: {
+                                            $regex: searchString,
+                                            $options: '-i',
+                                        },
+                                    },
+                                ],
+                            }
```
petrjasek, line 79: seems like we have multiple versions of doing this query, would be good to put this in a helper and make it consistent everywhere, we're searching also `email/first_name/last_name` in other places

scripts/core/ui/components/SelectUser.tsx
tomaskikutis: What do you mean by hard-coding? Using pixel values or using Inline styles? Inline styles work the same as styles from the style sheet file and will be accessible.

This is a self-contained component so even if I moved them to a style sheet file, those styles wouldn't be used anywhere else, so I don't see a reason to move them.

scripts/core/SuperdeskReactComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+        this.abortControllers = [];
+
+        this.asyncHelpers = {
+            fetch: (input: RequestInfo, init?: RequestInit): Promise<Response> => {
+                const controller = new AbortController();
+
+                this.abortControllers.push(controller);
+
+                // wrapping in another promise in order to handle AbortError in this component.
+                return new Promise((resolve, reject) => {
```
petrjasek, line 28: wouldn't this work the same if you do `return fetch(...)` ?

tomaskikutis: Nope. If a rejected promise is not returned in the catch block, the next `.then` gets executed.

```
new Promise(() => {
    throw new Error('aaa');
})
    .then(() => {
        console.log('log1');
    })
    .catch((err) => {
        console.log('catch', err.message);
    })
    .then(() => {
        console.log('log2');
    })
```

```
> catch aaa
> log2
```

If abort error occurs I want the promise chain to stop.

scripts/core/ui/components/SelectUser.tsx
petrjasek: I mean the absolute size using pixel value

scripts/core/SuperdeskReactComponent.tsx
petrjasek: isn't this how promises work? you can return `Promise.reject` in the catch to stop the chain if you want

tomaskikutis: It is how they work, and I don't like it :smile: If I return `Promise.reject`, it will display an error in the console about an unhandled rejection and will likely send it to sentry too which shouldn't be done. Code that is behaving correctly should not produce errors.

I could add a `catch` handler when consuming the method like this:
```
componentDidMount() {
    if (this.props.selectedUserId != null) {
        this.asyncHelpers.httpRequestJsonLocal<IUser>({
            method: 'GET',
            path: `/users/${this.props.selectedUserId}`,
        }).then((selectedUser) => {
            this.setState({selectedUser});
        }).catch(() => {
            // < ----------------------------------------------------------------
        });
    }
}
```

But this goes against the encapsulation principle - abort error is an implementation detail of `asyncHelpers.httpRequestJsonLocal` and implementation details should not escape the abstraction layer. It would also add boilerplate which the component is created to avoid.

petrjasek: np, makes sense. would be good to avoid that duplication:
```ts
const promiseWrapper = (promise) => new Promise((resolve, reject) => {
    promise.then(resolve)
        .catch((err) => ...);
};

return promiseWrapper(fetch(...));
return promiseWrapper(httpRequestJsonLocal...);
```

scripts/core/ui/components/SelectUser.tsx
tomaskikutis: You are right, preferred font size set in the browser is ignored when using pixels for font-sizes. I've updated it to use rem, but it isn't helping at the moment because there's a CSS rule that sets font-size to `10px` for the `html` tag, so changing that browser preference does nothing for Superdesk. Changing that 10px rule to a percent value would fix it in the sense that browser preference would be respected, but our styles [wouldn't look nice](https://user-images.githubusercontent.com/2076953/104027170-4c935d80-51c7-11eb-87df-be6d3a1674c3.png).
I'll create a separate ticket to address this.

tomaskikutis: Added SDESK-5742

petrjasek (approved):

PR #3749, opened 2020-12-29
Improve expiry options for the authentication token of the Content API in the subscriber config. [SDESK-5707]

GyanP:


scripts/apps/publish/views/subscribers.html
```diff
@@ -265,12 +268,9 @@ <h3 class="modal__heading" translate ng-hide="subscriber._id">Add New Subscriber
...
                                             <select class="sd-line-input__select"
                                                 ng-model="tokens.ttl"
+                                                ng-disabled="tokens.neverExpire"
                                                 title="{{ :: 'Token expiry' | translate }}">
-                                                <option value="7" translate>1 week</option>
-                                                <option value="14" translate>2 weeks</option>
-                                                <option value="30" translate>1 month</option>
-                                                <option value="180" translate>6 months</option>
-                                                <option value="365" translate>1 year</option>
+                                                <option ng-repeat="(key, value) in tokens.days" value="{{value}}" translate>{{key}}</option>
```
pablopunk, line 273: now that values are dynamic we lose the `translate` feature, don't we?

GyanP: Yes, you are right.

pablopunk (approved):

scripts/apps/publish/controllers/SubscriberTokenController.ts
```diff
@@ -52,8 +52,7 @@ export function SubscriberTokenController($scope, api, $rootScope) {
     this.generate = (ttl) =>
         api.save('subscriber_token', {
             subscriber: subscriber._id,
-            expiry: !this.neverExpire ? expiry(ttl) : null,
-            never_expire: this.neverExpire,
+            expiry_days: this.neverExpire ? null : ttl,
```
petrjasek, line 55: I think this shouldn't be ever `null` but rather `0` for no expiry

scripts/apps/publish/views/subscribers.html
```diff
@@ -236,8 +236,11 @@ <h3 class="modal__heading" translate ng-hide="subscriber._id">Add New Subscriber
...
                                     <div class="sd-list-item__column sd-list-item__column--grow sd-list-item__column--no-border">
                                         <div class="sd-list-item__row">
-                                            <span class="sd-overflow-ellipsis sd-list-item--element-grow"><code contenteditable="true" spellcheck="false">{{ :: token._id}}</code></span>
-                                            <span class="sd-list-item__text-label">{{ :: 'expires:' | translate }} <time sd-datetime data-date="token.expiry" data-from-now="true"></time></span>
+                                            <span class="sd-overflow-ellipsis sd-list-item--element-grow"><code contenteditable="false" spellcheck="false">{{ :: token._id}}</code></span>
+                                            <span class="sd-list-item__text-label">{{ :: 'expires:' | translate }}
+                                                <time ng-if="token.expiry" sd-datetime data-date="token.expiry" data-from-now="true"></time>
+                                                <span ng-if="token.never_expire" translate>Never Expire</span>
```
petrjasek, line 242: I guess here it should check `ng-if="token.expiry == null"`

petrjasek: and it might say just `Never` there is already `expires:` before

PR #3751, opened 2021-01-04
SDESK-5712 ADD:  ability to force unlock asset.

dev-tsoni:
Ability to force unlock asset having privilege to manage_sams_assets.

### DIFF
--- a/scripts/extensions/sams/src/components/assets/assetListPanel.tsx
+++ b/scripts/extensions/sams/src/components/assets/assetListPanel.tsx
@@ -20,7 +20,6 @@ interface IProps {
     onItemClicked(asset: IAssetItem): void;
     onItemDoubleClicked?(asset: IAssetItem): void;
     updateSelectedAssetIds(asset: IAssetItem): void;
-    forceUnlockAsset?(asset: IAssetItem): void;
 }
 
 export class AssetListPanel extends React.PureComponent<IProps> {
@@ -29,7 +28,6 @@ export class AssetListPanel extends React.PureComponent<IProps> {
         this.onItemClick = this.onItemClick.bind(this);
         this.onItemDoubleClick = this.onItemDoubleClick.bind(this);
         this.onUpdateSelectedAssetIds = this.onUpdateSelectedAssetIds.bind(this);
-        this.onForceUnlock = this.onForceUnlock.bind(this);
     }
 
     onItemClick(asset: IAssetItem) {
@@ -44,10 +42,6 @@ export class AssetListPanel extends React.PureComponent<IProps> {
 
     onUpdateSelectedAssetIds(asset: IAssetItem) {
         this.props.updateSelectedAssetIds(asset);
-    }
-
-    onForceUnlock(asset: IAssetItem) {
-        this.props.forceUnlockAsset!(asset);
     }
 
     render() {
--- a/scripts/extensions/sams/src/components/assets/assetListPanel.tsx
+++ b/scripts/extensions/sams/src/components/assets/assetListPanel.tsx
@@ -2,7 +2,7 @@
 import * as React from 'react';
 
 // Types
-import {IAssetItem, ASSET_LIST_STYLE, IAssetCallback} from '../../interfaces';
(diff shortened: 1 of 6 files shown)


scripts/extensions/sams/src/components/assets/assetListPanel.tsx
```diff
@@ -20,6 +20,7 @@ interface IProps {
     onItemClicked(asset: IAssetItem): void;
     onItemDoubleClicked?(asset: IAssetItem): void;
     updateSelectedAssetIds(asset: IAssetItem): void;
+    forceUnlockAsset?(asset: IAssetItem): void;
```
MarkLark86, line 23: Is there a scenario when `forceUnlockAsset` is not provided to `AssetListPanel` on render? (as the `?` tells the compiler this value is optional).

Otherwise the `?` can be removed from `IProps` and `this.props.forceUnlockAsset!(asset)` can be changed to `this.props.forceUnlockAsset(asset)`

scripts/extensions/sams/src/components/assets/assetListPanel.tsx
```diff
@@ -44,10 +46,21 @@ export class AssetListPanel extends React.PureComponent<IProps> {
...
+    onForceUnlock(asset: IAssetItem) {
+        this.props.forceUnlockAsset!(asset);
+    }
+
     render() {
         const {gettext} = superdeskApi.localization;
         const {assertNever} = superdeskApi.helpers;
 
+        if (superdeskApi.privileges.hasPrivilege('sams_manage_assets')) {
+            this.props.actions!.push({
```
MarkLark86, line 58: React properties should not be modified directly.

The `SamsWorkspaceComponent` already constructs a list of actions to provide, it might be better to manage these list of actions there instead.

scripts/extensions/sams/src/utils/assets.ts
```diff
@@ -40,6 +40,14 @@ export function getAction(assetCallback: IAssetCallback): IAssetAction {
             onSelect: assetCallback.onSelect,
             isAllowed: (asset) => asset._created != null,
         };
+    case ASSET_ACTIONS.FORCE_UNLOCK:
+        return {
+            id: ASSET_ACTIONS.FORCE_UNLOCK,
+            label: gettext('Force Unlock'),
+            icon: 'unlocked',
+            onSelect: assetCallback.onSelect,
+            isAllowed: (asset) => asset._created != null,
```
MarkLark86, line 49: `isAllowed` should probably use `isAssetLocked` as well so that we only show this action on locked items

scripts/extensions/sams/src/components/assets/assetPreviewPanel.tsx
```diff
@@ -83,6 +85,29 @@ export class AssetPreviewPanelComponent extends React.PureComponent<IProps> {
...
+                onSelect: this.onDownloadSingleAssetCompressedBinary,
+            },
+            {
+                action: ASSET_ACTIONS.DELETE,
+                onSelect: this.onDeleteAsset,
+            },
+            ];
+
+        if (superdeskApi.privileges.hasPrivilege('sams_manage_assets')) {
+            actions!.push({
```
MarkLark86, line 104: The `!` should not be required here, as `actions` is initialised with a value directly before this statement

MarkLark86 (changes requested):

scripts/extensions/sams/src/components/assets/assetListPanel.tsx
dev-tsoni: Yes it is being used in selectAssetModal.

MarkLark86: Now that this action is controlled by `SamsWorkspaceComponent` this can be removed, along with the class level function `onForceUnlock`

dev-tsoni: Removed :v:  .

MarkLark86 (approved):

PR #3756, opened 2021-01-12
update protractor

petrjasek:


package.json
```diff
@@ -147,7 +146,7 @@
     "karma-ng-html2js-preprocessor": "^1.0.0",
     "karma-sourcemap-loader": "^0.3.7",
     "karma-webpack": "^2.0.13",
-    "protractor": "5.4.2",
+    "protractor": "^7.0.0",
```
tomaskikutis, line 149: Could you lock the version?

Why the update, has something stopped working?

tomaskikutis (approved):

package.json
petrjasek: I couldn't run it locally, it was always complaining about webdriver not being updated, that seems to work better now.
the version is locked in the lock file.

tomaskikutis: Yep, I had issues with webdriver not being updated too and running the update wouldn't help because there would be 2 versions of webdriver. One as a root module and one in protractor.

tomaskikutis (approved):

PR #3757, opened 2021-01-12
avoid watching all files in extensions when running karma

petrjasek:


karma.conf.js
```diff
@@ -35,6 +35,9 @@ module.exports = function(config) {
             debug: false,
             progress: false,
             // quiet: true
+            watchOptions: {
+                ignored: /extensions/,
```
tomaskikutis, line 39: is this a regex rule that would ignore any file/folder that has "extensions" in the path? Could you set a relative path instead that wouldn't affect multiple files/directories? There may be none now, but I would go with a more stable solution anyway

tomaskikutis (approved):

PR #3759, opened 2021-01-13
SDESK-5719 ADD: Confirmation window for deleting assets.

dev-tsoni:


scripts/extensions/sams/src/store/assets/actions.ts
```diff
@@ -253,16 +253,32 @@ export function deleteAsset(asset: IAssetItem): IThunkAction<void> {
...
+            const selectedAssets = getSelectedAssetItems(getState());
+            const assetName = (selectedAssets.length === 1) ? selectedAssets[0].name : undefined;
+
+            return openDeleteConfirmationModal(assetName, selectedAssets.length)
+                .then((response: boolean) => {
+                    if (response === true) {
+                        selectedAssets.map((selectedAsset) =>
+                            dispatch(deleteAsset(selectedAsset))
+                                .then(() => {
+                                    dispatch(closeMultiActionBar());
```
MarkLark86, line 275: Won't this dispatch the `closeMultiActionBar` for every Asset deleted? If that is the case, It might be better to run this once after all Promises have returned. i.e.
```js
return Promise.all(
    selectedAssets.map(
        (selectedAsset) => dispatch(deleteAsset(selectedAsset))
    )
).finally(() => {
    dispatch(closeMultiActionBar())
});
```

MarkLark86 (approved):

PR #3763, opened 2021-01-15
fix(assignments): Fixed issue with assignments calculation

pavlovicnemanja:


scripts/apps/master-desk/components/AssignmentsComponent.tsx
```diff
@@ -69,7 +69,10 @@ export class AssignmentsComponent extends React.Component<IProps, IState> {
         let total = 0;
 
         if (this.state.deskData && this.state.deskData[desk._id]) {
-            this.state.deskData[desk._id].forEach((item) => total += item ? item.count : 0);
+            this.state.deskData[desk._id].forEach((item) =>
+                this.assignmentStages.some((stage) =>
```
tomaskikutis, line 73: Why are you using `some` when its return value is never used? `some` stops iterating if `true` is returned. If you only care about one item, it would be more clear to use `find`.

scripts/apps/master-desk/components/AssignmentsComponent.tsx
```diff
@@ -70,7 +70,7 @@ export class AssignmentsComponent extends React.Component<IProps, IState> {
 
         if (this.state.deskData && this.state.deskData[desk._id]) {
             this.state.deskData[desk._id].forEach((item) =>
-                this.assignmentStages.some((stage) =>
+                this.assignmentStages.find((stage) =>
```
tomaskikutis, line 73: Can you also simplify it and split it in 2 steps? First find the assignment stage(without modifying anything) and then use that stage to increment the total.

tomaskikutis (approved):

PR #3764, opened 2021-01-18
update ui-framework to 2.3.1

MarkLark86:
The changes in the ui-framework are needed for Planning 2.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,8 @@
     "react-test-renderer": "^16.13.1",
     "request": "^2.88.2",
     "superdesk-code-style": "1.5.0",
-    "typescript-eslint-parser": "^18.0.0"
+    "typescript-eslint-parser": "^18.0.0",
+    "webdriver-manager": "12.1.7"
   },
   "resolutions": {
     "moment": "2.20.1"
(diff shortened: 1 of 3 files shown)


tomaskikutis:
It should also be updated for the [dateTimeField extension](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/extensions/datetimeField/package.json#L17)

MarkLark86:
I had to add `webdriver-manager` back to `devDependencies` as e2e tests were failing to run, with the exception being raised here: https://github.com/superdesk/superdesk-client-core/blob/develop/run-end-to-end-tests.js#L6

petrjasek (approved):

package.json
```diff
@@ -152,7 +152,8 @@
     "react-test-renderer": "^16.13.1",
     "request": "^2.88.2",
     "superdesk-code-style": "1.5.0",
-    "typescript-eslint-parser": "^18.0.0"
+    "typescript-eslint-parser": "^18.0.0",
+    "webdriver-manager": "12.1.7"
```
tomaskikutis, line 156: Why are you adding "webdriver-manager"? I see it was removed not long ago in #3756.

tomaskikutis: Never mind, I see you already added a comment in the PR regarding this.

PR #3765, opened 2021-01-18
FIX: Errors when saving desk after changing stage details. [SDESK-5683]

GyanP:


scripts/apps/desks/controllers/DeskConfigController.ts
```diff
@@ -95,17 +95,20 @@ export function DeskConfigController($scope, $controller, notify, desks, modal)
...
-                    _.merge($scope.desk.orig, res);
-                    return true;
-                }, (response) => {
-                    if (angular.isDefined(response.data._message)) {
-                        notify.error(gettext('Error: ' + response.data._message));
-                    } else {
-                        notify.error(gettext('There was a problem, desk not saved. Refresh Desks.'));
-                    }
-                    return false;
+                // First fetch the latest updated desk and then save it.
```
petrjasek, line 98: so the desk got updated meanwhile by other user or by current one? it seems more like this should be handled in the place where desk got updated, otherwise you might be overriding other user changes you haven't seen

tomaskikutis (approved):

PR #3774, opened 2021-01-22
Allow search in labels dropdown

pablopunk:
SDBELGA-439

### DIFF
--- a/package.json
+++ b/package.json
@@ -153,7 +153,7 @@
     "request": "^2.88.2",
     "superdesk-code-style": "1.5.0",
     "typescript-eslint-parser": "^18.0.0",
-    "webdriver-manager": "12.1.7"
+    "webdriver-manager": "12.1.8"
   },
   "resolutions": {
     "moment": "2.20.1"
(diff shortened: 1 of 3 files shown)


tomaskikutis (approved):

package.json
```diff
@@ -176,7 +176,7 @@
     "start-test-server": "cd test-server && docker-compose build && docker-compose up -d",
     "stop-test-server": "cd test-server && docker-compose stop",
     "start-client-server": "http-server dist -p 9000 -s &",
-    "stop-client-server": "fuser -k 9000/tcp",
+    "stop-client-server": "([[ \"$(uname)\" == \"Darwin\" ]] && lsof -i tcp:9000 2>/dev/null | grep -v PID | awk '{print $2}') | xargs kill || fuser -k 9000/tcp",
```
tomaskikutis, line 179: I'd keep this either linux only or write a proper cross platform script using node/python that would be really cross platform including windows.

pablopunk: why linux only? we have developers using mac and linux, I don't think we need to execute e2e tests on windows

tomaskikutis: Because most developers are using linux and our infrastructure code is not cross platform. It's easier to maintain a linux-only(or any other single platform) setup. I think you are the only one using a mac for working on client core, so you could instead add a global alias like `@kille2eserver` so it works for you and we don't add extra code to the project that would need to be maintained.

tomaskikutis (approved):

PR #3776, opened 2021-01-25
Accessibility improvements for articles list

tomaskikutis:
SDESK-5764


package.json
```diff
@@ -29,6 +29,7 @@
   "main": "scripts/index.js",
   "dependencies": {
     "@metadata/exif": "github:superdesk/exif#431066d",
+    "@superdesk/end-to-end-testing-helpers": "1.0.7",
```
pablopunk, line 32: shouldn't this be on `devDependencies`?

tomaskikutis: There are many entries that should be in `devDependencies` - webpack, webpack-dev-server, types, grunt packages. The problem is that we publish source to npm and not a bundle. After fetching that source, parent repository then uses the build scripts of this repository. When packages are installed, `devDependencies` aren't installed and it breaks the build. I'm not 100% sure about this case if it would work in `devDependencies`, but it was in `dependencies` section before and I left it here after updating the package name.

pablopunk: yeah I remember that I just thought we don't really use e2e helpers in client repos

pablopunk (approved):

spec/content_spec.ts
```diff
@@ -30,7 +30,7 @@ describe('content', () => {
     // wait a bit after sending keys to body
     function pressKey(key) {
         browser.actions().sendKeys(key).perform();
-        browser.sleep(50);
+        browser.sleep(1000);
```
pablopunk, line 33: is it necessary to be that long? don't we have tests where `pressKey` is called multiple times?

pablopunk (approved):

spec/content_spec.ts
tomaskikutis: Probably not. I think I added this when testing a flaky test and if I remember correctly it didn't help. I'll try restoring it.

pablopunk (approved):

PR #3777, opened 2021-01-25
avoid category_label, should be just category

petrjasek:
SDESK-5750


tomaskikutis:
As I remember `category_label` was added to get a translated category name. I assume now the `category` itself would be already translated?

petrjasek:
@tomaskikutis that's right

tomaskikutis (approved):

PR #3778, opened 2021-01-27
SDESK-5679 ADD: Error message while adding duplicate set name

dev-tsoni:
Added error while creating a duplicate set.


scripts/extensions/sams/src/utils/api.ts
```diff
@@ -30,9 +30,17 @@ export function getApiErrorMessage(error: IAPIError): string {
         return API_ERRORS[error.error](error);
     }
     if (error.errors?.name != null) {
```
MarkLark86, line 32: This will only respond to validation errors with the `name` field. (see https://superams.readthedocs.io/en/latest/errors/resource.html for a reference to this type of error).

It would also be better to move this error handling to the `API_ERRORS` dictionary under the `04001` error code, so that we can easily determine what error codes are handled by superdesk.

scripts/extensions/sams/src/utils/api.ts
```diff
@@ -30,9 +30,17 @@ export function getApiErrorMessage(error: IAPIError): string {
...
-        return superdeskApi.localization.gettext('Error[{{number}}]: Name requried', {
-            number: error.error,
-        });
+        for (let key of error.errors.name) {
+            if (key === 'unique') {
+                return superdeskApi.localization.gettext('Error[{{number}}]: Name not unique', {
+                    number: error.error,
+                });
+            } else if (key === 'required') {
```
MarkLark86, line 38: There should be another `else` statement here to catch all other types of validation errors, using the `key` as the type of validation error

MarkLark86 (changes requested):

scripts/extensions/sams/src/utils/api.ts
MarkLark86: The `error.errors` dictionary contains a mapping where the key is the name of the field that failed validation. Here `name` is the field that failed. This dictionary could contain other fields as well. So you would need to iterate over the keys as well

MarkLark86 (approved):

PR #3779, opened 2021-01-27
Fix hyperlink to desk after fetching ingested item

pablopunk:
SDESK-5747

### DIFF
--- a/scripts/apps/search/components/FetchedDesksInfo.tsx
+++ b/scripts/apps/search/components/FetchedDesksInfo.tsx
@@ -62,35 +62,30 @@ export class FetchedDesksInfo extends React.Component<any, any> {
     render() {
         const items = [];
 
-        items.push(React.createElement('dt', {
-            key: 'dt',
-            style: {paddingRight: '5px'},
-        }, gettext('fetched in')));
-
-        if (this.state.desks.length) {
-            const desk = this.state.desks[0];
-            const name = this.formatDeskName(desk.desk.name);
-
-            items.push(
-                <dd key="dd1">
-                    {
-                        desk.isUserDeskMember ?
-                            <a onClick={this.openDesk(desk)}>{name}</a> :
-                            <span className="container">{name}</span>
-                    }
-                </dd>,
-            );
-
-            if (this.state.desks.length > 1) {
-                items.push(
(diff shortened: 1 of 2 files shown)


scripts/apps/search/components/FetchedDesksInfo.tsx
```diff
@@ -71,22 +71,27 @@ export class FetchedDesksInfo extends React.Component<any, any> {
...
 
-            items.push(React.createElement('dd', {key: 'dd1'}, desk.isUserDeskMember ?
-                React.createElement('a', {onClick: this.openDesk(desk)}, name) :
-                React.createElement('span', {className: 'container'}, name),
-            ));
+            items.push(
+                <dd key="dd1">
+                    {
+                        desk.isUserDeskMember ?
+                            <a onClick={this.openDesk(desk)}>{name}</a> :
```
tomaskikutis, line 78: If there's no `href` it should be a button. Besides being semantically incorrect it also breaks keyboard navigation. Links without `href` attribute don't receive focus.

scripts/apps/search/components/FetchedDesksInfo.tsx
```diff
@@ -71,22 +71,27 @@ export class FetchedDesksInfo extends React.Component<any, any> {
             const desk = this.state.desks[0];
             const name = this.formatDeskName(desk.desk.name);
 
-            items.push(React.createElement('dd', {key: 'dd1'}, desk.isUserDeskMember ?
-                React.createElement('a', {onClick: this.openDesk(desk)}, name) :
-                React.createElement('span', {className: 'container'}, name),
-            ));
+            items.push(
```
tomaskikutis, line 74: It would be cleaner to "interpolate" inside JSX instead of creating a separate array holding JSX items. Since there is only one item, keys wouldn't be needed either.

```
return (
            <div>
                <dl>
                    {
                        desk.isUserDeskMember
                            ? (<a onClick={this.openDesk(desk)}>{name}</a>)
                            : (<span className="container">{name}</span>)
                    }
                </dl>
            </div>
        );
```

pablopunk:
@tomaskikutis my bad, DesksDropdown is already a <dd>

tomaskikutis (approved):

PR #3780, opened 2021-01-28
Make ISubject['source'] optional

pablopunk:


tomaskikutis (approved):

scripts/extensions/auto-tagging-widget/src/data-transformations.ts
```diff
@@ -63,7 +63,7 @@ export function getExistingTags(article: IArticle): IServerResponse {
                         name,
                         description,
                         qcode,
-                        source,
+                        source: source ?? '',
```
tomaskikutis, line 66: It's not a good practice to use fake values like here. Empty string is not a valid source. It would be much better to simply make it optional in other interfaces in this extension.

tomaskikutis (approved):

PR #3781, opened 2021-01-29
SDESK-5744 ADD: Sams sets to stay in sync with database.

dev-tsoni:


scripts/extensions/sams/src/extension.ts
```diff
@@ -12,6 +12,7 @@ const extension: IExtension = {
 
         Object.assign(superdeskApi, superdesk);
 
+        // Then using the following to test (inside the extension.activate function):
```
MarkLark86, line 15: This comment is no longer needed

scripts/extensions/sams/src/notifications/index.ts
```diff
@@ -0,0 +1,30 @@
+import {superdeskApi} from '../apis';
+import {onSetCreated, onSetUpdated, onSetDeleted} from './sets';
+
+const websocketNotificationMap: {[key: string]: (event: any) => void} = {
```
MarkLark86, line 4: It would be good to have these callback functions strongly typed. i.e. add the following to the `interfaces.ts` file:

```ts
export interface ISAMSBaseEvent {
    // Every event from SAMS should contain the following
    user_id: string;
    session_id: string;
    extension: 'sams';

    // These attributes are provided with most events from SAMS
    item_id?: string;
    _etag?: string;
}

export type ISAMSWebsocketEvent = CustomEvent<ISAMSEvent>;
```

You can then use `ISAMSWebsocketEvent` in `websocketNotificationMap` and in the function signature of these callbacks (i.e. in `onSetCreated`, `onSetUpdated` and `onSetDeleted`)

scripts/extensions/sams/src/notifications/index.ts
```diff
@@ -0,0 +1,30 @@
...
+    'sams:set:created': onSetCreated,
+    'sams:set:updated': onSetUpdated,
+    'sams:set:deleted': onSetDeleted,
+};
+
+let websocketDeregistrationArray: Array<() => void> = [];
+
+export function registerWebsocketNotifications() {
+    if (websocketDeregistrationArray.length > 0) {
+        return;
```
MarkLark86, line 14: Add a comment here as to why we're returning early without registering any listeners

scripts/extensions/sams/src/notifications/sets.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export function onSetUpdated(event: any) {
+    const {notify} = superdeskApi.ui;
+    const store = getStore();
+    const user_id = superdeskApi.session.getCurrentUserId();
+
+    store?.dispatch<any>(loadSets());
+    if (user_id !== event.detail.extra.user_id) {
+        const item_id = store?.getState().sets.selectedSetId;
```
MarkLark86, line 18: Use selectors when accessing a value from the redux state, i.e. `src/store/sets/selects.ts: getSelectedSetId`

scripts/extensions/sams/src/notifications/sets.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        if (item_id === event.detail.extra.item_id) {
+            store?.dispatch(closeSetContentPanel());
+            notify.info('Set was updated by another user.');
```
MarkLark86, line 22: Missing `gettext`

scripts/extensions/sams/src/notifications/sets.ts
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+}
+
+export function onSetDeleted(event: any) {
+    const {notify} = superdeskApi.ui;
+    const store = getStore();
+    const user_id = superdeskApi.session.getCurrentUserId();
+
+    store?.dispatch<any>(loadSets());
```
MarkLark86, line 32: To reduce server load, we don't need to reload all `Sets`. Instead we can simply remove this `Set` from the redux store, using the action `store/sets/actions.ts: removeSetInStore` instead

scripts/extensions/sams/src/notifications/sets.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (user_id !== event.detail.extra.user_id) {
+        const item_id = store?.getState().sets.selectedSetId;
+
+        if (item_id === event.detail.extra.item_id) {
+            store?.dispatch(closeSetContentPanel());
+            notify.info('Set was deleted by another user.');
```
MarkLark86, line 38: Missing `gettext`

MarkLark86 (changes requested):

MarkLark86 (approved):

PR #3785, opened 2021-02-02
Improve extensions loading

tomaskikutis:
Enables importing superdesk api from extensions synchronously, without having to pass it from `activate` method through factory functions.

Loading from the parent repository would look like this:

```
startApp(
    [
        {
            id: 'helloWorld',
            load: () => import('superdesk-core/scripts/extensions/helloWorld/dist/src/extension').then((res) => res.default),
        }
    ],
    {},
    {},
);
 ```

### DIFF
--- a/scripts/core/register-extensions.tsx
+++ b/scripts/core/register-extensions.tsx
@@ -1,11 +1,11 @@
 import {flatMap, noop} from 'lodash';
 import {getSuperdeskApiImplementation} from './get-superdesk-api-implementation';
 import {AuthoringWorkspaceService} from 'apps/authoring/authoring/services/AuthoringWorkspaceService';
-import {IExtension} from 'superdesk-api';
+import {IExtension, ISuperdesk} from 'superdesk-api';
 import {extensions as extensionsWithActivationResult} from 'appConfig';
 
 export function registerExtensions(
-    extensions: Array<IExtension>,
+    extensionLoaders: Array<{id: string; load(): Promise<IExtension>}>,
     superdesk,
     modal,
     privileges,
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by tomaskikutis** [comment]
Just to have a working example I have set API instances on a `window` which is not ideal since all extensions have access to API instances of other extensions. I've tried setting API instance as an exported constant and later import it from client-core but it's no better in terms of access control and didn't work due to build config(webpack couldn't find things). I think it'd be better to generate a module for each extension in `node_modules` directory and import the API from there. Any other ideas?

File: scripts/core/register-extensions.tsx [hunk 1]
```diff
@@ -16,54 +16,65 @@ export function registerExtensions(
...
+
+            pages.forEach((page) => {
+                superdesk
+                    .activity(page.url, {
+                        label: page.title,
+                        priority: 100,
+                        category: superdesk.MENU_MAIN,
+                        adminTools: false,
+                        controller: noop,
+                        template: '<sd-extension-page></<sd-extension-page>',
```
petrjasek, line 75: seems like typo in the template

petrjasek (commented):
I'm not sure if possibility to access other extension api instance is an issue really, is there any security concern? or is the idea to avoid some potential errors?

tomaskikutis:
I guess I'm just being pedantic. Using `window` is not a good practice in general to structure code, but this is an exception not the rule, so it wouldn't hurt. Regarding security, storing variables on window allows browser extensions to read variables and potentially steal some API or auth tokens. On the other hand, extensions can probably inject a script that would overwrite XMLHttpRequest methods and read anything that is sent over the network anyway.

I guess I can keep the current `window` based implementation then.

pablopunk (approved):

PR #3790, opened 2021-02-05
SDESK-5761 ADD: Sams assets to stay in sync with database.

dev-tsoni:


scripts/extensions/sams/src/notifications/assets.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export function onAssetCreated(event: ISAMSWebsocketEvent) {
+    const store = getStore();
+    const item_id = event.detail.extra.item_id!;
+    const params = getAssetSearchParams(store?.getState());
+    const listStyle = getAssetListStyle(store?.getState());
+
+    return samsApi.assets.query(params, listStyle)
+        .then((response) => {
+            Object.keys(response._items).forEach((element: any) => {
```
MarkLark86, line 28: `response._items` is an `Array<IAssetItem>`, so you can simply run:

```
response._items.forEach((item: IAssetItem) => {
    // logic
})
```

It's also best to avoid the use of `any`. This makes it easier for developers to understand the code, and allows the compiler to pick up on problems.

MarkLark86 (approved):

PR #3791, opened 2021-02-08
feat(master-desk): Removed old master desk

pavlovicnemanja:


scripts/apps/desks/views/main.html
```diff
@@ -1,40 +0,0 @@
...
-
-        <div class="desk-container" ng-if="view === 'tasks'" ng-repeat="desk in desks._items | orderBy: 'name'">
-            <div class="desk" ng-include="'scripts/apps/desks/views/desk-tasks.html'"></div>
-        </div>
-
-        <div class="desk-container" ng-if="view === 'users'" ng-repeat="desk in desks._items | orderBy: 'name'">
-            <div class="desk" ng-include="'scripts/apps/desks/views/desk_users.html'"></div>
-        </div>
-        <div class="desk-container" ng-if="view === 'sluglines'" ng-repeat="desk in desks._items | orderBy: 'name'">
-            <div class="desk" ng-include="'scripts/apps/desks/views/slugline.html'"></div>
```
tomaskikutis, line 27: You should also check these included views(and their child views) and remove them too if they are not used anywhere else.

tomaskikutis: I see you have deleted these views, but not their dependencies that are not used anywhere else.

You should delete it recursively - open the file that you want to delete and see if it depends on something else, a directive for example. See if that directive is used somewhere else - if not mark it for deletion, but also look at the dependencies of that directive and so on until you get to a leaf node in the dependency tree. It may sound like a lot of work, but it will probably be only a few extra files in reality.

tomaskikutis: Looks like you forgot `sd-task-status-items` in `/scripts/apps/desks/views/desk-tasks.html`.

pavlovicnemanja: done 👍

tomaskikutis (approved):

PR #3792, opened 2021-02-08
Fix hyperlink ui with button tag

pablopunk:
SDESK-5747


scripts/apps/search/components/FetchedDesksInfo.styles.scss
```diff
@@ -0,0 +1,11 @@
+@import '~variables.scss';
+
+.fetched-desks-info-component {
+  button {
```
tomaskikutis, line 4: How about adding a class name on a button itself so styles don't depend on element type?

tomaskikutis (approved):

PR #3794, opened 2021-02-09
Display authors in the list view

tomaskikutis:
SDBELGA-442

DEPENDS ON: https://github.com/superdesk/superdesk-ui-framework/pull/540


pablopunk:
conflicts ⚠️ 🔥

scripts/apps/search/components/fields/authors.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+        const rolesByUserId = authors.reduce((acc, item) => {
+            acc[item.userId] = item.roleId;
+
+            return acc;
+        }, {});
+
+        const userIds = authors.map(({userId}) => userId);
+
+        const getUsers = () => this.asyncHelpers.httpRequestJsonLocal<IRestApiResponse<IUser>>({
```
petrjasek, line 41: I think we should handle this in some layer and not send a query for every item in the list, we could debounce that query but keep all the requested ids and then fetch it just once.

styles/sass/tables.scss
```diff
@@ -7,7 +7,7 @@
 // BASE TABLES
 // -----------------
 
-table {
+.table {
```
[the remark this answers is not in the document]
tomaskikutis: I wanted to use a table to show authors and roles, but it had lots of default styles that I would have to overwrite. Instead I changed those rules so they only apply to elements with a class name instead of being global. I also did the same in the UI framework and updated current markup to rely on these classes instead of global styles.

Yep, the rules should be merged, I'll update the PR.

scripts/apps/search/components/fields/authors.tsx
tomaskikutis: I have added such a layer on a separate branch and will submit a PR after this is merged.

petrjasek (approved):

PR #3796, opened 2021-02-10
FIX: Media contacts cannot be saved when Country is added to the address. [SDESK-5786]

GyanP:
https://github.com/superdesk/superdesk-core/pull/2139

### DIFF
--- a/scripts/apps/contacts/components/ItemContainer.tsx
+++ b/scripts/apps/contacts/components/ItemContainer.tsx
@@ -74,7 +74,7 @@ export class ItemContainer extends React.Component<any, any> {
         }
 
         if (item.country) {
-            addressInfo.push(item.country);
+            addressInfo.push(item.country?.name);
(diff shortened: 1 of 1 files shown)


scripts/apps/contacts/components/ItemContainer.tsx
```diff
@@ -74,7 +74,7 @@ export class ItemContainer extends React.Component<any, any> {
         }
 
         if (item.country) {
-            addressInfo.push(item.country);
+            addressInfo.push(item.country?.name);
```
tomaskikutis, line 77: If it doesn't have a `name` property, `undefined` will be pushed which would probably not desired behavior. Modify the condition above instead to check if name exists.

```
if (item.country?.name != null) {
    addressInfo.push(item.country.name);
}
```

tomaskikutis (approved):

PR #3798, opened 2021-02-10
[INGEST] [SETTINGS] Don't assign url_id if it already exists

jerome-poisson:
SDESK-5200


tomaskikutis (approved):
This works, but it might be better to use `!= null` so it checks against both - `null` and `undefined`

jerome-poisson:
@tomaskikutis thanks, I'm updating it

tomaskikutis (approved):

PR #3800, opened 2021-02-10
Revert "Improve extensions loading"

MarkLark86:
Reverts superdesk/superdesk-client-core#3785, due to systemic build failures


tomaskikutis:
What was the error message? I should have told you that you need to remove `id` from the [planning extension](https://github.com/superdesk/superdesk-planning/blob/develop/client/planning-extension/src/extension.ts#L101), but there should not be any other issues.

tomaskikutis:
Registration of extensions also should be updated [like here](https://github.com/superdesk/superdesk/pull/3220/files), but I have submitted PRs for all customer repos.

PR #3802, opened 2021-02-12
use new field to detect if user is online

petrjasek:
SDESK-5809

requires https://github.com/superdesk/superdesk-core/pull/2142

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -278,5 +278,7 @@ function getDefaults(grunt) {
             'pt_BR',
             'pl',
         ],
+
+        userOnlineMinutes: 15,
     };
 }
--- a/scripts/apps/users/controllers/UserListController.ts
+++ b/scripts/apps/users/controllers/UserListController.ts
@@ -1,4 +1,7 @@
+import moment from 'moment';
 import {gettext, getUserSearchMongoQuery} from 'core/utils';
+import {appConfig} from 'appConfig';
+import {serverFormat} from 'core/datetime/datetime';
 
 UserListController.$inject = ['$scope', '$location', 'api', 'lodash', 'session', 'usersService'];
 export function UserListController($scope, $location, api, _, session, usersService) {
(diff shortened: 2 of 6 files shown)


webpack.config.js
```diff
@@ -278,5 +278,7 @@ function getDefaults(grunt) {
             'pt_BR',
             'pl',
         ],
+
+        userOnlineMinutes: 15,
```
tomaskikutis, line 282: What about checking websocket status on the server instead? If the websocket is connected - it would mean the user is online.

tomaskikutis (approved):

webpack.config.js
petrjasek: if you keep browser tab opened you would still appear to be online, while you might not be doing anything for hours

tomaskikutis: Sounds accurate to me. "online" means connected, no?

petrjasek: yep that would mean browser/device is online, but doesn't say much about the user, so not sure how useful that would be actually.

tomaskikutis: It depends what the use case is. In most software "online" indicator is so you know if the user is reachable and can respond. I have the slack tab open and have not done anything there for hours, but I am reachable. Is the use case different here?

petrjasek: lets ask @karelpetrak I don't see the original requirements anywhere, not sure if the slack usecase works well for newsroom

MarkLark86: Some applications use 3 states to determine a users' online status:
1. offline
2. online/active
3. online/inactive

IMO, this `userOnlineMinutes` could also be used on the back-end at some point. Would it be beneficial to have this setting in `settings.py` and exposed to the front-end via the `app.client_config` attribute?

petrjasek: I wasn't sure what would be the usecase on the server. It would make sense to let user set their status maybe, now we're tracking last activity time, but that would be something for next release

PR #3811, opened 2021-02-19
feat(pinning): Save pinned tabs in user preference

pavlovicnemanja:


scripts/apps/authoring/widgets/widgets.ts
```diff
@@ -229,6 +252,51 @@ function WidgetsManagerCtrl(
         }
     };
 
+    $scope.pinWidget = (widget: IWidget) => {
+        if ($scope.pinnedWidget) {
+            $scope.pinnedWidget.pinned = false;
+        }
+
+        if (!superdesk.pinnedWidgetResized && widget && !$scope.pinnedWidget) {
```
tomaskikutis, line 260: Can you use a local variable in the file for  `pinnedWidgetResized` instead of setting it on `superdesk`?

scripts/core/ui/ui.ts
```diff
@@ -842,11 +842,21 @@ function splitterWidget(superdesk, $timeout, $rootScope) {
...
 
-            const workspace = element,
-                authoring = element.next('#authoring-container'),
+            let workspace, authoring, container;
+
+            const initializeContainers = () => {
+                workspace = element ? element :
+                    angular.element('#workspace-container');
+
+                authoring = element.next('#authoring-container');
```
tomaskikutis, line 851: Why are you using `.next()` to get an element by ID? Since IDs are unique, wouldn't `angular.element('#authoring-container')` would achieve the same?

pavlovicnemanja: I was thinking that next should search for smaller amount of files (because it will always be next to it) and `angular.element` will go through complete DOM

tomaskikutis: Getting an element by ID is very fast, you certainly don't need to worry about performance here. I doubt that `angular.element` would check every node. It is probably using native browser methods under the hood like `document.getElementById` / `document.querySelector` which are optimized by the browser.

tomaskikutis (approved):

PR #3816, opened 2021-02-23
Related entities

tomaskikutis:
SDBELGA-442

A static `getRelatedEntities` method can now be added to any react component that is rendering a field in the list view. Before the list view is rendered, it will batch and fetch all related entities and make them available in props of the component.

I didn't convert all fields to use it yet. Even though it batches the ids of entities of the same type and only uses a single request per entity type, the request count would still increase a bit since we have entities of different types. I'm planning to implement a caching layer and only then convert the rest of the components to use `getRelatedEntities`.


scripts/core/itemList/items-list-limited.tsx
```diff
@@ -14,31 +15,47 @@ interface IProps {
...
                 .then((items) => {
-                    this.setState({items});
+                    getRelatedEntities(
+                        items,
+                        this.state.relatedEntities,
+                        this.abortController.signal,
+                    ).then((relatedEntities) => {
+                        this.setState({
+                            items,
+                            relatedEntities: mergeRelatedEntities(this.state.relatedEntities, relatedEntities),
```
petrjasek, line 50: seems like there is always merge happening after return from `getRelatedEntities`, given that those are available in the call I would probably rather do the merge inside `getRelatedEntities` before returning

petrjasek (commented):
seems like this actually puts together fetching of related entities which is pretty generic with the articles lists, could we have this separated and pass config what needs to be fetched as param? there could be a helper for the list doing the same thing as now. and actually this should probably get access to the component which needs the data so it could call it's setState method in case something gets updated later, or is the idea to fetch it once and leave it like that?

tomaskikutis:
> could we have this separated and pass config what needs to be fetched as param?

You mean `getRelatedEntities` function? Yep, I can move out reading component data and pass the result as a parameter.

Regarding updating related entities when they change, I forgot to handle it. Will add websocket listeners.

scripts/apps/search/components/ItemListAngularWrapper.tsx
```diff
@@ -50,6 +65,20 @@ export class ItemListAngularWrapper extends React.Component<IProps, IState> {
...
+        this.eventListenersToRemoveBeforeUnmounting = [];
+
+        this.handleContentChanges = throttleAndCombineArray(
+            (changes) => {
+                getAndMergeRelatedEntitiesUpdated(
+                    this.state.relatedEntities,
+                    changes,
+                    this.abortController.signal,
+                ).then((relatedEntities) => {
+                    this.setState({relatedEntities});
```
petrjasek, line 77: would it actually return the same instance if there were no changes to related entities? and thus setState wouldn't force a re-render on each event

tomaskikutis: A new instance is returned every time. If there are no changes to an entity - websocket events don't fire and this code doesn't run.

petrjasek (approved):

scripts/apps/search/components/ItemListAngularWrapper.tsx
tomaskikutis: Right, it fires when any entities change, not only related ones. As you pointed out there is filtering and in case related entities didn't change, the same instance is returned.

scripts/core/getRelatedEntities.ts
```diff
@@ -0,0 +1,136 @@
...
+        acc[change.resource] = acc[change.resource].add(change.itemId);
+
+        return acc;
+    }, {});
+
+    if (Object.keys(entitiesToFetch).length > 0) {
+        return fetchRelatedEntities(entitiesToFetch, abortSignal)
+            .then((result) => mergeRelatedEntities(currentEntities, result));
+    } else {
+        return Promise.resolve({});
```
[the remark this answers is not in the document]
tomaskikutis: Yep, I have found that the hard way by investigating e2e failtures ;D

PR #3819, opened 2021-02-26
fix pages from extensions not being visible in the menu

petrjasek:
and handle privileges for those.

SDESK-5766

### DIFF
--- a/scripts/extensions/annotationsLibrary/src/extension.tsx
+++ b/scripts/extensions/annotationsLibrary/src/extension.tsx
@@ -81,7 +81,7 @@ var extension: IExtension = {
     activate: (superdesk: ISuperdesk) => {
         const {gettext} = superdesk.localization;
 
-        return Promise.resolve({
+        return superdesk.privileges.getOwnPrivileges().then(() => ({
             contributions: {
                 editor3: {
                     annotationInputTabs: [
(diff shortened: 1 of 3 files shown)


scripts/extensions/annotationsLibrary/src/extension.tsx
```diff
@@ -101,6 +101,7 @@ var extension: IExtension = {
                         title: gettext('Annotations Library'),
                         url: '/annotations-library',
                         component: getAnnotationsLibraryPage(superdesk),
+                        privileges: ['concept_items'],
```
tomaskikutis, line 104: I would like to keep the API minimal and avoid adding properties when possible. Extensions can already check for privileges so we can write it like this:

```
return Promise.resolve({
    contributions: {
        // ...
        pages: hasPrivilege('concept_items') ? [
            {
                title: gettext('Annotations Library'),
                url: '/annotations-library',
                component: getAnnotationsLibraryPage(superdesk),
            },
        ]: [],
    },
});
```

petrjasek: not sure this is possible, extensions is loaded before the user session is established

petrjasek: or this should actually do something like:
```ts
return superdesk.privileges.getOwnPrivileges().then(() => {
    contributions: ...
});
```
?

petrjasek: that seems working, I don't like that boilerplate as using a privilege with page is pretty common, so this would be rather incomplete api for a page in my opinion

tomaskikutis: It's not ideal that retrieving privileges doesn't work synchronously, and some boilerplate is required, but that's easy to fix. I'll do that next time I have some time for code improvements. If you look at all possible [contribution points](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L72) that an extension can plug into, it is likely that a lot of them might need to be limited depending on privileges, but I think it's much cleaner to do it inside the extension rather than implementing it for every type of contribution point.

tomaskikutis (approved):

PR #3821, opened 2021-03-01
Keep old visual state while loading data

tomaskikutis:
SDCP-483

### DIFF
--- a/scripts/core/ArticlesListByQuery.tsx
+++ b/scripts/core/ArticlesListByQuery.tsx
@@ -4,10 +4,10 @@ import React from 'react';
 import {ArticlesListV2} from './ArticlesListV2';
 import {IRestApiResponse, IArticle} from 'superdesk-api';
 import {httpRequestJsonLocal} from './helpers/network';
-import {flatMap, once} from 'lodash';
+import {flatMap} from 'lodash';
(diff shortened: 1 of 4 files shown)


scripts/core/ArticlesListByQuery.tsx
```diff
@@ -106,11 +109,7 @@ class ArticlesListByQueryComponent extends React.PureComponent<IPropsInner, ISta
                     <ArticlesListV2
                         itemCount={itemCount}
                         pageSize={this.props.query.max_results}
-                        loadItems={(from, to) => this.loadItems(from, to).then(({_items}) => {
-                            this.markAsInitializedOnce();
-
-                            return _items;
-                        })}
+                        loadItems={(from, to) => this.loadItems(from, to).then(({_items}) => _items)}
```
petrjasek, line 112: seems like all this does now is `loadItems={this.loadItems}`

petrjasek (approved):

scripts/core/ArticlesListByQuery.tsx
tomaskikutis: It destructures `_items` from the response.

petrjasek: I see now

PR #3823, opened 2021-03-02
As a user I want to see a list of items 'marked for me' in Personal space. [SDBELGA-455]

GyanP:


tomaskikutis:
All code related to "mark for user" functionality should be inside the extension. There's no API at the moment to add sections to personal space, so it should be added. Here's a [code sample](https://user-images.githubusercontent.com/2076953/109635815-16c98000-7b4b-11eb-9e1c-dc8ee715725b.png) of how it would look.

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -53,10 +52,31 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
...
     this.activeFilterTags = {};
 
+    Promise.all(
+        Object.values(extensions)
+            .map(
+                (extension) =>
+                    extension.activationResult?.contributions?.personalSpace?.getSections?.(),
+            ),
+    ).then((res) => {
+        const response = _.flattenDeep(res);
```
tomaskikutis, line 62: It's better to only import the thing you need. `import {flattenDeep} from 'lodash'` That way bundle size is smaller, because it's not importing the entire library.

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -53,10 +52,31 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
     };
     this.activeFilterTags = {};
 
+    Promise.all(
+        Object.values(extensions)
+            .map(
+                (extension) =>
+                    extension.activationResult?.contributions?.personalSpace?.getSections?.(),
```
tomaskikutis, line 59: You probably need a default value here to avoid `undefined` -  `.getSections?.() ?? []`

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -53,10 +52,31 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
...
+            ),
+    ).then((res) => {
+        const response = _.flattenDeep(res);
+
+        response.forEach((resp) => {
+            if (resp?.type === 'markedForMe') {
```
tomaskikutis, line 65: If you use a default that I suggested in the other comment, `resp` will never be null, thus you don't need this condition. Also it's not good that specific strings from an extension are used in the core. The purpose of extensions is decoupling. The code in the core should not make any assumptions about what is in extension code. It should only implement the API.

scripts/extensions/markForUser/src/extension.tsx
```diff
@@ -23,6 +30,23 @@ const extension: IExtension = {
                 globalMenuHorizontal: [getMarkedForMeComponent(superdesk)],
                 articleListItemWidgets: [getDisplayMarkedUserComponent(superdesk)],
                 authoringTopbarWidgets: [getDisplayMarkedUserComponent(superdesk)],
+                personalSpace: {
+                    getSections: () => superdesk.session.getCurrentUser().then((user) => {
+                        const items: Array<IPersonalSpaceSection> = [
+                            {
+                                type: 'markedForMe',
```
tomaskikutis, line 37: 1. It should work to change the value of the `type` and still have it working. It wouldn't be possible now, because that exact string is [hardcoded in the core](https://github.com/superdesk/superdesk-client-core/pull/3823/files#diff-7ee9b9f524733890c3aca8f73439db650f27c47bb417643a8a8cb8c51aa06a2eR110). You should remove that hardcoding and read the data in registered extensions instead.
2. It would be good to rename `type` to `id` if it doesn't do anything else except identification.

scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -719,7 +719,9 @@ export function MonitoringGroup(
                             searchCriteria.source.size = PAGE_SIZE;
                         }
                     }
-                } else if (scope.group != null && scope.group.type === 'personal') {
+                } else if (scope.group != null
+                    && (scope.group.type === 'personal'
+                        || scope.group.type === 'markedForMe')) {
```
tomaskikutis, line 724: When you implement extensions API interfaces in the core, the implementation must no rely on anything in extension code - only the interface of the API. Here you are hardcoding an ID, which means if I changed the ID in the extension, it would stop working.

To avoid that, you could check IDs of all registered extensions for personal section, similarly as you do in `filterQueryByCardType`

scripts/apps/monitoring/services/CardsService.ts
```diff
@@ -7,8 +7,24 @@ import {
     SENT_OUTPUT,
     SCHEDULED_OUTPUT,
 } from 'apps/desks/constants';
-import {appConfig} from 'appConfig';
-import {IMonitoringFilter} from 'superdesk-api';
+import {appConfig, extensions} from 'appConfig';
+import {IMonitoringFilter, IPersonalSpaceSection} from 'superdesk-api';
+
+export function getExtensionSections() {
```
tomaskikutis, line 13: `getSections` isn't returning a promise, so you should not be using `Promise.all` and make this function return an array instead.

scripts/apps/monitoring/services/CardsService.ts
```diff
@@ -76,33 +94,42 @@ export function CardsService(search, session, desks, $location) {
 
     function filterQueryByCardType(query, queryParam, card: ICard) {
         let deskId;
+        const extensionSection = getExtensionSections();
+
+        extensionSection.then((response) => {
+            const data = response.find((res: IPersonalSpaceSection) => res?.id === card.type);
```
tomaskikutis, line 100: > `res?.id`

Is it possible that `res.id` would be `null`?

I thought it wasn't, because compiler forces the extension to set a string and inside `getExtensionSections` you have a default value of an empty array which gets filtered out by `flattenDeep`.

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -48,29 +52,45 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
     };
     this.activeFilterTags = {};
 
+    Promise.all(
```
tomaskikutis, line 55: What about using `getExtensionSections` here?

tomaskikutis (commented):
If you are doing some clean up work like lint fixes or replacing lodash imports it would be easier to review if that was in a separate commit.

GyanP:
> If you are doing some clean up work like lint fixes or replacing lodash imports it would be easier to review if that was in a separate commit.

sure, I will keep that in mind.

scripts/apps/monitoring/services/CardsService.ts
GyanP: Previously I was getting it, but after making the latest changes (API return array), I think I can use flatMap and that will remove the empty array.

scripts/apps/monitoring/controllers/AggregateCtrl.ts
GyanP: Good Catch.

scripts/apps/monitoring/services/CardsService.ts
tomaskikutis:
```
_.flattenDeep([1, [2, 3], []])
> (3) [1, 2, 3]
```

`flattenDeep` removes empty arrays too. You wasn't using a default value before and that's why you were getting it.

scripts/apps/monitoring/services/CardsService.ts
```diff
@@ -7,8 +7,19 @@ import {
     SENT_OUTPUT,
     SCHEDULED_OUTPUT,
 } from 'apps/desks/constants';
-import {appConfig} from 'appConfig';
-import {IMonitoringFilter} from 'superdesk-api';
+import {appConfig, extensions} from 'appConfig';
+import {IMonitoringFilter, IPersonalSpaceSection} from 'superdesk-api';
+
+export function getExtensionSections() {
+    const section = flatMap(
```
tomaskikutis, line 14: It should be `sections` because it's an array. Actually you don't need this variable at all, you can simply return it's value.

scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -710,7 +711,17 @@ export function MonitoringGroup(
              * return {promise} list of items
              */
             function apiquery(searchCriteria, applyProjections) {
+                const personalGroup = ['personal', 'sent'];
```
tomaskikutis, line 714: If it's an array, naming should be in plural - `personalGroups`

scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -710,7 +711,17 @@ export function MonitoringGroup(
...
                 var provider = 'search';
+                const personalSectionIds = [];
```
tomaskikutis, line 716: You can simplify this

`const personalSectionIds = getExtensionSections().map(({id}) => id);`

scripts/apps/monitoring/services/CardsService.ts
GyanP: got it.

scripts/apps/monitoring/services/CardsService.ts
```diff
@@ -76,33 +89,40 @@ export function CardsService(search, session, desks, $location) {
 
     function filterQueryByCardType(query, queryParam, card: ICard) {
         let deskId;
+        const extensionSection = getExtensionSections();
+
+        extensionSection.forEach((response) => {
+            if (response.id === card.type) {
+                query.filter(response.query);
```
tomaskikutis, line 96: `query.filter` should never be called multiple times. It wouldn't be, because you are checking the ID, but it is still confusing that it is called inside a loop. Instead use `find` to get a single item, check if it's found and only call then.

tomaskikutis (approved):

PR #3825, opened 2021-03-02
Fix multi select for bulk actions in refactored views

tomaskikutis:
SDESK-5846

### DIFF
--- a/scripts/core/MultiSelectHoc.tsx
+++ b/scripts/core/MultiSelectHoc.tsx
@@ -1,27 +1,29 @@
 import React from 'react';
 import {OrderedMap, Set} from 'immutable';
 import {
-    IArticle,
     IWebsocketMessage,
     IResourceUpdateEvent,
     IResourceDeletedEvent,
 } from 'superdesk-api';
 import {addWebsocketEventListener} from './notification/notification';
-import {ARTICLE_RELATED_RESOURCE_NAMES} from './constants';
 import {throttleAndCombineSet} from './itemList/throttleAndCombine';
-import {generateTrackByIdentifier} from 'apps/search/services/SearchService';
 
-export interface IMultiSelectOptions {
-    selected: OrderedMap<string, IArticle>;
-    select(item: IArticle): void;
-    selectMultiple(items: OrderedMap<string, IArticle>): void;
-    unselect(id: string): void;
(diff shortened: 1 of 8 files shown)


scripts/core/MultiSelectHoc.tsx
```diff
@@ -86,7 +98,7 @@ export class MultiSelectHoc extends React.PureComponent<IProps, IState> {
     handleContentChanges(resource: string, id: string) {
         // Unselect items that no longer match the query.
 
-        if (ARTICLE_RELATED_RESOURCE_NAMES.includes(resource) && this.state.selected.has(id)) {
+        if (this.props.resourceNames.includes(resource) && this.state.selected.has(id)) {
```
petrjasek, line 101: imo it should be using `getId` when checking `state.selected`, shouldn't it?

tomaskikutis: It's a long story :upside_down_face:

Ideally, I should not need `getId` at all, and would use the `_id` field that is present for every resource.

Since that `_id` field is not unique for published items as we already discussed, it doesn't work here. I'm trying to work around it using `getId` which does the job in most cases.

Here's it's messy, because I assume that IDs received from a websocket notification are the same(have version appended like `generateTrackByIdentifier` does) as `getId` would produce if I had a full object(which I don't).

If IDs from websocket notifications don't have version appended, then I'm a bit out of options. I may be able to restructure `handleContentChanges` so it's passed as a prop from a parent component and could fetch all the items with that duplicate id from the websocket notification and then use `getId`, but that feels like a sandwich of hacks.

The way it's implemented now would only misbehave when users select published items for bulk actions and those published items would be modified by someone else and would no longer match the query that is used to render the list.

It'd be good to fix the root cause - that duplicate ID issue so I could get rid of off `getId` workaround. I ended up not using the fix you did [here](https://github.com/superdesk/superdesk-core/pull/2112), because I realized that even if I add that query parameter, it would not work. It would work for brand new code, but I wanted to use existing components, that have their own child components that don't know about the query parameter and wouldn't work in both scenarios.

petrjasek: I see, we should plan for the fix without any param probably for some future version then.

petrjasek (approved):

PR #3830, opened 2021-03-07
Fix menu items from extensions not being visible

tomaskikutis:
SDESK-5859


MarkLark86 (approved):

petrjasek:
@tomaskikutis in order to add this to 2.2 should we revert https://github.com/superdesk/superdesk-client-core/pull/3828 ?

tomaskikutis:
The fix is not specific to the change how extensions are registered and would work without that PR reverted, but it would be easier to merge if we reverted.

PR #3835, opened 2021-03-10
SDESK-5581 ADD: Tags through Autocomplete in assetEditorPanel.

dev-tsoni:
tags_sams
Adding new Tags through Enter key is not added.

### DIFF
--- a/scripts/extensions/sams/src/components/assets/assetEditor.tsx
+++ b/scripts/extensions/sams/src/components/assets/assetEditor.tsx
@@ -1,6 +1,9 @@
 // External modules
 import * as React from 'react';
 import {connect} from 'react-redux';
+import {OrderedMap} from 'immutable';
+import nextId from 'react-id-generator';
+import {noop} from 'lodash';
 
 // Types
 import {ASSET_STATE, IAssetItem, ISetItem} from '../../interfaces';
@@ -11,7 +14,7 @@ import {superdeskApi} from '../../apis';
 import {getActiveSets} from '../../store/sets/selectors';
 
 // UI
-import {FormLabel, Input, Option, Select} from 'superdesk-ui-framework/react';
+import {FormLabel, Input, Option, Select, Autocomplete} from 'superdesk-ui-framework/react';
 import {FormGroup, FormRow} from '../../ui';
 
 // Utils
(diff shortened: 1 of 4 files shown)


scripts/extensions/sams/src/components/assets/assetEditor.tsx
```diff
@@ -141,6 +169,42 @@ class AssetEditorComponent extends React.PureComponent<IProps> {
                         </FormRow>
                     </FormGroup>
                 )}
+                {this.fieldEnabled('tags') && (
+                    <FormGroup>
+                        <FormRow>
+                            <Autocomplete
+                                key={nextId.name}
```
MarkLark86, line 176: It doesn't look like this `key` is being changed. After clicking on a suggested item from the popup, the text field in Autocomplete should clear.

scripts/extensions/sams/src/components/assets/assetEditor.tsx
```diff
@@ -141,6 +169,42 @@ class AssetEditorComponent extends React.PureComponent<IProps> {
...
+                                label={gettext('Tags')}
+                                value={''}
+                                keyValue="name"
+                                items={[]}
+                                search={(searchString, callback) => {
+                                    let cancelled = false;
+
+                                    httpRequestJsonLocal<IAutoTaggingSearchResult>({
```
MarkLark86, line 184: This might be better off in `src/api/assets.ts`, and accessed in UI components through the api

`samsApi.assets.searchTags(searchString)`

scripts/extensions/sams/src/components/assets/assetPreviewPanel.tsx
```diff
@@ -182,6 +183,12 @@ export class AssetPreviewPanelComponent extends React.PureComponent<IProps> {
                                 <FormLabel text={gettext('Set')} style="light" />
                                 <Text>{this.props.setName}</Text>
                             </FormRow>
+                            <FormRow>
+                                <FormLabel text={gettext('Tags')} style="light" />
+                                {this.props.asset.tags?.map((tag) => (
+                                    <Tag key={this.props.asset?.tags.indexOf(tag)} text={tag.name} onClick={noop} />
```
MarkLark86, line 189: Does the `Tag` UI component have a way to control if the `x` icon can not be shown? As removing a Tag in Preview is not supported

dev-tsoni: I don't think we can remove that currently using any option available.

PR #3837, opened 2021-03-15
avoid refreshing the list for search providers

petrjasek:
external content should not be affected by superdesk events.

SDCP-491

### DIFF
--- a/scripts/apps/search/directives/SearchResults.ts
+++ b/scripts/apps/search/directives/SearchResults.ts
@@ -21,7 +21,7 @@ SearchResults.$inject = [
 const HEX_REG_EXP = /[a-f0-9]{24}/;
 
 function isObjectId(value) {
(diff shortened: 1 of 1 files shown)


scripts/apps/search/directives/SearchResults.ts
```diff
@@ -171,6 +171,11 @@ export function SearchResults(
              * Schedule an update if it's not there yet
              */
             function queryItems(event?, data?) {
+                if (isObjectId(scope.search.repo.search)) {
```
tomaskikutis, line 174: I hope this works for the initial query.

tomaskikutis (approved):

scripts/apps/search/directives/SearchResults.ts
petrjasek: we don't use it for initial query

PR #3840, opened 2021-03-16
feat(iMatrics): Added ability to configure fields and order of fields in superdesk.config.js

pavlovicnemanja:
Example configuration in superdesk.config.js:
```
iMatricsFields: {
    imatrics_category: {
        name: 'Category',
        order: 1,
    },
    imatrics_topic: {
        name: 'Topics',
        order: 2,
    },
    topics: {
        name: 'Media topics',
        order: 3,
    }
}
```

### DIFF
--- a/scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
+++ b/scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
@@ -33,6 +33,13 @@ interface IProps {
     article: IArticle;
 }
 
+interface IIMatricsFields {
+    [key: string]: {
+        name: string;
+        order: number;
+    };
+}
+
 type IEditableData = {original: IAutoTaggingResponse; changes: IAutoTaggingResponse};
 
 interface IState {
@@ -48,6 +55,10 @@ function tagAlreadyExists(data: IEditableData, qcode: string): boolean {
     return data.changes.analysis.has(qcode);
 }
 
+function hasConfig(key: string, iMatricsFields: IIMatricsFields) {
+    return iMatricsFields[key] != null;
+}
+
 export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
     const {preferences} = superdesk;
     const {httpRequestJsonLocal} = superdesk;
@@ -61,7 +72,7 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
     return class AutoTagging extends React.PureComponent<IProps, IState> {
         private isDirty: (a: IAutoTaggingResponse, b: Partial<IAutoTaggingResponse>) => boolean;
         private _mounted: boolean;
-        private iMatricsFields = superdesk.instance.config.iMatricsFields ?? [];
+        private iMatricsFields = superdesk.instance.config.iMatricsFields ?? {};
 
         constructor(props: IProps) {
             super(props);
@@ -171,16 +182,8 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
                 data,
             );
         }
-        getGroupName(group: string) {
-            const {vocabularyLabels} = this.state;
-
-            if (this.iMatricsFields[group]) {
-                return this.iMatricsFields[group].name;
-            } else if (vocabularyLabels && vocabularyLabels.get(group)) {
-                return vocabularyLabels.get(group);
-            }
-
-            return group;
+        getGroupName(group: string, vocabularyLabels: Map<string, string>) {
+            return this.iMatricsFields[group]?.name ?? vocabularyLabels?.get(group) ?? group;
         }
         reload() {
             this.setState({data: 'not-initialized'});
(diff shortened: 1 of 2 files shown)


scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -61,6 +61,7 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
     return class AutoTagging extends React.PureComponent<IProps, IState> {
         private isDirty: (a: IAutoTaggingResponse, b: Partial<IAutoTaggingResponse>) => boolean;
         private _mounted: boolean;
+        private iMatricsFields = superdesk.instance.config.iMatricsFields ?? [];
```
tomaskikutis, line 64: `iMatricsFields` is an object in `superdesk.config.js`, the default should be an object too instead of an array.

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -388,8 +389,7 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
                                 const others = items.filter((tag) => isEntity(tag) === false);
                                 const othersGrouped = others.groupBy((tag) => tag.group.value);
                                 const othersGroupedAndSorted = othersGrouped.sortBy(
-                                    (_, key) => key!.toString().toLocaleLowerCase(),
-                                    (a, b) => a.localeCompare(b),
+                                    (_, key) => this.iMatricsFields[key].order ?? -1,
```
tomaskikutis, line 392: I have added sorting in order to prevent groups switching places after an update from the server. For this to work, sorting has to be deterministic and return the same order given the same items.

It looks like your implementation would only work if all possible fields are configured. Can you improve it so it works both without configuration or with partial configuration(for example 2 fields configured out of total of 5)?

tomaskikutis: The back-end returns an object, not an array so it can't be sorted. We could assume that if you iterate object keys, the order would be the same as in JSON that the object was constructed from, but it will not always be the case.

What you can do is to apply sorting by name, that was implemented before in this file, and sort it again using `iMatricsFields` config.

pavlovicnemanja: yep, that was my idea also. Tnx!

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
petrjasek: it would be probably simpler if it is a list in the config so what you see there is what you get in the authoring,
rather than having to check those order values

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -388,10 +399,20 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
 
                                 const others = items.filter((tag) => isEntity(tag) === false);
                                 const othersGrouped = others.groupBy((tag) => tag.group.value);
-                                const othersGroupedAndSorted = othersGrouped.sortBy(
-                                    (_, key) => this.iMatricsFields[key].order ?? -1,
+
+                                const othersGroupedAndSortedByConfig = othersGrouped
+                                    .filter((_, key) => !!this.iMatricsFields[key])
+                                    .sortBy((_, key) => this.iMatricsFields[key].order,
```
tomaskikutis: You are only passing `comparatorValueMapper` here. It'd be good to pass comparator function too, so it's clear how items are compared. The function should be `(a, b) => a - b`

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -388,10 +399,20 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
...
                                 );
 
+                                const othersGroupedAndSortedNotInConfig = othersGrouped
+                                    .filter(item => !othersGroupedAndSortedByConfig.includes(item))
+                                    .sortBy((_, key) => key!.toString().toLocaleLowerCase(),
```
tomaskikutis, line 409: no need for an exclamation mark here ---> `key!.toString()`

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -388,10 +399,20 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
...
-                                const othersGroupedAndSorted = othersGrouped.sortBy(
-                                    (_, key) => this.iMatricsFields[key].order ?? -1,
+
+                                const othersGroupedAndSortedByConfig = othersGrouped
+                                    .filter((_, key) => !!this.iMatricsFields[key])
+                                    .sortBy((_, key) => this.iMatricsFields[key].order,
                                 );
 
+                                const othersGroupedAndSortedNotInConfig = othersGrouped
+                                    .filter(item => !othersGroupedAndSortedByConfig.includes(item))
```
tomaskikutis: It'd be cleaner to make a `function hasConfig(key) { return this.iMatricsFields[key] != null) }`. That way, you would not need to iterate the list in order to check the value(.includes does the iteration).

The lists will be small here, so it's just a nice to have. If the lists were big, it'd be a performance issue.

tomaskikutis (approved):

PR #3841, opened 2021-03-17
Sdesk 5581 add tags to asset metadata

dev-tsoni:


scripts/core/superdesk-api.d.ts
```diff
@@ -1104,6 +1109,7 @@ declare module 'superdesk-api' {
         patchRaw<T extends IBaseRestApiResponse>(endpoint, id: T['_id'], etag: T['_etag'], patch: Partial<T>): Promise<T>;
         delete<T extends IBaseRestApiResponse>(endpoint, item: T): Promise<void>;
         uploadFileWithProgress<T>(endpoint: string, data: FormData, onProgress: (event: ProgressEvent) => void): Promise<T>;
+        searchTags(endpoint, params?: Dictionary<string, any>): Promise<IAutoTaggingSearchResult>;
```
MarkLark86, line 1112: This API entry is specific to `SAMS`, so it should belong in the `SAMS API` instead:
`/scripts/extensions/sams/src/api/assets.ts`

scripts/extensions/sams/src/components/assets/assetEditor.tsx
```diff
@@ -1,6 +1,9 @@
 // External modules
 import * as React from 'react';
 import {connect} from 'react-redux';
+import {OrderedMap} from 'immutable';
```
MarkLark86, line 4: Why use `OrderedMap` instead of an `Array`

scripts/extensions/sams/src/components/assets/assetEditor.tsx
```diff
@@ -24,11 +27,34 @@ interface IProps {
     sets: Array<ISetItem>;
     fields?: Array<keyof IAssetItem>;
 }
+interface ITagUi {
```
MarkLark86, line 30: You could use `IAssetItem['tags']` here instead of re-declaring the same interface

scripts/extensions/sams/src/components/assets/assetEditor.tsx
```diff
@@ -141,6 +168,41 @@ class AssetEditorComponent extends React.PureComponent<IProps> {
                         </FormRow>
                     </FormGroup>
                 )}
+                {this.fieldEnabled('tags') && (
+                    <FormGroup>
+                        <FormRow>
+                            <Autocomplete
+                                key={nextId.name}
```
MarkLark86, line 175: This `key` needs to change with every update, so that the search box's text is cleared after a user selects a Tag to be added

scripts/extensions/sams/src/components/assets/assetEditor.tsx
```diff
@@ -141,6 +168,41 @@ class AssetEditorComponent extends React.PureComponent<IProps> {
...
+                                label={gettext('Tags')}
+                                value={''}
+                                keyValue="name"
+                                items={[]}
+                                search={(searchString, callback) => {
+                                    let cancelled = false;
+
+                                    superdeskApi.dataApi.searchTags('sams/assets/tags', {
```
MarkLark86, line 183: This would become `samsApi.assets.searchTags(searchString + '*')

scripts/extensions/sams/src/components/assets/uploadAssetModal.tsx
```diff
@@ -128,7 +128,11 @@ export class UploadAssetModalComponent extends React.Component<IUploadAssetModal
 
         for (field in asset) {
             if (['_id', 'length'].includes(field) === false) {
-                data.append(field, asset[field] as string);
+                if (typeof asset[field] === 'object') {
+                    data.append(field, JSON.stringify([asset[field]]));
```
MarkLark86, line 132: Why is the value being supplied as an array? i.e.
```typescript
JSON.stringify([asset[field]])
```
vs
```typescript
JSON.stringify(asset[field])
```

The value being sent to the server should be the same value as stored in the local Asset, in a jsonified string

scripts/extensions/sams/src/components/assets/assetEditor.tsx
dev-tsoni: It is being converted into array below, where this function is called, line 188.

dev-tsoni:
```
description: ""
filename: "City7.jpg"
length: 1234818
mimetype: "image/jpeg"
name: "City7.jpg"
set_id: "602d04c5ee05b7bb601ebabf"
state: "draft"
tags: {name: "item6", code: "item6"}
```
This is the response without [ ].

scripts/extensions/sams/src/components/assets/assetEditorPanel.tsx
```diff
@@ -69,16 +71,59 @@ export class AssetEditorPanelComponent extends React.PureComponent<IProps, IStat
...
 
     onChange<K extends keyof IAssetItem>(field: K, value: IAssetItem[K]) {
-        this.setState((prevState: IState) => ({
-            updates: {
-                ...prevState.updates,
-                [field]: value,
-            },
-            isDirty: true,
-        }));
+        if (field === 'tags') {
```
MarkLark86, line 79: This would be better moved into the `AssetEditor` itself, otherwise this code (and `addTag`, `removeTag`) would need to be duplicated to `UploadAssetModal` and `EditAttachmentModal` as well

scripts/extensions/sams/src/components/assets/assetEditorPanel.tsx
```diff
@@ -149,12 +194,27 @@ export class AssetEditorPanelComponent extends React.PureComponent<IProps, IStat
...
                             ]}
                         />
+                        <FormGroup>
+                            <FormRow>
+                                {this.state.updates?.tags?.map((tag) => (
+                                    <Tag
+                                        key={this.state.updates?.tags?.indexOf(tag)}
+                                        text={tag.name}
+                                        onClick={() => {
+                                            this.removeTag(this.state.updates?.tags?.indexOf(tag)!);
```
MarkLark86, line 212: Again, this would be better off as a function in this class/component, and in turn that function will call `this.props.onChange` with the new tag value

scripts/extensions/sams/src/components/assets/assetEditor.tsx
MarkLark86: I'm still confused here, `tags` should always be an Array.

scripts/extensions/sams/src/components/assets/uploadAssetModal.tsx
MarkLark86: This code is very specific to `tags`. If the value is an `object`, `JSON.stringify` should suffice for all data types. `tags` should always be an array, so it should work here without encasing in `[` and `]`

scripts/extensions/sams/src/components/assets/assetEditorPanel.tsx
dev-tsoni: In onChange,we pass the tag, but can not pass the operation to be performed. I'll have to pass a string with it that will specify that the tag is needed to be added or removed. I'll be adding the code in commit though.

scripts/extensions/sams/src/components/assets/assetEditorPanel.tsx
dev-tsoni: `assetEditorPanel`, `uploadAssetModal` and `editAttachmentModal`, all have their own custom `onChange` and states, I think I'll have to probably add `addTag` and `removeTag` separately to them all, customised according to their states.

scripts/extensions/sams/src/components/assets/uploadAssetModal.tsx
```diff
@@ -186,7 +190,7 @@ export class UploadAssetModalComponent extends React.Component<IUploadAssetModal
 
         return (
             <FileUploadModal
-                dark={true}
+                dark={false}
```
MarkLark86, line 193: Why was this changed from dark to light?

dev-tsoni: The tags were not being visible in dark.

scripts/extensions/sams/src/components/assets/assetEditor.tsx
```diff
@@ -20,27 +22,119 @@ import {getHumanReadableFileSize} from '../../utils/ui';
 interface IProps {
     asset: Partial<IAssetItem>;
     disabled?: boolean;
-    onChange(field: string, value: string): void;
+    onChange/* <K extends keyof IAssetItem> */(field: string, value: any): void;
```
MarkLark86, line 25: What's wrong with the types on this function prop? i.e. can you un-comment so it's like this:
```typescript
onChange<K extends keyof IAssetItem>(field: K, value: IAssetItem[K]): void;
```

dev-tsoni: Ohh, my bad. removing it.

MarkLark86 (approved):

PR #3844, opened 2021-03-26
FIX: Search providers: credentials are not stored when default view is selected. [SDESK-5880]

GyanP:


scripts/apps/search-providers/directive.ts
```diff
@@ -41,6 +41,12 @@ export default function SearchProviderConfigDirective(searchProviderService, not
              * Upserts the selected search provider.
              */
             $scope.save = function() {
+                // check for config
+
+                if ($scope.origProvider?.config && $scope.provider?.config) {
```
tomaskikutis, line 46: What about other fields? If I added `config2` property, would the same issue occur? It would be good to write code that would solve the issue no matter the field name.

tomaskikutis (approved):

PR #3845, opened 2021-03-26
fix monitoring group firing too many requests

petrjasek:
wait for query to finish before starting another one

### DIFF
--- a/scripts/apps/monitoring/directives/MonitoringGroup.ts
+++ b/scripts/apps/monitoring/directives/MonitoringGroup.ts
@@ -141,21 +141,21 @@ export function MonitoringGroup(
                 );
             }
 
-            function scheduleQueryFn(event?, data?) {
-                return queryItems(event, data, {auto: (data && data.force) ? 0 : 1})
-                    .finally(() => {
-                        scope.$applyAsync();
-                    });
-            }
-
             let queryPromise = null;
+            // Using debounce to run it only once for multiple events
+            // happening at around same time, plus wait for query to finish
+            // before sending another query so if server is slow to respond
+            // it will slow down requests.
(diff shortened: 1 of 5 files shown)


scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -141,14 +141,24 @@ export function MonitoringGroup(
...
                     .finally(() => {
                         scope.$applyAsync();
                     });
             }
 
-            const scheduleQuery = throttle(scheduleQueryFn, 1000);
+            let queryPromise = null;
+            const scheduleQuery = debounce((event, data) => {
+                if (queryPromise == null) {
+                    queryPromise = scheduleQueryFn(event, data)
```
tomaskikutis, line 154: I would call `queryItems` here directly and remove `scheduleQueryFn` since it doesn't do anything, except one line in `.finally`

scripts/apps/monitoring/directives/MonitoringGroup.ts
```diff
@@ -141,14 +141,24 @@ export function MonitoringGroup(
...
+            function scheduleQueryFn(event?, data?) {
+                return queryItems(event, data, {auto: (data && data.force) ? 0 : 1})
                     .finally(() => {
                         scope.$applyAsync();
                     });
             }
 
-            const scheduleQuery = throttle(scheduleQueryFn, 1000);
+            let queryPromise = null;
+            const scheduleQuery = debounce((event, data) => {
```
tomaskikutis, line 152: It would be good to add a comment here, explaining how is it different from debouncing only.

tomaskikutis (approved):

PR #3846, opened 2021-03-26
Drop resource intensive request that was only used as a workaround

tomaskikutis:
REQUIRES changes to unlock item after publishing on the back-end.


petrjasek:
can you keep that unlock call for now? not sure when we have that backend fix

petrjasek (approved):

PR #3847, opened 2021-03-31
Extensions API improvements required for implementing for assignments overview in planning extension

tomaskikutis:
SDNTB-674


tasks/get-extension-directories-sync.js
```diff
@@ -2,21 +2,76 @@
...
+     *
+     * Here's a sample of extension registration code that the regex is targeting:
+        {
+            id: 'planning-extension',
+            load: () =>
+                import('superdesk-planning/client/planning-extension/dist/planning-extension/src/extension')
+                    .then((res) => res.default),
+        }
+     */
+    var extensionRegistrationPattern = /id:\W*'(.*)',\W*load:\W*import\('(.*\/dist.+)'/g;
```
petrjasek, line 63: this seems to be a bit error prone to me, what about using some json file with that config? `extension.json` or something? why do we actually need an id and not use package name?

tomaskikutis: I agree that it's not a stable solution, but it shouldn't break often either, especially since we don't change index files often.

It's easier to setup if extensions are registered from a single place. I agree that `extension.json` would be a less error prone solution, but we would lose typescript support. It's very convenient to be able to jump to the interface and see all available extension config options. Not many extensions have their config options yet, but there will be more of them as we move more things to extensions.

The problem with using folder name as extension ID is that it needs to be passed to client-side code and it can't access the file system to retrieve the folder name.

scripts/core/with-live-query.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    source: JSON.stringify(toElasticQuery(query)),
+                },
+            },
+        ).then((data) => {
+            this.setState({data: data, loading: false});
+        });
+    }
+
+    handleContentChanges(changes: Array<IResourceChange>) {
+        if (this.state.loading === true) {
```
petrjasek, line 95: if I get it right it will ignore any changes done while it's loading data right?

scripts/core/with-live-query.tsx
```diff
@@ -0,0 +1,148 @@
...
+        super(props);
+
+        this.state = {loading: true};
+
+        this.eventListenersToRemoveBeforeUnmounting = [];
+
+        this.fetchItems = this.fetchItems.bind(this);
+        this.handleContentChanges = this.handleContentChanges.bind(this);
+
+        this.handleContentChangesThrottled = throttleAndCombineArray(
```
petrjasek, line 30: I think throttle with 300 is not going to work, it needs to wait longer when the response takes longer, and refreshing every 300ms is too often anyhow, maybe we should have it configurable.

petrjasek (commented):
I'm not sure if we should be using throttle,

scripts/core/with-live-query.tsx
tomaskikutis: Yes, and that is a problem actually, because we would end up with old data. Instead I should probably simply call the throttled version so it queues another update.

scripts/core/with-live-query.tsx
tomaskikutis: I'll make it a second then.

https://github.com/superdesk/superdesk-client-core/pull/3847#discussion_r623072075 this would resolve the issue with server response taking longer than throttle wait. If new events arrive and it is still loading, it would get throttled again, which works like a queue that would only be released after the long running request has finished.

scripts/core/with-live-query.tsx
petrjasek: but when do you set loading to true again? it seems like that's only set in constructor so all further updates should be fetching from the server

tomaskikutis: Good catch. I'll set it on every request.

tasks/generate-placeholder-file-for-extension-styles.js
```diff
@@ -0,0 +1,6 @@
+const fs = require('fs');
+
+fs.writeFileSync(
+    'styles/extension-styles.generated.css',
```
petrjasek, line 4: wouldn't it be possible to handle that on webpack level? we can create entry point with multiple input files

tomaskikutis: I'm not sure it's possible, since the locations of extensions and their CSS files are specified only in the root repo. It would also probably not work well with the goal to distribute client-core to npm pre-built(instead of source as we do now), which would require isolating its build process from root repo and planning/analytics modules.

scripts/core/with-live-query.tsx
```diff
@@ -0,0 +1,159 @@
...
+                        ...data,
+                        _items: res,
+                        _meta: {
+                            ...data._meta,
+                            total: data._meta.total - diff,
+                        },
+                    },
+                });
+
+                this.updatingRequestInProgress = false;
```
petrjasek, line 135: this might be better in finally otherwise it will stop updating on error

tasks/build-extensions/compile-e2e-tests.js
```diff
@@ -0,0 +1,29 @@
...
+    protractorInstalled = true;
+} catch (e) {
+    protractorInstalled = false;
+}
+
+if (protractorInstalled) {
+    directories.forEach(({extensionRootPath}) => {
+        try {
+            execSync(
+                `cd ${extensionRootPath} && npm run compile-e2e`,
```
petrjasek, line 21: I think there should be `--if-present` otherwise the error message would be false positive when there are no tests.
and when it fails it should probably return an error at the end so we see that on CI

tasks/build-extensions/namespace-css-from-extensions.js
```diff
@@ -1,10 +1,18 @@
+function getModule(name) {
```
petrjasek, line 1: I think this can use some comment

tasks/verify-client-api-changes.js
```diff
@@ -0,0 +1,22 @@
...
+function getDirectories(path) {
+    return fs.readdirSync(path).filter((file) => {
+        return fs.statSync(path + '/' + file).isDirectory();
+    });
+}
+
+const allExtensions = 'scripts/extensions';
+
+getDirectories(allExtensions).forEach((extensionDir) => {
+    execSync(
```
petrjasek, line 18: this would only print an error, but won't make the process fail

tasks/build-extensions/compile-e2e-tests.js
tomaskikutis: I have added `--if-present` flag, but in general, e2e tests aren't running on extensions. They were running at some point in the past, but the implementation was very minimal and due to some incompatibilities we've disabled it. I'll get to it sometime when refactoring stuff.

tasks/verify-client-api-changes.js
petrjasek: did it? I had to add [this](https://github.com/superdesk/superdesk-client-core/blob/develop/run-end-to-end-tests.js#L59) otherwise failing tests were green on github actions / travis.

tomaskikutis: In your case it was because of promises. Promises swallow errors by default. At least they used to. I remember being annoyed by it often. It looks like it improved since promises first appeared. I don't know whether the spec was changed or browsers started logging in anyway.

petrjasek: ok if you're sure it works that's fine

petrjasek (approved):

PR #3849, opened 2021-04-05
As a user I want to see a schedule time even before the item is published. [SDBELGA-446]

GyanP:


scripts/apps/search/components/fields/scheduledDateTime.tsx
```diff
@@ -14,15 +15,23 @@ class ScheduledDateTime extends React.PureComponent<IProps> {
...
+        const scheduledState = ['scheduled'];
 
-        if (this.props.item.state != null && this.props.item.state === 'scheduled' && scheduled != null) {
+        if (appConfig?.features?.showPublishSchedule) {
+            scheduledState.push('in_progress', 'submitted');
+        }
+
+        if (this.props.item.state != null && scheduledState.includes(this.props.item.state) && scheduled != null) {
             const datetime = this.datetime;
+            const title = `Article is scheduled for ${datetime.longFormat(scheduled)}`;
```
petrjasek, line 28: this is not translatable

scripts/apps/search/components/fields/scheduledDateTime.tsx
```diff
@@ -14,15 +15,23 @@ class ScheduledDateTime extends React.PureComponent<IProps> {
     }
     render() {
         const scheduled = this.props.item.archive_item
-            ? this.props.item.archive_item.schedule_settings.utc_publish_schedule : null;
+            ? this.props.item.archive_item.schedule_settings.utc_publish_schedule
+            : appConfig?.features?.showPublishSchedule ? this.props.item?.publish_schedule : null;
```
petrjasek, line 19: actually not sure why `scheduled` which seems to be the value should depend on the feature config

scripts/apps/search/components/fields/scheduledDateTime.tsx
```diff
@@ -16,7 +17,7 @@ class ScheduledDateTime extends React.PureComponent<IProps> {
     render() {
         const scheduled = this.props.item.archive_item
             ? this.props.item.archive_item.schedule_settings.utc_publish_schedule
-            : appConfig?.features?.showPublishSchedule ? this.props.item?.publish_schedule : null;
+            : this.props.item?.publish_schedule || null;
```
petrjasek, line 20: still not sure it's ok here, `utc_publish_schedule` is in utc, but `publish_schedule` should be local time, so those are different, can you explain how that works pls?

GyanP: okay, I got your point, Actually, we are not getting `schedule_settings` for the `in_progress` or `submitted` items, so that's why I used `publish_schedule` but I think we need `utc_publish_schedule` for `in_progress` or `submitted` items as well, am I right?

petrjasek (approved):

PR #3853, opened 2021-04-07
fix list rendering error when profile is missing

petrjasek:
SDBELGA-507


tomaskikutis:
Why would profile be missing? It there's an ID, it must be in the database. I think we only soft-delete content profiles from the DB, no?

petrjasek:
not sure how it happened, but that's the way it is now on some instances

tomaskikutis (approved):
It'd be good to connect to those instances and inspect the DB.

PR #3855, opened 2021-04-07
fix suggestions popups in editor3

petrjasek:
instead of redux internals pass store explicitly

SDESK-5871


GyanP (approved):

tomaskikutis:
It's strange that 2 providers are needed. It worked in the past with one. Did you try bisecting and seeing what broke?

petrjasek:
oh yes, I did bisect. we updated the redux version and that internal behaviour where store was set in the context changed there, so I've made it explicit now.

PR #3856, opened 2021-04-09
FIX: Issues with the list in Spike view [SDESK-5923]

GyanP:


petrjasek:
can you pls explain how does it fix it? it seems to be it should reload even more now

GyanP:
> can you pls explain how does it fix it? it seems to be it should reload even more now

We were getting the value in decimal and we are checking for the exact value, for example, we were getting the scroll height as 2497 but offset + scrollTop 2497.385 something like that, so that's why I have to make it round off.

petrjasek:
I mean it will call this `this.loadMore();` more often right? because it wasn't matching without rounding. now with rounding that `reachedBottom` can be actually true

GyanP:
> I mean it will call this `this.loadMore();` more often right? because it wasn't matching without rounding. now with rounding that `reachedBottom` can be actually true

Ya right, it will call load more, if we reached the bottom and that's what's the issue specified in the ticket that **there was no pagination** so after this changes it will give us the pagination.

petrjasek (approved):

PR #3857, opened 2021-04-12
Issues with the list in Spike view [SDESK-5923]

GyanP:
Resolve constant loading.

### DIFF
--- a/scripts/core/itemList/LazyLoader.tsx
+++ b/scripts/core/itemList/LazyLoader.tsx
@@ -63,6 +63,8 @@ export class LazyLoader<T> extends React.Component<IProps<T>, IState<T>> {
     private reloadAllItems() {
         const MAX_PAGE_SIZE = 200; // back-end limit
 
+        // The size doesn't contain the new items.
+        // so we need to reload the items and for that we need a default value as 1.
         const loadedItemsCount = this.state.items.size > 0 ? this.state.items.size : 1;
 
         const pages = new Array(Math.ceil(loadedItemsCount / MAX_PAGE_SIZE)).fill(null).map((_, i) => {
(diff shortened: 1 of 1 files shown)


scripts/core/itemList/LazyLoader.tsx
```diff
@@ -63,7 +63,7 @@ export class LazyLoader<T> extends React.Component<IProps<T>, IState<T>> {
     private reloadAllItems() {
         const MAX_PAGE_SIZE = 200; // back-end limit
 
-        const loadedItemsCount = this.state.items.size;
+        const loadedItemsCount = this.state.items.size > 0 ? this.state.items.size : 1;
```
petrjasek, line 66: can you pls add some comment to this? not sure why you do it

GyanP:
We need a default value of 1 because if we get 0 then it will generate an Array with 0 values [here](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/itemList/LazyLoader.tsx#L68)  which we don't want, else it will break down [here](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/itemList/LazyLoader.tsx#L84) with an error that `cannot reduce an empty array.`

scripts/core/itemList/LazyLoader.tsx
GyanP: Done.

petrjasek:
imo if the size is 0 it should just return, there is no point to do any reloading:
```ts
if (loadedItemsCount === 0) {
    return;
}
```

GyanP:
> imo if the size is 0 it should just return, there is no point to do any reloading:

We need to reload it because of the new item. The size doesn't contain the new items, it's just the size of loadedItemsCount (that already present) and for the first time it would be 0 because there are no items present, so If the new item is being sent so it should reload.

petrjasek:
I see, can you add that info to the code?

GyanP:
> I see, can you add that info to the code?

Done.

petrjasek:
shouldn't we add +1 all the time then to get new items? and not just when it was empty? because it doesn't check for the total number of items so if it would go from MAX_PAGE_SIZE count to MAX_PAGE_SIZE + 1 that extra item would be me missing right?

petrjasek (approved):

PR #3860, opened 2021-04-19
FIX: Publish button appears for items in personal space. [SDESK-5942]

GyanP:


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -141,7 +141,7 @@ export function AuthoringDirective(
 
             function checkShortcutButtonAvailability(personal = false) {
                 if (personal && appConfig?.features?.publishFromPersonal) {
-                    return $scope.item.state !== 'draft' || $scope.dirty;
+                    return $scope.item.state !== 'draft' && $scope.dirty;
                 }
                 return $scope.item.task && $scope.item.task.desk && $scope.item.state !== 'draft' || $scope.dirty;
```
petrjasek, line 146: I think here it's pretty much same issue, as they see the button when `publishFromPersonal` is not enabled

GyanP: The issue is for personal space and now they are not able to see the button when `publishFromPersonal` is not enabled.

petrjasek: yes the issue is for personal space with publish from personal disabled, so it goes to the return line
and there it's true if `$scope.dirty` is true.

GyanP: But we have a check over [here](https://github.com/superdesk/superdesk-client-core/blob/429e2d3e4ea0c50d7d7d41b767ef53949d9e8a98/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts#L156) so If it is set to disabled or false then its never going to call.

petrjasek: yep but there it's checking `customAuthoringTopbar` which they have enabled on cp, but they don't use `publishFromPersonal`

petrjasek (approved):

PR #3862, opened 2021-04-19
Do not display the 'missing link' for corrections. [SDESK-5928]

GyanP:


scripts/apps/authoring/authoring/directives/AuthoringHeaderDirective.ts
```diff
@@ -174,7 +174,9 @@ export function AuthoringHeaderDirective(
                             if (items && items._items.length && !getNoMissingLink()) {
                                 // if takes package is missing or not rewrite of.
                                 scope.missing_link = !scope.item.rewrite_of &&
-                                    !scope.item.rewritten_by;
+                                    !scope.item.rewritten_by
+                                    && scope.action !== 'correct' && !scope.isCorrection
+                                    && !scope.item.correction_sequence;
```
tomaskikutis, line 179: It'd be better to either only use a single variable to determine whether it's a correction or to extract the check to a function. When there are many conditions and multiple checks are used for the same thing it becomes hard to understand which lines belong to a group and which not.

tomaskikutis (approved):

PR #3863, opened 2021-04-19
Fix sticky toolbar for editor3

tomaskikutis:
SDESK-5765


tomaskikutis:
Yep, forgot about it. Added it now.

petrjasek (approved):

MarkLark86:
@petrjasek This commit https://github.com/superdesk/superdesk-client-core/commit/e98aa703da8ef026c4ec99b104faaea045c76ad2 will need to be cherry-picked as well into any version this is back-ported to (to remove the `superdesk-planning` submodule)

scripts/apps/authoring/views/article-edit.html
```diff
@@ -641,7 +641,7 @@
         data-single-line="field.field_options.single"
         data-value="item.extra[field._id]"
         data-bind-to-value="compareView"
-        data-scroll-container=".page-content-container"
+        data-scroll-container=".ppage-content-container--scrollable"
```
petrjasek, line 644: seems like typo `.PPage`

tomaskikutis: here's the fix https://github.com/superdesk/superdesk-client-core/pull/3867

PR #3864, opened 2021-04-20
As a user I want to have the Coming Up field active when I create a new story [SDBELGA-466]

GyanP:
https://github.com/superdesk/superdesk-belga/pull/305


tomaskikutis:
They say in the ticket that when a new story is created with a template that has a field with a date-time saved, it should not use the saved date-time, but work as it would if the date-time was toggled on - that is to initialize with a current date.

GyanP:
> They say in the ticket that when a new story is created with a template that has a field with a date-time saved, it should not use the saved date-time, but work as it would if the date-time was toggled on - that is to initialize with a current date.

ya, but the issue was when I am toggling the datetime i.e **for toggling on**, the save button is disabled, If I make any further changes then I get the save button enabled, and if then If I save it, and edit the same template, I get the datetime as **off**.

tomaskikutis:
The change you did is fine, I know it wasn't working to save a date-time in a template before. Your changes fix the first requirement in the ticket.

The second requirement is that when a new article is created using a template, it should display, not a date that is saved in the template, but a date that would result from toggling date-time input from "off" to "on".

To implement this in a clean way, I would add 2 additional properties to `ICustomFieldType`:
```
templateEditorComponent?: React.ComponentType<IEditorComponentProps<IConfig>>;
prepare?(article: IArticle): IArticle;
```

Template editor component would work the same as `editorComponent`, except it would only be displayed in template edit view. We could add a radio button whether it is desired to store a specific date in the template, or only mark it for initialization. Instead of a date, we could store `{initialize: true}` or something similar.

`prepare` would get called when an article is being created and would change the value from `{initialize: true}`, to an expected date.

I could help you out with this second part if you prefer.

scripts/core/superdesk-api.d.ts
```diff
@@ -1816,6 +1816,15 @@ declare module 'superdesk-api' {
         config: IConfig;
     }
 
+    export interface ITemplateEditorComponentProps<IValue, IConfig> {
+        item: IArticle;
+        value: IValue;
+        setValue: (value: IValue) => void;
+        readOnly: boolean;
+        config: IConfig;
+        radioValue?: string,
```
tomaskikutis, line 1825: It's a generic interface meant to suit any authoring field, there should be no fields in the object that are specific to one authoring field.

It looks to me like you should be able to compute radio value from field value.

scripts/apps/authoring/authoring/authoring-custom-field.tsx
```diff
@@ -9,6 +9,7 @@ interface IProps {
     field: IArticleField;
     editable: boolean;
     onChange: (field: IArticleField, value: any) => any;
+    template: any;
```
tomaskikutis, line 12: Can you specify the type? It looks like there is no interface yet. You should add one in that case.

scripts/apps/authoring/authoring/authoring-custom-field.tsx
```diff
@@ -70,13 +72,26 @@ export class AuthoringCustomField extends React.PureComponent<IProps, IState> {
...
         return (
             <div>
-                <FieldType.editorComponent
-                    item={item}
-                    value={this.state.value}
-                    setValue={(value) => this.setValue(value)}
-                    readOnly={!editable}
-                    config={field.custom_field_config}
-                />
+                {this.props.template == null ?
```
tomaskikutis, line 75: Shouldn't it determine which component to render, depending on whether `templateEditorComponent` exists? Not based on the current value of `template`?

scripts/extensions/datetimeField/src/getTemplateDateTimeField.tsx
```diff
@@ -0,0 +1,230 @@
+import {ISuperdesk, ITemplateEditorComponentProps} from 'superdesk-api';
+import * as React from 'react';
+import set from 'date-fns/set';
+import format from 'date-fns/format';
+import addMinutes from 'date-fns/addMinutes';
+import {IDateTimeFieldConfig} from './extension';
+import {DatePickerISO, TimePicker, Button, Switch} from 'superdesk-ui-framework/react';
+import {Radio, CheckGroup} from 'superdesk-ui-framework';
+
+interface ITemplateDateTimeFieldState {
```
tomaskikutis, line 10: It's a local interface that isn't exported anywhere, you can name it simply `IState`.

scripts/extensions/datetimeField/src/getTemplateDateTimeField.tsx
```diff
@@ -0,0 +1,230 @@
...
+    radioValue: string | null;
+    previousValue: any;
+}
+
+function isDateValue(value: any) {
+    return !!Date.parse(value);
```
tomaskikutis, line 16: It's better to avoid using `!!` because it can mean different things. For example, I thought this was a null/undefined check when it's probably not. `Date.parse(value) !== NaN` is much clearer.

scripts/extensions/datetimeField/src/getTemplateDateTimeField.tsx
```diff
@@ -0,0 +1,230 @@
...
+        onRadioValueChange(radioValue: any) {
+            this.setState({radioValue});
+            if (radioValue === 'setCurrentDate') {
+                this.props.setValue(currentDateTime);
+            } else {
+                this.props.setValue(this.state.previousValue);
+            }
+        }
+
+        render() {
```
tomaskikutis, line 49: In case user has selected to set a static date-time, can you reuse the component from `src/getDateTimeField.tsx` instead of copy-pasting the code from it?

scripts/extensions/datetimeField/src/getTemplateDateTimeField.tsx
```diff
@@ -0,0 +1,230 @@
...
+import * as React from 'react';
+import set from 'date-fns/set';
+import format from 'date-fns/format';
+import addMinutes from 'date-fns/addMinutes';
+import {IDateTimeFieldConfig} from './extension';
+import {DatePickerISO, TimePicker, Button, Switch} from 'superdesk-ui-framework/react';
+import {Radio, CheckGroup} from 'superdesk-ui-framework';
+
+interface ITemplateDateTimeFieldState {
+    radioValue: string | null;
```
tomaskikutis, line 11: Use union types where possible: `radioValue: 'setCurrentDate' | 'showDate' | null;`

scripts/apps/authoring/authoring/authoring-custom-field.tsx
GyanP: we have defined the `templateEditorComponent` in customFieldTypes [here](https://github.com/superdesk/superdesk-client-core/blob/32395021e0da0fe595d2248a60ada727ec232d18/scripts/extensions/datetimeField/src/extension.tsx#L50), so don't you think it's gonna be present every time?

scripts/extensions/datetimeField/src/getTemplateDateTimeField.tsx
GyanP: I thought of the same but then I found out, I cannot use that because suppose the value is a string, so if I use `!!Date.parse(value)` then I would get `false`, but If I use something like this `Date.parse(value) !== NaN` it will return `true`.

scripts/extensions/datetimeField/src/getTemplateDateTimeField.tsx
GyanP: we have not copied exactly the same, if you see [here](https://github.com/superdesk/superdesk-client-core/pull/3864/files/32395021e0da0fe595d2248a60ada727ec232d18#diff-12d88161b515852811fe7dbc60312ec50e13d30c6aeb4a5d4d620c837d15261bR58) we have set some values and have some [checks](https://github.com/superdesk/superdesk-client-core/pull/3864/files/32395021e0da0fe595d2248a60ada727ec232d18#diff-12d88161b515852811fe7dbc60312ec50e13d30c6aeb4a5d4d620c837d15261bR113) as well.

scripts/apps/authoring/authoring/authoring-custom-field.tsx
tomaskikutis: You have marked it [optional](https://github.com/superdesk/superdesk-client-core/pull/3864/files#diff-ca49bae7f62187bd0909482a419226feafbf06df59d2d06c2031cd298e7d62aaR1845) in the interface. And that's a good thing. In most cases, when fields are static, using `editorComponent` will work fine when configuring a template.

`templateEditorComponent` would be needed for advanced functionality like dynamic templates, but when an extension doesn't need that advanced functionality, we shouldn't be making it harder for the extension author by requiring it.

scripts/extensions/datetimeField/src/getTemplateDateTimeField.tsx
tomaskikutis: My bad, I thought that `NaN === NaN` would evaluate to `true`, but it is `false` :D

It should work to use `isNaN(Date.parse('2020-01-01')) !== true`

scripts/extensions/datetimeField/src/getTemplateDateTimeField.tsx
tomaskikutis: I'm not saying you copied exactly, but it's always better to reuse when possible, even if you have to add a few configurable things in the original component(which you might not even need to do here).

Regarding the first example, you don't need to modify the component and can call `onRadioValueChange` in a `setValue` callback:

```
<DateTimeField
    setValue={(val) => {
        if (val == null) {
            // ...
        } else {
            // ...
            this.onRadioValueChange('setCurrentDate');
        }
    }}
/>
```

scripts/extensions/datetimeField/src/getTemplateDateTimeField.tsx
```diff
@@ -14,7 +14,7 @@ interface IState {
 export function getTemplateDateTimeField(superdesk: ISuperdesk) {
     const {Spacer} = superdesk.components;
     const currentDateTime = '{{ now|iso_datetime }}';
-    const DateTimeField: any = getDateTimeField(superdesk);
+    const DateTimeField = getDateTimeField(superdesk);
```
tomaskikutis, line 17: Why did you set it to `any`? Can you go over this extension and check that `any` is not used where not really needed?

I was searching for usages of a prop using IDE, and it found none because of this.

scripts/extensions/datetimeField/src/getTemplateDateTimeField.tsx
```diff
@@ -0,0 +1,100 @@
...
+                                label: 'Choose a date',
+                            },
+                        ]}
+                        onChange={(val) => this.onRadioValueChange(val)}
+                    />
+                </CheckGroup>
+            );
+
+            return (
+                <Spacer type="horizontal" align="center" spacing="medium">
```
tomaskikutis, line 81: Can you check the alignment in the UI? Something seems off a bit.

tomaskikutis (approved):

PR #3871, opened 2021-04-23
sync user login/logout status on master desk

petrjasek:
SDESK-5817

uses https://github.com/superdesk/superdesk-core/pull/2199

### DIFF
--- a/scripts/apps/master-desk/components/UserListComponent.tsx
+++ b/scripts/apps/master-desk/components/UserListComponent.tsx
@@ -4,19 +4,19 @@ import classNames from 'classnames';
 import {UserAvatar} from 'apps/users/components/UserAvatar';
 import {IDesk, IUserRole, IUser} from 'superdesk-api';
 
-export interface IUserExtra extends IUser {
+export interface IUserExtra {
+    user: IUser;
     data: {
         assigned: number;
         locked: number;
     };
-    authors: Array<string>;
 }
 
 interface IProps {
     desk: IDesk;
     role: IUserRole;
     users: Array<IUserExtra>;
-    onUserSelect(user: IUserExtra): void;
+    onUserSelect(user: IUser): void;
 }
 
 export class UserListComponent extends React.Component<IProps, {}> {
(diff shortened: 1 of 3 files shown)


pavlovicnemanja (approved):

tomaskikutis:
Does this only affect user avatar component? If so, it'd be better to handle it there.

petrjasek:
it does, but this is rather about the state which should be imo higher than in the avatar component, we can have same user on many desks and there is no need to refetch the user more than once

tomaskikutis:
I agree that re-fetching data multiple times for the same user is not good. But handling this in every parent component that renders user avatar isn't a good solution either.

Ideally, there would be a central data store that would handle such updates. Since we don't have it yet, what about putting a short-lived cache object inside user avatar component? It would avoid fetching the same user multiple times and would allow moving the code to user avatar component instead of putting it everywhere where avatar is used.

petrjasek:
will think about it, but would rather put there some initial store than complicate the components which should be using it.

tomaskikutis:
I actually meant to put the cache object, not in the component, but at the file level. Something simple like

```
var cache = {};

function getUser(id) {
    if (cache[id] != null) {
        return Promise.resolve(cache[id]);
    } else {
        fetch(...).then((res) => {
            cache[id] = res;

            return res;
        })
    }
}
```

Initial store would work too.

petrjasek:
and what about the listener? that would be defined on each component?

petrjasek:
@tomaskikutis I would rather leave it as it is now, there are like 1500 instances on cp of that user avatar component so don't want to complicate that one

tomaskikutis:
> there are like 1500 instances on cp of that user avatar component

Good, that means that you would fix the issue in all places, not only in 1/1500. ;)

To me, addressing it, seems an easy choice:
* Current implementation isn't clean - it violates separation of concerns and introduces undesirable coupling. (master desk should be ignorant of how avatar component works and not try to affect it)
* The fix is only available in one place
* It's relatively easy to implement it right. I wouldn't say adding a cache object is either risky or adds much coupling, so it doesn't seem like a complication to me.

petrjasek:
I mean there are 1500 avatar components rendered at the same time, and with this approach it's handled for all in 1 place.
and while now it's only checking the update status of the user, in general it should keep user info fresh because that's what master desk is showing, so imo it's not violating separation of concerns

tomaskikutis (approved):

PR #3873, opened 2021-04-27
fix toggle all on new privilege without privileges populated

petrjasek:
SDESK-5811


scripts/core/superdesk-api.d.ts
```diff
@@ -523,7 +523,9 @@ declare module 'superdesk-api' {
     export interface IUserRole extends IBaseRestApiResponse {
         _id: string;
         name: string;
-        privileges: any;
+        privileges?: {
+            [privilege: string]: number;
```
tomaskikutis, line 527: You can also use a number value as a type `[privilege: string]: 1;`. I think there are no other values.

tomaskikutis (approved):

PR #3875, opened 2021-05-04
disallow "publish", and "send  to" options in bulk actions menu for published items

tomaskikutis:
SDESK-5959

### DIFF
--- a/scripts/apps/search/controllers/get-bulk-actions.tsx
+++ b/scripts/apps/search/controllers/get-bulk-actions.tsx
@@ -26,7 +26,8 @@ export function getBulkActions(
 
     const {isLocked, isLockedByOtherUser, isPublished} = sdApi.article;
 
-    const noneLocked = articles.every((article) => !isLocked(article));
+    const noneLocked = articles.some((article) => isLocked(article)) !== true;
+    const nonePublished = articles.some((article) => isPublished(article)) !== true;
 
     if (articles.every(({state}) => state === ITEM_STATE.INGESTED)) {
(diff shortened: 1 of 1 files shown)


petrjasek (approved):

scripts/apps/search/controllers/get-bulk-actions.tsx
```diff
@@ -26,7 +26,8 @@ export function getBulkActions(
 
     const {isLocked, isLockedByOtherUser, isPublished} = sdApi.article;
 
-    const noneLocked = articles.every((article) => !isLocked(article));
+    const noneLocked = articles.some((article) => isLocked(article)) !== true;
```
petrjasek, line 29: this is actually rather `someUnlocked` than `noneLocked`

tomaskikutis: `articles.some((article) => isLocked(article))` means `at least one is locked`.

If it is `not true`, that `at least one is locked`, then none are locked.

petrjasek: I've missed that `!== true` at the end, not very readable imo

petrjasek (approved):

scripts/apps/search/controllers/get-bulk-actions.tsx
tomaskikutis: I thought `some` would perform better than `every`, but turns out it's the same. `every` stops checking after the first negative response.

PR #3876, opened 2021-05-10
Revert "Revert "Revert "Revert "Improve extensions loading (#3785)" (…

petrjasek:
…#3800)" (#3803)" (#3828)"

This reverts commit bf84c88dc04718e15c33a481e6ee963b4d52fbc4.

@tomaskikutis after updating develop from release 2.2 that revert commit got it, so trying to restore the state before, can you pls check this is the latest version we should have in develop?


scripts/core/register-extensions.tsx
```diff
@@ -69,41 +64,55 @@ export function registerExtensions(
...
+            )
+                .forEach(registerPage);
 
-        flatMap(
-            activationResults,
-            (activationResult) => activationResult.contributions?.workspaceMenuItems ?? [],
-        )
-            .forEach(registerWorkspaceMenu);
-    }).then(() => {
-        dispatchInternalEvent('extensionsHaveLoaded', true);
```
tomaskikutis, line 107: `extensionsHaveLoaded` event should be kept.

scripts/core/register-extensions.tsx
```diff
@@ -5,8 +5,12 @@ import {IExtension, IPage, IWorkspaceMenuItem, IExtensionActivationResult, ISupe
 import {extensions as extensionsWithActivationResult} from 'appConfig';
 import {dispatchInternalEvent} from './internal-events';
 
+interface IExtensionModule {
```
tomaskikutis, line 8: The boilerplate in question is 27 chars long `.then((res) => res.default)`. I'd rather keep the interface cleaner and use the boilerplate. It's only needed in the index/config file which is rarely edited anyway.

petrjasek: it has to be set in every repo on every branch, that's not so rare imo

tomaskikutis: Maybe you are setting things up more frequently, but even then, it's 10 seconds to type. That's one of the stranger abstractions I've seen, but go ahead if it helps :)

petrjasek: :smile:

PR #3877, opened 2021-05-10
fix(iMatrics): Fixed issue with sorting entities in iMatrics

pavlovicnemanja:
```
iMatricsFields: {
    entities: {
        organisation: {
            name: 'Organisations',
            order: 3,
        },
        person: {
            name: 'People',
            order: 2,
        },
        place: {
            name: 'Places',
            order: 1,
        },
    },
    others: {
        topics: {
            name: 'Media topics',
            order: 1,
        },
        imatrics_category: {
            name: 'Category',
            order: 3,
        },
        imatrics_topic: {
            name: 'Topics',
            order: 4,
        }
    }
}
```

### DIFF
--- a/scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
+++ b/scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
@@ -413,7 +413,8 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
                                 const othersGrouped = others.groupBy((tag) => tag.group.value);
 
                                 let allGrouped = OrderedMap<string, any>();
-                                othersGrouped.map((tags, groupId) => {
+
+                                othersGrouped.map((tags, groupId) =>
                                     allGrouped = allGrouped.set(groupId,
                                         <ToggleBoxNext
                                             key={groupId}
@@ -432,12 +433,12 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
                                                     );
                                                 }}
                                             />
-                                        </ToggleBoxNext>
-                                    )
-                                })
+                                        </ToggleBoxNext>,
+                                    ),
+                                );
 
                                 if (entitiesGroupedAndSorted.size > 0) {
-                                    allGrouped = allGrouped.set('entities', 
+                                    allGrouped = allGrouped.set('entities',
                                         <ToggleBoxNext
                                             title={this.getGroupName('entities', vocabularyLabels)}
                                             style="circle"
@@ -465,7 +466,7 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
                                                     />
                                                 </div>
                                             )).toArray()}
-                                        </ToggleBoxNext>
+                                        </ToggleBoxNext>,
                                     );
                                 }
 
@@ -477,7 +478,8 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
                                 const allGroupedAndSortedNotInConfig = allGrouped
                                     .filter((_, key) => !hasConfig(key, this.iMatricsFields.others));
 
-                                const allGroupedAndSorted = allGroupedAndSortedByConfig.concat(allGroupedAndSortedNotInConfig);
+                                const allGroupedAndSorted = allGroupedAndSortedByConfig
+                                    .concat(allGroupedAndSortedNotInConfig);
 
                                 return (
                                     <React.Fragment>
--- a/scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
+++ b/scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
@@ -414,8 +414,8 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
 
                                 let allGrouped = OrderedMap<string, JSX.Element>();
 
-                                othersGrouped.map((tags, groupId) =>
-                                    allGrouped = allGrouped.set(groupId,
+                                othersGrouped.forEach((tags, groupId) =>
+                                    tags && groupId ? allGrouped = allGrouped.set(groupId,
                                         <ToggleBoxNext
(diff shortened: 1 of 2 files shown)


scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -395,26 +395,91 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
...
+
+                                const entitiesGroupedAndSorted = entitiesGroupedAndSortedByConfig
+                                    .concat(entitiesGroupedAndSortedNotInConfig);
 
                                 const others = items.filter((tag) => isEntity(tag) === false);
                                 const othersGrouped = others.groupBy((tag) => tag.group.value);
 
-                                const othersGroupedAndSortedByConfig = othersGrouped
-                                    .filter((_, key) => hasConfig(key, this.iMatricsFields))
-                                    .sortBy((_, key) => this.iMatricsFields[key].order,
```
tomaskikutis, line 408: Could you explain what wasn't working with previous implementation?

It looks to me that you have added sorting specifically for entities, but I thought it would be enough to modify how `entitiesGroupedAndSorted` or `othersGroupedAndSorted` are put together. Why were changes needed below this line?

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -395,26 +395,91 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
...
+                                let allGrouped = OrderedMap<string, any>();
+
+                                othersGrouped.map((tags, groupId) =>
```
tomaskikutis, line 417: return value of `map` isn't used

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -395,26 +395,91 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
...
+                                const entitiesGroupedAndSorted = entitiesGroupedAndSortedByConfig
+                                    .concat(entitiesGroupedAndSortedNotInConfig);
 
                                 const others = items.filter((tag) => isEntity(tag) === false);
                                 const othersGrouped = others.groupBy((tag) => tag.group.value);
 
-                                const othersGroupedAndSortedByConfig = othersGrouped
-                                    .filter((_, key) => hasConfig(key, this.iMatricsFields))
-                                    .sortBy((_, key) => this.iMatricsFields[key].order,
+                                let allGrouped = OrderedMap<string, any>();
```
tomaskikutis, line 415: Avoid using `any`. In this case, the type would be `JSX.Element`. It better thought to avoid putting jsx elements to arrays and render directly. It's easier to keep track of element order this way.

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
pavlovicnemanja: issue is that group of entites are not sorted and included, because in old implementation we showed all groups and entities are added at bottom of page. Now we need to include sorting of entities group also, and to be included in config (NTB want to add entitie group at top of list

pavlovicnemanja: image

and this group should on top:
image

tomaskikutis: What about the second question?

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
pavlovicnemanja: what is not used? both `tags` and `groupId` is used it code below

pavlovicnemanja: this is iteration for `othersGrouped` to be stored in `allGrouped` OrderedMap

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
pavlovicnemanja: we need to sort entities as group and items in entities group separately

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
tomaskikutis: > what is not used?

The return value. `map` is a function that transforms an array/iterable, using the provided function and returns the new array. You never save that new array to a variable. If you only want to iterate the array/iterable, and not transform it, you should be using `forEach` instead.

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -414,8 +414,8 @@ export function getAutoTaggingComponent(superdesk: ISuperdesk, label: string) {
 
                                 let allGrouped = OrderedMap<string, JSX.Element>();
 
-                                othersGrouped.map((tags, groupId) =>
-                                    allGrouped = allGrouped.set(groupId,
+                                othersGrouped.forEach((tags, groupId) =>
+                                    tags && groupId ? allGrouped = allGrouped.set(groupId,
```
tomaskikutis, line 418: That ternary is a bit confusing, it'd be more readable to use a simple if
```
othersGrouped.forEach((tags, groupId) => {
if (tags != null && groupId != null) {
    allGrouped = allGrouped.set(groupId, ...);
}
});
```

Can `tags` or `groupId` be `null` though? That map shouldn't have empty items

tomaskikutis (approved):

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
pavlovicnemanja: Nope, but is showing error it is possible null, so just to avoid warning.

PR #3878, opened 2021-05-11
fix: UserAvatar throwing error on logout

MarkLark86:
This is causing failures in Planning e2e


tomaskikutis:
It looks a bit cleaner with this change, but I was referring to not changing this component at all, and handling the null check in the parent component on the planning side. Would that work?

MarkLark86:
The exception being thrown is not from the Planning module, but upon logout. Look at the console and you will see uncaught exceptions being thrown. The E2E software we use there fail a test if there is an uncaught exception thrown

tomaskikutis:
I see. I have checked what's the root cause of it and submitted https://github.com/superdesk/superdesk-client-core/pull/3883. It fixes the avatar error too.

PR #3883, opened 2021-05-14
Avoid errors when logging out

tomaskikutis:
It used to set user object to `null` before logging out and cause errors in the UI. That's because when logged out, not only log-in page is rendered, but the rest application too, even if hidden. I have tried in the past not to render it at all instead of hiding it, but it's hard to do. Since we reload the page after logging out anyway, there's no benefit in clearing user object.

### DIFF
--- a/scripts/core/auth/auth.ts
+++ b/scripts/core/auth/auth.ts
@@ -196,12 +196,6 @@ export default angular.module('superdesk.core.auth', [
             modal,
         ) {
             $rootScope.logout = function() {
-                function doLogout() {
-                    $rootScope.$broadcast(SESSION_EVENTS.LOGOUT);
-                    localStorage.clear();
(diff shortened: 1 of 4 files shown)


scripts/core/auth/auth.ts
```diff
@@ -215,7 +216,7 @@ export default angular.module('superdesk.core.auth', [
 
                 if (canLogout) {
                     api.auth.getById(session.sessionId).then((sessionData) => {
-                        api.auth.remove(sessionData).then(replace, replace);
+                        api.auth.remove(sessionData).then(doLogout, doLogout);
```
petrjasek, line 219: this looks like `.finally(doLogout);`

petrjasek (approved):

petrjasek (approved):

PR #3886, opened 2021-05-19
Build tools

tomaskikutis:
Move extension building to a separate package. [Here's how](https://user-images.githubusercontent.com/2076953/118807545-bd9afe00-b8a8-11eb-97a9-4f10c9fb6671.png) it would be enabled in root repositories.

There's also a task to convert .po to .json files, that will allow including translations from extensions.


petrjasek:
so how would it get the po files from planning and combine that with po files from core?

tomaskikutis:
I'm still working on that part. In order to get translations from extensions, I'd need to get extension directories. That was already implemented in extension build scripts and you asked about removing `build-extensions.js` from the root repository so I first moved the build scripts and will now implement combining translations.

MarkLark86 (approved):

MarkLark86:
Looks okay to me

petrjasek:
that `file:../../` path in package.json from the example goes where?

tomaskikutis:
It's not only an example, it's used to build translations before running e2e tests.
`file: build-tools` is the same as `file: ./build-tools` - it looks for a directory and `package.json` in that directory.

petrjasek:
yep I'm just wondering how to reference the build tools in eg superdesk-cp repo

tomaskikutis:
Ah, you mean that example. I will release a npm package after this PR is merged.

petrjasek (approved):

PR #3887, opened 2021-05-20
Sdesk 5958 one asset can not be added as attachment in one article multiple times

dev-tsoni:


scripts/extensions/sams/src/api/assets.ts
```diff
@@ -100,6 +100,17 @@ function querySetIds(source: IRootElasticQuery, params: IAssetSearchParams) {
     }
 }
 
+function queryAssetIds(source: IRootElasticQuery, params: IAssetSearchParams) {
+    if (params.assetIds != null && params.assetIds.length > 0) {
```
MarkLark86, line 104: It would be more meaningful to use a param name that indicates the Assets will be excluded (also if we later need a param to filter by Asset IDs)

dev-tsoni: Agreed. May be `excludedAssetIds` ?

scripts/extensions/sams/src/apps/samsAttachmentsWidget.tsx
```diff
@@ -108,10 +108,19 @@ class SamsAttachmentsWidgetComponent extends React.PureComponent<IProps> {
...
+            let media = attachment.media as IMedia;
+
+            attachmentMediaIds = attachmentMediaIds.concat(media._id);
+        });
+
         this.props.setAssetSearchParams({
             sizeTo: superdeskApi.instance.config.attachments_max_size / 1048576, // bytes -> MB
             states: [ASSET_STATE.PUBLIC, ASSET_STATE.INTERNAL],
             setIds: this.props.activeSets.map((set) => set._id),
+            assetIds: attachmentMediaIds,
```
MarkLark86, line 123: This can be simplified into a single line assignment:
```ts
excludeAssetIds: this.props.attachments.map((attachment) => (
    typeof attachment.media === 'string' ?
        attachment.media :
        attachment.media._id
)),
```

MarkLark86 (approved):

PR #3888, opened 2021-05-20
keep users data outside of react components

petrjasek:
SDESK-5817

built this as a POC with flux for handling users data in one place
which could be used by other components later. that seems to integrate
better into our app than redux.

this seems easier to work with than redux as it can register stores any
time rather than once at the bootstrap. we could use that for basic entities
like users/desks/roles etc and keep the rest of the state in components
which are top level views like MasterDeskApp, which could still define
its own store which would be registered when the component mounts.

what do you think @superdesk/client ?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -245,6 +245,22 @@
       "resolved": "https://registry.npmjs.org/@types/eslint-visitor-keys/-/eslint-visitor-keys-1.0.0.tgz",
       "integrity": "sha512-OCutwjDZ4aFS6PB1UZ988C4YgwlBHJd6wCeQqaLdmadZ/7e+w79+hbMUFC1QXDNCmdyoRfAFdm0RypzwR+Qpag=="
     },
+    "@types/fbemitter": {
+      "version": "2.0.32",
+      "resolved": "https://registry.npmjs.org/@types/fbemitter/-/fbemitter-2.0.32.tgz",
+      "integrity": "sha1-jtIE2g9U6cjq7DGx7skeJRMtCCw=",
+      "dev": true
+    },
+    "@types/flux": {
+      "version": "3.1.9",
+      "resolved": "https://registry.npmjs.org/@types/flux/-/flux-3.1.9.tgz",
+      "integrity": "sha512-bSbDf4tTuN9wn3LTGPnH9wnSSLtR5rV7UPWFpM00NJ1pSTBwCzeZG07XsZ9lBkxwngrqjDtM97PLt5IuIdCQUA==",
+      "dev": true,
+      "requires": {
+        "@types/fbemitter": "*",
+        "@types/react": "*"
+      }
+    },
     "@types/hoist-non-react-statics": {
       "version": "3.3.1",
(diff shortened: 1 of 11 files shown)


tomaskikutis:
> :warning: The Flux project is in maintenance mode and there are many more sophisticated alternatives available (e.g. Redux, MobX) and we would recommend using them instead. - https://github.com/facebook/flux

I have never used flux, so I don't know the differences well. After a quick [look](https://stackoverflow.com/a/32916602) I don't see significant upsides.

> it can register stores any time rather than once at the bootstrap

With redux, there should only be one store anyway. It may be required to register reducers all at once at bootstrap, but why would it be a problem?

We should also think further than caching entities. As we replace more angular parts with react, we will probably need to introduce application state other than entities, that could be shared between components and extensions.

I'm not sure yet on which library would work best. I like redux, but am not sure what would be the best way to optimize re-renders. For example, if user object changes, avatar component re-renders even if the change was in a property that is not used in that component. Reselect is used in planning, but I don't like it at all. I would rather write `shouldComponentUpdate` manually for each component :smile: Maybe we should look into `MobX` for such global application state.

For now, regarding only caching entities I would stick to redux.

petrjasek:
@tomaskikutis not sure what you mean with that re-rendering - calling render method of a component or manipulating dom?

tomaskikutis:
I meant when components update automatically in response to changed props and `render` method is called by react internals. For example, see [this fiddle](https://jsfiddle.net/gqd0jtcp/). User component only uses name property, but always re-renders(see console output) when `lastlogin` changes, even though it's not used in the component. It could be fixed by passing only a single property that is needed instead of the whole object, but it would probably not work with bigger components when they need to pass entire objects to children.

petrjasek:
I see, I'm not sure those re-renders would be an issue in general, but if we measure it's a problem somewhere we can handle it either via `shouldComponentUpdate` or change the props not to pass the whole object but only the data it's using maybe.

tomaskikutis:
> I'm not sure those re-renders would be an issue in general

It wouldn't when the application is small(or react part is small in our case). I wouldn't be so sure that it wouldn't become an issue when it grows bigger. I believe there might be a legitimate reason, and not just a fad why people are using [redux selectors](https://redux.js.org/introduction/learning-resources#selectors).

> if we measure it's a problem somewhere we can handle it either via shouldComponentUpdate or change the props not to pass the whole object

It would work here and there, but I don't think it's a good solution architecturally, because it would increase coupling significantly. We would have to switch to designing interfaces not only according to business logic, but also to handle performance optimizations. It's useful to be able to pass big objects to components, have them use whatever they need, and still not lose out on performance.

I'm happy with redux for entities, but I would research mobx in the future before adding application level state to the store. I don't know if it can optimize re-renders in nested components*, but if it could, it would be awesome: we wouldn't have to profile for performance issues and would get great performance by default, even on low end hardware.

* Like the fiddle before, but assuming there's a wrapper component for `User` that doesn't do anything, except forwards props. It'd be great if mobx could prevent re-rendering of that wrapper at the same time when re-render of `User` is prevented.

PR #3889, opened 2021-05-21
SDESK-5968 ADD: Filter assets by tags

dev-tsoni:


dev-tsoni:
filtertags

scripts/extensions/sams/src/components/assets/assetFilterPanel.tsx
```diff
@@ -79,6 +87,54 @@ export class AssetFilterPanel extends React.PureComponent<IProps, IState> {
         };
     }
 
+    setLocalAssetTagSearchParams(tag: IAssetTag, method?: string) {
```
MarkLark86, line 90: This code is very similar to the code we already have in the `AssetEditor` component.

To help maintain this code (and reduce duplicated code), it would be ideal if we had a React component for these Asset Tag Inputs.

Using a similar interface to the UI-Framework input fields, the props could include:
```ts
interface IProps {
    label?: string;  // defaults to `gettext('Tags')
    value?: Array<IAssetTag>; // initial value
    onChange(value: Array<IAssetTag>): void; // provides the final Array value to be stored with the Asset
    disabled?: boolean;
    existingTagsOnly?: boolean;
}
```

This new component would handle internally the following:
* Searching for tags (i.e. via `samsApi.assets.searchTags`)
* Using `addTag` and `removeTag` to modify this components internal state (and calling `onChange` with the new Array)
* Rendering the Input field
* Rendering the list of Tags

There were some improvements done to this input type in `AssetEditor`, so it might be good to look at both of these implementations to come up with the final implementation

scripts/extensions/sams/src/components/assets/assetFilterPanel.tsx
```diff
@@ -79,6 +87,54 @@ export class AssetFilterPanel extends React.PureComponent<IProps, IState> {
...
+            const oldState = preState;
+            const tags: Array<IAssetTag> = preState.localSearchParams.tags! || [];
+            const index = tags.findIndex((tag) => {
+                return tag.code === value.code;
+            });
+
+            if (index === -1) {
+                tags!.push(value);
+                return {
+                    localSearchParams: {
```
MarkLark86, line 109: Does this override any other search params currently set?

scripts/extensions/sams/src/utils/assets.ts
```diff
@@ -141,3 +149,18 @@ export function isAssetLockedByCurrentUser(asset: Partial<IAssetItem>): boolean
         return false;
     }
 }
+
+export function toClientFormat(response: IAutoTaggingSearchResult): OrderedMap<string, IAssetTag> {
```
MarkLark86, line 153: Now that this function is used in multiple places, a more descriptive name would help. As we're converting a search result to Tags, maybe something like `convertTagSearchResultToAssetTags`?

scripts/extensions/sams/src/components/assets/assetFilterPanel.tsx
dev-tsoni: No. It only updates the existing tags with new ones appended to them.

scripts/extensions/sams/src/components/assets/assetFilterPanel.tsx
dev-tsoni: In here for `setLocalAssetTagSearchParams`, the issue is that states are diff for both filter and editor, due to which I did not use same function for both of them.
Rest things are mostly same.

scripts/core/helpers/url.ts
```diff
@@ -52,7 +52,22 @@ export const urlParams: ISuperdesk['browser']['location']['urlParams'] = {
     setString: (field, value) => {
         setUrlParameter(field, value?.length > 0 ? value : null);
     },
+    // Tags
+    getStringArray: (field) => {
```
MarkLark86, line 56: These two helper functions should be generic and deal with an Array of strings only, so they can be used elsewhere (outside SAMS). So the interface would be something like the following:
```typescript
interface urlParams {
    getStringArray(field: string): Array<string> | undefined;
    setStringArray(field: string, value: Array<string>): void;
}
```

Then in SAMS code you would convert before/after using these helper functions. i.e.

```typescript
// Get Tags from URL
const tags = (urlParams.getStringArray('tags') ?? []).map((tag) => ({
    code: tag,
    name: tag,
}));

// Set URL param for Tags
urlParams.setStringArray(
    'tags',
    (tags ?? []).map((tag) => tag.code)
);
```

scripts/extensions/sams/src/components/assets/assetFilterPanel.tsx
MarkLark86: The `onChange` prop of this new component would always provide an `Array<IAssetTag>`. It is then up to the consuming component (`AssetEditor` or `AssetFilterPanel`) to store these tags whichever way is required. This way the functionality is generic inside this new component, and we use `onChange` to control where the tags are stored

scripts/core/superdesk-api.d.ts
```diff
@@ -1413,6 +1413,11 @@ declare module 'superdesk-api' {
         getMediaId(attachment: IAttachment): IMedia['_id'];
     }
 
+    export interface IAssetTag {
```
MarkLark86, line 1416: With your latest changes, `IAssetTag` is no longer required in core

dev-tsoni: updated.

scripts/extensions/sams/src/components/assets/assetFilterPanel.tsx
MarkLark86: This improvement will be added at a later date

MarkLark86 (approved):

PR #3891, opened 2021-05-21
Fix master desk users state via redux

petrjasek:
SDESK-5817

replaces https://github.com/superdesk/superdesk-client-core/pull/3888 now using redux instead of flux


scripts/core/data/users/UserActions.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    initUsers(data: Array<IUser>) {
+        SuperdeskDispatcher.dispatch({
+            type: UserActionTypes.INIT_USERS,
+            payload: data,
+        });
+    },
+};
+
+export default Actions;
```
tomaskikutis, line 21: I'd use named exports everywhere. It enforces consistency of naming across the app and still allows for overwriting to handle cases when the name is taken by another variable.

scripts/core/data/users/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+interface IUserState {
+    entities: {[id: string]: IUser};
+}
+
+const initialState: IUserState = {
+    entities: {},
+};
+
+export default function(state = initialState, action: IAction): IUserState {
```
tomaskikutis, line 13: It's better to use a union type for `IAction`. That way, constants for actions are no longer needed and payloads are type-checked:

```
type IAction =
    {type: 'INIT_USERS', payload: string}
    | {type: 'UPDATE_USERS', payload: number};
```

scripts/apps/desks/services/DesksFactory.ts
```diff
@@ -133,6 +134,7 @@ export function DesksFactory($q, api, preferencesService, userList, notify,
                     _.each(result, (user) => {
                         self.userLookup[user._id] = user;
                     });
+                    UserActions.initUsers(self.users._items);
```
tomaskikutis, line 137: Could you fetch users to redux store on application start and instead of `userList.getAll()` above, get them from the redux store? It'd be good not to rely on angular services for new code.

tomaskikutis: What about this one? ^

petrjasek: I would rather do it in a separate PR if we go for this redux approach, it won't change the master desk code

tomaskikutis (approved):

petrjasek:
I've managed to override the latest changes with some other code, so @tomaskikutis can you pls check it again just in case I've missed something?

scripts/core/data/users/index.tsx
```diff
@@ -26,7 +26,7 @@ export function users(state = initialState, action: IUserAction): IUserState {
     }
 
     case 'UPDATE_ENTITY':
-        if (action.payload.resource === 'users') {
+        if (action.payload.resource === 'users' && state.entities[action.payload._id] !== action.payload.data) {
```
tomaskikutis, line 29: When would this action fire with old data? Is there any use in comparing object references rather than data?

tomaskikutis (approved):
I'd consider changing state shape from `state.users.entities` to `state.entities.users` so it's more clear that everything inside entities is simply cache and not related to application state.

scripts/core/data/users/index.tsx
petrjasek: it's related to the [cache PR](https://github.com/superdesk/superdesk-client-core/pull/3892), if there are multiple events triggering updates of the same user it might get called with same object when it gets resolved, so this will just update the state 1x instead of multiple times

tomaskikutis: Could you handle it earlier before it reaches the reducer? It'd be best to debounce the callback that is triggered by a websocket event so we aren't dispatching actions that might get discarded.

scripts/core/data/index.ts
```diff
@@ -38,17 +40,27 @@ addWebsocketEventListener(
...
-            store.dispatch({
-                type: 'UPDATE_ENTITY',
-                payload: {
-                    resource,
-                    _id,
-                    data,
-                },
+        const updateKey = `${resource}/${_id}`;
+
+        if (pendingUpdates[updateKey]) {
```
tomaskikutis, line 45: By ignoring a newer event, we'd end up with old data.

How about adding an optional argument to `addWebsocketEventListener` to specify the throttling?

petrjasek: it's not old data necessarily, but maybe this whole thing is just premature optimization, will drop it for now

tomaskikutis: If there is a new notification, it means that something has changed, and by ignoring that change, some properties would be out of date. If you are lucky with race conditions, you might accidentally receive up to date data, but there would still be times when it wouldn't work.

tomaskikutis (approved):

PR #3894, opened 2021-05-25
FIX: Clear filters in masterdesk search does not clear the search boxes. [SDESK-5862]

GyanP:


scripts/apps/master-desk/MasterDesk.tsx
```diff
@@ -46,6 +46,7 @@ interface IState {
     planning: boolean;
     deskFilter: string;
     filters: any;
+    clearFilterDisplay: boolean;
```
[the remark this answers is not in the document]
GyanP: if we checked button with inspecting `IState['filters']` then whenever we do change in any fields like **slugline, byline, headline** button will show instantly without click on `Apply Filter` button.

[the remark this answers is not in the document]
GyanP: Sorry, My bad, I think I miss understood the comment. I have made some changes that will address this comment.

scripts/apps/master-desk/components/FilterBarComponent.tsx
```diff
@@ -1,11 +1,13 @@
 import {gettext} from 'core/utils';
+import _ from 'lodash';
 import React from 'react';
 
 import {IFilter} from './FilterPanelComponent';
 
 interface IProps {
     filters: IFilter;
     onFilterChange(filters: any): void;
+    setClearAllFilters(): void;
```
tomaskikutis, line 10: It'd be good to limit coupling between filter-bar and other components.

For example, we want to clear both - applied filters and editable filters in the sidebar. To make things less coupled, this component shouldn't have any information about the sidebar filters. Using the following interface would accomplish it:

```
interface IProps {
    filters: IFilter;
    removeFilter(id: string): void;
    removeAllFilters(): void;
}
```

scripts/apps/master-desk/components/FilterBarComponent.tsx
```diff
@@ -1,11 +1,13 @@
 import {gettext} from 'core/utils';
+import _ from 'lodash';
```
tomaskikutis, line 2: To reduce bundle size, only import individual functions.

`import {isObject} from 'lodash';`

scripts/apps/master-desk/MasterDesk.tsx
```diff
@@ -116,6 +116,7 @@ export class MasterDesk extends React.Component<{}, IState> {
                 filters[key] = null;
             }
         }
+        this.setState({filters: {}});
```
tomaskikutis, line 119: If you are setting filters to an empty object, you shouldn't need the loop above?

scripts/apps/master-desk/components/FilterBarComponent.tsx
```diff
@@ -6,18 +6,17 @@ import {IFilter} from './FilterPanelComponent';
...
-    removeFilter(item: any): void {
-        this.props.onFilterChange({...this.props.filters, [item]: []});
+    removeFilter(id: string): void {
+        this.props.removeFilter(id);
     }
 
     clearFilters(): void {
-        this.props.setClearAllFilters();
-        this.props.onFilterChange({});
+        this.props.removeAllFilters();
```
tomaskikutis: Why not call functions from props directly in the render method instead of having additional functions which don't do anything but forward arguments?

scripts/apps/master-desk/MasterDesk.tsx
GyanP: Since there's a reference of the Filter panel component `state.filter` in the parent component, I created a shallow copy of parent components and then changed the object keys' value. Which eventually updated the Filter panel component state.filter. So, using the single parent component function inside one child, I was able to clear the state of another. But yes, that was the bad approach, so I've changed it now.

scripts/apps/master-desk/components/FilterPanelComponent.tsx
```diff
@@ -71,7 +93,32 @@ export class FilterPanelComponent extends React.Component<IProps, IState> {
...
+            byline: [],
+        };
+
+        this.setState({filter: filter});
+
+        this.props.onFilterChange({});
+    }
+
     render() {
+        const isDisplayClearFilters = () => {
```
tomaskikutis, line 109: It can be written in a shorter form:

```
const showClearFiltersButton = Object.keys(this.props.filters).some((key) => this.props.filters[key].length !== 0);
```

Regarding performance, it wouldn't have an impact here, but it's a good practice to stop iterating as soon as you find the value that you are looking for. When you set `isDisplay = true;` on line 115, there is no point in continuing the loop anymore because you already have the result.

scripts/apps/master-desk/components/FilterPanelComponent.tsx
```diff
@@ -34,6 +36,26 @@ export class FilterPanelComponent extends React.Component<IProps, IState> {
         this.handleInputChange = this.handleInputChange.bind(this);
         this.addFilter = this.addFilter.bind(this);
         this.clearDeskFilter = this.clearDeskFilter.bind(this);
+        this.clearFilters = this.clearFilters.bind(this);
+    }
+
+    componentDidUpdate(prevProps: IProps) {
```
tomaskikutis, line 42: When state is tightly connected to props and needs to be updated when props change, it's cleaner to use a key instead. You wouldn't need `componentDidUpdate` and the component would re-initialize when filters change.

tomaskikutis: Yep, it does initialize the state again, but since your state is derived from props, that shouldn't be problem if you add a few lines in the constructor to initialize state from props. I'm not 100% sure, but looks like it should work.

PR #3896, opened 2021-05-27
add tooltips to imatrics tags

petrjasek:
SDNTB-692

### DIFF
--- a/scripts/extensions/auto-tagging-widget/src/tsconfig.json
+++ b/scripts/extensions/auto-tagging-widget/src/tsconfig.json
@@ -1,11 +1,6 @@
 {
     "extends": "superdesk-code-style/tsconfig-strict",
     "compilerOptions": {
-        "outDir": "../dist/src",
-        "skipLibCheck": true,
-        "types": []
-    },
-    "exclude": [
-        "../node_modules"
-    ]
+        "outDir": "../dist/src"
+    }
 }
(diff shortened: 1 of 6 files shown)


scripts/extensions/auto-tagging-widget/src/types.ts
```diff
@@ -3,6 +3,8 @@ export interface ITagUi {
     description?: string;
     qcode: string;
     source?: string;
+    original_source?: string;
+    aliases?: Array<string>;
```
tomaskikutis, line 7: Could you add comments on what those are?

scripts/extensions/auto-tagging-widget/src/tsconfig.json
```diff
@@ -1,6 +1,11 @@
 {
     "extends": "superdesk-code-style/tsconfig-strict",
     "compilerOptions": {
-        "outDir": "../dist/src"
-    }
+        "outDir": "../dist/src",
+        "skipLibCheck": true,
```
tomaskikutis, line 5: What about installing [types](https://github.com/react-bootstrap/react-bootstrap-typescript)? If that library doesn't work, you could try importing it as `any` so it doesn't affect other libraries.

```
import __OverlayTrigger from 'react-bootstrap/OverlayTrigger';

const OverlayTrigger = __OverlayTrigger as unknown as any;
```

scripts/extensions/auto-tagging-widget/src/types.ts
petrjasek: it's actually aliases :wink: so other names which are used for the concept

scripts/extensions/auto-tagging-widget/src/tsconfig.json
petrjasek: will revert those changes, I had some issues but that got fixed when updating typescript

tomaskikutis (approved):

PR #3897, opened 2021-05-27
support installing extensions outside of client core repository using build tools

tomaskikutis:
Initially I have published the planning extension to npm, because of the issue of not being able to find API typescript definitions when building (superdesk-core is a devDependency of planning extension so it can be referenced during development). It wouldn't work to use the npm package with `planning-mvp` that is used for building test instances on fireq, so I'm copying the API typescript definitions to all extensions.

### DIFF
--- a/build-tools/package-lock.json
+++ b/build-tools/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@superdesk/build-tools",
-  "version": "1.0.6",
+  "version": "1.0.7",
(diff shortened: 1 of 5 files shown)


petrjasek:
what about pushing those types as a package itself?

tomaskikutis:
We could push it as separate package, but we would then be more likely to bump into conflicts with versions not matching between packages.

petrjasek (approved):

PR #3903, opened 2021-06-04
FIX: It is possible to publish a story from authoring desk via 'quick publishing buttons'. [SDESK-5988]

GyanP:


petrjasek (commented):
imo would be good to also implement some backend check for this

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -186,10 +187,7 @@ export function AuthoringDirective(
                                 $scope.stage = result;
                             });
 
-                        desks.fetchDeskById($scope.origItem.task.desk).then((desk) => {
-                            $scope.deskName = desk.name;
-                            $scope.deskType = desk.desk_type;
-                        });
+                        getDesk();
```
tomaskikutis, line 190: Functions that are called `get{Something}` shouldn't **write** to variables outside the function. `setDesk` would be a better name.

tomaskikutis (approved):

PR #3905, opened 2021-06-04
SDESK-6012 FIX: SAMS advance search not working.

dev-tsoni:


scripts/extensions/sams/src/components/assets/assetFilterPanel.tsx
```diff
@@ -149,7 +149,7 @@ export class AssetFilterPanel extends React.PureComponent<IProps, IState> {
             name: undefined,
             filename: undefined,
             description: undefined,
-            tags: undefined,
+            tags: [],
```
MarkLark86, line 152: This does't cover all scenarios, such as when assigning search params from the URL params (where it's possible that `tags` could be set an `undefined` value).

Supporting the `IAssetSearchParams.tag` attribute to be undefined would be more resilient.

The only place I see `tags == undefined` not currently being supported is when setting search parms: https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/extensions/sams/src/api/assets.ts#L360.

`params.tags!.map` is used here, where as `(params.tags ?? []).map` would support an `undefined` value as well

MarkLark86 (approved):

PR #3909, opened 2021-06-07
implement system messages UI

petrjasek:
SDESK-5969 SDESK-5659

### DIFF
--- a/scripts/apps/system-messages/index.ts
+++ b/scripts/apps/system-messages/index.ts
@@ -1,0 +1,41 @@
+import {gettext} from 'core/utils';
+import {reactToAngular1} from 'superdesk-ui-framework';
+import {SystemMessagesSettingsComponent} from './components/system-messages-settings';
+import {SystemMessagesComponent} from './components/system-messages';
+import {IBaseRestApiResponse, IUser} from 'superdesk-api';
+
+export interface ISystemMessage extends IBaseRestApiResponse {
+    type: 'warning' | 'alert' | 'primary' | 'success';
+    is_active: boolean;
+    message_title: string;
+    message: string;
+    user_id: IUser['_id'];
+}
+
+export const RESOURCE = 'system_messages';
+
+angular.module('superdesk.apps.system-messages', [
+    'superdesk.core.activity',
+])
+    .component('sdSystemMessagesSettings', reactToAngular1(
+        SystemMessagesSettingsComponent,
+        [],
+        [],
+    ))
+
+    .component('sdSystemMessages', reactToAngular1(
+        SystemMessagesComponent,
+        [],
+        [],
+    ))
+
+    .config(['superdeskProvider', function config(superdesk) {
+        superdesk.activity('/settings/system-messages', {
+            label: gettext('System Message'),
+            template: '<sd-system-messages-settings></sd-system-messages-settings>',
+            category: superdesk.MENU_MAIN,
+            priority: 1000,
+            adminTools: true,
+            privileges: {[RESOURCE]: 1},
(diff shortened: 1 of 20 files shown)


scripts/core/superdesk-api.d.ts
```diff
@@ -897,6 +898,9 @@ declare module 'superdesk-api' {
 
         field: string;
 
+        /** default value */
+        value?: string;
```
tomaskikutis, line 902: Is this used anywhere?

scripts/core/ui/components/generic-form/form-field.tsx
```diff
@@ -43,9 +46,13 @@ export function getFormFieldPreviewComponent(
     item: {readonly [key: string]: any},
     formFieldConfig: IFormField,
     options: { showAsPlainText?: boolean } = {},
-): JSX.Element {
+): JSX.Element | null {
     const Component = getFormFieldComponent(formFieldConfig.type);
 
+    if (Component == null) {
```
tomaskikutis, line 52: Why would it be `null`?

scripts/core/ui/components/generic-form/from-group.tsx
```diff
@@ -37,6 +37,10 @@ export class FormViewEdit extends React.Component<IProps> {
                         } else if (isIFormField(item)) {
                             const FieldComponent = getFormFieldComponent(item.type);
 
+                            if (FieldComponent == null) {
```
tomaskikutis, line 40: Why would it be `null`?

scripts/core/ui/components/generic-form/form-field.tsx
petrjasek: I was adding a hidden field, which wouldn't be rendered, but at the end I'm not using it

tomaskikutis: I'm not getting it. How would a "hidden field" work here? From what I understand, all field types should have a corresponding component that would render the input field.

petrjasek: it was like `<input type="hidden" value="...">`, was using it for `user_id`, but then I've moved that handling to the server

scripts/apps/system-messages/index.ts
```diff
@@ -0,0 +1,41 @@
...
+import {SystemMessagesSettingsComponent} from './components/system-messages-settings';
+import {SystemMessagesComponent} from './components/system-messages';
+import {IBaseRestApiResponse, IUser} from 'superdesk-api';
+
+export interface ISystemMessage extends IBaseRestApiResponse {
+    type: 'warning' | 'alert' | 'primary' | 'success';
+    is_active: boolean;
+    message_title: string;
+    message: string;
+    user_id: IUser['_id'];
```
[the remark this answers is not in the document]
petrjasek: it's always populated by server

scripts/core/ui/components/generic-form/form-field.tsx
tomaskikutis: I think `component_parameters` can be used for holding extra info that you tried to put inside a hidden field. Or if it doesn't work in some scenario, maybe it'd be worth adding "hidden field" input type. I'd drop this for now so we don't carry dead code that we would be afraid to remove in the future.

scripts/core/superdesk-api.d.ts
petrjasek: I had it for that hidden field, can remove it now

tomaskikutis (approved):

PR #3912, opened 2021-06-11
[SDBELGA-539] Update ui-framework to 2.4.3

MarkLark86:


MarkLark86:
@petrjasek Bumping of the ui-framework is required for `bel8` release (due to the `BelgaCoverage` extension).

It turns out that using 2 separate versions of the ui-framework (and importing the bundles CSS files) causes UI issues.

petrjasek (approved):

tomaskikutis:
@petrjasek I'd be good to drop ui-framework from devDependencies in belga extension. In order to compile extensions, we are installing them in dev mode and such collisions are likely in the future if we keep ui-framework dependency in more than one place.

PR #3919, opened 2021-06-18
update ui framework

MarkLark86:
This is required for SDESK-5801


tomaskikutis (approved):

petrjasek:
adding that to 2.3, there is a fix for SDESK-6060

PR #3920, opened 2021-06-18
FIX: New corrections workflow: items get stuck during corrections workflow [SDCP-551]

GyanP:
Prevent the multi-time same API calling on the correct button.

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
@@ -124,6 +124,7 @@ export function AuthoringDirective(
             $scope.mediaFieldVersions = {};
             $scope.refreshTrigger = 0;
             $scope.isPreview = false;
+            $scope.isCorrectionDisabled = false;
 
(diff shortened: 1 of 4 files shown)


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -124,6 +124,7 @@ export function AuthoringDirective(
             $scope.mediaFieldVersions = {};
             $scope.refreshTrigger = 0;
             $scope.isPreview = false;
+            $scope.isCorrectionDisabled = false;
```
tomaskikutis, line 127: It's disabled only when in progress right? It'd be clearer to call it `isCorrectionInProgress`

GyanP: Yes, you are right.
> It'd be clearer to call it isCorrectionInProgress

Firstly, I thought of the same name. Let me just rename it.

tomaskikutis (approved):

tomaskikutis (approved):

PR #3921, opened 2021-06-18
spiked view - fix blinking

tomaskikutis:
SDESK-6038

It was caused by using absolute positioning for the loader. It would cover one item while loading and when loading was quick it used to appear as if items are blinking.

### DIFF
--- a/scripts/core/data/index.ts
+++ b/scripts/core/data/index.ts
@@ -29,16 +29,11 @@ addWebsocketEventListener(
     (event: IWebsocketMessage<IResourceUpdateEvent>) => {
         const {resource, _id} = event.extra;
 
-        const RESOURCE_BLACKLIST = [
-            'auth',
-            'audit',
-
-            // these are not collections
-            'archive_spike',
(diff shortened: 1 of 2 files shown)


petrjasek (approved):

scripts/core/data/index.ts
```diff
@@ -32,6 +32,10 @@ addWebsocketEventListener(
         const RESOURCE_BLACKLIST = [
```
MarkLark86, line 32: IMO, this should rather be an `allow list` rather than a `deny list`.
I am seeing requests to the server that are not needed, with the `UPDATE_ENTITY` dispatched to redux.

image

petrjasek: yep makes sense

petrjasek: @tomaskikutis can you make that change in this PR?

tomaskikutis: Sure

petrjasek (approved):

PR #3922, opened 2021-06-18
store aliases and original source on item subjects

petrjasek:
SDNTB-692


scripts/core/superdesk-api.d.ts
```diff
@@ -1901,5 +1901,7 @@ declare module 'superdesk-api' {
         source?: string;
         translations?: {};
         altids?: {[key: string]: string};
+        aliases?: Array<string>;
```
tomaskikutis, line 1904: Can you add a comment describing the difference between source and original_source?

tomaskikutis (approved):

PR #3924, opened 2021-06-22
Validate required fields for generic form

tomaskikutis:
SDESK-6057

### DIFF
--- a/scripts/core/ui/components/generic-form/has-value.tsx
+++ b/scripts/core/ui/components/generic-form/has-value.tsx
@@ -1,0 +1,35 @@
+import {assertNever} from 'core/helpers/typescript-helpers';
+import {isIFormGroup, isIFormField, FormFieldType} from './interfaces/form';
+import {IFormField, IFormGroup} from 'superdesk-api';
+import {stripHtmlTags} from 'core/utils';
+
+export function hasValue(fieldConfig: IFormField, value: any): boolean {
+    const type: FormFieldType = fieldConfig.type;
+
+    switch (type) {
+    case FormFieldType.textSingleLine:
+        return typeof value === 'string' && value.trim().length > 0;
+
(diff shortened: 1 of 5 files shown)


petrjasek (approved):

scripts/core/ui/components/generic-form/has-value.tsx
```diff
@@ -0,0 +1,35 @@
+import {assertNever} from 'core/helpers/typescript-helpers';
+import {isIFormGroup, isIFormField, FormFieldType} from './interfaces/form';
+import {IFormField, IFormGroup} from 'superdesk-api';
+import {stripHtmlTags} from 'core/utils';
+
+export function hasValue(fieldConfig: IFormField, value: any): boolean {
```
petrjasek, line 6: I'm not sure about this whole structure where if you add a new field type you have to modify bunch of files, would it be possible to define the field type with methods like hasValue, getPreview, etc. ?

tomaskikutis: It would definitely be easier to work with, if it was a single object.

On the other hand, the inconvenience is quite minimal. I have just added a new field regarding instance configs PR. I added an entry to the main `enum`, ran lint and typescript threw errors everywhere that something's missing. It took a few minutes to go over it and add a new case.

I'll check it out in that UI configs PR if there's an easy way to improve it.

petrjasek: it's a bit different probably if you know the code and just fix those parts, for me when I've added a new type it was like:
**why are there so many errors now???**

tomaskikutis: A comment above `enum` declaration explaining the process of adding new fields would probably already half-solve it. Anyway, I'll look into putting it all in a single object and will add a comment if I can't do it easily.

PR #3930, opened 2021-06-24
always load selected value in autocomplete

tomaskikutis:
SDBELGA-552

### DIFF
--- a/scripts/core/ui/components/Autocomplete.tsx
+++ b/scripts/core/ui/components/Autocomplete.tsx
@@ -7,6 +7,7 @@ import {IBaseRestApiResponse, IRestApiResponse} from 'superdesk-api';
 interface IProps<T extends IBaseRestApiResponse> {
     placeholder: string;
     query(searchString): Promise<IRestApiResponse<T>>;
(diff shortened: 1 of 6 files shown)


scripts/core/ui/components/Autocomplete.tsx
```diff
@@ -18,6 +19,7 @@ interface IProps<T extends IBaseRestApiResponse> {
 interface IState<T> {
     fetchedItems?: Array<T>;
     loading: boolean;
+    selectedT: T | null;
```
petrjasek, line 22: what does `T` stand for in `selectedT` ?

tomaskikutis: For the generic object, same as in `Array<T>`

petrjasek: is that an idiom? if not I would go for something like `selectedItem`, looks like a typo to me

tomaskikutis: I don't think we use it much. I'll change it :)

petrjasek (approved):

PR #3934, opened 2021-06-30
FIX: Abbreviations are not working. [SDESK-6062]

GyanP:


scripts/core/spellcheck/spellcheck.ts
```diff
@@ -8,7 +8,7 @@ import {debounce, once} from 'lodash';
 SpellcheckService.$inject = ['$q', 'api', 'dictionaries', '$rootScope', '$location', 'lodash', 'preferencesService'];
 function SpellcheckService($q, api, dictionaries, $rootScope, $location, _, preferencesService) {
     var PREFERENCES_KEY = 'spellchecker:status',
-        lang,
+        lang = 'en',
```
tomaskikutis, line 11: Would abbreviations work in another language with this code? I think it'd better to make sure that a correct language is being set without hard-coding values

scripts/core/editor3/store/index.ts
```diff
@@ -193,15 +195,20 @@ export default function createEditorStore(
     if (spellcheck != null) {
         // after we have the dictionary, force update the editor to highlight typos
         spellcheck.getDict().finally(() => store.dispatch(forceUpdate()));
-
         spellcheck.getAbbreviationsDict().then((abbreviations) => {
             store.dispatch(setAbbreviations(abbreviations || {}));
         });
     }
 
+    editor3Stores.push(store);
```
petrjasek, line 203: when does it remove stores? it seems like it would be pushing all the time, that might use quite some memory eventually

petrjasek (approved):

PR #3935, opened 2021-06-30
[SDESK-5801] SAMS: Use data-theme for dark mode

MarkLark86:


petrjasek:
I would use [Context](https://reactjs.org/docs/context.html) for this one

MarkLark86:
> I would use [Context](https://reactjs.org/docs/context.html) for this one

Fritz has improved controlling the theme of components in the UI-Framework. If any node has a `data-theme` attribute, then all child nodes will use that theme.

This PR removes the previous implementation which required to pass `theme` to all render components.
I am unsure as to what benefits using `Context` would provide here, given the usage of this new `data-theme` attribute.

petrjasek:
I see, so it's handled on css level

petrjasek (approved):

PR #3943, opened 2021-07-08
Instance configuration in user interface

tomaskikutis:
SDESK-5991

The plan is to generate JSON schema from TypeScript interfaces and then generate a user interface from that schema. Values from user interface would get saved to the database and downloaded by front-end on next page reload.

I'm using https://github.com/YousefED/typescript-json-schema for generating a JSON shema. For generating user interface I was considering using https://github.com/json-editor/json-editor but decided to go with our [generic form](https://github.com/superdesk/superdesk-client-core/tree/develop/scripts/core/ui/components/generic-form) since it has most of what is needed.

Our [current interface](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L1687) for settings is big, badly grouped and is missing type definitions so I'm using a test interface(`ICoreSettings`) for development.

I have added a script(`generate-instance-configuration-schema`) to `@superdesk/build-tools` to generate JSON schema from that interface. It is saved to `scripts/instance-settings.ts`. In `scripts/apps/instance-configuration/components/InstanceConfiguration.tsx`, JSON schema is converted into a format that then can be used by our generic form in order to render the user interface.

To do next:
* ~Add back-end endpoints that would allow reading and writing values from UI settings editor.~
* ~Connect UI settings editor to use that endpoint.~
* ~Fix privileges for that endpoint~
* Start moving settings from `superdesk.config.js`. In case there are lots of usages for a specific config option, it might be useful to create a function to map the new format to `appConfig` format, so usages don't have to be changed. For new settings and the ones that are easy to move we should not be using `appConfig`, and instead create a new variable that implements the interface that settings editor is being generated from.
* Add missing field types to generic form.
* ~Fix translations of labels. In the `"translations" `objects in [scripts/instance-settings.ts](https://github.com/superdesk/superdesk-client-core/blob/3893353f4ce1b4db089e75d9a295a984e20a614f/scripts/instance-settings.ts) labels are wrapped in `gettext` as they are in the interfaces, which is in camelCase format. Generation of translations needs to be modified(in build-tools) so camelCased labels are split into multiple words and converted to lower case.~
* ~[generic form] Improve styles for nested groups. For example, "editor" group is nested under "authoring", but it's not visible in the UI(see the screenshot).~
* ~Load configs synchronously on startup~
* ~Update all generic ui components to show description~
* ~Fix object flattening - preserve arrays~
* ~Improve array support - add string, number, boolean and union types~
* Implement `anyOf` e.g. `Array<string | {firstName: string}>`
* ~Implement Arrays of objects~
* ^ Fix validation for arrays of objects
* Implement merging interfaces(or generated JSON schemas) from extensions
* ~bug: empty form values for numbers can be saved even if required~
* ~ui improvement - when errors are shown for a field, hide hint message. Otherwise both pieces of text go on top of each other.~
* ~validation should fail when array field has empty values~


MarkLark86:
> Fix labels. We use camelCase for field names, it needs to be split into multiple words and converted to lower case in build-tools script before generating JSON schema.

@tomaskikutis Using this method for labels, how would we support translation into different languages?

tomaskikutis:
There are `"translations"` objects in the generated file [scripts/instance-settings.ts](https://github.com/superdesk/superdesk-client-core/blob/3893353f4ce1b4db089e75d9a295a984e20a614f/scripts/instance-settings.ts). I'm suggesting to only modify the part wrapped in `gettext`. I'll clarify it in the PR description.

scripts/instance-settings.generated.ts
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable quotes, comma-dangle */
```
petrjasek, line 1: should this be commited in repo if it's generated?

scripts/apps/instance-configuration/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    ))
+
+    .config(['superdeskProvider', function config(superdesk) {
+        superdesk.activity('/settings/instance-configuration', {
+            label: gettext('Instance configuration'),
+            // tslint:disable-next-line: max-line-length
+            template: '<div class="sd-scrollable-inside-absolute-container"><sd-instance-configuration-settings></sd-instance-configuration-settings></div>',
+            category: superdesk.MENU_MAIN,
+            priority: 1000,
+            adminTools: true,
```
petrjasek, line 30: do we show these to administrators only?

scripts/instance-settings.generated.ts
tomaskikutis: Good point, I've removed it.

scripts/apps/instance-configuration/index.ts
tomaskikutis: No, these only describe position in UI. I have added `privileges: {users: 1},`, but it would probably be better to use another privilege, since this is about instance config, not users. Do you know what privilege could I use?

petrjasek: we would have to add a new one, atm on the backend it checks if user is type administrator only

petrjasek (approved):

PR #3948, opened 2021-07-15
Disable the package feature for Belga via config. [SDBELGA-558]

GyanP:


petrjasek:
afaik the idea was to set that config on superdesk-belga repo, are there any changes needed in core really?

GyanP:
> afaik the idea was to set that config on superdesk-belga repo, are there any changes needed in core really?

Yes, As per the description of a ticket which was Belga does not make use of the "packages" and thus the respective actions and items are creating confusion in the UI, So what I understood from this is that we need to disable the **Create Package options** from MultiActions Bar and 3dot Menu of a story as well as the package filter from Monitoring View and Spike view for the Belga. Therefore, I have added a new configuration which is `packageDisable`, and made the changes in the client-core.

petrjasek:
I see, there is also this config to hide create package btn but that doesn't remove the other options,
so would be good to connect it:
https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/workspace/content/views/sd-content-create.html#L43

petrjasek:
or rather merge this options into single config

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -53,6 +54,16 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
...
 
+    if (appConfig.features.hideCreatePackage) {
+        availableFileTypes.forEach((file) => {
+            if (file.type !== 'composite' && file.type !== 'highlight-pack') {
+                this.fileTypes.push(file);
+            }
+        });
+    } else {
+        this.fileTypes = availableFileTypes;
+    }
```
petrjasek, line 65: imo this would be easier to read using `.filter`
```ts
this.fileTypes = availableFileTypes.filter(
    (fileType) => (
        fileType.type !== 'composite' ||
        fileType.type !== 'highlight-pack' ||
        appConfig.features.hideCreatePackage !== false
    ));
```

scripts/core/constants.ts
```diff
@@ -1,6 +1,18 @@
...
+
 // Used for determining whether to update articles list upon receiving a websocket event.
 export const ARTICLE_RELATED_RESOURCE_NAMES = [
     'archive',
     'archive_spike',
     'archive_unspike',
 ];
+
+export const ITEM_TYPES = [
+    {type: 'text', label: gettext('text')},
```
petrjasek, line 11: gettext won't work here as it's called before translations can be loaded

scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -54,15 +53,12 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
...
-            }
-        });
-    } else {
-        this.fileTypes = availableFileTypes;
-    }
+    this.fileTypes = this.fileTypes.filter(
+        (fileType) => (
+            appConfig.features.hideCreatePackage ?
+                fileType.type !== 'composite' && fileType.type !== 'highlight-pack' :
+                fileType
```
petrjasek, line 60: it would be more readable if you put here `true`, if the config is not set there is no filtering

petrjasek (commented):
imo we should avoid that duplication, there is same code in 2 places

scripts/apps/monitoring/controllers/AggregateCtrl.ts
GyanP: You mean something like this: 
`appConfig.features.hideCreatePackage ?
                fileType.type !== 'composite' && fileType.type !== 'highlight-pack' :
                true`
 or something like this:
`appConfig.features.hideCreatePackage === true ?
                fileType.type !== 'composite' && fileType.type !== 'highlight-pack' :
                fileType`

petrjasek: I mean the first one

scripts/core/ArticlesListByQueryWithFilters.tsx
```diff
@@ -192,9 +180,19 @@ export class ArticlesListByQueryWithFilters extends React.PureComponent<IProps,
...
+                {type: 'composite', label: gettext('package')},
+                {type: 'highlight-pack', label: gettext('highlights package')},
+                {type: 'video', label: gettext('video')},
+                {type: 'audio', label: gettext('audio')},
+            ];
+
             const options: Array<IFileTypeOption> = [
                 filterAll,
-                ...getItemTypes().map((itemType) => {
+                ...getItemTypes(item_types).map((itemType) => {
```
petrjasek, line 195: why do you pass those types?

GyanP: > why do you pass those types?

To remove duplicates I made a function **getItemTypes** in utils.ts which takes all available filters as an argument and returned required filters based on the **hideCreatePackage** setting.

petrjasek: yep but seems that what you pass there is pretty much the same always, isn't it?

GyanP: No, Not exactly, Actually when it calls in `ArticlesListByQueryWithFilters.tsx` we did not pass **all filter** because it comes from the `filterAll` object. Therefore we need an argument to pass but I have made some changes and after that changes, we need not have to pass any argument.

PR #3950, opened 2021-07-22
Search on sign-off in monitoring view (marked for user) [SDBELGA-417]

GyanP:


petrjasek:
would it be possible to handle it on server? imo it's not needed to complicate client for it

GyanP:
> would it be possible to handle it on server? imo it's not needed to complicate client for it

Yes, I think we can handle it on the server.

GyanP:
Handled on the server side.

PR #3951, opened 2021-07-26
FIX: Edit button has to be pressed twice to edit an unpublished item in output stage. [SDESK-6091]

GyanP:


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -246,7 +246,11 @@ export function AuthoringDirective(
             $scope.edit = function edit() {
                 if ($scope.origItem.state === 'unpublished') {
                     api.update('archive', $scope.origItem, {state: 'in_progress'})
-                        .then((updated) => authoringWorkspace.edit(updated));
+                        .then((updated) => {
+                            $scope.close().then(() => {
```
petrjasek, line 250: can you add a comment why it's needed?

petrjasek (approved):

tomaskikutis:
I thought unpublished items are not supposed to be in the output stage in the first place?

> `// for updating and refreshing the item everywhere,`

@GyanP what does "everywhere" include? Shouldn't the item update in response to a websocket message? It'd be much better than to trigger closing/opening.

PR #3953, opened 2021-07-29
implement auto refreshing for masterdesk

petrjasek:
- implement request queue
- implement data provider using the queue
- fix assignments to use single query for all desks

SDESK-5858

uses https://github.com/superdesk/superdesk-core/pull/2246

### DIFF
--- a/scripts/core/helpers/request-queue.ts
+++ b/scripts/core/helpers/request-queue.ts
@@ -1,0 +1,198 @@
+import {httpRequestJsonLocal} from './network';
+
+export interface IRequestParams {
+    method: 'GET' | 'POST';
+    endpoint: string;
+    data?: any;
+    params?: any;
+    id?: string;
+    abortSignal?: AbortSignal;
+}
+
+export enum RequestPriority {
+    LOW,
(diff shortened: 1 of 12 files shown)


pavlovicnemanja (approved):

scripts/core/helpers/request-queue.ts
```diff
@@ -0,0 +1,198 @@
...
+    concurrency = 2;
+    queue: Array<Request> = [];
+    updateTimeout: number | null = null;
+
+    add(params: IRequestParams, provider: any, priority: RequestPriority = RequestPriority.LOW) {
+        const req = new Request(params, provider, priority);
+        const prev = this.queue.find((_req) => _req.isEqual(req) && !_req.isFinished());
+
+        if (prev != null) {
+            return prev.promise;
```
tomaskikutis, line 34: Shouldn't existing request be cancelled and a new one initialized in order to avoid old data being returned?

petrjasek: that should be before we get the data, but I could check if it's pending and not started yet

PR #3954, opened 2021-08-02
FIX: Limit further Typing does not work for 'Extended Headline' field. [SDESK-6109]

GyanP:


petrjasek:
what was the problem?

GyanP:
> what was the problem?

When we select **Limit further typing** or **HighLight characters** from the **Headline** field then we get the `headline` value but in the case of the **Extended Headline** field, we get `extra>headline_extended` value. Therefore, I have split the `headline_extended` from `extra>headline_extended`.

petrjasek:
> > what was the problem?
> 
> When we select **Limit further typing** or **HighLight characters** from the **Headline** field then we get the `headline` value but in the case of the **Extended Headline** field, we get `extra>headline_extended` value. Therefore, I have split the `headline_extended` from `extra>headline_extended`.

shy is that a problem? it seems like it's the correct id for the field we use elsewhere

GyanP:
> > > what was the problem?
> > 
> > 
> > When we select **Limit further typing** or **HighLight characters** from the **Headline** field then we get the `headline` value but in the case of the **Extended Headline** field, we get `extra>headline_extended` value. Therefore, I have split the `headline_extended` from `extra>headline_extended`.
> 
> why is that a problem? it seems like it's the correct id for the field we use elsewhere

When we select **Limit further typing** or **Highlight characters** from the **Headline** field we get the value `headline` and which is available as a key in object get from `event.detail?.[CHARACTER_LIMIT_UI_PREF]`. Thus, it set chosen value for the field but in the case of **Extended Headline** we get the value `extra>headline_extended` which is not available in the object. That's why I have split the `extra` keyword from `extra>headline_extended` and get `headline_extended` which matches with the key of the object and set its value which we choose.

petrjasek (approved):

PR #3958, opened 2021-08-05
As an editor I like to see and search based on all keywords coming with the ingest sources AFP, ANP and dpa. [SDBELGA-555]

GyanP:


petrjasek:
what does this new config do?

GyanP:
> what does this new config do?

we have initialized the **original metadata** with empty items in `vocabulary.json`. Therefore it does not display the original metadata field on UI. We might get the same requirement in the future that has empty items. That's the reason I made a new config `metadata_with_no_items` in which we can specify the metadata names which we wanted to show on the UI even if they don't contain any items. However, I think we have one more approach that what if we keep saving the metadata in the original metadata in vocabulary. For now, we are just saving it in the article as a subject, what we can do is we can save it in the original metadata vocabulary. What are your thoughts on this?

petrjasek:
I wouldn't make this a global config, what about setting that cv in vocabularies.json as readonly and display the values for such cvs without dropdown?

tomaskikutis (approved):

PR #3965, opened 2021-08-12
FIX: Save button is not activated on editing the caption for fetched media placed in Media Gallery. [SDESK-6067]

GyanP:


tomaskikutis:
What was the issue?

It'd also be good to always add conditions explicitly when using non boolean values in boolean operations:
i.e. to use `if (a != null && b != null)` instead of `if (a && b)`

GyanP:
> What was the issue?
> 
> It'd also be good to always add conditions explicitly when using non boolean values in boolean operations:
> i.e. to use `if (a != null && b != null)` instead of `if (a && b)`

When we first drag and drop an ingested image item into **Media Gallery** then the **decription_text** key is not present there.

tomaskikutis:
> When we first drag and drop an ingested image item into Media Gallery then the decription_text key is not present there.

I meant to ask what was the issue with the code? I see you added an `OR` condition, but it's not clear to me which use case is it addressing?

GyanP:
> > When we first drag and drop an ingested image item into Media Gallery then the decription_text key is not present there.
> 
> I meant to ask what was the issue with the code? I see you added an `OR` condition, but it's not clear to me which use case is it addressing?

If `associations` are present then we have **description_text** key and its value which satisfy if `(typeof scope.item.associations?.[association]?.[field] === 'string')` condition and allow adding description text but when **description_text** key is not present in `associations` then condition does not satisfy. For this scenario, I just added `OR` condition to check if `associations` present and **description_text** key and its value are not `null` then the **description_text** key will be added.

tomaskikutis (approved):

PR #3966, opened 2021-08-15
Fix closing articles from bottom toolbar

tomaskikutis:
SDESK-6037


petrjasek (approved):
just not sure if the modal fix should be part of this PR

tomaskikutis:
Do you mean replacement of unsaved changes prompt? I found the code hard to understand so I've restructured it. I don't think there's a better time to fix it than when already fixing a bug related to the same code.

PR #3968, opened 2021-08-16
tweak master desk auto updating

petrjasek:
- update on spike/translate/duplicate/etc
- check for changes on second+ response and only call handler if so
- only reuse previous request if it wasn't started

SDESK-5858

### DIFF
--- a/scripts/apps/master-desk/components/OverviewComponent.tsx
+++ b/scripts/apps/master-desk/components/OverviewComponent.tsx
@@ -56,7 +56,9 @@ export class OverviewComponent extends React.Component<IProps, IState> {
             },
             {
                 archive: {
+                    create: true,
                     update: [
+                        'state',
                         'task.desk',
                         'task.stage',
                         'slugline',
--- a/scripts/core/helpers/data-provider.ts
+++ b/scripts/core/helpers/data-provider.ts
@@ -7,11 +7,17 @@ import {
     IListenTo,
     IResourceCreatedEvent,
     IResourceDeletedEvent,
+    IRestApiResponse,
 } from 'superdesk-api';
 
-import {debounce} from 'lodash';
+import {debounce, isEqual} from 'lodash';
 import {requestQueue, RequestPriority} from './request-queue';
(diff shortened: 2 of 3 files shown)


scripts/apps/master-desk/components/OverviewComponent.tsx
```diff
@@ -56,7 +56,9 @@ export class OverviewComponent extends React.Component<IProps, IState> {
             },
             {
                 archive: {
+                    create: true,
                     update: [
+                        'state',
```
tomaskikutis, line 61: It'd be better fields didn't have to be hardcoded, but would instead listen for changes on query fields, like in `with-live-query.tsx` for example.

scripts/core/helpers/data-provider.ts
```diff
@@ -21,6 +27,7 @@ export class DataProvider implements IDataProvider {
     listeners: Array<() => void> = [];
 
     private scheduleUpdate: () => void;
+    private cache: IRestApiResponse<any>;
```
tomaskikutis, line 30: It's be better to add [some generic types](https://user-images.githubusercontent.com/2076953/129571676-66865acb-4ad2-494e-b1a7-9626513e7de9.png) instead of `any`.

scripts/apps/master-desk/components/OverviewComponent.tsx
petrjasek: here it's doing aggregations and none of those fields are actually returned, not sure how to best specify that

tomaskikutis: It'd be good to have this list of fields in a single place.

If we are sending fields on which to aggregate from client to server, then we could use the same fields(in client code) to decide whether to refresh data.

If we don't send the fields by which to aggregate and server and it is hardcoded on the server, it would be better to use a custom websocket event that would only fire when aggregated fields change.

petrjasek: will give it a try and might introduce it in another PRs

tomaskikutis (approved):

PR #3973, opened 2021-08-24
Display an error message when gmail ingest fails to authenticate

tomaskikutis:
SDESK-5794


jerome-poisson:
@tomaskikutis would it be possible to have an alert with a `X` close button which stays until user click on the `X`? This way we can be sure that user read message and has time to see the instructions.

jerome-poisson (approved):

PR #3981, opened 2021-09-06
FIX: The 'publish schedule' alert does not work for all publish options. [SDBELGA-574]

GyanP:


scripts/apps/authoring/authoring/directives/SendItem.tsx
```diff
@@ -119,6 +120,10 @@ export function SendItem($q,
             scope.metadata = metadata.values;
 
             scope.publish = function() {
+                confirmQuickPublish([scope.item]).then(publishButtonAction);
+            };
+
+            function publishButtonAction() {
```
tomaskikutis, line 126: I would inline the code instead of creating a separate function. `scope.publish` is already linked to a button click, so the naming of `publishButtonAction` is not very accurate.

Could you also rename `confirmQuickPublish` and to `confirmPublish`? Because it's no longer only for quick publish.

scripts/apps/authoring/authoring/directives/SendItem.tsx
```diff
@@ -120,10 +120,14 @@ export function SendItem($q,
             scope.metadata = metadata.values;
 
             scope.publish = function() {
-                confirmQuickPublish([scope.item]).then(publishButtonAction);
+                if (appConfig?.features?.confirmPublishModal) {
```
tomaskikutis, line 123: `confirmDueDate` would be a better name

scripts/apps/authoring/authoring/directives/SendItem.tsx
```diff
@@ -7,7 +7,7 @@ import {AuthoringWorkspaceService} from '../services/AuthoringWorkspaceService';
 import {appConfig, extensions} from 'appConfig';
 import {IExtensionActivationResult, IArticle} from 'superdesk-api';
 import {ITEM_STATE} from 'apps/archive/constants';
-import {confirmQuickPublish} from '../services/quick-publish-modal';
+import {confirmQuickPublish as confirmPublish} from '../services/quick-publish-modal';
```
tomaskikutis, line 10: Can you rename the original function instead of creating an alias?

GyanP: > Can you rename the original function instead of creating an alias?

Thanks, i will do changes as per your suggestion.

tomaskikutis (approved):

PR #3983, opened 2021-09-07
extract translation strings from extensions

tomaskikutis:
SDESK-6167


build-tools/src/extensions/extract-translations.js
```diff
@@ -0,0 +1,47 @@
...
+                        context: 3,
+                    },
+                }),
+            ]);
+
+        for (const _path of pathsAbsolute) {
+            jsParser.parseFilesGlob(`${_path}/**/*.@(ts|js|tsx|jsx)`);
+        }
+
+        extractor.savePotFile(`${extensionRootPath}/translations.generated.pot`);
```
petrjasek, line 41: so there would be a file generated for each extension?

tomaskikutis: Only for those that specify it in `package.json`. There's no need to specify it for those extensions that are in `superdesk-client-core/scripts/extensions`, because they get picked up by the main translation extractor.

end-to-end-testing-helpers/index.ts
```diff
@@ -100,13 +100,13 @@ export function hover(elem: ElementFinder) {
...
     fileInput: ElementFinder,
-    fileNames: Array<string>, // relative to spec/test-files folder
+    fileNames: Array<string>, // relative to e2e/client/specs/test-files folder
 ) {
     fileInput.sendKeys(
         fileNames
             .map(
                 (relativePath) =>
-                    path.resolve(__dirname, '../../../../spec/test-files/' + relativePath),
+                    path.resolve(__dirname, '../../../../../../e2e/client/specs/test-files/' + relativePath),
```
petrjasek, line 109: going 6 levels up looks a bit strange, shouldn't it get some base path as argument rather?

petrjasek:
does this affect superdesk-planning in any way?

end-to-end-testing-helpers/index.ts
tomaskikutis: I agree it is a bit strange. I could have used less levels, but I think it's more clear if I go all the way up to the root path. Since we now have a dedicated npm package for e2e tests exclusively, I think I could remove `end-to-end-testing-helpers` as a separate package and move it to e2e dir. The path would be simpler and we wouldn't need to bother with publishing to npm and versioning. What do you think?

tomaskikutis:
> does this affect superdesk-planning in any way?
It doesn't at the moment, because it's not specified in `package.json` in the planning project. If we wanted, we could remove planning specific translation string extraction code, add an option to `package.json` and it could then use the extraction from build tools.

end-to-end-testing-helpers/index.ts
petrjasek: yep moving that to e2e sounds good

tomaskikutis: In the end I didn't move it, because extensions will and do need the helpers and they are not supposed to import things outside their extension directory. I have changed the function so it takes an absolute and I don't need to go 6 level up (which would not have worked when used from extensions anyway).

petrjasek (approved):

PR #3984, opened 2021-09-08
FIX: My Search pane changes once I click on an item and stop the automatic refresh of the list. [SDBELGA-501]

GyanP:
fixed the issue of a double click on an item

### DIFF
--- a/scripts/apps/search/controllers/SearchMenuController.ts
+++ b/scripts/apps/search/controllers/SearchMenuController.ts
@@ -57,7 +57,8 @@ export default function SearchMenuController(
 
     const initActiveProvider = () => {
         if ($location.path() === '/search') {
-            if (!$location.search()._id && !$location.search().item) { // prevent from changing active provider during click on article
+            // prevent from changing active provider during click on article
+            if (!$location.search()._id && !$location.search().item) {
                 this.activeProvider = this.providers.find(
                     (provider) => isEqual($location.search(), getSearchParams(provider)),
                 );
--- a/scripts/apps/search/controllers/SearchMenuController.ts
+++ b/scripts/apps/search/controllers/SearchMenuController.ts
@@ -57,12 +57,11 @@ export default function SearchMenuController(
 
     const initActiveProvider = () => {
         if ($location.path() === '/search') {
-            this.activeProvider = this.providers.find(
-                (provider) => isEqual(
-                    _.omit($location.search(), ['_id', 'item', 'action']),
-                    getSearchParams(provider),
-                ),
-            );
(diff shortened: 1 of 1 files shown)


scripts/apps/search/controllers/SearchMenuController.ts
```diff
@@ -56,11 +57,12 @@ export default function SearchMenuController(
...
     const initActiveProvider = () => {
         if ($location.path() === '/search') {
-            if (!$location.search()._id) {
-                this.activeProvider = this.providers.find(
-                    (provider) => isEqual($location.search(), getSearchParams(provider)),
-                );
-            }
+            this.activeProvider = this.providers.find(
+                (provider) => isEqual(
+                    _.omit($location.search(), ['_id', 'item', 'action']),
```
tomaskikutis, line 62: It would be better to pick the values you need from query parameters instead of omitting.

GyanP: > It would be better to pick the values you need from query parameters instead of omitting.

The reason for using omitting here is When we do a single click on the article then **$location.search()** object added one more key that is **_id**. when we do double click on the article then **$location.search()** object added two more keys that are **item** and **action**. so that's why I used omit method to remove the respective key from the **$location.search()** object to compare with the **getSearchParams(provider)** object.

tomaskikutis: Wouldn't it work to pick keys conditionally as well? If you check that `_id` is present, you pick one set of keys, if not - other.

scripts/apps/search/controllers/SearchMenuController.ts
```diff
@@ -56,7 +57,7 @@ export default function SearchMenuController(
 
     const initActiveProvider = () => {
         if ($location.path() === '/search') {
-            if (!$location.search()._id) {
+            if (!$location.search()._id && !$location.search().item) {
```
tomaskikutis, line 60: Can you add a comment in the code describing what this line does and when is it relevant?

tomaskikutis (approved):

scripts/apps/search/controllers/SearchMenuController.ts
```diff
@@ -56,7 +57,7 @@ export default function SearchMenuController(
 
     const initActiveProvider = () => {
         if ($location.path() === '/search') {
-            if (!$location.search()._id) {
+            if (!$location.search()._id && !$location.search().item) { // prevent from changing active provider
```
tomaskikutis, line 60: It would help more if the comment answered questions "why?" and "under what circumstances?"

tomaskikutis (approved):

PR #3985, opened 2021-09-08
always fetch all templates when needed

petrjasek:
there is no pagination in the ui

SDNTB-709


tomaskikutis:
How is this fixed? Did you simply remove the limit and return everything from the server?

petrjasek:
yes exactly. not everything but the limit will be set higher on ntb

tomaskikutis (approved):

PR #3988, opened 2021-09-13
[SDNTB-688] Expose TextEditor3 through SuperdeskAPI

MarkLark86:


tomaskikutis:
It would be better to keep `generic-form` encapsulated and not move its [private interfaces](https://github.com/superdesk/superdesk-client-core/pull/3988/files#diff-ca49bae7f62187bd0909482a419226feafbf06df59d2d06c2031cd298e7d62aaR1064) to `superdesk-api` or [import internal modules](https://github.com/superdesk/superdesk-client-core/pull/3988/files#diff-113b1a3cf188c0e5e988e211946d1ae406a38198b44f7b15d7ff70d220e65a30R80) outside of `generic form`.

Could you instead expose to the API either `Editor3Standalone` or create a wrapper component and expose that?

tomaskikutis (approved):

PR #3992, opened 2021-09-16
FIX: To Desk feature fails when article was opened from search or monitoring widget. [SDESK-6161]

GyanP:


GyanP:
For the monitoring widget, the `scope.$on` is not listening for the event `$rootScope.$broadcast` so I  used this approach finally. Further suggestions are welcomed.

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -1032,13 +1032,26 @@ export function AuthoringDirective(
...
 
-                $rootScope.$broadcast('item:nextStage', {
-                    stage: stageList[stageIndex],
-                    itemId: $scope.item._id,
-                    selectedStage: selectedStage,
-                    selectedDesk: selectedDesk,
-                    item: $scope.item,
-                });
+                if ($location.$$path === '/workspace') {
+                    api.save('move', {}, {
```
MarkLark86, line 1036: We have a hook for to run before sending in extensions (https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L107)

Is this hook relevant here, and will it still be fired?

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -1032,13 +1032,26 @@ export function AuthoringDirective(
...
+                        task: {desk: selectedDesk._id, stage: stageList[stageIndex]._id},
```
MarkLark86, line 1037: I'm also concerned about duplication of code. Would it be better if we had this either in the SuperdeskApi, or in the internal api (https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/api/article.ts).

This way the actual sending of content (api calls) could be managed in the one place

MarkLark86 (commented):
@tomaskikutis Any ideas here?

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -1032,13 +1032,26 @@ export function AuthoringDirective(
...
                 }
 
-                $rootScope.$broadcast('item:nextStage', {
-                    stage: stageList[stageIndex],
-                    itemId: $scope.item._id,
-                    selectedStage: selectedStage,
-                    selectedDesk: selectedDesk,
-                    item: $scope.item,
-                });
+                if ($location.$$path === '/workspace') {
```
tomaskikutis, line 1035: It'd be good to avoid URL-specific behavior. I assume your intention is to check whether next stage is available, and you are using URL as a metric, but it'd be better to do a direct check, by checking `scope.currentDesk` or something similar.

It's also not clear to me why `'/workspace'` is used. Judging from the ticket, it looks like sending works from monitoring, and doesn't work from search, so I'd expect a special case to be added to handle sending from search.

As Mark pointed out, `beforeSend` hooks need to be fired, which it wouldn't be in this case. I also agree on having the sending logic in one place, but it would be quite hard to add it to internal API, because internal API is just a plain JS file with exported stateless functions that doesn't have access to any angular state. Instead I suggest implementing this in `scope.send` function in `scripts/apps/authoring/authoring/directives/SendItem.tsx`(after `beforeSend` is ran).

GyanP:
Task: To Desk not working from widget.
The **scope.$on** is not listening for the event **$rootScope.$broadcast** (Don’t know why this is happening)

Approach: Replace **$scope.$on** with **$rootScope.$on**.

Problem with this approach: API called more than one time so some error messages display on the UI.

Another approach: I added **a key with boolean value** in **$rootScope.$broadcast** and executing code conditionally under this defined boolean value and becomes false after the first execution but it fails when we apply **To Desk** on two articles one after one then it works for one article only.

@mark @tomas I hope now the task is clear to you guys, I would appreciate it if you can give me some suggestions for this problem.

tomas:
Please don't mention me again in this repository. Thanks

GyanP:
> Please don't mention me again in this repository. Thanks



> Please don't mention me again in this repository. Thanks

Sorry, my mistake. Thanks

tomaskikutis:
I understand now that the issue is not only in search, but in monitoring widget too. My suggestion remains the same - avoid url specific behavior and try putting the logic to `scope.send` function in `scripts/apps/authoring/authoring/directives/SendItem.tsx` so it is in one place, and `beforeSend` hook gets called.

If you have issues with events, try using native JavaScript events instead of angular events.

PR #3993, opened 2021-09-21
Remove imatrics tags when parents cant be found

tomaskikutis:
SDNTB-705

imatrics errors

### DIFF
--- a/scripts/core/ui/components/alert.scss
+++ b/scripts/core/ui/components/alert.scss
@@ -1,0 +1,17 @@
+.alert-extended {
+    display: flex!important;
+    width: 100%!important;
+    justify-content: space-between!important;
+    align-items: center!important;
+    padding-left: 1.2rem!important;
+    padding-right: 1.2rem!important;
+    margin-bottom: 0!important;
+}
+
+.alert-extended--title {
+    font-size: 1.1em!important;
+}
+
+.alert-extended--message {
+    font-size: 1em!important;
+}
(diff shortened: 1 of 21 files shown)


tomaskikutis:
I extracted inline styles to classes, but it doesn't seem any better to me. When inline classes are used, at least it's easy to refactor, because you don't need to search for other usages. I don't think anyone should be modifying CSS without reading component code anyway, so I don't see how it helps.

petrjasek:
It was more about things like font sizes, imo that should be defined on global level and not in each component, not sure how you would keep the UI consistent if each component defines it's own styles. in general this component should be part of ui framework imo, not the imatrics widget but the error message it uses for example

tomaskikutis:
> imo that should be defined on global level and not in each component, not sure how you would keep the UI consistent if each component defines it's own styles

Yep, I agree. I moved alert styles to core component that is exposed via extensions API.

scripts/core/ui/components/alert.scss
```diff
@@ -0,0 +1,17 @@
+.alert-extended {
+    display: flex!important;
+    width: 100%!important;
+    justify-content: space-between!important;
+    align-items: center!important;
+    padding-left: 1.2rem!important;
```
MarkLark86, line 6: These could be replaced with the `spacing` classes from the UI-Framework:
https://ui-framework.superdesk.org/#/components/spacing

In this case you would add the classes:
`sd-padding-x--1-5` and  `sd-margin-b--0`

This has the benefit that if the UI-Framework ever changes the `$sd-base-increment` variable, views like this would continue to be consistent with the rest of the UI-Framework

petrjasek (approved):

PR #4000, opened 2021-09-28
SDESK-6145 ADD: As an author full screen view of image is required.

dev-tsoni:


scripts/extensions/sams/src/components/assets/assetImagePreviewFullScreen.tsx
```diff
@@ -0,0 +1,209 @@
...
+        return (
+            <Modal
+                id="AssetImagePreviewFullscreenModal"
+                size="fullscreen"
+                closeModal={this.props.closeModal}
+            >
+                {/* Left section, containing image preview. */}
+
+                <section
+                    style={{
```
MarkLark86, line 52: We should avoid using inline `style` attributes, and try and use the UI-Framework as much as possible.

Could you use the classes used from monitoring's fullscreen media preview?
https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/archive/views/media-view.html

scripts/core/superdesk-api.d.ts
```diff
@@ -257,6 +257,12 @@ declare module 'superdesk-api' {
         order: number,
     }
 
+    export interface IDimension {
```
MarkLark86, line 260: This metadata is specific to SAMS, and `IAssetItem`, so it would be better off going into `scripts/extensions/sams/src/interfaces.ts`.

The interface also has the following schemas (as based on schema from SAMS https://github.com/superdesk/sams/blob/develop/src/clients/python/sams_client/schemas/assets.py#L48):
```typescript
export interface IAssetRenditionArgs {
    width: number;
    height: number;
    keep_proportions: boolean;
}

export interface IAssetRendition {
    _media_id: string;
    width: number; // actual width
    height: number; // actual height
    params: IAssetRenditionArgs; // arguments used when the rendition was generated
    versioncreated: string;
    filename: string;
    length: number;
}

export interface IAssetItem extends IVersionInformation {
    ....
    renditions: Array<IAssetRendition>;
}
```

PR #4005, opened 2021-10-04
Fix for newly added stages

tomaskikutis:
SDESK-6156


scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -259,7 +259,22 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
...
      **/
     function initGroups(settings) {
-        self.groups = settings.groups.length > 0 ? settings.groups : getDefaultGroups(settings);
+        self.groups =
+            (settings.groups.length > 0 ? settings.groups : getDefaultGroups(settings))
+                .filter((card) => {
+                    if (card.type === 'stage') { // filter out deleted stages
+                        var stage = self.stageLookup[card._id];
+
+                        if (stage != null) {
```
petrjasek, line 268: this if/else block looks like `return stage != null;`

petrjasek (approved):

PR #4010, opened 2021-10-11
[SDNTB-669] feat(iMatrics): Added iMatrics field in authoring header

pavlovicnemanja:
- [x] Create extension function
- [x] Display data in header from extension point
- [x] Apply styles in header list

### DIFF
--- a/scripts/extensions/auto-tagging-widget/src/tag-list.tsx
+++ b/scripts/extensions/auto-tagging-widget/src/tag-list.tsx
@@ -76,13 +76,17 @@ export function getTagsListComponent(superdesk: ISuperdesk): React.ComponentType
                 ));
 
                 return (
-                    !inline ? <div style={{paddingLeft: level === 0 || inline ? 0 : 14}}>
-                        { treeNodesMap }
-                    </div> : <React.Fragment>
-                        {level === 0 ? '' : ' | '}
-                        { treeNodesMap }
-                    </React.Fragment>
-                   
+                    !inline ? (
+                        <div style={{paddingLeft: level === 0 || inline ? 0 : 14}}>
+                            { treeNodesMap }
+                        </div>
+                    ) : (
+                        <React.Fragment>
+                            {level === 0 ? '' : ' | '}
+                            { treeNodesMap }
+                        </React.Fragment>
+                    )
+
                 );
             }
 


petrjasek:
we need this rather on the release/2.2.1 branch, can you change the base?

scripts/extensions/auto-tagging-widget/src/header-auto-tagging.tsx
```diff
@@ -0,0 +1,90 @@
...
+            const allGroupedAndSortedByConfig = allGrouped
+                .filter((_, key) => hasConfig(key, this.iMatricsFields.others))
+                .sortBy((_, key) => this.iMatricsFields.others[key].order,
+                    (a, b) => a - b);
+
+            const allGroupedAndSortedNotInConfig = allGrouped
+                .filter((_, key) => !hasConfig(key, this.iMatricsFields.others));
+
+            const allGroupedAndSorted = allGroupedAndSortedByConfig
+                .concat(allGroupedAndSortedNotInConfig);
```
tomaskikutis, line 83: Can you extract this code into a reusable piece so it's not duplicated in multiple places? I would also consider not putting JSX into a map, but grouping raw objects and outputting JSX only in the return block of the `render` method.

scripts/core/superdesk-api.d.ts
```diff
@@ -84,6 +91,7 @@ declare module 'superdesk-api' {
             articleGridItemWidgets?: Array<React.ComponentType<{article: IArticle}>>;
             authoringTopbarWidgets?: Array<React.ComponentType<{article: IArticle}>>;
             authoringSideWidgets?: Array<IAuthoringSideWidget>;
+            authoringHeaderComponents?: Array<AuthoringHeaderItem>;
```
tomaskikutis, line 94: Can you use `customFieldTypes` instead? We shouldn't add multiple API extension points when one already exists that does 90% of what's needed. The problem with `customFieldTypes` in this case is that they have to be configured via content profiles. We could add `always-on` option to the interface and the same extension point should work then.

```
export interface ICustomFieldType<IConfig> {
        id: string;
        label: string;
        editorComponent: React.ComponentType<IEditorComponentProps<IConfig>>;
        previewComponent: React.ComponentType<IPreviewComponentProps>;
        configComponent?: React.ComponentType<IConfigComponentProps<IConfig>>;

        /**
         * 'content-profiles'   - field will only appear in authoring when configured in content profile.
         * 'always-on'          - field will always appear in authoring.
         *                      configuration via content profiles will not be available.
         */
        configuration?: 'content-profiles' | 'always-on'; // defaults to 'content-profiles'
    }
```

pavlovicnemanja: but how to specify component that will be in header? Extend it? and add `headerComponent`?

tomaskikutis: You would need to provide `editorComponent` and inside the extension you would configure it to be in read-only mode. Components defined in `previewComponent` are only rendered in the [preview pane ](https://user-images.githubusercontent.com/2076953/138265289-f1080998-ce3a-4d7a-b49b-04e2c03700fb.png)

PR #4014, opened 2021-10-18
Add error boundaries to every item in the list and also to every field in list item [to develop]

tomaskikutis:
SDBELGA-590


petrjasek:
copy of https://github.com/superdesk/superdesk-client-core/pull/4014

PR #4015, opened 2021-10-19
[SDESK-6201] Fix sticky toolbar requiring a second render of Editor3Component`

MarkLark86:
Otherwise the `editorNode` prop in `ToolbarComponent.computeState` will be undefined on first render

### DIFF
--- a/scripts/core/editor3/components/toolbar/index.tsx
+++ b/scripts/core/editor3/components/toolbar/index.tsx
@@ -41,7 +41,6 @@ class ToolbarComponent extends React.Component<any, IState> {
     constructor(props) {
         super(props);
 
-        this.scrollContainer = $(props.scrollContainer || window);
         this.onScroll = this.onScroll.bind(this);
 
         this.computeState = this.computeState.bind(this);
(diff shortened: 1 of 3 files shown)


scripts/core/editor3/components/toolbar/index.tsx
```diff
@@ -126,6 +125,7 @@ class ToolbarComponent extends React.Component<any, IState> {
     }
 
     componentDidMount() {
+        this.scrollContainer = $(this.props.scrollContainer || window);
```
tomaskikutis, line 128: It's better not to set a default. Otherwise it would make it impossible to have it non-sticky, and may break the UI for existing components that do not have it set are exist in another scroll container.

tomaskikutis (commented):
Was there something that wasn't working because of that 2nd render or this is simply an improvement?

MarkLark86:
There is a bug here that required a second render for sticky toolbar to work.

This works in Authoring, as the field is rendered twice on initial display for some reason.

In the Planning form, it was only rendered once on initial display. Hence the sticky toolbar required a re-render for the sticky toolbar to work (reference to the DraftJS div element)

PR #4023, opened 2021-11-03
feat(iMatrics): Improved stylings on tags in iMatrics header

pavlovicnemanja:


scripts/extensions/auto-tagging-widget/src/tag-list.tsx
```diff
@@ -45,7 +45,8 @@ export function getTagsListComponent(superdesk: ISuperdesk): React.ComponentType
                         <Tag
                             key={item.qcode}
                             text={item.name}
-                            shade={savedTags.has(item.qcode) ?
+                            readOnly={readOnly}
```
tomaskikutis, line 48: Consider improving the component in UI framework that it's always read-only when `onClick` is undefined. You would have to update the usage here too, to pass `undefined` to `onClick` instead of `noop`.

tomaskikutis (approved):

PR #4025, opened 2021-11-08
Authoring react

tomaskikutis:


scripts/apps/authoring-react/manage-editor3-inside-authoring-react.tsx
```diff
@@ -0,0 +1,106 @@
...
+     */
+    const customFields: Array<ICustomFieldType<IValue, unknown, RawDraftContentState>> = [
+        {
+            id: 'editor3',
+            label: gettext('Editor3 (authoring-react)'),
+            editorComponent: Editor3Component,
+            // previewComponent: Editor3Component
+            previewComponent: () => null,
+
+            fromStorageValue: (rawContentState, config, onChange) => {
```
petrjasek, line 62: I'm not sure if this interface will be enough, the editor might need also associations

tomaskikutis: I've changed the API a bit since then. Here's the new one

```
storeValue?(fieldId: string, article: IArticle, value: IValue): IArticle;
retrieveStoredValue?(fieldId: string, article: IArticle): IValue;
```

If associations are needed, `retrieveStoredValue` could be implemented as follows:
```
retrieveStoredValue: (fieldId, article) => {
    const rawContentState = article.fields_meta?.[fieldId]?.['draftjsState'][0];

    const store = createEditorStore(
        {
            editorState: rawContentState ?? convertToRaw(ContentState.createFromText('')),
            // ...
        },
        ng.get('spellcheck'),
        true,
    );

    return {
        store,
        contentState: store.getState().editorState.getCurrentContent(),
        associations: getAssociations(article),
    };
},
```

It might only be a challenge if 2 or more fields wanted to use associations, then we might consider passing it directly, but until now current interface would work in my opinion.

PR #4027, opened 2021-11-11
display all desk in internal destinations dropdown

tomaskikutis:
SDCP-584

We had agreed earlier to show only 200 desks at a time, but since we have all desks loaded now, I'm updating the component to use the desk service now.

### DIFF
--- a/scripts/core/ui/components/generic-form/input-types/desk_single_value.tsx
+++ b/scripts/core/ui/components/generic-form/input-types/desk_single_value.tsx
@@ -1,30 +1,42 @@
-import {IDesk} from 'superdesk-api';
+import {IDesk, IRestApiResponse} from 'superdesk-api';
 import {getSelectSingleValueAutoComplete} from './select_single_value_autocomplete';
 import {gettext} from 'core/utils';
-import {dataApi} from 'core/helpers/CrudManager';
+import ng from 'core/services/ng';
 
 export const DeskSingleValue = getSelectSingleValueAutoComplete({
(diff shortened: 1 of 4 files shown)


scripts/core/ui/components/generic-form/input-types/desk_single_value.tsx
```diff
@@ -1,30 +1,42 @@
...
-    ),
-    queryById: (id) => dataApi.findOne<IDesk>('desks', id),
+    query: (searchString: string) => {
+        const desks = ng.get('desks').desks._items;
+        const searchStringLower = searchString.toLocaleLowerCase();
+
+        const matches = desks.filter((desk: IDesk) => desk.name.toLocaleLowerCase().includes(searchStringLower));
+        const size = matches.length;
+
+        const response: IRestApiResponse<IDesk> = {
```
petrjasek, line 14: I wonder if we should required in autocomplete to get the api response, imo that's kind of implementation detail of the query and for autocomplete is just needs a list of items

tomaskikutis: Since it's an async autocomplete it probably needs the amount of total items to handle pagination. I assume that's the reason I added it initially. The code of that autocomplete component is pretty messy and will be removed, so I'd rather not spend time changing the interface and updating all the usages now.

petrjasek: :ok:

petrjasek (approved):

PR #4028, opened 2021-11-16
As a user I want to have the Coming Up field active when I create a new story. [SDBELGA-466]

GyanP:
Remove Always choose current date and Choose a date radio button from Coming Up field in the template


scripts/core/superdesk-api.d.ts
```diff
@@ -2125,6 +2125,7 @@ declare module 'superdesk-api' {
         setValue: (value: IValue) => void;
         readOnly: boolean;
         config: IConfig;
+        template?: ITemplate;
```
tomaskikutis, line 2128: Is it used anywhere? It seems there's only a check if it's `null`, but actual value is never read.

tomaskikutis (commented):
The requirements in the ticket seem to me the same as was implemented in #3864. Is something not working as was implemented there? I've ran this PR and I no longer see an option when editing a template to use [a dynamic date](https://user-images.githubusercontent.com/2076953/142014610-9e5f3cdd-a32a-4880-bab1-15ba0539fadb.png). If I were to believe what's displayed in the UI, I would assume it would always initialize a new article with the same hardcoded date, which is probably not desired behavior?

petrjasek:
yep the requirement is the same to initialize that field based on current time always, there should be no option to select any specific date in the template, only toggle that field on/off

tomaskikutis:
@petrjasek  We had a discussion with Gideon yesterday and changed the requirements a bit. The updates are in the implementation section in the ticket. What you said remains true when time offset is configured. When there's no time offset, we'd show the date input and when an article is created using that template, it would have that exact date that was saved, not current + offset like it is when offset is set. The part with no offset is nice-to-have and not required for this ticket though

scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
```diff
@@ -0,0 +1,37 @@
...
+        extends React.PureComponent<IEditorComponentProps<string | null, IDateTimeFieldConfig> & IPropsAdditional> {
+        render() {
+            const checkbox = (
+                <Switch
+                    value={this.props.value != null}
+                    onChange={(value) => {
+                        if (value) {
+                            const initialConfig = this.props.config.initial_offset_minutes
+                            this.props.setValue(`{{ now|iso_datetime(${initialConfig}) }}`)
+                            notify.success(`Time offset is configured to be ${initialConfig} minutes`)
```
petrjasek: not sure we need this, but if we do we should translate it

GyanP: @tomaskikutis suggested that, it would be good to show the message to the user.

scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
```diff
@@ -21,7 +23,7 @@ export function getToggleDateTimeField(superdesk: ISuperdesk) {
                             const initialConfig = this.props.config.initial_offset_minutes;
 
                             this.props.setValue(`{{ now|iso_datetime(${initialConfig}) }}`);
-                            notify.success(`Time offset is configured to be ${initialConfig} minutes`);
+                            notify.success(gettext(`Time offset is configured to be ${initialConfig} minutes`));
```
petrjasek, line 25: in this case we shouldn't use the string literal as that would be looking for a different string for each value, rather it should be a param:
```
notify.success(gettext(`Time offset is configured to be {{x}} minutes`, {x: initialConfig}));
```

scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
tomaskikutis: Could you change this to display a message below the checkbox instead of a notification? And also to use the message from the ticket? It explains how it works in more detail.

petrjasek (approved):

tomaskikutis:
@GyanP why did you merge the PR without addressing my last few comments? Can you open another PR and do the changes there?

GyanP:
> @GyanP why did you merge the PR without addressing my last few comments? Can you open another PR and do the changes there?

@tomaskikutis I have done the changes in last commit.

tomaskikutis:
[Last changes](https://user-images.githubusercontent.com/2076953/142859872-f3cc0a8d-7ed5-463d-ab09-bd0a62f04fa2.png) I see are from 3 days ago. I [posted](https://github.com/superdesk/superdesk-client-core/pull/4028#discussion_r754123256) [comments](https://github.com/superdesk/superdesk-client-core/pull/4028#discussion_r754123256) a few hours ago :)

GyanP:
> [Last changes](https://user-images.githubusercontent.com/2076953/142859872-f3cc0a8d-7ed5-463d-ab09-bd0a62f04fa2.png) I see are from 3 days ago. I [posted](https://github.com/superdesk/superdesk-client-core/pull/4028#discussion_r754123256) [comments](https://github.com/superdesk/superdesk-client-core/pull/4028#discussion_r754107899) a few hours ago :)

Ah, My mistake, I totally missed it, really sorry, I'll create a new PR and do the changes over there.

PR #4030, opened 2021-11-24
As a user I want to have the Coming Up field active when I create a new story. [SDBELGA-466]

GyanP:
address this comment https://github.com/superdesk/superdesk-client-core/pull/4028#discussion_r754123256

### DIFF
--- a/scripts/apps/authoring/authoring/authoring-custom-field.tsx
+++ b/scripts/apps/authoring/authoring/authoring-custom-field.tsx
@@ -80,6 +80,7 @@ export class AuthoringCustomField extends React.PureComponent<IProps, IState> {
                             setValue={(value) => this.setValue(value)}
                             readOnly={!editable}
                             config={field.custom_field_config}
+                            template={this.props.template}
                         />
                     ) :
                     (
@@ -89,6 +90,7 @@ export class AuthoringCustomField extends React.PureComponent<IProps, IState> {
                             setValue={(value) => this.setValue(value)}
                             readOnly={!editable}
                             config={field.custom_field_config}
+                            template={this.props.template}
                         />
                     )
                 }
--- a/scripts/apps/authoring/authoring/authoring-custom-field.tsx
+++ b/scripts/apps/authoring/authoring/authoring-custom-field.tsx
@@ -80,7 +80,6 @@ export class AuthoringCustomField extends React.PureComponent<IProps, IState> {
                             setValue={(value) => this.setValue(value)}
                             readOnly={!editable}
                             config={field.custom_field_config}
-                            template={this.props.template}
                         />
                     ) :
                     (
@@ -90,7 +89,6 @@ export class AuthoringCustomField extends React.PureComponent<IProps, IState> {
                             setValue={(value) => this.setValue(value)}
                             readOnly={!editable}
                             config={field.custom_field_config}
-                            template={this.props.template}
                         />
                     )
                 }
--- a/scripts/apps/authoring/views/authoring-header.html
+++ b/scripts/apps/authoring/views/authoring-header.html
@@ -519,7 +519,7 @@
             </div>
 
             <!-- custom fields from extensions API -->
-            <div class="authoring-header__item"
+            <div class="authoring-header__item authoring-header__item-start"
                 ng-repeat="field in fields track by field._id"
                 ng-if="field.custom_field_type != null && editor[field._id].section === 'header'"
                 sd-width="{{ editor[field._id].sdWidth || 'full' }}"
@@ -527,7 +527,7 @@
                 order="{{ editor[field._id].order }}"
                 ng-attr-title="{{schema[field._id].readonly === true ? readOnlyLabel : ''}}"
                 sd-validation-error="error[field._id.toLowerCase()]">
-                <label class="authoring-header__item-label" ng-style="template.data.extra.DueBy !== null && {'margin-bottom':'35px'}">{{field.display_name}}</label>
+                <label class="authoring-header__item-label">{{field.display_name}}</label>
                 <div class="authoring-header__input-holder">
                     <div class="authoring-header__hint" ng-if="helper_text[field._id]">{{field.helper_text}}</div>
                     <sd-authoring-custom-field
--- a/scripts/apps/authoring/views/authoring-header.html
+++ b/scripts/apps/authoring/views/authoring-header.html
@@ -527,7 +527,7 @@
                 order="{{ editor[field._id].order }}"
                 ng-attr-title="{{schema[field._id].readonly === true ? readOnlyLabel : ''}}"
                 sd-validation-error="error[field._id.toLowerCase()]">
-                <label class="authoring-header__item-label">{{field.display_name}}</label>
+                <label class="authoring-header__item-label" ng-style="template.data.extra.DueBy !== null && {'margin-bottom':'35px'}">{{field.display_name}}</label>
                 <div class="authoring-header__input-holder">
                     <div class="authoring-header__hint" ng-if="helper_text[field._id]">{{field.helper_text}}</div>
                     <sd-authoring-custom-field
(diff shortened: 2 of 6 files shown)


scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
```diff
@@ -8,34 +8,45 @@ interface IPropsAdditional {
...
-                            ));
                         } else {
                             this.props.setValue(null);
                         }
                     }}
                 />
             );
+            const messageText = gettext(
+                `Time offset is configured to be {{minutes}} minutes for this field. When an article is created
+                based on this template, it's value will initialize to creation time {{offset}} minutes`,
```
petrjasek, line 31: I would say `creation time plus {{offset}} minutes`

scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
```diff
@@ -8,34 +8,45 @@ interface IPropsAdditional {
...
-                            notify.success(gettext('Time offset is configured to be {{minutes}} minutes',
-                                {minutes: initialConfig},
-                            ));
                         } else {
                             this.props.setValue(null);
                         }
                     }}
                 />
             );
+            const messageText = gettext(
```
tomaskikutis, line 29: It'd be better to use placeholders that don't repeat the words around it.

Instead of `offset is {{offset}}` or `configured to {{minutes}} minutes`, prefer

`offset is {{x}}`, `configured to {{y}} minutes and {{z}} seconds`

It will be clearer for translators this way.

I have noticed that it was not possible to use the same placeholder multiple times in the same translation and pushed a commit to your branch to fix this.

scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
```diff
@@ -8,34 +8,45 @@ interface IPropsAdditional {
...
             return (
-                <div>
-                    {checkbox}
-                </div>
+                <>
+                    <div>
+                        {checkbox}
+                    </div>
+                    <div>
+                        {this.props.template?.data?.extra?.DueBy &&
```
tomaskikutis, line 41: We can't hardcode field names, especially in an extension! Shame on @petrjasek for not pointing this out in his review :smile:

To solve this, we could pass `templateValue` property that would refer to a value only of that field, unlike `template` prop that passes the entire template.

scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
```diff
@@ -8,34 +8,45 @@ interface IPropsAdditional {
 }
 
 export function getToggleDateTimeField(superdesk: ISuperdesk) {
-    const {notify} = superdesk.ui;
     const {gettext} = superdesk.localization;
 
     return class ToggleDateTimeField
         extends React.PureComponent<IEditorComponentProps<string | null, IDateTimeFieldConfig> & IPropsAdditional> {
```
tomaskikutis, line 14: Type type should be `React.PureComponent<ITemplateEditorComponentProps<...`, not `React.PureComponent<IEditorComponentProps<...`. I see `ITemplateEditorComponentProps` got deleted, so it should be restored. Also remove `template` prop from `ITemplateEditorComponentProps` - editor should already receive a value with template applied if needed.

scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
```diff
@@ -8,34 +8,45 @@ interface IPropsAdditional {
...
-                    {checkbox}
-                </div>
+                <>
+                    <div>
+                        {checkbox}
+                    </div>
+                    <div>
+                        {this.props.template?.data?.extra?.DueBy &&
+                            (
+                                <span className="authoring-header__hint">
```
tomaskikutis, line 43: Avoid using class names from core. It's best if extensions use either their own CSS or react components from superdeskApi/ui-framework. You can add `index.css` in the `src` directory and use [getClass](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/extensions/markForUser/src/get-marked-for-me-component.tsx#L159) method from extensions api, to get the class name. That extra method is needed, because class names from extensions are namespaced to each extension to avoid naming conflicts.

scripts/apps/authoring/authoring/authoring-custom-field.tsx
```diff
@@ -80,6 +80,7 @@ export class AuthoringCustomField extends React.PureComponent<IProps, IState> {
                             setValue={(value) => this.setValue(value)}
                             readOnly={!editable}
                             config={field.custom_field_config}
+                            template={this.props.template}
```
tomaskikutis, line 83: can you remove template prop from here & `ITemplateEditorComponentProps` interface?

scripts/apps/authoring/views/authoring-header.html
```diff
@@ -527,7 +527,7 @@
                 order="{{ editor[field._id].order }}"
                 ng-attr-title="{{schema[field._id].readonly === true ? readOnlyLabel : ''}}"
                 sd-validation-error="error[field._id.toLowerCase()]">
-                <label class="authoring-header__item-label">{{field.display_name}}</label>
+                <label class="authoring-header__item-label" ng-style="template.data.extra.DueBy !== null && {'margin-bottom':'35px'}">{{field.display_name}}</label>
```
tomaskikutis, line 530: why is that extra margin needed? Could we add it without hardcoding the field name ID?

scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
GyanP: > 

We have removed getTemplateDateTimeField.tsx file. I think we can use `IEditorComponentProps` because `ITemplateEditorComponentProps` and `IEditorComponentProps` both are seems to be same interface so any specific reason to use I`TemplateEditorComponentProps` here?

scripts/apps/authoring/views/authoring-header.html
GyanP: > 

When Comin Up field switch is ON then due to message display alignment between Coming Up label and switch is mismatched that's why to match that sequence extra margin provided

scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
tomaskikutis: It seems to me that `getTemplateDateTimeField` was replaced with `getToggleTemplateDateTimeField.tsx`, but the usage remains the same - it's a component that is only used for managing the template and is not used in the editor, right?

> any specific reason to use ITemplateEditorComponentProps here?

To reduce code coupling. It's cleaner from code quality point of view, to contain features so they don't know about each other and can't access each other's code or data. Isolating functionality in one place, or at least in as little places as possible makes code more stable and easier to reason about. Depending on requirements it is sometimes not possible, but we should isolate them when possible as it is now.

scripts/apps/authoring/views/authoring-header.html
GyanP: > 
comingUp

tomaskikutis: This is with that extra CSS that fixes it? I was wondering how it looked before the fix

GyanP: > 

No, This scenario is without extra CSS

tomaskikutis: I see. There are 2 problems with your solution
* hardcoding of `dueDate` field - it means that it would break if date time field was configured with another ID
* margin is hardcoded in pixels. Even the same text might take more or less lines in another language and the design would break

You could avoid both issues by adding `align-items: start;` to the parent element(with class "authoring-header__item").

scripts/extensions/datetimeField/src/getToggleTemplateDateTimeField.tsx
tomaskikutis: @GyanP I see you forgot to address this one. Can you change the placeholder to `x`, `n` or something similar to avoid the same words("minutes") next to each other? (`configured to be {{minutes}} minutes`)

tomaskikutis (approved):

PR #4033, opened 2021-12-02
As a user I want to have the Coming Up field active when I create a new story in save as a template. [SDBELGA-466]

GyanP:


scripts/apps/templates/controllers/CreateTemplateController.ts
```diff
@@ -32,9 +34,23 @@ export function CreateTemplateController(
...
+    this.dateTimeFields = null;
 
     activate();
 
+    function itemData() {
+        const _item = JSON.parse(JSON.stringify(templates.pickItemData(item)));
+
+        self.dateTimeFields?.forEach((field) => {
+            if (_item.extra[field._id]) {
+                const initialOffset = field.custom_field_config.initial_offset_minutes;
```
tomaskikutis, line 46: `initial_offset_minutes` config is an implementation detail of an extension. To maintain proper encapsulation, we can only implement the extensions API from core code(what extensions expose via `IExtensionActivationResult` interface). We can not put code in the core that is aware of implementation details of any particular extension.

For example, if you want custom fields from extensions to be able to set initial value when creating a template, you could add [an extension point similar to this](https://user-images.githubusercontent.com/2076953/144410684-446454d1-0379-42a4-afb4-945dc635eba3.png), and do it inside the extension then.

tomaskikutis (commented):
Can you post functional changes in PR description? There's a ticket for this, but this is a 2nd PR for the same ticket if we don't count the one addressing comments from the first one. Were there issues with first implementation, did requirements change?

GyanP:
> Can you post functional changes in the PR description? There's a ticket for this, but this is a 2nd PR for the same ticket if we don't count the one addressing comments from the first one. Were there issues with the first implementation, did requirements change?

Actually, previous PR changes are working fine, there is one new requirement like when we create an article with a template containing the coming up field then choose any date in the future for it  and save the article as a template. after this when we create a new article with that new template then coming up field should behave like it is just turned on(means current datetime + offset time)

tomaskikutis:
If I understand correctly, the issue is that when you "save as template" from authoring, it will save the date that is selected via date picker, and not the dynamic date template that we want to get saved?

If that's the case, adding `onTemplateCreate(): any` option that I mentioned in another comment should work and allow us to keep the logic inside the extension.

GyanP:
> If I understand correctly, the issue is that when you "save as template" from authoring, it will save the date that is selected via date picker, and not the dynamic date template that we want to get saved?

yep, that was the case.

scripts/apps/templates/controllers/CreateTemplateController.ts
```diff
@@ -32,8 +35,25 @@ export function CreateTemplateController(
     this.types = templates.types;
     this.createTypes = _.filter(templates.types, (element) => element._id !== 'kill');
     this.save = save;
+    this.dateTimeFields = null;
 
     activate();
+    function itemData() {
+        const _item = JSON.parse(JSON.stringify(templates.pickItemData(item)));
+
+        self.dateTimeFields?.forEach((field) => {
```
tomaskikutis, line 44: `onTemplateCreate` must be triggered for all extensions not only date-times.

scripts/apps/templates/controllers/CreateTemplateController.ts
```diff
@@ -32,8 +35,25 @@ export function CreateTemplateController(
...
+            if (_item.extra[field._id]) {
+                Object.values(extensions).forEach(({activationResult}) => {
+                    if (activationResult.contributions?.customFieldTypes) {
+                        if (activationResult.contributions?.customFieldTypes[0]?.onTemplateCreate) {
```
tomaskikutis, line 48: It has to run `onTemplateCreate` for each extension, not only the first one. Not only run it, but get the return value and patch it. As I've mentioned in another comment, you can't mutate items from inside extensions.

Also, 2 arguments are passed to  `onTemplateCreate`, when in the interface(superdesk-api.d.ts) there are no arguments defined.

scripts/core/superdesk-api.d.ts
```diff
@@ -2160,6 +2160,8 @@ declare module 'superdesk-api' {
         previewComponent: React.ComponentType<IPreviewComponentProps>;
         configComponent?: React.ComponentType<IConfigComponentProps<IConfig>>;
         templateEditorComponent?: React.ComponentType<ITemplateEditorComponentProps<IConfig>>;
+
+        onTemplateCreate?(): any;
```
tomaskikutis, line 2164:
```TypeScript
/**
 * Takes field value as an argument and returns modified or same value.
 */
onTemplateCreate?(value: any, config: IConfig): any;
```

Can you use the following interface? It should function in a similar way as `entities.article.onPatchBefore`

scripts/extensions/datetimeField/src/extension.tsx
```diff
@@ -32,6 +32,13 @@ function getDateTimePreviewComponent(superdesk: ISuperdesk) {
     };
 }
 
+function onTemplateCreate(item: any, field: any) {
+    const initialOffset = field.custom_field_config.initial_offset_minutes;
+
+    item.extra[field._id] = `{{ now|add_timedelta(minutes=${initialOffset})|iso_datetime }}`;
```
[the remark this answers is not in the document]
GyanP: I see, what if I just return the value from `onTemplateCreate` function.
For example: `return {{ now|add_timedelta(minutes=${initialOffset})|iso_datetime }}`

scripts/apps/templates/controllers/CreateTemplateController.ts
GyanP: Did you mean that we should create the `onTemplateCreate?(field: any): any` interface inside the `IExtensionActivationResult` interface?

scripts/extensions/datetimeField/src/extension.tsx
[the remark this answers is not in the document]
tomaskikutis: Yep, I think that's the best solution - to only return new value of that field.

scripts/apps/templates/controllers/CreateTemplateController.ts
tomaskikutis: No. The problem is that you call `onTemplateCreate` inside `forEach` loop. If there are no date time fields, it will never be called, even though it should.

GyanP: As I can see our main goal is that `onTemplateCreate `should be called from all the extensions independently. so what if we define on`TemplateCreate?(field: any): any` in `IExtensionActivationResult` interface then we can use this function by defining in any extension. Is there any approach where we can reflect this function in every extension?

tomaskikutis: Our main goal is to keep implementation details of date-time field inside the extension. Extensions won't call `onTemplateCreate` directly, but will only provide a function, which will be called by core before creating the template.

Keeping `onTemplateCreate` in `ICustomFieldType` is the best for this use case in my opinion, because it's directly related to custom field. If one extension had 5 custom fields, it could provide 5 `onTemplateCreate` functions conveniently. If we needed a global `onTemplateCreate` that is independent from custom fields, we could as well add it to `IExtensionActivationResult['entities']`, but since we are only dealing with a custom field now, I don't think we need it yet.

scripts/core/superdesk-api.d.ts
```diff
@@ -2160,6 +2160,8 @@ declare module 'superdesk-api' {
         previewComponent: React.ComponentType<IPreviewComponentProps>;
         configComponent?: React.ComponentType<IConfigComponentProps<IConfig>>;
         templateEditorComponent?: React.ComponentType<ITemplateEditorComponentProps<IConfig>>;
+
+        onTemplateCreate?(field: IConfig): any;
```
tomaskikutis, line 2164: https://github.com/superdesk/superdesk-client-core/pull/4033#discussion_r763888174

Can you also pass existing field value to the function?

scripts/apps/templates/controllers/CreateTemplateController.ts
```diff
@@ -32,8 +35,27 @@ export function CreateTemplateController(
...
+    function itemData() {
+        const _item = JSON.parse(JSON.stringify(templates.pickItemData(item)));
+
+        self.dateTimeFields?.forEach((field) => {
+            if (_item.extra[field._id]) {
+                Object.values(extensions).forEach(({activationResult}) => {
+                    if (activationResult.contributions?.customFieldTypes) {
+                        activationResult.contributions?.customFieldTypes?.forEach((customField) => {
+                            if (customField?.onTemplateCreate) {
+                                _item.extra[field._id] = customField.onTemplateCreate(field);
```
tomaskikutis, line 50: Let's say I had 2 `onTemplateCreate` functions defined. One appends "1" to the value, and other appends "2" to existing value. In the result, it should be both "1" and "2" appended. Current implementation would only use the result of the last iteration.

scripts/apps/templates/controllers/CreateTemplateController.ts
tomaskikutis: To solve the issue I commented on in the first comment in this thread, you need to remove iteration on datetime fields(`self.dateTimeFields?.forEach`) and instead iterate extensions. For each extension you should get it's config and pass as an argument when calling `onTemplateCreate`

tomaskikutis (approved):

PR #4034, opened 2021-12-03
fix manual refresh v2

tomaskikutis:
SDESK-6314

Simplified logic. If the config is enabled, it will now always show refresh button if items were added, removed or updated in a way that caused order of items to change. Scroll position or user session will no longer affect the feature.

### DIFF
--- a/scripts/apps/search/services/SearchService.ts
+++ b/scripts/apps/search/services/SearchService.ts
@@ -91,6 +91,10 @@ export function generateTrackByIdentifier(
  * It is shown if items are added/removed or order has changed.
  */
 export function showRefresh(currentItems: Array<IArticle>, newItems: Array<IArticle>) {
+    if (newItems.length !== currentItems?.length) {
+        return true;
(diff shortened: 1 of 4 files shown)


scripts/apps/search/services/SearchService.ts
```diff
@@ -86,6 +86,23 @@ export function generateTrackByIdentifier(
     return getTrackByIdentifier(item._id, item.state !== ITEM_STATE.INGESTED ? item._current_version : null);
 }
 
+/**
+ * Determine if refresh button needs to be shown.
+ * It is shown if items are added/removed or order has changed.
+ */
+export function showRefresh(currentItems: Array<IArticle>, newItems: Array<IArticle>) {
+    for (let i = 0; i < newItems.length; i++) {
```
petrjasek, line 94: would be good to check lengths as well, otherwise if last item is removed it won't catch it

petrjasek (approved):

PR #4035, opened 2021-12-07
show a refresh icon independently from who is viewing the list

tomaskikutis:
SDESK-6314


tomaskikutis:
I don't think we need this after all. It's fine if the list is getting updated for the same user that triggered the change.

PR #4036, opened 2021-12-08
fix displaying of signal in preview/authoring

petrjasek:
SDESK-6198


GyanP (approved):

tomaskikutis:
it'd be good to check if it's also fixed in full preview aka "print preview" (scripts/apps/authoring/preview/fullPreview.tsx)

PR #4037, opened 2021-12-09

GyanP:


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -214,7 +214,7 @@ export function AuthoringDirective(
                 if (!$scope.item.task.desk) {
                     return false;
                 }
-                const desk = desks.getItemDesk($scope.item);
+                const desk = _.find(desks.userDesks, (item) => item._id === $scope.item.task.desk);
```
petrjasek, line 217: why is that method not working? can we fix it?

GyanP: > why is that method not working? can we fix it?

when we refresh the page and at the time of the call of **getItemDesk** method, `desks.deskLookup` has no value so this method returns **null**. All desks of the user are present in **userDesks** only that's why I get desk from there.

petrjasek: shouldn't we rather call the method when it's actually loaded? it just seems like a code duplication and would only work if the item is on a desk where user is a member (which might not be the case)

GyanP: > shouldn't we rather call the method when it's actually loaded? it just seems like a code duplication and would only work if the item is on a desk where user is a member (which might not be the case)

**getItemDesk** is defined in `DeskFactory.ts` which is injected in `AuthoringDirective.ts` as a service. when we open or create the article then this service is loaded with all data so **getItemDesk** works properly but in the case of the refreshing page during the call of **getItemDesk** it's not loaded with data so it returns null. I found this approach only is it ok or could you tell me any other approach for this? 

```
if (desk) {
	$scope.deskName = desk.name;
        $scope.deskType = desk.desk_type;
} else {
	desks.fetchDeskById($scope.origItem.task.desk).then((_desk) => {
		$scope.deskName = _desk.name;
		$scope.deskType = _desk.desk_type;
	});
}
```

petrjasek: I think we could do something like:
```
desks.initialize().then(() => {
    getDeskStage();
    getCurrentTemplate();
});
```

petrjasek (approved):

PR #4038, opened 2021-12-13
fix manual refresh v3 - update total item count even if auto-refresh is disabled

tomaskikutis:
SDESK-6314


tomaskikutis:
I misunderstood the requirements, it should remain as is.

PR #4040, opened 2021-12-17

GyanP:


petrjasek (approved):

petrjasek:
btw what issue does it fix?

GyanP:
> btw what issue does it fix?

Actually, when we save the item and open it then if we refresh the page, those publish buttons are reappearing but when we change some content automatically those are disappeared. so to fix it.

GyanP:
> btw what issue does it fix?

Quick button variables like `$scope.publishEnabled` and `$scope.publishAndContinueEnabled` etc. does not get updated value because `$scope.$watch` is called before calling `getDeskStage` and `getCurrentTemplate` functions.

PR #4041, opened 2021-12-20
allow scheme to be populated for place

petrjasek:
SDNTB-722


scripts/extensions/auto-tagging-widget/src/types.ts
```diff
@@ -7,5 +7,6 @@ export interface ITagUi {
     aliases?: Array<string>;
     altids: {[key: string]: string};
     parent?: string;
+    scheme?: string;
```
tomaskikutis, line 10: could you add a comment regarding what is it used for?

tomaskikutis (approved):

PR #4042, opened 2021-12-22

GyanP:


GyanP:
related to SDESK-6323, there is one option that is not addressed in the last PR that is **select the item and the publish button appears on top for authoring desk** so we have to remove it from there also.

tomaskikutis (approved):

PR #4044, opened 2021-12-27
FIX: Possibility of selecting no language brings strange behavior of language filter. [SDESK-6111]

GyanP:


tomaskikutis (approved):

petrjasek:
hi @GyanP seems like some tests are failing

tomaskikutis:
regarding CI, it fails on every PR now. My PR to update ui-framework should fix it. Regarding e2e, it seems tests can't even start to back-end issues

PR #4046, opened 2022-01-04
FIX: problems saving  image item's metadata after editing. [SDESK-6316]

GyanP:


scripts/core/upload/image-crop-directive.ts
```diff
@@ -318,6 +318,7 @@ export default angular.module('superdesk.core.upload.imagecrop', [
                         if (!_.isEqual(self.poi, poi)) {
                             angular.extend(self.poi, poi);
                             self.onChange();
+                            $rootScope.$broadcast('isDirtyChange', self.onChange());
```
tomaskikutis, line 321: can you use a more specific event name? like `point-of-interest--updated`

petrjasek (approved):

PR #4047, opened 2022-01-04
As a SD admin I want to update the metadata section by uploading (or referencing) a settings file. [SDESK-6331]

GyanP:


scripts/apps/vocabularies/views/vocabulary-config-modal.html
```diff
@@ -288,4 +288,72 @@ <h3 class="modal__heading" ng-show="!vocabulary._id"><span translate>Create</spa
             </button>
         </div>
     </div>
+
+    <div sd-modal data-model="vocabulary" class="modal--large" ng-if="vocabulary.field_type === 'upload' ">
```
petrjasek, line 292: would be good to have this part in a standalone component so we can reuse it in other places later (eg for content types)

scripts/apps/vocabularies/controllers/VocabularyConfigController.ts
```diff
@@ -60,6 +62,30 @@ export function VocabularyConfigController($scope: IScope, $route, $routeParams,
...
     };
 
+    /**
+     * Download vocabulary in json file.
+     *
+     * @param {Object} vocabulary
+     */
+    $scope.downloadVocabulary = (vocabulary: IVocabulary) => {
+        const _vocabulary = JSON.stringify(vocabulary);
+        const _vocabularyUri = 'data:application/json;charset=utf-8,' + encodeURIComponent(_vocabulary);
```
tomaskikutis, line 72: Could you extract file downloading logic to a separate function? Something like [`downloadBlob`](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/helpers/utils.tsx#L97)

scripts/apps/vocabularies/views/vocabulary-config-modal.html
tomaskikutis: I agree. And let's write new code in react instead of angular

scripts/core/utils.ts
```diff
@@ -213,3 +213,15 @@ export function isScrolledIntoViewVertically(element: HTMLElement, container: HT
 
     return topVisible && bottomVisible;
 }
+
+export function downloadFile(data: {}, fileName: string, extensionValue: string, htmlElementId: string) {
```
tomaskikutis, line 217: Can you drop `htmlElementId` argument, and create an element in this function instead? Also avoid using jQuery where it's easy to do so.

scripts/apps/vocabularies/components/drop-zone.tsx
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+import {IDropZoneComponentProps} from 'superdesk-api';
+import {gettext} from 'core/utils';
+
+interface IState {
+    hover: boolean;
+}
+
+interface IJQueryDragEvent extends JQueryEventObject {
```
tomaskikutis, line 9: Can you use native JS instead of jQuery in this file?

scripts/apps/vocabularies/components/UploadComplete.tsx
```diff
@@ -0,0 +1,11 @@
+import {gettext} from 'core/utils';
+import React from 'react';
+
+export const UploadComplete = () => {
+    return (
+        <div className="file-uploaded">
+            <span><i className="icon-ok icon--green" /></span>
+            <p>{gettext('Your upload was successfull')}</p>
```
tomaskikutis, line 8: There's a typo here, should be `successful`

scripts/apps/vocabularies/components/UploadConfigModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+                files: [],
+            };
+        }
+
+        onAddFiles(files) {
+            this.setState({files: Array.from(files || [])});
+        }
+
+        uploadFile() {
+            const notify = ng.get('notify');
```
tomaskikutis, line 41: Can you import and use `notify` constant from `scripts/core/notify/notify.tsx`? It avoids direct dependency on angular

scripts/apps/vocabularies/components/UploadConfigModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+                } else if (res._error) {
+                    notify.error(gettext(res._error._message));
+                }
+            }).catch((error: any) => {
+                notify.error(gettext(error._message));
+            });
+        }
+
+        render() {
+            const modalInformationLabel = `Please be aware that uploading external config files
```
tomaskikutis, line 62: Can you wrap it in `gettext` so it can be translated?

scripts/apps/vocabularies/components/UploadConfigModalInformation.tsx
```diff
@@ -0,0 +1,15 @@
...
+import {gettext} from 'core/utils';
+
+export const UploadconfigModalInformation = ({label}) => {
+    return (
+        <div className="sd-alert sd-alert--hollow sd-alert--primary">
+            <span>
+                <i className="big-icon--info big-icon--blue" style={{margin: '0 0.6rem 0 -0.3rem'}} />
+            </span>
+            <span>
+                {gettext(label)}
```
tomaskikutis, line 11: Only inline strings can be passed to `gettext`, not variables. Drop `gettext` here and make sure that the value that comes from props is already translated in the same place where it's defined.

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
```diff
@@ -46,6 +47,7 @@ interface IScope extends IScopeConfigController {
 }
 
 const idRegex = '^[a-zA-Z0-9-_]+$';
+const RESOURCE = 'upload/config-file';
```
tomaskikutis, line 50: This is not used anywhere

scripts/apps/vocabularies/views/vocabulary-config-modal.html
```diff
@@ -1,5 +1,5 @@
 <form name="editForm">
-    <div sd-modal data-model="vocabulary" class="modal--x-large">
+    <div sd-modal data-model="vocabulary" class="modal--x-large" ng-if="vocabulary.field_type !== 'upload' ">
```
tomaskikutis, line 2: Is this `ng-if` still needed after components were moved to react?

GyanP: Good catch, we don't need it.

scripts/apps/vocabularies/components/UploadConfigModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+    return class UploadConfigModal extends React.PureComponent<IProps, IState> {
+        constructor(props: IProps) {
+            super(props);
+
+            this.state = {
+                files: [],
+            };
+        }
+
+        onAddFiles(files) {
```
tomaskikutis, line 37: You forgot to bind `this` in the constructor

scripts/apps/vocabularies/components/UploadConfigModalInformation.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import {gettext} from 'core/utils';
+
+export const UploadconfigModalInformation = ({label}) => {
```
tomaskikutis, line 4: Capitalization is incorrect. Should be `UploadConfigModalInformation`

scripts/apps/vocabularies/components/UploadConfigModal.tsx
```diff
@@ -0,0 +1,115 @@
+import * as React from 'react';
+import {Button} from 'core/ui/components';
+import ng from 'core/services/ng';
```
tomaskikutis, line 3: `ng` is imported but not used.

scripts/apps/vocabularies/components/UploadConfigModalInformation.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import {gettext} from 'core/utils';
```
tomaskikutis, line 2: `gettext` is imported, but not used.

scripts/core/utils.ts
```diff
@@ -213,3 +213,17 @@ export function isScrolledIntoViewVertically(element: HTMLElement, container: HT
 
     return topVisible && bottomVisible;
 }
+
+export function downloadFile(data: IVocabulary, fileName: string) {
```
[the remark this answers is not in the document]
GyanP: I see you point now, will do it.

tomaskikutis (approved):

PR #4050, opened 2022-01-14
As an editor I want to deschedule various items at once

GyanP:
Related to SDBELGA-603, Minor change: A confirmation message should be shown indicating the number of items that were de-scheduled.

### DIFF
--- a/scripts/apps/search/controllers/get-multi-actions.tsx
+++ b/scripts/apps/search/controllers/get-multi-actions.tsx
@@ -352,7 +352,7 @@ export function getMultiActions(
                     }),
                 )).then(() => {
                 if (errors.length === 0) {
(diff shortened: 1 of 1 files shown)


scripts/apps/search/controllers/get-multi-actions.tsx
```diff
@@ -349,7 +352,7 @@ export function getMultiActions(
                     }),
                 )).then(() => {
                 if (errors.length === 0) {
-                    notify.success(gettext('All Articles have been descheduled'));
+                    notify.success(gettext(`${success.length} Articles have been descheduled`));
```
petrjasek, line 355: not sure if gettext works like that, imo it should be like
```
gettext('{{count}} articles descheduled', {count: success.length})
```

petrjasek (approved):

PR #4052, opened 2022-01-18
pressing ctrl+z after 2 specific NTB macros worked erases the Abstract field [SDESK-6354]

GyanP:


scripts/core/editor3/components/toolbar/InlineStyleButtons.tsx
```diff
@@ -42,14 +42,20 @@ export class InlineStyleButtonsComponent extends React.Component<any, any> {
 
     render() {
         const {editorFormat, editorState} = this.props;
-        const currentStyle = editorState.getCurrentInlineStyle();
+        const currentStyle = () => {
+            try {
```
tomaskikutis, line 46: why do you use `try` here? Does `getCurrentInlineStyle` throw an error?

tomaskikutis:
Could you explain what was the issue and the fix?

scripts/core/editor3/components/toolbar/InlineStyleButtons.tsx
aalokrmb: Yes, it throws an error: 

```
Uncaught TypeError: Cannot read property 'getLength' of undefined
    at getInlineStyleForNonCollapsedSelection (EditorState.js:538)
    at EditorState.getCurrentInlineStyle (EditorState.js:211)
    at InlineStyleButtonsComponent.render (InlineStyleButtons.tsx:78)
    at finishClassComponent (react-dom.development.js:17039)
    at updateClassComponent (react-dom.development.js:16994)
    at beginWork$1 (react-dom.development.js:18505)
    at HTMLUnknownElement.callCallback (react-dom.development.js:347)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:397)
    at invokeGuardedCallback (react-dom.development.js:454)
    at beginWork$$1 (react-dom.development.js:23217)
    at performUnitOfWork (react-dom.development.js:22208)
    at workLoopSync (react-dom.development.js:22185)
    at renderRoot (react-dom.development.js:21878)
    at runRootCallback (react-dom.development.js:21554)
    at eval (react-dom.development.js:11353)
    at unstable_runWithPriority (scheduler.development.js:643)
```

so that's why I put a try catch block.

GyanP: Yes, it throws an error: 
```
Uncaught TypeError: Cannot read property 'getLength' of undefined
    at getInlineStyleForNonCollapsedSelection (EditorState.js:538)
    at EditorState.getCurrentInlineStyle (EditorState.js:211)
    at InlineStyleButtonsComponent.render (InlineStyleButtons.tsx:78)
    at finishClassComponent (react-dom.development.js:17039)
    at updateClassComponent (react-dom.development.js:16994)
    at beginWork$1 (react-dom.development.js:18505)
    at HTMLUnknownElement.callCallback (react-dom.development.js:347)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:397)
    at invokeGuardedCallback (react-dom.development.js:454)
    at beginWork$$1 (react-dom.development.js:23217)
    at performUnitOfWork (react-dom.development.js:22208)
    at workLoopSync (react-dom.development.js:22185)
    at renderRoot (react-dom.development.js:21878)
    at runRootCallback (react-dom.development.js:21554)
    at eval (react-dom.development.js:11353)
    at unstable_runWithPriority (scheduler.development.js:643)
```
so that's why I added a try and catch block.

tomaskikutis: Did you find out the root cause of it? It would seem to me, an empty set would be returned if there are no inline styles.

GyanP: > an empty set

I thought the same but it seems like the problem is in DraftJS library, so I think It would be better, if we just put try and catch.

tomaskikutis: An exception might also mean issues with editorState, so it'd be good to understand the reason for the exception since the same issue might also affect other parts of the code. Is it easy to reproduce? I could look into it as well.

GyanP:
> Could you explain what was the issue and the fix?

The issue is only when we have Link button active in our abstract field, As soon as we add the link button, our Link toolbar calls and it calls the [getSelectedEntityData](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/editor3/components/links/LinkToolbar.tsx#L34).

Now this `getSelectedEntityData` calls `getSelectedEntityKey` for the key in which we are calling `getSelectedBlock` which is basically a DraftJS method but it returns null, I am not sure why but it keeps returning null, and that's why we get the errors.

tomaskikutis:
I have taken a look at it. The issue is that when `getSelectedBlock` is called, it receives a corrupted `editorState` that is passed as an argument. You'll see that when I check current selection vs content blocks, selection is referencing blocks that do not exist. I suspect the code that applies macro might be at fault.

To resolve the issue, ensure that editorState is not corrupted  - that is blocks that are referenced by the selection must exist. When this is fixed, I expect there will be no exceptions thrown and code that you added here to handle the exceptions should be removed.

https://user-images.githubusercontent.com/2076953/151355506-230a4bc9-5a33-48e4-8f31-4c0be11aa5b4.mp4

tomaskikutis:
closing in favor of https://github.com/superdesk/superdesk-client-core/pull/4061

PR #4053, opened 2022-01-21
[SDESK-6365] improve: Patch user resource when updating last active

MarkLark86:
@petrjasek Was playing around today with this, seems to work fine on my local. Not sure if this can be easily backported to 2.3 but might be worth looking at

### DIFF
--- a/scripts/core/data/index.ts
+++ b/scripts/core/data/index.ts
@@ -34,9 +34,22 @@ const resourcesInStore = [
 addWebsocketEventListener(
     'resource:updated',
     (event: IWebsocketMessage<IResourceUpdateEvent>) => {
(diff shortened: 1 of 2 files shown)


petrjasek (approved):
nice, I was thinking about pushing the timestamp from server but it should be pretty close to current time

MarkLark86:
> nice, I was thinking about pushing the timestamp from server but it should be pretty close to current time

I was thinking that too, but that would add more websocket notifications, and as you said it will be close enough. Even if it's a massive 5 second difference, the use case in the front end wouldn't make any difference.

tomaskikutis (approved):
If there were more frequently updated fields, a better solution would be to store a subset of user interface that would only have most frequently used fields.

PR #4059, opened 2022-02-01
Predefined text fields

tomaskikutis:
SDFID-632

predefined-field--settings

predefined-field--authoring


petrjasek:
would it be possible to use it with existing custom text field? I think they already use that field just want to add templates to it

scripts/core/superdesk-api.d.ts
```diff
@@ -947,29 +954,37 @@ declare module 'superdesk-api' {
         direction: 'ascending' | 'descending';
     }
 
+    export interface ICrudManagerResponse<T> {
```
petrjasek, line 957: not sure about the naming, what does crud manager response mean

petrjasek (commented):
I think it would be better to make it possible to add predefined values to existing custom text fields

scripts/core/superdesk-api.d.ts
tomaskikutis: `CRUD` stands for "create, read, update, delete" https://en.wikipedia.org/wiki/Create,_read,_update_and_delete

In this case, crud manager is an interface for storage of lists. I initially created it for generic list view(used in annotations library, internal destinations page), which stores data via HTTP API calls. This time, I wanted to reuse the same UI from generic list view, but the data I wanted to manage wasn't backed by HTTP API collections, but was a simple array. To make it work, I created an implementation of crud manager interface that works with an array.

tomaskikutis:
> I think it would be better to make it possible to add predefined values to existing custom text fields

This is a bit tricky, because this predefined field is implemented as a react component against extensions API, and current custom text fields are angular based, including management UI.

What about removing current custom text field, renaming `predefined text field` to `custom text field` and writing an upgrade script to update a vocabulary with a new config format?

scripts/core/superdesk-api.d.ts
petrjasek: I know the crud, I was rather concerned about the manager response because that crud manager is not responding to anything, is it?

petrjasek:
> > I think it would be better to make it possible to add predefined values to existing custom text fields
> 
> This is a bit tricky, because this predefined field is implemented as a react component against extensions API, and current custom text fields are angular based, including management UI.
> 
> What about removing current custom text field, renaming `predefined text field` to `custom text field` and writing an upgrade script to update a vocabulary with a new config format?

lets maybe first test this for the current usecase and we will see how it works there, maybe it would be premature optimization

scripts/core/superdesk-api.d.ts
tomaskikutis: Correct, crud manager itself is not responding to anything. Its methods do - `read`, `sort`, `refresh`, `removeFilter`, `goToPage`. The interface defines common data format that these methods respond with. I don't mind the name, feel free to suggest another which is clearer to you :)

tomaskikutis: `ICrudManagerState` has more fields than `ICrudManagerResponse` which wouldn't be available in crud manager methods. I think the interfaces themselves are fine and I doubt that it would work to simplify them easily. What we could do is rename if it would improve clarity.

petrjasek: ok, would probably call it ICrudManagerData then just to avoid calling it response when it's not

petrjasek (approved):

PR #4060, opened 2022-02-03
As an SD user manager, I want to indicate whether an author wants to be publicly visible or hidden [SDBELGA-605]

GyanP:


scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -1217,7 +1217,8 @@ export function MetadataService(api, subscribersService, vocabularies, $rootScop
                             role: role.qcode,
                             name: role.name,
                             parent: user._id,
-                            sub_label: user.display_name});
+                            sub_label: user.display_name,
+                            public: user.public});
```
tomaskikutis, line 1221: Can you add the property and comment its purpose in the interface? https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L172

tomaskikutis (approved):

PR #4062, opened 2022-02-08
fix translations popup

tomaskikutis:
SDBELGA-615


scripts/core/ui/components/popup.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    render() {
+        return (
+            <div
+                ref={(el) => {
+                    this.wrapperEl = el;
+
+                    if (el != null) {
+                        if (el != null) {
+                            this.positionOnce(el);
```
petrjasek, line 81: not sure why doing 2x `if (el != null)` and actually why not call this from componentDidMount?

tomaskikutis: Just double checking to be sure :D It's interesting how I missed that. Regarding `componentDidMount`, I think `ref` function runs after `componentDidMount`, so I wouldn't have the element inside `componentDidMount`. I'm not 100% sure, so I'll double check that too ;D

petrjasek (approved):

GyanP (approved):

PR #4067, opened 2022-02-17
As an SD user manager, I want to indicate whether an author wants to be publicly visible or hidden [SDBELGA-605]

GyanP:


GyanP:
Related to #SDBELGA-605,
when we activate checkbox `private` and save the user. after that when we are creating a new article and adding that user as an author, the `private` key flag is not displaying so to resolve this issue.

PR #4069, opened 2022-02-18
merge hotfix/2.3.4 into release/2.4

petrjasek:


petrjasek:
I'm making sure we're not missing in 2.4 any fix we did for 2.3.x, so pls @tomaskikutis @GyanP check if all your changes here are actually supposed to be in the latest version. thx

PR #4073, opened 2022-03-03
Do not limit header height

tomaskikutis:
SDESK-6364

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16542,6 +16542,11 @@
         "minimalistic-assert": "^1.0.0"
       }
     },
+    "web-animations-js": {
+      "version": "2.3.2",
+      "resolved": "https://registry.npmjs.org/web-animations-js/-/web-animations-js-2.3.2.tgz",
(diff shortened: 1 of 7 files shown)


petrjasek (approved):

fritzSF (commented):
Whitout limiting the header hight there will be no transition. I'm against killing general UX features because of requests of one customer or some edge cases. IMO this is generally not the right way to go.

fritzSF (changes requested):
Limiting the height to a large value (e.g. 100vh, 2000px or similar) might be sufficient to solve the problem and still preserve some transitioning while opening the header.

tomaskikutis:
https://user-images.githubusercontent.com/2076953/156609316-fb94322c-dabd-4cef-bb12-8a5ca5faaab2.mp4

fritzSF (approved):

fritzSF:
Niice! Looks much better than before.

pavlovicnemanja (approved):

PR #4078, opened 2022-03-11
As a Belga editor I like to see more info in the content pane of the Belga archive preview [SDBELGA-627]

GyanP:


scripts/apps/search/components/preview-subject.tsx
```diff
@@ -23,7 +24,7 @@ export class PreviewSubject extends React.PureComponent<IProps> {
                             key={subj.scheme + ':' + subj.qcode}
                             className="tag-label"
                             title={subj.name} // longer names might not fit the area
-                        >{subj.name}</span>,
+                        >{getVocabularyItemNameTranslated(subj, this.props.item.language)}</span>,
```
petrjasek, line 27: this should be probably also used for the `title` param

PR #4080, opened 2022-03-17
Improve the Belga archive integration [SDBELGA-606]

GyanP:


petrjasek (dismissed):

petrjasek:
what's the issue you're trying to solve here?

GyanP:
> what's the issue you're trying to solve here?

Those changes are related to this PR https://github.com/superdesk/superdesk-belga/pull/384
Previously, we have a single selection for types in search params for Belga Archive items. According to the requirement, inside the search params, we should be able to fetch the Belga Archive item using multiple types like "Text", "Alert" both so to achieve it.
Screenshot from 2022-03-14 13-54-53

petrjasek:
but why can't we handle that in the client repo?

petrjasek:
imo we shouldn't have a code here for a specific field of a client

GyanP:
> imo we shouldn't have a code here for a specific field of a client

I see, you mean we should handle it on belga side and not here?

scripts/apps/search/directives/SearchResults.ts
```diff
@@ -424,8 +424,21 @@ export function SearchResults(
...
+                        var item_id = item.extra ? item.extra.itemid : false;
+
+                        if (item_id && item.guid) {
+                            criteria.params = {
+                                'detailed_info': {
+                                    'newsItemId': item.extra.itemid,
+                                    'guid': item.guid,
+                                }};
+                            return api.query(getProvider(criteria), criteria).then((item_) => {
+                                processPreview(item_._items[0]);
```
petrjasek, line 436: I think we should check there is actually an item returned (and probably fallback to `processPreview(item);` call, the api might not work

scripts/apps/search/directives/SearchResults.ts
```diff
@@ -424,8 +424,21 @@ export function SearchResults(
...
                 if (item) {
                     if (item._type === 'externalsource') {
-                        processPreview(item);
-                        return;
+                        var item_id = item.extra ? item.extra.itemid : false;
+
+                        if (item_id && item.guid) {
+                            criteria.params = {
+                                'detailed_info': {
+                                    'newsItemId': item.extra.itemid,
```
petrjasek, line 432: why do we need this? is guid not enough? this is again client specific (actually belga archive specific)

scripts/apps/search/directives/SearchResults.ts
GyanP: Ok, so can we do something like that?
```
return api.query(getProvider(criteria), criteria).then((item_) => {
           if (item_){
                item = item_._items[0] 
           }
           processPreview(item);    
});
```

scripts/apps/search/directives/SearchResults.ts
GyanP: IMO we need newsItemId to get a detailed response, so only guid is not enough.  However, we can rename key **newsItemId** to **item_id** like - 
```
'detailed_info': {
       'item_id': item.extra.itemid,
       'guid': item.guid,
}
```

scripts/apps/search/directives/SearchResults.ts
petrjasek: `_items[0]` could still be undefined, and there could be an error from the external server

scripts/apps/search/directives/SearchResults.ts
petrjasek: what is the difference between those 2 ids?

GyanP: guid is `newsObjectId`, which is used to call the normal API which fetches only the main information and
item_id is `newsItemId` which is used to call the detailed API which fetches proper details like the item contains related articles or not, proper authors' details, etc for Belga Archive.

petrjasek: I see, could we use one to fetch the other? or maybe if we send that one as `preview_id` or something and based on that we would fetch the preview or not, otherwise this still looks belga specific in the client code

petrjasek (approved):

PR #4082, opened 2022-03-22
Fix Not possible, from the Monitoring, to save any Template with the Coming Up closed [SDBELGA-637]

GyanP:


scripts/apps/templates/controllers/CreateTemplateController.ts
```diff
@@ -33,19 +33,20 @@ function applyMiddleware(_item: IArticle, content, vocabularies): Promise<IArtic
...
-                        ...itemNext,
-                        extra: {
-                            ...itemNext.extra,
-                            [fieldId]: customField.onTemplateCreate(
-                                itemNext?.extra?.[fieldId],
-                                config,
-                            ),
-                        },
-                    };
+                    if (!itemNext?.extra?.[fieldId] === null) {
```
tomaskikutis, line 36: What was the issue and solution here? I would rather always call `onTemplateCreate` to give the extension the opportunity to set a default.

GyanP: suppose we have a template with the Coming up field active then if we create an article based on the template, switch the Coming Upfield off and use the 3 dots menu, do "save as template" and update the existing template. 
So the expected behavior is that the Coming Upfield should be disabled in the template. currently, if the coming upfield is disabled it always comes as enabled so to fix this issue.

GyanP: - suppose we have a template with the Coming up field active then if we create an article based on the template, switch the Coming Upfield off and use the 3 dots menu, do "save as template" and update the existing template. 
So the expected behavior is that the Coming Upfield should be disabled in the template. currently, if the coming upfield is disabled it always comes as enabled so to fix this issue. 
- and also current behavior is just the opposite of the ticket **Coming Upfield disappears from template after the template is updated by Save as template from authoring [SDBELGA-628]**

tomaskikutis: I see. Could you instead modify date time extension to return `null`, but still call `onTemplateCreate` here?

GyanP: yep I will do that.

scripts/apps/templates/controllers/CreateTemplateController.ts
```diff
@@ -10,7 +10,7 @@ import {IArticle, ICustomFieldType, IVocabulary} from 'superdesk-api';
  */
 function applyMiddleware(_item: IArticle, content, vocabularies): Promise<IArticle> {
     // Custom field types with `onTemplateCreate` defined. From all extensions.
-    const fieldTypes: {[id: string]: ICustomFieldType<any, any>} = {};
+    const fieldTypes: { [id: string]: ICustomFieldType<any, any> } = {};
```
tomaskikutis, line 13: seems like there are no changes left in this file, can you discard whitespace changes?

tomaskikutis: I meant the entire file. There are still whitespace changes

tomaskikutis (approved):

PR #4086, opened 2022-04-06
Fix Content profile dialog: After checking a few widgets, user is able to click on Save button multiple times which invokes error [SDESK-6408]

GyanP:


scripts/apps/workspace/content/controllers/ContentProfilesController.ts
```diff
@@ -150,6 +150,7 @@ export function ContentProfilesController($scope, $location, notify, content, mo
         var e = $scope.editing;
         var diff = {};
 
+        $scope.ngForm.$dirty = false;
```
tomaskikutis, line 153: Can you use a separate property like `savingInProgress`?

scripts/apps/workspace/content/views/profile-settings.html
```diff
@@ -146,7 +146,7 @@ <h3 class="modal__heading">{{ :: 'Editing' | translate }} "{{ editing.form.label
         </form>
     </div>
     <div class="modal__footer">
-        <button id="profile-update" class="btn btn--primary" ng-disabled="editForm.$invalid || !editForm.$dirty" ng-click="ctrl.update()" translate>Save</button>
+        <button id="profile-update" class="btn btn--primary" ng-disabled="editForm.$invalid || !editForm.$dirty || !ctrl.savingInProgress" ng-click="ctrl.update()" translate>Save</button>
```
tomaskikutis, line 149: > `|| !ctrl.savingInProgress`

Should exclamation mark be here? It should be disabled when saving **IS** in progress.

GyanP: my bad, yes you are right I have added it mistakenly.

scripts/apps/workspace/content/controllers/ContentProfilesController.ts
```diff
@@ -158,7 +159,8 @@ export function ContentProfilesController($scope, $location, notify, content, mo
 
         content.updateProfile(e.original, diff)
             .then(refreshList.bind(this, false), reportError)
-            .then(this.toggleEdit.bind(this, null));
+            .then(this.toggleEdit.bind(this, null))
+            .then(this.savingInProgress = false);
```
tomaskikutis, line 163: A function should be passed to `.then`, not expression. It would be `() => {this.savingInProgress = false}` in this case.

tomaskikutis (approved):

PR #4091, opened 2022-05-03
Best adapted and same content view for all the text [SDBELGA-471]

GyanP:


scripts/apps/archive/views/preview.html
```diff
@@ -46,6 +46,12 @@
               <span ng-if="item._type === 'archived'" class="stage" tooltip="{{:: 'Archived from' | translate}} {{deskName}} / {{ stage }}" tooltip-placement="bottom" ng-show="stage"><b>{{:: 'Archived from' | translate}} {{deskName}}</b> / {{ stage }}</span>
               <span ng-if="item._type !== 'archived'" class="stage" tooltip="{{deskName}} / {{ stage }}" tooltip-placement="bottom" ng-show="stage"><b>{{deskName}}</b> / {{ stage }}</span>
             </div>
+            <div class="preview-header__flex-row preview-header__flex-row--single-line" ng-if="subjectPreviewFields.length">
+              <div class="tag-label" ng-repeat="author in selected.preview.authors">{{author.role}} : {{author.sub_label}}</div>
+            </div>
+            <div class="preview-header__flex-row preview-header__flex-row--single-line" ng-if="checkDuplicateKeywords">
+              <div class="tag-label" ng-repeat="keyword in checkDuplicateKeywords">{{keyword}}</div>
```
petrjasek, line 53: does it call the function automatically?

scripts/apps/archive/directives/MediaPreview.ts
```diff
@@ -115,6 +115,17 @@ export function MediaPreview(api, $rootScope, desks, superdesk, content, storage
             scope.$on('$destroy', () => {
                 elem.off('dragstart');
             });
+
+            scope.checkDuplicateKeywords = function() {
```
petrjasek, line 119: not sure about the naming and the usage later being it a function, what about something like:
```ts
scope.uniqueKeywords = scope.selected.preview.keywords.filter((keyword) => ...);
```

scripts/apps/archive/views/preview.html
GyanP: Yes it is

GyanP:
Because in some cases, some keywords are similar to subjects, so IMO we only need to preview those keywords that are not present in subjects. or if we preview all keywords, we will see duplicate values in the preview like this : 
SDBELGA-471

petrjasek:
maybe we would rather need some labels there so it's more obvious what is what

scripts/apps/archive/views/preview.html
```diff
@@ -46,6 +46,14 @@
               <span ng-if="item._type === 'archived'" class="stage" tooltip="{{:: 'Archived from' | translate}} {{deskName}} / {{ stage }}" tooltip-placement="bottom" ng-show="stage"><b>{{:: 'Archived from' | translate}} {{deskName}}</b> / {{ stage }}</span>
               <span ng-if="item._type !== 'archived'" class="stage" tooltip="{{deskName}} / {{ stage }}" tooltip-placement="bottom" ng-show="stage"><b>{{deskName}}</b> / {{ stage }}</span>
             </div>
+            <div class="preview-header__flex-row preview-header__flex-row--single-line" ng-if="selected.preview.authors">
+              <span class="inline-label">Authors</span><br />
```
petrjasek, line 50: we would need gettext for those labels

scripts/apps/search/components/preview-subject.tsx
```diff
@@ -16,18 +16,30 @@ export class PreviewSubject extends React.PureComponent<IProps> {
...
+                    subjects.push(<div>
                         <span
-                            key={subj.scheme + ':' + subj.qcode}
-                            className="tag-label"
-                            title={getVocabularyItemNameTranslated(subj, this.props.item.language)}
-                            // longer names might not fit the area
-                        >{getVocabularyItemNameTranslated(subj, this.props.item.language)}</span>,
+                            key={subject?.scheme + ':' + subject?.qcode}
+                            className="inline-label"
+                        >{subject?.scheme}</span><br /></div>,
```
petrjasek, line 31: we would need to get the actual cv name here

scripts/apps/archive/views/preview.html
```diff
@@ -46,13 +46,13 @@
...
+            <div class="tag-main" ng-if="selected.preview.authors">
+              <span class="inline-label">{{preview_headings.author}}:</span>
+              <span class="tag-label" ng-repeat="author in selected.preview.authors">{{author.role}}:{{author.sub_label}}</span>
             </div>
-            <div class="preview-header__flex-row preview-header__flex-row--single-line" ng-if="selected.preview.keywords">
-              <span class="inline-label">Keywords</span><br />
-              <div class="tag-label" ng-repeat="keyword in selected.preview.keywords" title="Keywords">{{keyword}}</div>
+            <div class="tag-main" ng-if="selected.preview.keywords">
+              <span class="inline-label"><br/>{{preview_headings.keywords}}:</span>
+              <span class="tag-label" ng-repeat="keyword in selected.preview.keywords" title="Keywords">{{keyword}}</span>
```
petrjasek, line 55: not sure we need the title, but if so it should be the same as label (translated)

scripts/apps/search/components/preview-subject.tsx
```diff
@@ -24,11 +24,11 @@ export class PreviewSubject extends React.PureComponent<IProps> {
                     /* Adding the Subject heading only for
                     first time so that seems scheme subjects are grouped together.*/
 
-                    subjects.push(<div>
+                    subjects.push(
                         <span
-                            key={subject?.scheme + ':' + subject?.qcode}
+                            key={field.display_name}
```
petrjasek, line 29: you can probably use `field._id` which should be unique, `display_name` might not be and that would break the UI

GyanP: I also had this thing in my mind but for example, we have one field called `Packages`  if we use `_id ` for that one then on UI we have a label **Service-product**, that's why I have used display_name here so that on UI we have a label **Packages**, is that ok?

petrjasek: yep for label it's ok, but this is `key` which is not visible and must be unique

scripts/apps/archive/views/preview.html
```diff
@@ -48,11 +48,11 @@
             </div>
             <div class="tag-main" ng-if="selected.preview.authors">
               <span class="inline-label">{{preview_headings.author}}:</span>
-              <span class="tag-label" ng-repeat="author in selected.preview.authors">{{author.role}}:{{author.sub_label}}</span>
+              <span class="tag-label" ng-repeat="author in selected.preview.authors" title="{{author.role}}:{{author.sub_label}}">{{author.role}}:{{author.sub_label}}</span>
             </div>
             <div class="tag-main" ng-if="selected.preview.keywords">
               <span class="inline-label"><br/>{{preview_headings.keywords}}:</span>
-              <span class="tag-label" ng-repeat="keyword in selected.preview.keywords" title="Keywords">{{keyword}}</span>
+              <span class="tag-label" ng-repeat="keyword in selected.preview.keywords" title="{{keyword}}">{{keyword}}</span>
```
petrjasek, line 55: I think in general we don't need title for span if it's the same as the actual value, not sure if we use it anywhere else?

GyanP: actually, I'm added it to maintain the consistency like we have a title for subject-related fields also(here it is https://github.com/superdesk/superdesk-client-core/blob/36822018045b10c9d8c838d34a6bd6c6b4d106f5/scripts/apps/search/components/preview-subject.tsx#L38) so that while content-preview we have a title for author and storytags like we have for subjects(Belga-keywords etc)

scripts/apps/archive/directives/MediaPreview.ts
```diff
@@ -115,6 +116,11 @@ export function MediaPreview(api, $rootScope, desks, superdesk, content, storage
             scope.$on('$destroy', () => {
                 elem.off('dragstart');
             });
+
+            scope.preview_headings = {
```
petrjasek, line 120: just wondering why using this instead of `| translate` in the template?

petrjasek (approved):

PR #4092, opened 2022-05-04
throttle updates search view

tomaskikutis:
SDESK-6409

Prevents frequent re-rendering when there are many websocket events at once from different ingests

### DIFF
--- a/scripts/apps/search/directives/SearchResults.ts
+++ b/scripts/apps/search/directives/SearchResults.ts
@@ -1,4 +1,4 @@
-import _ from 'lodash';
+import _, {throttle} from 'lodash';
 import {gettext} from 'core/utils';
 import {appConfig} from 'appConfig';
 import {ISearchOptions, showRefresh} from '../services/SearchService';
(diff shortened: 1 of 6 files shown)


scripts/apps/search/directives/SearchResults.ts
```diff
@@ -94,6 +94,45 @@ export function SearchResults(
             var criteria = search.query(getSearch(), queryOptions).getCriteria(true),
                 oldQuery = _.omit(getSearch(), '_id');
 
+            let fetchInProgress = false;
+
+            /**
+             * Schedule an update
+             */
+            const queryItems = throttle(
```
petrjasek, line 102: I think we would only need to throttle the call to `_queryItems` otherwise there might be events which we ignore anyhow getting in a way of events which should trigger an update.

tomaskikutis: I don't see how it could work incorrectly in any scenario. Do you? If so, I'll adjust it.

We'll eventually delete this code together with everything else remaining in angular so I would rather not fix what's not broken in this case :D

petrjasek: yep you're right, for a moment I thought there might be some events which would trigger update on those external ones but that's not the case. my bad

petrjasek (approved):

PR #4095, opened 2022-05-13
Expose body editor's link pre-populated protocol as setting in config file [SDESK-6227]

GyanP:


scripts/core/editor3/components/links/LinkInput.tsx
```diff
@@ -176,7 +176,7 @@ export class LinkInputComponent extends React.Component<any, any> {
                         onChange={(e) => {
                             this.setState({url: e.target.value});
                         }}
-                        placeholder={gettext('Insert URL')}
+                        placeholder={gettext('https://')}
```
petrjasek, line 179: I don't think we need `gettext` for this

petrjasek (approved):

PR #4098, opened 2022-05-20
Improvements for the top Assignments and Mark for me widgets on top panel [SDESK-6431]

GyanP:


scripts/core/ui/components/TopMenuDropdownButton.tsx
```diff
@@ -20,6 +20,7 @@ export class TopMenuDropdownButton extends React.PureComponent<{
 
         return (
             <button
+                title="Marked for me"
```
petrjasek, line 23: gettext missing

scripts/core/superdesk-api.d.ts
```diff
@@ -1025,6 +1025,7 @@ declare module 'superdesk-api' {
         noPadding?: boolean;
         justifyContent?: string;
         bold?: boolean;
+        title?: string;
```
tomaskikutis, line 1028: This is a layout interface, it'd be good not to include unrelated properties if possible. Can you use the title in the child component?

GyanP: I thought the same but if we use the title in the child property, In this case, we have to declare it in the `<Slugline>` and `<Headline>` Component, So does it make sense to you? Another problem is, that when we hover over the article it will not show the tooltip, it will show the tooltip only if we hover over the slugline or headline which I think is wrong.

tomaskikutis: I took a look at the interface again, and I think adding a title is reasonable because ellipsis is set in the same interface.

tomaskikutis (approved):

PR #4099, opened 2022-05-24
improve showRefresh function, avoid race conditions

tomaskikutis:
SDESK-6409

### DIFF
--- a/scripts/apps/search/directives/SearchResults.ts
+++ b/scripts/apps/search/directives/SearchResults.ts
@@ -97,10 +97,15 @@ export function SearchResults(
             let fetchInProgress = false;
 
             /**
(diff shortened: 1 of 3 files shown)


scripts/apps/search/directives/SearchResults.ts
```diff
@@ -246,19 +253,22 @@ export function SearchResults(
                 }
 
                 return api.query(getProvider(criteria), criteria).then((items) => {
+                    if (initiatedAt < cancelAllBeforeTime) {
+                        return;
```
petrjasek, line 257: might be good to reject here so we stop the chain if any

petrjasek (approved):

petrjasek (approved):

PR #4100, opened 2022-05-26
Improvements for the top Assignments and Mark for me widgets on top panel[SDESK-6431]

GyanP:


scripts/core/ui/components/TopMenuDropdownButton.tsx
```diff
@@ -2,14 +2,14 @@
...
 import classNames from 'classnames';
-import {gettext} from 'core/utils';
 
 export class TopMenuDropdownButton extends React.PureComponent<{
     onClick: () => void;
     disabled?: boolean;
     active: boolean;
     pulsate: boolean;
     'data-test-id'?: string;
+    title?: string;
```
tomaskikutis, line 12: is it only for tooltip? If so, tooltip would be a clearer name that wouldn't get confused with button text

tomaskikutis (approved):

tomaskikutis (approved):

PR #4102, opened 2022-05-31
fix patch-package main directory resolution

petrjasek:
it was going into my home due to some assumptions about
node_modules and parent directory, imo it will be more reliable
to start with working directory from which the script is called.

### DIFF
--- a/tasks/patch-package.js
+++ b/tasks/patch-package.js
@@ -20,7 +20,7 @@ function copyFolderSync(from, to) {
 let copied = false;
 
 const clientCoreRoot = path.join(__dirname, '../');
-const maybeParentModulePath = path.join(clientCoreRoot, '../../');
+const maybeParentModulePath = process.cwd();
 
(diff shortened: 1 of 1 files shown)


tomaskikutis:
It has to go up to account for a scenario when client is installed as a node module and is located in `superdesk/client/node_modules/superdesk-client-core`. I suspect your patch would break things. Does it crash for you? It shouldn't be a problem that it checks your home directory, doesn't find `node_modules` there and falls back to `clientCoreRoot`

petrjasek:
if it's inside `superdesk/client` you run the script from client folder so it will be the cwd anyhow. and I actually seem to have `node_modules` in my home dir that's how I found it

tomaskikutis (approved):
Build scripts are tricky to deal with and especially test on multiple configurations. I would simply leave the current version and delete node_modules from home, but if you're feeling lucky, go ahead with the fix ;D

PR #4103, opened 2022-05-31
fix working directory in postinstall script

petrjasek:


petrjasek:
it's getting complicated, so reverting for now
https://github.com/superdesk/superdesk-client-core/pull/4104

PR #4106, opened 2022-06-07
Issues with the multiedit feature [SDBELGA-651]

GyanP:


scripts/apps/authoring/styles/multiedit.scss
```diff
@@ -20,9 +20,13 @@ $top-offset: 48px;
 }
 
 .multiedit-screen {
+    height: 90%;
```
tomaskikutis, line 23: why 90% and not 100%? Is this to account for bottom toolbar? If so, it'd be better to use `calc` function to subtract the bottom toolbar height.

GyanP: I have made some changes so that it looks good, I have align the burger menu and remove and save button on the same line.

tomaskikutis (approved):

PR #4108, opened 2022-06-14
[SDESK-6441] Ability to register new routing rule handlers

MarkLark86:


scripts/api/ingest.ts
```diff
@@ -0,0 +1,61 @@
...
+                day_of_week: ['MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT', 'SUN'],
+                hour_of_day_from: null,
+                hour_of_day_to: null,
+                _allDay: true,
+            },
+        };
+    },
+};
+
+function getIngestRuleHandlers() {
```
petrjasek, line 37: I would rather fetch those from the server

MarkLark86: Good point, I'll do add the endpoint tomorrow. Though I will still need part of this code, to allow extensions to register the UI that handles the specific ingest rule. Otherwise we'll need the specific UI's for each ingest rule in core

petrjasek (approved):

PR #4110, opened 2022-06-29
Fix afp news ml parser: not all subjects of an item are displayed in the metadata preview of an afp news item [SDBELGA-654]

devketanpro:


scripts/apps/archive/directives/MediaMetadata.ts
```diff
@@ -51,18 +51,18 @@ export function MediaMetadata(userList, archiveService, metadata, $timeout) {
             }
 
             scope.getLocaleName = function(terms, scheme) {
-                const term = terms.find((element) => element.scheme === scheme);
+                const filteredTerms = terms.filter((element) => element.scheme === scheme);
 
-                if (!term) {
+                if (!filteredTerms) {
                     return 'None';
```
petrjasek, line 57: this should probably return an array and that `None` can use gettext

devketanpro: Yup we can, 
I have 2 solutions for that :-
```
if (filteredTerms.length === 0) {
    return gettext("None")
 }
```
or
```
if (!filteredTerms) {
    return []
 }
``` 

Please let me know which one is best for our goal.
Thanks

devketanpro: yes i have updated the changes please let me know if want any changes. 
thanks.

tomaskikutis (approved):

PR #4114, opened 2022-07-19
rebuild package lock files with node 14

tomaskikutis:


petrjasek (commented):
would be also good to switch node version for github actions

petrjasek (approved):

PR #4115, opened 2022-07-28
[SDESK-6543] fix(contacts): trim string values before saving

MarkLark86:


scripts/apps/contacts/components/Form/ContactFormContainer.tsx
```diff
@@ -133,8 +133,30 @@ export class ContactFormContainer extends React.PureComponent<IProps, IState> {
...
+            if (typeof sanitizedValue === 'string') {
+                sanitizedValue = sanitizedValue.trim();
+
+                if (origContact[key] === undefined && sanitizedValue.length === 0) {
+                    return;
+                }
+            } else if (
+                sanitizedValue instanceof Array &&
+                sanitizedValue.length > 0 &&
+                typeof sanitizedValue[0] === 'string'
```
tomaskikutis, line 147: Since you call .trim later on all items, maybe it's better to check that all are strings:
`sanitizedValue.every((x) => typeof x === 'string')`

scripts/apps/contacts/components/Form/ContactFormContainer.tsx
```diff
@@ -133,8 +133,30 @@ export class ContactFormContainer extends React.PureComponent<IProps, IState> {
         let diff: any = {};
 
         each(this.state.currentContact, (value, key) => {
-            if (!isEqual(origContact[key], value)) {
-                extend(diff, {[key]: value});
+            let sanitizedValue = value;
```
tomaskikutis, line 136: Could you add a type for this value, or comment what are known possibilities?

scripts/apps/contacts/components/Form/ContactFormContainer.tsx
```diff
@@ -133,8 +133,30 @@ export class ContactFormContainer extends React.PureComponent<IProps, IState> {
...
+                    return;
+                }
+            } else if (
+                sanitizedValue instanceof Array &&
+                sanitizedValue.length > 0 &&
+                typeof sanitizedValue[0] === 'string'
+            ) {
+                sanitizedValue = (sanitizedValue as Array<string>)
+                    .map((arrayValue: string) => arrayValue.trim())
+                    .filter((arrayValue: string) => arrayValue.length);
```
tomaskikutis, line 151: It'd be better to make a function passed to `filter` return a boolean:
`.filter((arrayValue: string) => arrayValue.length > 0);`

tomaskikutis (approved):

PR #4116, opened 2022-08-08
[SDAAP-22] Fix missing link badge gone missing

marwoodandrew:


MarkLark86:
@petrjasek This bug was introduced in `v2.3` with the PR https://github.com/superdesk/superdesk-client-core/pull/3862

Should we back port this to `hotfix/2.4.1`?

petrjasek:
@MarkLark86 yep we should

MarkLark86 (approved):

PR #4120, opened 2022-08-24
Fix : 'Go to items' button for planning ingest takes the user to global search instead of planning section [SDESK-6560]

devketanpro:


scripts/apps/ingest/directives/IngestSourcesContent.ts
```diff
@@ -594,8 +595,16 @@ export function IngestSourcesContent(ingestSources, api, $location,
                             fulltext: '',
                         };
 
+                        var filter_value = '';
+
+                        if (contentTypes.includes('event')) {
+                            filter_value = 'EVENT';
+                        }
+                        if (contentTypes.includes('planning')) {
+                            filter_value = 'PLANNING';
```
MarkLark86, line 604: If the `contentTypes` includes both `event` & `planning`, this would have to be changed to `filter_value = 'COMBINED'`

MarkLark86 (approved):

PR #4121, opened 2022-09-02
[SDAAP-27] Missing qcode values in filter conditions were causing errors

marwoodandrew:


scripts/apps/content-filters/controllers/FilterConditionsController.ts
```diff
@@ -37,8 +37,12 @@ export function FilterConditionsController($scope, contentFilters, notify, modal
 
                 _.each(values, (value) => {
                     var v = _.find(allValues, (val) => val[valueField].toString() === value);
-
-                    $scope.filterCondition.values.push(v);
+                    if (v !== undefined) {
+                        $scope.filterCondition.values.push(v);
+                    }
+                    else {
+                        notify.warning(gettext('WARNING unknown filter condition value : ') + value);
```
MarkLark86, line 44: Adding the params inside the `gettext` call can help by providing a context to translators.
i.e.
```ts
gettext('WARNING unknown filter condition value: {{ value }}', {value: value})
```

marwoodandrew: OK, will change, lint is complaining as well!

MarkLark86 (approved):

PR #4122, opened 2022-09-14
Fix : Ingest dashboard does not count event and planning items [SDESK-6587]

devketanpro:


scripts/apps/ingest/directives/IngestUserDashboard.ts
```diff
@@ -27,8 +27,18 @@ export function IngestUserDashboard(api, userList, privileges, moment) {
                     },
                 };
 
-                api.ingest.query(criteria).then((result) => {
-                    scope.ingested_count = result._meta.total;
+                scope.item.content_types.map((types) => {
```
petrjasek, line 30: not sure with the map here, seems like if you have selected 4 types (text/picture/audio/video) this would run the query 4x

scripts/apps/ingest/directives/IngestUserDashboard.ts
```diff
@@ -27,9 +27,17 @@ export function IngestUserDashboard(api, userList, privileges, moment) {
                     },
                 };
 
-                api.ingest.query(criteria).then((result) => {
-                    scope.ingested_count = result._meta.total;
-                });
+                if (scope.item.content_types.includes('event') || scope.item.content_types.includes('planning')) {
+                    const param = scope.item.content_types.includes('event') ? 'events' : 'planning';
```
petrjasek, line 31: not sure if param is a best name, something like resource would be more descriptive

scripts/apps/ingest/directives/IngestUserDashboard.ts
```diff
@@ -27,9 +27,17 @@ export function IngestUserDashboard(api, userList, privileges, moment) {
...
+
+                    api.query(param, criteria).then((result) => {
+                        scope.ingested_count = result._meta.total;
+                    });
+                } else {
+                    api.ingest.query(criteria).then((result) => {
```
petrjasek, line 37: you could use `api.query('ingest', ...)` here and avoid some duplication

MarkLark86: This whole change can be reduced down by assigning a value `resource`, using that as first argument to `api.query` and setting the value to either `events`, `planning` or `ingest` based on the `content_types` included in the ingest config

petrjasek (approved):

PR #4123, opened 2022-09-20
Event and planning ingests appear in global search under provider but will not show results [SDESK-6546]

devketanpro:


scripts/apps/search/directives/SearchParameters.ts
```diff
@@ -241,7 +241,19 @@ export function SearchParameters($location, asset, tags, metadata, common, desks
              */
             function fetchProviders() {
                 ingestSources.fetchAllIngestProviders().then((items) => {
-                    scope.providers = items;
+                    const filteredItems = items.filter((item) => {
+                        const content_type = item.content_types.find((type, index) => {
```
petrjasek, line 245: I'm not what the code does really, you put into content_type type which is not planning related and then you check if that type is included in the same list? I guess it will be always true if the content_type is not null. but imo more readable would be something like
```typescript
scope.providers = items.filter((provider) => provider.content_types.some((type) => type !== 'event' && type !== 'planning'))`
```

petrjasek (approved):

PR #4124, opened 2022-09-23
Fix : Lock on the sign-off on story not being respected [SDESK-6593]

devketanpro:


scripts/apps/authoring/views/article-edit.html
```diff
@@ -585,7 +585,7 @@
            id="sign_off"
            ng-model="item.sign_off"
            ng-change="autosave(item)"
-           ng-disabled="!_editable || editSignOff == null || schema.sign_off.readonly === true"
+           ng-disabled="!editSignOff && true"
```
petrjasek, line 588: not sure why this `&& true` is there?

petrjasek (approved):

PR #4125, opened 2022-10-03
fix dragging chrome 106

tomaskikutis:


scripts/apps/archive/services/DragItemService.ts
```diff
@@ -51,14 +51,14 @@ export function DragItemService() {
 
         dt.effectAllowed = 'link';
 
-        if (item.renditions && item.renditions.thumbnail) {
-            const img = document.createElement('img');
-            const div = getThumbnailPlaceholder();
-            const rendition = item.renditions.thumbnail;
+        // if (item.renditions && item.renditions.thumbnail) {
```
petrjasek, line 54: can you pls add a comment that it's crashing chrome 106

PR #4132, opened 2022-10-11
Multi edit save improvement [SDESK-6531]

devketanpro:


scripts/apps/authoring/multiedit/multiedit.ts
```diff
@@ -221,6 +221,7 @@ function MultieditArticleDirective(authoring, content, multiEdit, lock, $timeout
                         () => {
                             scope.$applyAsync(() => {
                                 InitializeMedia.initMedia(scope);
+                                scope.save();
```
petrjasek, line 224: I'm not sure if calling save on autosave is good idea, that would generate too many versions. and how does it fix the issue?

devketanpro: Actually IMO the problem is that when we back to multi edit editor, we load the auto saved changes but it is not preserving our data. that's why I have added `scope.save()` under the autosave.

petrjasek: so the autosave is not working?

devketanpro: As multi-edit only preserves and reflects those changes which are saved using save() not autosave(), for doing so we have to use save() with it.
if we talk about main editor , there is a functionality when we close the editor they show popup to save these changes or not. but in multiedit we have not , so I have add save() in the autosave() in the multiedit.

petrjasek: yep I see, but calling save creates a new version and is something that should be triggered by user action and not via autosave. so that doesn't work as a fix imo

devketanpro: Ok I got your point, Let me try to find another solution and let you know. Thanks for update.

PR #4134, opened 2022-10-13
Macros widget refactoring

thecalcc:


scripts/apps/authoring-react/macros/macros.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+const getLabel = () => gettext('Macros widget');
+
+interface IState {
+    macros: Array<IMacro> | null;
+    article: {
+        saved: IArticle | null,
+        autosaved: IArticle | null,
+    };
+    displayGrouped: boolean | null;
```
tomaskikutis, line 37: When there is some special meaning like here, add a comment in the interface like
`displayGrouped: boolean | null; // null when grouping can't be toggled. It can't be toggled when there are no groups.`

scripts/apps/authoring-react/macros/macros.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+            Object.keys(groupedMacros).sort().forEach((key) => {
+                groupedOrdered[key] = groupedMacros[key].sort((a, b) => a.label.localeCompare(b.label));
+            });
+            groupedOrdered['Miscallaneous'] = this.state.macros.filter((x) => x.group == null);
+        }
+
+        const RunMacroButton = (macro: IMacro) => {
+            return (
+                <div key={macro.name} style={{paddingTop: 4}}>
```
tomaskikutis, line 145: It's not a good practice to use key outside of iteration. Better convert this to an actual component and use it like that:
 `<RunMacroButton key={macro.name} macro={macro} />`

PR #4135, opened 2022-10-18
Display description of (ingested) audio and video files also in the preview [SDESK-6618]

devketanpro:


scripts/apps/archive/views/preview.html
```diff
@@ -151,6 +151,8 @@
 
             <div class="text dateline" order="{{editor['dateline'].order}}" ng-show="selected.preview.dateline" sd-html-preview="selected.preview.dateline.text"></div>
 
+            <div class="text body-text" order="{{editor['event_description'].order}}" ng-show="selected.preview.description_text" sd-html-preview="selected.preview.description_text" item="selected.preview"></div>
```
petrjasek, line 154: not sure about that `event_description` for order? why is that?

devketanpro: Because IMO if we use `event_description` for order, the `item.description` it looks like this.
SDESK-6618 (1)

petrjasek: I would probably rather do this like for [image description](https://github.com/superdesk/superdesk-client-core/pull/4135/files#diff-32d5353b9ef1fdafaa57a2b7353e61c62ce7a2647a7f616138a74603ff602727R136) so it would be always after the media player

petrjasek (changes requested):
that `media_description` is a specific field, it would be better to work the same like description for images

petrjasek (approved):

PR #4145, opened 2022-11-24
[TGA-38] Pass field details when rendering custom fields

MarkLark86:


scripts/core/superdesk-api.d.ts
```diff
@@ -2154,6 +2155,7 @@ declare module 'superdesk-api' {
     export interface IPreviewComponentProps {
         item: IArticle;
         value: any;
+        field: IVocabulary;
```
tomaskikutis, line 2158: It'd be cleaner to only expose field_id. On the other hand, you're only using it to retrieve a label, which shouldn't be done inside the extenion so maybe it's not needed to expose it at all.

tomaskikutis (commented):
It's not a good idea to use `archive` collection for this. Users will then show up in articles search, mess up statistics data etc. Since this would be a form with 10 fields or so, I'd rather create a new UI page via extensions API(via `contributions.pages`), and implement a custom form that'd work against a dedicated DB collection.

If we need to ship it very quickly, maybe we can ship this approach and start working on an upgrade that would implement it properly.

scripts/core/superdesk-api.d.ts
MarkLark86: Please see my comment in the `superdesk-tga` repo regarding the labels.

MarkLark86: In regards to the `field` param, I agree it may be cleaner to only expose the bare minimum that **current** extensions are using, but I do not see an issue with passing the entire `IVocabulary` to the extension considering that data belongs to said extension and is not outside of it's scope anyway.

tomaskikutis: When designing the API I'm thinking about many use cases, not only about current ones, but I can't think of why would an input field need a vocabulary. Custom fields by design are not supposed to render labels themselves, and for field configs, we have dedicated config interface and component. Also not all fields may have an associated vocabulary - for example for rundown items(that are built on top of refactored authoring) I'm using inline JSON as a content profile and there are no associated vocabularies in that case.

I think an easy and clean solution here, if one really needed a vocabulary, would be to pass field ID and fetch the vocabulary inside field editor/preview component.

scripts/apps/authoring/authoring/authoring-custom-field.tsx
```diff
@@ -89,6 +89,7 @@ export class AuthoringCustomField extends React.PureComponent<IProps, IState> {
                             setValue={(value) => this.setValue(value)}
                             readOnly={!editable}
                             config={field.custom_field_config}
+                            field_id={field._id}
```
petrjasek, line 92: looks like python, `fieldId` might be better

petrjasek (approved):

PR #4151, opened 2022-12-02
Remove formatting does not work in body_html [SDESK-6692]

devketanpro:


tomaskikutis:
If TypeScript types are missing for functions that you are modifying I'd be good to add them. Can you do this?

devketanpro:
> If TypeScript types are missing for functions that you are modifying I'd be good to add them. Can you do this?

Yeah sure

scripts/core/editor3/helpers/selection/blockKeys.ts
```diff
@@ -1,8 +1,12 @@
-export const getBlockKeys = (contentState, start, end) => [
-    ...contentState
+export interface IContentState {
```
petrjasek, line 1: imo would be better to just use imported `ContentState` here instead of making things up

petrjasek (approved):

PR #4157, opened 2022-12-07
Toolbar export action modal

thecalcc:


scripts/apps/authoring-react/toolbar/export-modal.tsx
```diff
@@ -0,0 +1,126 @@
...
+            return httpRequestJsonLocal({
+                method: 'POST',
+                path: '/export',
+                payload: {
+                    item_ids: [this.props.article._id],
+                    validate: this.state.validate,
+                    format_type: this.state.selectedFormatter,
+                },
+            }).then((response: any) => {
+                return window.open(response.url);
```
petrjasek, line 66: we might also close the modal once the download is triggered

scripts/apps/authoring-react/toolbar/export-modal.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+        return (
+            <Modal
+                size="medium"
+                onHide={this.props.closeModal}
+                visible
+                zIndex={1050}
+                headerTemplate={gettext('Export')}
+            >
+                <Spacer v gap="32">
```
petrjasek, line 86: would be good to use some constants for such numbers like 32 so it's consistent and defined in 1 place

petrjasek: I see

tomaskikutis:
reviewed

PR #4160, opened 2022-12-13
fix svg import

petrjasek:


scripts/custom.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
```
tomaskikutis, line 1: it's probably better to put it to globals file where we already have some extensions registered https://github.com/superdesk/superdesk-client-core/blob/broadcasting/scripts/globals.d.ts#L83

pavlovicnemanja (approved):

tomaskikutis (approved):

PR #4163, opened 2022-12-21
Fix : code is shown instead of a proper label inside filter condition drop down list [SDESK-6762]

devketanpro:


tomaskikutis:
> Fix : filter condition drop down list contains reserved keywords

What are "reserved keywords"?

scripts/apps/workspace/helpers/getLabelForFieldId.ts
```diff
@@ -4,9 +4,10 @@ import ng from 'core/services/ng';
 export const getLabelForFieldId = (fieldId, vocabularies) => {
     const labelMap = GET_LABEL_MAP();
     const field = vocabularies.find((obj) => obj._id === fieldId);
+    const categories = vocabularies.find((v) => v._id === 'categories');
 
-    if (fieldId === 'anpa_category') {
-        return vocabularies.find((v) => v._id === 'categories').display_name;
+    if (fieldId === 'anpa_category' && categories) {
```
tomaskikutis, line 9: Only use boolean expressions inside `if` statements. `categories` is a vocabulary object, not a boolean, so you should use `categories != null` inside `if` stamenent - which will evaluate to a boolean.

Also `categories` is not the best naming choice since the variable represents a single vocabulary while variable name is in plural. A better name would be `categoriesVocabulary`.

devketanpro:
> > Fix : filter condition drop down list contains reserved keywords
> 
> What are "reserved keywords"?

In the sense that we displayed an unnecessary element in the drop down like this:-
filter_condition_dropdown

tomaskikutis:
I've looked more into the issue - it turns out that due to a missing vocabulary a label inside the dropdown couldn't be retrieved.

I've opened https://github.com/superdesk/superdesk-client-core/pull/4169 to deal with incomplete instance configuration.

PR #4164, opened 2022-12-23
Toolbar compare article versions

thecalcc:


tomaskikutis:
add status and author

tomaskikutis:
ui improvements needed

tomaskikutis:
TODO: move panel to a separate component

tomaskikutis:
all good, except we need to reorder props and improve the UI to use a table so we can support dynamic header height

PR #4165, opened 2022-12-23
make the name of of the routing rule available for editing in the second step of the form [SDESK-6774]

devketanpro:


fritzSF (changes requested):
Hey @devketanpro , please don't use new CSS for common elements that are alreadt defined.
The structure should be like this:
`<div class="button-group button-group--end button-group--comfort" role="group">
   <button class="btn" tabindex="0">Cancel</button>
   <button class="btn btn--primary" tabindex="0">Save</button>
   <button class="btn btn--primary" tabindex="0">Done</button>
</div>`
inside the 'modal__body-footer' in this case
Additionaly this will be needed to make it work:
```
.modal.modal--tabs .modal__body-footer {
    display: flex;
}
```
you can add this in `/superdesk-client-core/styles/sass/sf-additional.scss`

scripts/apps/ingest/views/settings/ingest-routing-content.html
```diff
@@ -121,12 +121,12 @@ <h3 class="modal__heading" translate ng-hide="editScheme._id" translate>Add New
...
             </div>
             <div class="modal__body-footer">
-                <div class="modal_footer_btn">
-                    <div class="cancelBtn" ng-if="rule">
-                        <button class="btn" ng-click="cancelRule(rule)" translate>Cancel</button>
-                        <button class="btn btn--primary" ng-click="save()" ng-disabled="!rule.name" translate>Save</button>
-                    </div>
-                    <button class="btn btn--primary" ng-click="done()" ng-disabled="rule" translate>Done</button>
+                <div class="button-group button-group--end button-group--comfort" role="group">
+                    <span ng-if="rule">
```
fritzSF, line 125: An inline element (span in this case),  should not be used as a container for block level elements (button, div, etc.). Use rather a div instead. Also, you can use a helper class on that element to avoid flex problems and the need for additional CSS, like so:
`<div class="sd-display--contents" ng-if="rule">`

styles/sass/sf-additional.scss
```diff
@@ -2361,4 +2361,14 @@ sd-multi-image-edit {
         justify-content: flex-end;
         gap: var(--gap--small);
     }
+}
+
+.modal.modal--tabs .modal__body-footer .button-group{
```
fritzSF, line 2366: I don't think you need all of this, most of it already comes from the framework. Only 'display: flex' is needed.
`.modal.modal--tabs .modal__body-footer .button-group {
    display: flex;
}`
If you add the helper class described in the previous comment you don't need the 'span' properties as well.

fritzSF (changes requested):

fritzSF (approved):

PR #4166, opened 2022-12-27
[SDAAP-56] Quirk in the headline character count counts the last spac…

marwoodandrew:
…e as 6 characters

### DIFF
--- a/scripts/apps/authoring/authoring/components/CharacterCount.tsx
+++ b/scripts/apps/authoring/authoring/components/CharacterCount.tsx
@@ -10,7 +10,7 @@ interface IProps {
 }
 
 export function getEditorFieldCharactersCount(text: string, html?: boolean) {
-    let input = text || '';
+    let input = text?.trim() ?? '';
 
     input = html ? cleanHtml(input) : input;
     input = input.replace(/\r?\n|\r/g, '');
--- a/scripts/apps/authoring/views/article-edit.html
+++ b/scripts/apps/authoring/views/article-edit.html
@@ -24,7 +24,7 @@
 
     <label for="title" class="field__label">{{label('headline')}}</label>
     <span ng-if="schema.headline.maxlength">
-        <sd-character-count data-item="item.headline" data-html="true" data-limit="schema.headline.maxlength"></sd-character-count>
+        <sd-character-count data-item="item.headline" data-html="false" data-limit="schema.headline.maxlength"></sd-character-count>
     </span>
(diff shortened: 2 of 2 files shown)


tomaskikutis:
It isn't clean to say the string is HTML when it's not :) I think trimming before counting characters would help. If not, post a snippet of text where it doesn't count correctly and I'll fix it.

tomaskikutis (approved):

marwoodandrew:
@tomaskikutis I think it is html, at least at the point the character count is being done it is :) I'm not saying it should be!

If you look at the [ticket](https://sofab.atlassian.net/jira/software/c/projects/SDAAP/boards/66?modal=detail&selectedIssue=SDAAP-56) you'll see a real world example of a string that demonstrates the issue with a video!

If you log into that PR's test instance instance find the story slugged ["BAD HEADLINE LENGTH"](https://sdcpr-4166.test.superdesk.org/#/workspace/monitoring?item=4ae226e5-7be4-4de9-ba1f-067aff4dc5f8&action=edit) and edit it, you'll note the length of headline "123<>4" is given as 6 (correct) and is indeed text, Now just go add a space at the end and magicaly the headline length becomes 18 which by some strange coincidence has the same length as the string `123&lt;&gt;4&nbsp;`

image

If you inspect the element you get the following, which shows the value being a bit html like.
`<div class="text-editor editor-type-html ng-isolate-scope medium-editor-element" tabindex="10" sd-text-editor-drop-zone="false" sd-focus-element="" data-append-element=".field" data-append-class="active" placeholder="Add content here" ng-show="type == 'text' || !type" spellcheck="false" contenteditable="true" data-medium-editor-element="true" role="textbox" aria-multiline="true" data-medium-editor-editor-index="5" medium-editor-index="d584b5b5-92b9-5498-a376-fabad068b553" data-medium-focused="true">123&lt;&gt;4&nbsp;</div>`

I was simply reverting the behavour back to how it was in 1.33, which incidentaly was wrong although it handles the `&nbsp;` it doesn't handle the `&lt; `or `&gt;` in I think [cleanHTML](https://github.com/superdesk/superdesk-client-core/blob/83f62bdd1771a6721529b68451388ae2c9e85c9f/scripts/apps/authoring/authoring/helpers.ts#L161) might need to do more!

tomaskikutis:
You're right, it does escape spaces to HTML entities which then causes wrong character count. I didn't notice it's about editor2.

PR #4167, opened 2022-12-28
Also handle semicolons (and other possible special characters) in the metadata mapping of image uploads [SDESK-6766]

devketanpro:


petrjasek (approved):

tomaskikutis:
@devketanpro I'm wondering what was the root cause and the solution? Why did you cast fields to a string? Weren't they  strings to begin with?

PR #4168, opened 2022-12-30
Fix : Locked items in package widget cannot be opened [SDESK-6706]

devketanpro:


scripts/apps/authoring/packages/packages.ts
```diff
@@ -29,7 +29,9 @@ function PackagesCtrl($scope, superdesk, api, search) {
         if (packageItem._type === 'published') {
             superdesk.intent('view', 'item', packageItem);
         } else {
-            superdesk.intent('edit', 'item', packageItem);
```
tomaskikutis, line 32: A function is called here to initiate editing. You should go to the implementation and fix the issue there

devketanpro: did you mean `superdesk.intent` needs to be changed?

tomaskikutis: `superdesk.intent` is a general function that is used for many things, but yes - you should track the callstack through the intent call until you get to code that is initiating the editing and do the change there.

scripts/apps/authoring/authoring/index.ts
```diff
@@ -216,8 +216,8 @@ angular.module('superdesk.apps.authoring', [
                     {action: 'list', type: 'archive'},
                     {action: 'edit', type: 'item'},
                 ],
-                additionalCondition: ['authoring', 'item', function(authoring, item) {
-                    return authoring.itemActions(item).edit;
+                additionalCondition: ['authoring', 'item', async function(authoring, item) {
```
tomaskikutis, line 219: For one we aren't using async/await in the codebase, so you'd need to rewrite this using promises. On the other hand, `itemActions` function is synchronous so neither `.then` nor `await` can be used.

tomaskikutis: you should probably step into use a debugger to step into `authoringWorkspace.edit` on line 213, go to the implementation and do the thing you tried to do in the initial PR version. I'm still not sure actually what's the issue and how you are trying to fix it. Can you comment on that too?

devketanpro: When we are trying to edit locked packages inside locked packages then We are not getting response on time from 
return `authoring.itemActions(item)` and because of that error which is not handled by us,and we are not able to reach at 
`authoringWorkspace.edit(data.item ? data.item : data);`  so, the main problem is not in this , problem lies in `return authoring.itemActions(item).`


image (12)

tomaskikutis: > We are not getting response on time

Which line in your opinion is asynchronous? I looked into it and it seems to me that we do get the response, it's simply `false` - I mean it refused to allow to edit it. I've looked into removing the condition in the first place, but it would likely change behavior somewhere else so I applied a targeted fix in packages.

PR #4172, opened 2023-01-03
avoid opening preview on focus when item is already selected

petrjasek:
SDESK-6777 SDNTB-784

[SDESK-6777]: https://sofab.atlassian.net/browse/SDESK-6777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDNTB-784]: https://sofab.atlassian.net/browse/SDNTB-784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/apps/search/components/ItemList.tsx
+++ b/scripts/apps/search/components/ItemList.tsx
@@ -151,8 +151,8 @@ export class ItemList extends React.Component<IProps, IState> {
         }
 
         if (event.type === 'focus' && item === this.getSelectedItem()) {
-            // when returning to same screen and item is already selected
-            // this should be noop
+            // when returning to browser tab focus can be triggered by browser
+            // and trigger select on already selected item
             return;
         }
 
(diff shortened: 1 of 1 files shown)


tomaskikutis:
Instead of handling it inside `select` function, couldn't we ensure that select function is not called at all in that case? Focusing an input in authoring shouldn't trigger focus on list item

petrjasek:
it's actually focus called on the item, it only happens when the item is locked so the focus doesn't go to authoring, then when the window is revisited the previously focused item gets the focus again

tomaskikutis:
I didn't see the video previously, now I understand the issue. Maybe make a comment more clear by specifying that by " returning to same screen" you mean browser tab changes and browser behavior of focusing element again.

tomaskikutis (approved):

PR #4173, opened 2023-01-09
Fix : ingest dashboard does not count incoming items [SDESK-6782]

devketanpro:


scripts/apps/ingest/directives/IngestUserDashboard.ts
```diff
@@ -27,14 +27,11 @@ export function IngestUserDashboard(api, userList, privileges, moment) {
...
-                var resource = 'ingest';
+                var resources = ['ingest', 'events', 'planning'];
 
-                if (scope.item.content_types.includes('event') || scope.item.content_types.includes('planning')) {
-                    resource = scope.item.content_types.includes('event') ? 'events' : 'planning';
-                }
-                api.query(resource, criteria).then((result) => {
-                    scope.ingested_count = result._meta.total;
-                });
+                resources.map((resource) => api.query(resource, criteria).then((result) => {
```
petrjasek, line 32: I would still do the filtering based on selected types as it would be mostly just one or the other, no need to query all for each subscriber

scripts/apps/ingest/directives/IngestUserDashboard.ts
```diff
@@ -27,8 +27,12 @@ export function IngestUserDashboard(api, userList, privileges, moment) {
                     },
                 };
 
-                var resources = ['ingest', 'events', 'planning'];
+                var resources = ['ingest'];
 
+                if (scope.item.content_types.includes('event') || scope.item.content_types.includes('planning')) {
```
petrjasek, line 32: imo splitting it would be more readable
```js
if (scope.item.content_types.includes('event')) {
    resources.push('events');
}
...

petrjasek (approved):

PR #4174, opened 2023-01-10
Fix : additional non default formatting options from content profile are not reflected in template [SDESK-6077]

devketanpro:


tomaskikutis:
to fix  this, you should get the `editor` options from a content profile rather than configs

devketanpro:
> to fix this, you should get the `editor` options from a content profile rather than configs

I also tried this approach but this will not working :-
```
               const editor = get(profile, 'editor',
                                  get(profile?.editor, contentType, constant.GET_DEFAULT_EDITOR()));
```

tomaskikutis:
I did test on sd-develop at the moment. As you said, after refresh, everything is shown correctly. I suspect the issue is that content profile is taken from the cache and we don't update that cache after modifying the content profile. The solution would then be to update the cache.

PR #4175, opened 2023-01-10
Add image suggestion feature to iMatrics widget

imatricsOpenSource:
Hi! This is our first time contributing to Superdesk! 

If you want to discuss any changes, we are able to hop on a call in our shared slack or discuss the changes in the comments. Happy reading!

### DIFF
--- a/scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
+++ b/scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
@@ -291,7 +291,10 @@ export class ImageTagging extends React.PureComponent<IProps, IState> {
                     triggerSelector="#image-suggestions-info-btn"
                     zIndex={999}
                 >
-                    {gettext('Image suggestions are based on generated tags and can be added to article content by dragging.')}
+                    {gettext(
+                        'Image suggestions are based on generated tags'
+                        + ' and can be added to article content by dragging.',
+                    )}
                 </Popover>
 
                 <div style={style}>
--- a/scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
+++ b/scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
@@ -1,0 +1,278 @@
+import * as React from "react";
+import { ISuperdesk } from "superdesk-api";
+import { ToggleBoxNext } from "superdesk-ui-framework";
+
+interface Tag {
+  title: string;
+  type: string;
+  pubStatus: boolean;
+  weight: number;
+}
+
+interface Image {
+  imageUrl: string;
+  thumbnailUrl: string;
+  id?: string;
+  headline?: string;
+  caption?: string;
+  credit?: string;
+  byline?: string;
+  source?: string;
+  dateCreated?: string;
+  archivedTime?: string;
+}
+
+interface ServerResponse {}
+
+interface Props {
+  superdesk: ISuperdesk;
+  data: any;
+  style?: React.CSSProperties;
+}
+
+const gridContainerStyle: React.CSSProperties = {
+  display: "grid",
+  gridTemplateColumns: "3fr 1fr",
+  gridColumnGap: "4px",
+  maxHeight: "320px",
+  overflow: "hidden",
+};
+
+const gridItemLeftStyle: React.CSSProperties = {
+  width: "100%",
+  height: "100%",
+  maxHeight: "inherit",
+};
+
+const gridItemRightStyle: React.CSSProperties = {
+  width: "100%",
+  height: "100%",
+  maxHeight: "inherit",
+};
+
+const imageContentStyle: React.CSSProperties = {
+  width: "100%",
+  maxHeight: "inherit",
+  overflowY: "auto",
+  display: "grid",
+  gridRowGap: "4px",
+};
+
+const imageStyle: React.CSSProperties = {
+  width: "100%",
+  height: "auto",
+};
+
+const imageWrapperStyle: React.CSSProperties = {
+  cursor: "pointer",
+  maxHeight: "100%",
+};
+
+const selectedCardStyle: React.CSSProperties = {
+  display: "flex",
+  flexDirection: "column",
+  height: "100%",
+  overflow: "hidden",
+  boxShadow: "1px 3px 6px 0 rgba(0,0,0,0.2)",
+  width: "100%",
+  borderRadius: "4px",
+};
+
+const cardStyle: React.CSSProperties = {
+  boxShadow: "1px 3px 6px 0 rgba(0,0,0,0.2)",
+  width: "100%",
+  borderRadius: "2px",
+};
+
+let isMounted: boolean = false;
+
+const ImageTaggingComponent = (props: Props) => {
+  const { superdesk, data, style } = props;
+  const { httpRequestJsonLocal } = superdesk;
+
+  const [showImages] = React.useState<boolean>(true);
+  const [isLoading, setIsLoading] = React.useState<boolean>(false);
+  const [fetchError, setFetchError] = React.useState<boolean>(false);
+  const [selectedImage, setSelectedImage] = React.useState<Image | null>(null);
+  const [images, setImages] = React.useState<Image[]>([]);
+
+  const fetchImages = async (item: Tag[], signal: AbortSignal) => {
+    setIsLoading(true);
+    httpRequestJsonLocal({
+      abortSignal: signal,
+      method: "POST",
+      path: "/ai_image_suggestions/",
+      payload: {
+        service: "imatrics",
+        item,
+      },
+    })
+      .then((data: any) => {
+        if (isMounted) {
+          setFetchError(false);
+          try {
+            setSelectedImage(data.result[0]);
+            setImages(data.result ?? []);
+          } catch {
+            setSelectedImage(null);
+            setImages([]);
+          }
+        }
+      })
+      .catch(() => {
+        setFetchError(true);
+      })
+      .finally(() => isMounted && setIsLoading(false));
+  };
+
+  const formatTags = (concepts: any) => {
+    let res: Tag[] = [];
+    for (const key in concepts) {
+      if (key === "subject") {
+        concepts[key].forEach((concept: any) => {
+          res.push({
+            title: concept.name,
+            type: "category",
+            pubStatus: true,
+            weight: 1,
+          });
+        });
+      } else {
+        concepts[key].forEach((concept: any) => {
+          res.push({
+            title: concept.name,
+            type: key,
+            pubStatus: true,
+            weight: 1,
+          });
+        });
+      }
+    }
+    return res;
+  };
+
+  const handleClickImage = (image: Image) => {
+    setSelectedImage(image);
+  };
+
+  const handleDragStart = (
+    event: React.DragEvent<HTMLDivElement>,
+    image: any
+  ) => {
+    try {
+      event.dataTransfer.setData(
+        "application/superdesk.item.picture",
+        JSON.stringify({
+          _id: image.id ?? "",
+          guid: image.id ?? "",
+          description_text: image.caption ?? "",
+          headline: image.headline ?? "",
+          original_source: image.source ?? "",
+          source: image.source ?? "",
+          versioncreated: image.archivedTime ?? "",
+          firstcreated: image.dateCreated ?? "",
+          pubstatus: "usable",
+          _type: "externalsource",
+          fetch_endpoint: "scanpix",
+          mimetype: "image/jpeg",
+          type: "picture",
+          renditions: {
+            thumbnail: {
+              href: image.thumbnailUrl ?? "",
+            },
+            viewImage: {
+              href: image.imageUrl ?? "",
+            },
+            baseImage: {
+              href: image.imageUrl ?? "",
+            },
+          },
+          byline: image.byline ?? "",
+          _created: image.dateCreated ?? "",
+        })
+      );
+    } catch {}
+  };
+
+  React.useEffect(() => {
+    const currentTags = data;
+    const formattedTags = formatTags(currentTags);
+    const controller = new AbortController();
+    fetchImages(formattedTags, controller.signal);
+    return () => controller.abort();
+  }, [data]);
+
+  React.useEffect(() => {
+    isMounted = true;
+    return () => {
+      isMounted = false;
+    };
+  }, []);
+
+  return isMounted ? (
+    <ToggleBoxNext
+      title={`image suggestions ${
+        isLoading ? "(...)" : fetchError ? "(error)" : `(${images.length})`
+      }`}
+      style="circle"
+      isOpen={showImages}
+      key="image-suggestion"
+    >
+      <div style={style}>
+        {isLoading ? (
+          <div style={{ display: "flex", alignItems: "center" }}>
+            <div className="spinner-big" />
+          </div>
+        ) : images?.length > 0 ? (
+          <div style={gridContainerStyle}>
+            <div style={gridItemLeftStyle}>
+              <figure style={selectedCardStyle}>
+                <div style={{ maxHeight: "72%" }}>
+                  <img
+                    style={imageStyle}
+                    alt=""
+                    src={selectedImage?.imageUrl}
+                    onDragStart={(e) => handleDragStart(e, selectedImage)}
+                  />
+                </div>
+                <figcaption
+                  style={{
+                    padding: "4px",
+                    overflow: "auto",
+                    color: "#000",
+                    backgroundColor: "rgba(255,255,255,0.75)",
+                    height: "100%",
+                  }}
+                >
+                  {selectedImage?.caption}
+                </figcaption>
+              </figure>
+            </div>
+            <div style={gridItemRightStyle}>
+              <div style={imageContentStyle}>
+                {images.map((image: Image, i: number) => (
+                  <div key={i} style={cardStyle}>
+                    <div style={imageWrapperStyle}>
+                      <img
+                        style={imageStyle}
+                        key={i}
+                        alt=""
+                        src={image.imageUrl}
+                        onClick={() => {
+                          handleClickImage(image);
+                        }}
+                        onDragStart={(e) => handleDragStart(e, image)}
+                      />
+                    </div>
+                  </div>
+                ))}
+              </div>
+            </div>
+          </div>
+        ) : null}
+      </div>
+    </ToggleBoxNext>
+  ) : null;
+};
+
+export default ImageTaggingComponent;
--- a/scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
+++ b/scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
@@ -219,9 +219,8 @@ export class ImageTagging extends React.PureComponent<IProps, IState> {
                         });
                     }
                 })
-                .catch((e: Error) => {
-                    superdesk.ui.alert('Failed to fetch image suggestions. Please, try again!\r'
-                    + JSON.stringify(e));
+                .catch(() => {
+                    superdesk.ui.alert('Failed to fetch image suggestions. Please, try again!');
                 })
                 .finally(() => this.setState({isLoading: false}));
         });
--- a/scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
+++ b/scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
@@ -4,6 +4,7 @@ import {ITagUi} from '../types';
 import {OrderedMap} from 'immutable';
 import {IServerResponse, ITagBase, toServerFormat} from '../adapter';
 import {ToggleBoxNext} from 'superdesk-ui-framework';
+import {debounce} from 'lodash';
 
 interface ITagInput {
     title: string;
@@ -38,7 +39,6 @@ interface IProps {
 interface IState {
     showImages: boolean;
     isLoading: boolean;
-    fetchError: boolean;
     selectedImage: IImage | null;
     images: Array<IImage>;
 }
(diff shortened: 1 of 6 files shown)


tomaskikutis:
Hi @imatricsOpenSource,

I didn't get to reviewing the code yet, but plan to do so soon. Ignore the failing `fire:www` task - the issue is on our side here.

tomaskikutis:
Regarding linting errros, you should see them by clicking "details" in "CI / test (pull_request) " row. Or running `npm run lint` locally. Currently the failing one is `ERROR: scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx[53, 36]: trailing whitespace`

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,278 @@
...
+  };
+
+  const handleDragStart = (
+    event: React.DragEvent<HTMLDivElement>,
+    image: any,
+  ) => {
+    try {
+      event.dataTransfer.setData(
+        "application/superdesk.item.picture",
+        JSON.stringify({
```
tomaskikutis, line 163: Specify the type of an object as `IArticle` before passing to `JSON.stringify`. This allows for TypeScript compiler to validate the properties.

```
const item: IArticle = {...};

event.dataTransfer.setData(
        "application/superdesk.item.picture",
        JSON.stringify(item),
)
```

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,278 @@
...
+  borderRadius: "2px",
+};
+
+let isMounted: boolean = false;
+
+const ImageTaggingComponent = (props: IProps) => {
+  const { superdesk, data, style } = props;
+  const { httpRequestJsonLocal } = superdesk;
+
+  const [showImages] = React.useState<boolean>(true);
```
tomaskikutis, line 91: Use class based components. We're not using hooks in the codebase.

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -47,13 +49,17 @@ interface IIMatricsFields {
     };
 }
 
-type IEditableData = {original: IAutoTaggingResponse; changes: IAutoTaggingResponse};
+type IEditableData = {
```
tomaskikutis, line 52: Could you revert lines that were modified, but nothing functional was actually changed, like here? Also whitespaces like added/removed newlines in other places.

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
```diff
@@ -47,13 +49,17 @@ interface IIMatricsFields {
...
+    original: IAutoTaggingResponse; 
+    changes: IAutoTaggingResponse;
+};
 
 interface IState {
     runAutomaticallyPreference: boolean | 'loading';
     data: 'not-initialized' | 'loading' | IEditableData;
     newItem: INewItem | null;
     vocabularyLabels: Map<string, string> | null;
+    showImageWidget: boolean;
```
tomaskikutis, line 62: This should be removed if it's not actually used. It's set to `true` in the constructor and can never be changed. In that case, simply render the component unconditionally.

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,278 @@
+import * as React from "react";
+import { ISuperdesk } from "superdesk-api";
+import { ToggleBoxNext } from "superdesk-ui-framework";
+
+interface ITag {
```
tomaskikutis, line 5: Is this the same entity as `ITagBase`? If so, let's use the existing interface

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,278 @@
...
+  credit?: string;
+  byline?: string;
+  source?: string;
+  dateCreated?: string;
+  archivedTime?: string;
+}
+
+interface IProps {
+  superdesk: ISuperdesk;
+  data: any;
```
tomaskikutis, line 27: Do not use `any`. Specify the actual type. Here and in other places in this PR.

tomaskikutis (changes requested):
Is there a specification/documentation somewhere describing how this should work? If not, maybe you could add it in the PR description?

It would also be useful to have a screen recording attached of the feature in action.

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
hanstchou: This is solved. It was hard to guess the payload, so I reused the article and changed what seemed relevant fields. Thanks for the feedback!

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
hanstchou: This is solved. Modified the functional component to the class component. Thanks for the feedback!

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
hanstchou: This is now reverted. Sorry about this. We didn't plan to modify it, it was because "ERROR: scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx[53, 36]: trailing whitespace" was shown. Thanks for the feedback!

scripts/extensions/auto-tagging-widget/src/auto-tagging.tsx
hanstchou: This has been refactored and renamed to "showImagesPreference", and uses the Superdesk preference setting similar to run automatically.

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
hanstchou: It's not same as ITagBase, it's in another format and is used in the request payload. This has been refactored to ITagInput. Thanks for the feedback!

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
hanstchou: This is now solved. IServerResponse from the adapter is now used. Thanks once again for the feedback! :)

hanstchou:
> Is there a specification/documentation somewhere describing how this should work? If not, maybe you could add it in the PR description?
> 
> It would also be useful to have a screen recording attached of the feature in action.

Hi, Tomas!

Currently, there is no official documentation. Below I attach a snippet of the extended widget.
Images should be suggested based on the tags, and the editors should be able to drag and drop any images on the article. It should function similarly to how the authors are currently attaching images in the articles, but faster and with ease.

Similar to the tagging, two additional variables need to be set in the config/env such as the image URL and credentials which we will provide upon request.

superdesk_image_suggestion

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,355 @@
...
+        }
+      }
+      return res;
+    }
+
+    handleClickImage = (image: IImage) => {
+      this.setState({ selectedImage: image });
+    }
+
+    handleDragStart = (
```
tomaskikutis, line 238: It only accesses `this.props.article` from the class. Move it out of a class and define as a standalone function. Pass `this.props.article` as an argument.

tomaskikutis (commented):
Thanks for explaining @hanstchou. Screenshots are especially useful.

> editors should be able to drag and drop any images on the article

from the code it seems it's only possible to drag the selected image and judging from the screenshot there are no visual clues for users to understand this. It would be good to either add a visual indication or make all images draggable, not only the selected one.

See my new comments on a class based component. Also finish fixing the lint issues. We have 2 linters and the second one only starts when first one doesn't report any issues. I did push a small change to make the first one pass. Main problems now are indentation style and usage of double quotes.

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,355 @@
...
+      event: React.DragEvent<HTMLDivElement>,
+      image: IImage | null,
+    ) => {
+      if (!image) {
+        return;
+      }
+      let item: IArticle;
+      try {
+        item = {
+          ...this.props.article,
```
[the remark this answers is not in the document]
hanstchou: Hi, Tomas!

I noticed this and we have now fixed the second lint issue. Additionally, the payload has been adjusted according to the backend changes.

The users can also drag the images on the list since they have a similar attribute (onDragStart) to the selected image. However, you have a point that there is no indication that they are draggable. However, the indicator feature has not been demanded/explained by the stakeholder.

From the UX perspective, how may we provide an indicator of the drag feature without needing to fill up more space in the widget for future development?

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,355 @@
...
+      clearTimeout(this.fetchTimeout);
+      if (
+        this._mounted &&
+        !this.state.isLoading &&
+        !this.isEqualTags(
+          this.formatTags(this.props.data),
+          this.state.prevInput,
+        )
+      ) {
+        this.fetchTimeout = setTimeout(() => {
```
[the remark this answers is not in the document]
hanstchou: The purpose of the timeout is to prevent immediate fetch upon tag changes. E.g. when manually removing 2 tags really quickly, we do not want 2 parallel requests. This is one way of preventing this without needing to cancel the first request.

hanstchou:
> from the code it seems it's only possible to drag the selected image and judging from the screenshot there are no visual clues for users to understand this. It would be good to either add a visual indication or make all images draggable, not only the selected one.
> 
> See my new comments on a class based component. Also finish fixing the lint issues. We have 2 linters and the second one only starts when first one doesn't report any issues. I did push a small change to make the first one pass. Main problems now are indentation style and usage of double quotes.


Hi, Tomas!

I noticed this and we have now fixed the second lint issue. Additionally, the payload has been adjusted according to the backend changes.

The users can also drag the images on the list since they have a similar attribute (onDragStart) to the selected image. However, you have a point that there is no indication that they are draggable. However, the indicator feature has not been demanded/explained by the stakeholder.

From the UX perspective, how may we provide an indicator of the drag feature without needing to fill up more space in the widget for future development?

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,355 @@
...
+      const { style } = this.props;
+      const { showImages, isLoading, fetchError, selectedImage, images } =
+        this.state;
+
+      if (!this._mounted) {
+        return null;
+      }
+      return (
+        <ToggleBoxNext
+          title={`image suggestions ${
```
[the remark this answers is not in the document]
hanstchou: Noted, thanks!

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,355 @@
...
+            this.setState({ fetchError: true });
+          })
+          .finally(() => this._mounted && this.setState({ isLoading: false }));
+      });
+    }
+
+    isEqualTags = (current: Array<ITagInput>, prev: Array<ITagInput>) =>
+      JSON.stringify(current) === JSON.stringify(prev)
+
+    formatTags = (concepts: IServerResponse) => {
```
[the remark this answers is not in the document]
hanstchou: Noted, thanks! Will move it outside the class in the next commit.

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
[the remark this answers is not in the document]
petrjasek: we allow external images, as long as it provides renditions with working urls it should be fine. ntb doesn't upload images in general, they use external images from scanpix

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,355 @@
...
+
+      this.state = {
+        showImages: true,
+        isLoading: false,
+        fetchError: false,
+        selectedImage: null,
+        images: [],
+        prevInput: [],
+      };
+      this._mounted = false;
```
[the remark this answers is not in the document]
hanstchou: Makes sense. _mounted will be removed.

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,355 @@
...
+      this.abortController.abort();
+      clearTimeout(this.fetchTimeout);
+    }
+
+    componentDidUpdate() {
+      clearTimeout(this.fetchTimeout);
+      if (
+        this._mounted &&
+        !this.state.isLoading &&
```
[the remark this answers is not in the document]
hanstchou: Adjusted. Thanks!

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,355 @@
...
+        !this.isEqualTags(
+          this.formatTags(this.props.data),
+          this.state.prevInput,
```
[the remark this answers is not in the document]
hanstchou: I see. Adjusted, thanks!

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
hanstchou: Thank you! handleDragStart has now been moved out of the class.

hanstchou:
Hi @tomaskikutis 

We think we have adjusted the code according to the feedback. Thank you!

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,355 @@
...
+      this.abortController.abort();
+      clearTimeout(this.fetchTimeout);
+    }
+
+    componentDidUpdate() {
+      clearTimeout(this.fetchTimeout);
+      if (
+        this._mounted &&
+        !this.state.isLoading &&
+        !this.isEqualTags(
```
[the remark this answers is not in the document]
hanstchou: Thank you!

hanstchou:
For the indicator, added a grab pointer on the images.

tomaskikutis:
> From the UX/UI perspective, do you have an idea of how may we provide an indicator of the drag feature without needing to fill up more space in the widget for future development?

One of the things I was about to suggest is using a relevant mouse cursor which you did already. Another useful thing to do would be to use a `ToggleBox` (not `ToggleBoxNext`) component from ui-framework which [has a badge property](https://user-images.githubusercontent.com/2076953/215345243-6d9b1a34-b9d7-4205-bb96-d0f27a6579dd.png). There you can pass any JSX and it will get displayed to the right of the title. There you could put an "i" icon(for information) and upon clicking it, display a popover describing the usage.

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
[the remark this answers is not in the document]
tomaskikutis: Can you use `debounce` or `throttle` from lodash instead? It accomplishes the same, but uses a more standard approach.

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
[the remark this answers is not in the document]
tomaskikutis: I was about to comment that usage of `gettext` is incorrect - you can't interpolate a string with JS and have to use `{{placeholder}}` syntax inside the string. On the other hand it would probably be better not to store `fetchError` in the state in the first place if the only purpose is to display it in the title. Better use `superdesk.ui.alert` to show an alert and suggest user to retry the action.

tomaskikutis:
When you push new commits, GitHub requires me to authorize running of tests, so it might look like lint issues are fixed when they're not. You can check it locally by running `npm run lint` in the project root.

hanstchou:
I've been running lint locally. It looks fine, and no error appears.

hanstchou:
We will make another commit with your suggested approach. Thanks for feedback.

hanstchou:
> 
> One of the things I was about to suggest is using a relevant mouse cursor which you did already. Another useful thing to do would be to use a `ToggleBox` (not `ToggleBoxNext`) component from ui-framework which [has a badge property](https://user-images.githubusercontent.com/2076953/215345243-6d9b1a34-b9d7-4205-bb96-d0f27a6579dd.png). There you can pass any JSX and it will get displayed to the right of the title. There you could put an "i" icon(for information) and upon clicking it, display a popover describing the usage.

hi @tomaskikutis 
When trying to replace ToggleBoxNext with ToggleBox from superdesk-ui-framwork following issue occurs. How do I get by this, and do you have a class or icon library for importing the "i"-icon?

Screenshot from 2023-01-31 16-05-01

tomaskikutis:
> When trying to replace ToggleBoxNext with ToggleBox from superdesk-ui-framwork following issue occurs.

Ensure you add `/react` at the end of your import: `import {IconButton} from 'superdesk-ui-framework/react';`

> do you have a class or icon library for importing the "i"-icon?

You can use `IconButton` with `info-sign`. Here's the list with all icons: https://ui-framework.superdesk.org/#/react/icon-font

hanstchou:
> Ensure you add `/react` at the end of your import: `import {IconButton} from 'superdesk-ui-framework/react';`

@tomaskikutis 
Still similar issue after importing from  'superdesk-ui-framework/react'.

hanstchou:
Hi @tomaskikutis

We have now replaced the ToggleBoxNext with a ToggleBox with a info button with popover showing the information. Hope everything is good now. Thank you!

scripts/extensions/auto-tagging-widget/src/ImageTaggingComponent/ImageTaggingComponent.tsx
```diff
@@ -0,0 +1,370 @@
...
+                        } else {
+                            return (
+                                <div style={gridContainerStyle}>
+                                    <div style={gridItemLeftStyle}>
+                                        <figure style={selectedCardStyle}>
+                                            <div style={{maxHeight: '72%'}}>
+                                                <img
+                                                    style={imageStyle}
+                                                    alt=""
+                                                    src={selectedImage?.imageUrl}
```
tomaskikutis, line 326: `<img` shouldn't be rendered at all if we don't have a url. Same for figcaption.

tomaskikutis (commented):
@hanstchou I did push a few improvements. I'd like you to fix one more thing comment and we're good to merge it then. Take a look at my changes and see if they're not breaking anything - I think they should. One more thing - ensure the server returns an empty array in `IImageServerResponse['result']` if no image suggestions are available.

tomaskikutis:
> Take a look at my changes and see if they're not breaking anything - I think they should

correction: by mistake I wrote that I expect my changes to break things. I meant to write that I do **NOT** expect them to break anything

hanstchou:
Hi @tomaskikutis 

Thank you for all the help and feedback!
I have ensured that images contains an URL by filtering the items with no containing URL in the response.

Also, the changes you made did not break anything, however, I could not test the method superdesk.ui.article.prepareExternalImageForDroppingToEditor since my local version is not on the latest branch.

We will modify the server to ensure return of an empty list if no images were found in the server repo.

tomaskikutis:
I did check it in action on one of our instances. It's good that you've filtered images that don't have a URL, but do update usage so optional chaining is not used then. Same for caption - either don't use optional chaining if it will always be present, or don't render `figcaption` node when caption is not present. See my [last comment](https://github.com/superdesk/superdesk-client-core/pull/4175#discussion_r1100412058)

If you updated the server to never return images that don't have a src, update the typescript interface on the client to make the `imageUrl` mandatory and remove client-side filtering.

hanstchou:
Hi @tomaskikutis !

Sorry that I missed the figcaption point. We have now fixed that figcaption is only rendered if there is a caption, and also removed unnecessary optional chaining in imageUrl.

On the other hand, we did only ensured that the server returns a result of an array. No filter of src in the server. Therefore we keep the filtering on the client.

Thank you for checking this through.

tomaskikutis:
> On the other hand, we did only ensured that the server returns a result of an array. No filter of src in the server. Therefore we keep the filtering on the client.

Can you filter it on the server instead?

Other than that, I don't have any other comments and we can merge it when unit tests pass.

hanstchou:
Hi @tomaskikutis !

We have now added the filter of images with no src on the server and removed the now unnecessary filter in the client. Let's pray for the unit test to pass! Thank you for everything!

tomaskikutis (approved):
Thanks @hanstchou

PR #4177, opened 2023-01-11
port send to desk changes from angular based authoring

thecalcc:


tomaskikutis:
will rename `send_to_desk_not_allowed`, make it optional
make available desks have a default value

tomaskikutis:
no need to merge, will cherry pick

TODO:

remove unused imports
send-to-tab make availableDesks private
destination-select
availableDesks prop add comment - defaults to allDesks
line 32 change to find and destruct

tomaskikutis:
replace filter with `has` to make checking for efficient scripts/core/interactive-article-actions-panel/utils/get-initial-destination.ts

scripts/core/interactive-article-actions-panel/actions/send-to-tab.tsx
```diff
@@ -38,7 +38,7 @@ export class SendToTab extends React.PureComponent<IProps, IState> {
         super(props);
 
         this.availableDesks = sdApi.desks.getAllDesks()
-            .filter((desk) => desk.send_to_desk_allowed)
+            .filter((desk) => desk?.send_to_desk_not_allowed === false)
```
tomaskikutis, line 40: first, you don't need to use optional chaining here, `desk` will always be an object. If you try accessing a non existing property of an object you simply get undefined, no error is being thrown - thus no need for optional chaining.

More importantly, your check should be `desk.send_to_desk_not_allowed !== true`, not `=== false`. You need to keep in mind that it would be `undefined` in some cases. Since we aren't enforcing it being mandatory, we'll not be using an upgrade script either

PR #4178, opened 2023-01-11
Handle the user message that a Metadata CV was updated in a less disturbing way [SDESK-6785]

devketanpro:


petrjasek (approved):

tomaskikutis:
I'm wondering @petrjasek why we are comparing it to session ID. Don't we want to show the message no matter which user modified a vocabulary?

petrjasek:
I think it's just checking if this current user is the one who edited the cv so then it doesn't need to tell him cv was updated. Question is if we need this at all.

PR #4179, opened 2023-01-11
Rename send_to_desk_not_allowed

thecalcc:


tomaskikutis (commented):
I saw end to end test fail for some reason. I noticed it's failing on sd-develop too, so your PR is not at fault. That being said, it's important to keep them passing. Usually we don't merge PRs with failing tests, so if it's failing it's likely some weird infrastructure issue that it was passing in a PR but started failing after merging. It sometimes happens with a chrome driver update for example. Just saying that if we notice the failure, we need to bring it up to each others attention to get it fixed.

In this case, I'll take care of it.

Regarding the PR itself, even if everything passes, only merge it when you have all related PRs ready for merging including the server ones

tomaskikutis (approved):

PR #4180, opened 2023-01-12
[SDESK-6655] fix(highlights): Fixed issue with blinking loader on highlights

pavlovicnemanja:


tomaskikutis:
Before ui-framework v3, everything was loaded after 3 dots loading animation. Now you changed it so after than loader, there's another, placeholder-style loader. Do we even need `SmoothLoader` then? Or we could remove it entirely? If yes, do so

pavlovicnemanja:
> Before ui-framework v3, everything was loaded after 3 dots loading animation. Now you changed it so after 3 dots loader, there's another, placeholder-style loader. Do we even need `SmoothLoader` then? Or we could remove it entirely? If yes, do so

Yes you are right. PR updated. We still need it on this place: `ItemListAngularWrapper`, but I removed it from other one

tomaskikutis (approved):
In general we shouldn't be removing `data-test-id` properties. They are used for e2e tests to find elements without relying too much on DOM structure. Even if they aren't used, it's good to keep them so we don't need to add them later when writing e2e tests.

We can keep the current PR as is, it's not a significant issue.

PR #4183, opened 2023-01-17
Highlights toolbar react

thecalcc:


scripts/core/helpers/show-popup-hoc.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface IProps {
+    children(toggle: (referenceElement: HTMLElement) => void): ReactNode;
+    placement: Placement;
+    Component: React.ComponentType<{closePopup(): void}>;
+    zIndex?: number;
+    closeOnHoverEnd?: boolean;
+    onClose?: () => void;
+}
+
+export class ShowPopoverHoc extends React.PureComponent<IProps> {
```
tomaskikutis, line 14: There's a naming convention for higher order components to start with `with`, like `WithPopover`. We can make exceptions, but conventional name sounds good to me, so maybe let's use that.

It'd be good also to have props camelCased `Component` -> `component`

scripts/apps/authoring-react/authoring-angular-integration.tsx
```diff
@@ -87,6 +89,41 @@ function getInlineToolbarActions(options: IExposedFromAuthoring<IArticle>): IAut
...
+                placement="right-end"
+                zIndex={1050}
+            >
+                {
+                    (togglePopup) => (
+                        <IconButton
+                            onClick={(event) =>
+                                togglePopup(event.target as HTMLElement)
+                            }
+                            id="select-highlights"
```
tomaskikutis, line 112: is ID used anywhere? I couldn't find it

scripts/apps/authoring-react/toolbar/highlights-management.tsx
```diff
@@ -0,0 +1,100 @@
...
+import {Card} from 'core/ui/components/Card';
+import {Spacer} from 'core/ui/components/Spacer';
+import {gettext} from 'core/utils';
+import React from 'react';
+import {IArticle, IHighlight} from 'superdesk-api';
+import {Button, Heading, IconButton, Text} from 'superdesk-ui-framework/react';
+
+interface IProps {
+    article: IArticle;
+    closePopup(): void;
```
tomaskikutis, line 12: Does this component need to know whether it's being rendered inside a popup or not? It doesn't look like it to me. Maybe we could call it simply "close"

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -197,19 +240,8 @@ interface IState {
...
-            <TranslateModal
-                closeModal={closeModal}
-                article={getItem()}
-            />
-        ));
-    },
-});
-
 export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapper, IState> {
+    highlightsPopupOpen: boolean;
```
tomaskikutis, line 244: I was about to say that it isn't used outside the component and we should make it `private`, but found that it isn't used anywhere at all so we should remove it ;D

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -348,21 +367,22 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
                                     const [authoringActionsFromExtensions, articleActionsFromExtensions] = res;
 
                                     return [
-                                        saveAsTemplate(item),
+                                        getSaveAsTemplate(item),
```
tomaskikutis, line 370: let's pass `getLatestItem` here too

scripts/apps/authoring-react/toolbar/highlights-management.tsx
```diff
@@ -0,0 +1,100 @@
...
+            this.setState({
+                ...this.state,
+                initialized: true,
+                selectedHighlighIds: res._items
+                    .filter((highlight) => this.props.article.highlights.includes(highlight._id)),
+            });
+        });
+    }
+
+    unmarkHighlight(highlightId: string): void {
```
tomaskikutis, line 48: It'd help to order component methods so it's easier to navigate. I order them the following way from bottom to top: `render` method, other react component life-cycle methods, custom methods and lastly the constructor at the top.

scripts/apps/authoring-react/toolbar/highlights-modal.tsx
```diff
@@ -0,0 +1,114 @@
...
+        if (this.state.initialized) {
+            this.setState({...this.state, isSaving: true});
+            sdApi.highlights.markItem(this.state.markedHighlights, this.props.article._id).then(() => {
+                this.props.closeModal();
+                dispatchInternalEvent('dangerouslyForceReloadAuthoring', undefined);
+            });
+        }
+    }
+
+    fetchArticleWithHighlights(): Promise<IArticle> {
```
tomaskikutis, line 63: not bound, on the other hand, it doesn't use anything from the class so it shouldn't be in the class anyway. You should use `sdApi.article.get` instead

scripts/apps/authoring-react/toolbar/highlights-modal.tsx
```diff
@@ -0,0 +1,114 @@
...
+interface IStateLoaded {
+    initialized: true;
+    availableHighlights: Array<IHighlight> | null;
+    markedHighlights: Array<string> | null;
+    isSaving: boolean;
+}
+
+type IState = IStateLoaded | IStateLoading;
+
+export default class HighlightsModal extends React.PureComponent<IProps, IState> {
```
tomaskikutis, line 29: let's use a named export instead. It's easier to grep usages

scripts/apps/authoring-react/toolbar/highlights-modal.tsx
```diff
@@ -0,0 +1,114 @@
...
+            >
+                <Spacer v gap="16">
+                    <MultiSelect
+                        onChange={(value: any) => {
+                            this.setState({
+                                ...state,
+                                markedHighlights: value.value.map(({_id}) => _id),
+                            });
+                        }}
+                        optionLabel="name"
```
tomaskikutis, line 93: not typed, use `nameof`

scripts/apps/authoring-react/toolbar/highlights-modal.tsx
```diff
@@ -0,0 +1,114 @@
...
+            <Modal
+                onHide={this.props.closeModal}
+                zIndex={1050}
+                size="small"
+                visible
+                headerTemplate={gettext('Highlights')}
+            >
+                <Spacer v gap="16">
+                    <MultiSelect
+                        onChange={(value: any) => {
```
tomaskikutis, line 87: would it work to get rid of `any` and replace it with a proper type?

thecalcc: Not really because the interface is defined as the value is a `string`.

PR #4185, opened 2023-01-18
Multi edit option in toolbar authoring react

thecalcc:


thecalcc:
Add a modal to choose the second article to enter multi-edit view before merging.

scripts/apps/authoring-react/toolbar/multi-edit-toolbar-action.tsx
```diff
@@ -0,0 +1,79 @@
...
+import {IArticle} from 'superdesk-api';
+import {MultiEditModal} from '../multi-edit-modal';
+import {Button, Modal, MultiSelect} from 'superdesk-ui-framework/react';
+import {Spacer} from 'core/ui/components/Spacer';
+import {showModal} from '@superdesk/common';
+import {gettext} from 'core/utils';
+import {sdApi} from 'api';
+
+interface IProps {
+    selectedArticles: Array<IArticle>;
```
tomaskikutis, line 11: `selectedArticles` -> `initiallySelectedArticles`

scripts/apps/authoring-react/toolbar/multi-edit-toolbar-action.tsx
```diff
@@ -0,0 +1,79 @@
...
+                contentPadding="none"
+                zIndex={1050}
+                onHide={this.props.onClose}
+                visible
+                headerTemplate={gettext('Select articles')}
+                size="medium"
+            >
+                <Spacer v gap="8" noWrap style={{padding: 10}}>
+                    <MultiSelect
+                        optionLabel="slugline"
```
tomaskikutis, line 40: use nameof - no magic strings

scripts/apps/authoring-react/toolbar/multi-edit-toolbar-action.tsx
```diff
@@ -0,0 +1,79 @@
...
+                        value={this.state.selectedArticles}
+                        onChange={(values) => {
+                            this.setState({selectedArticles: values as Array<IArticle>});
```
tomaskikutis, line 43: remove `as` assertion - the interface was fixed in ui-framework

tomaskikutis (commented):
try finding that special function for label
that should also fix bugs caused by label not being present
test again with articles that don't have a slugline neither other fields that the special function uses

thecalcc:
> try finding that special function for label that should also fix bugs caused by label not being present test again with articles that don't have a slugline neither other fields that the special function uses

Still need to check how to properly modify `optionLabel` prop of MultiSelect in primereact so that we have a function returning a string instead of just a string. That would allow us to be more flexible in choosing what to display as label.

thecalcc:
Changes before merging: 

https://sofab.atlassian.net/browse/SDESK-6837

Use the new optionLabel for MultiSelect component.

PR #4190, opened 2023-01-23
Dateline field implemented in React

thecalcc:
~~TODO: Implement patching of object fields~~


tomaskikutis:
create article with that field in angular then modify it in react and go back to angular and  

Dateline field PR
IDatelineValueOperational - add comment for day field format or if it's week days type them like that

PlacesService - (x) => geoNameToCity(x) 

data-layer.ts - Object patching is overriding objects that's why we need to add the full object to the diff for fields that are of object type

editor.tsx - add abort controller and function to cancel the request in searchOptions function

thecalcc:
> IDatelineValueOperational - add comment for day field format or if it's week days type them like that
We never actually fill that field so I will remove it.

thecalcc:
Use abort controller.

tomaskikutis (approved):

PR #4191, opened 2023-01-23
Fix incorect image directions when image is fliped and rotate

thecalcc:


tomaskikutis (approved):
don't forget to merge it at once with back-end changes

PR #4194, opened 2023-01-24
description_text field react

thecalcc:


scripts/apps/authoring-react/data-layer.ts
```diff
@@ -122,6 +123,28 @@ function getArticleContentProfile<T>(item: IArticle, fieldsAdapter: IFieldsAdapt
...
 
+        // TODO: write an upgrade script and remove hardcoding
+        // after angular based authoring is removed from the codebase
+        if (item.type === 'picture') {
+            const fieldConfig: IEditor3Config = {
+                editorFormat: [],
+                minLength: fakeScope.schema?.minlength,
+                maxLength: fakeScope.schema?.maxlength,
+                cleanPastedHtml: fakeScope.editor?.cleanPastedHTML,
+                singleLine: true,
```
tomaskikutis, line 134: is that single line in angular version?

scripts/apps/authoring-react/data-layer.ts
```diff
@@ -122,6 +123,28 @@ function getArticleContentProfile<T>(item: IArticle, fieldsAdapter: IFieldsAdapt
...
+                disallowedCharacters: [],
+            };
+
+            const description_text: IAuthoringFieldV2 = {
+                id: 'description_text',
```
tomaskikutis, line 139: Does it get written to the correct place? I'd think that without an adapter, it would get saved to IArticle>extra>description_text, when we want IArticle>description_text, no?

thecalcc: You're right, we'll need an adapter to write it to `article.description_text`

scripts/apps/authoring-react/data-layer.ts
thecalcc: I can't check it in sd-develop now, so I don't know.

tomaskikutis: why can't you check on sd-develop? Would it work to check locally then with angular based authoring?

thecalcc: Checked it on the local instance, we do allow multi line.

scripts/apps/authoring-react/data-layer.ts
```diff
@@ -122,6 +124,28 @@ function getArticleContentProfile<T>(item: IArticle, fieldsAdapter: IFieldsAdapt
             }
         }
 
+        // TODO: write an upgrade script and remove hardcoding
+        // after angular based authoring is removed from the codebase
+        if (item.type === 'picture') {
```
tomaskikutis, line 129: is it for picture only or all media including audio/video?

tomaskikutis: I mean in old authoring

thecalcc: Checked, we support audio and video too.

scripts/apps/authoring-react/data-layer.ts
```diff
@@ -122,6 +124,28 @@ function getArticleContentProfile<T>(item: IArticle, fieldsAdapter: IFieldsAdapt
...
+            };
+
+            const description_text: IAuthoringFieldV2 = {
+                id: DESCRIPTION_TEXT_FIELD_ID,
+                name: gettext('Description'),
+                fieldType: 'editor3',
+                fieldConfig,
+            };
+
+            contentFields = contentFields.set(description_text.id, description_text);
```
tomaskikutis, line 146: looks like config is defined twice, here and in the adapter. Which one gets used in the end? I'd think that the adapter one., Which means that maybe you should be able to do

```
contentFields = contentFields.set(description_text.id, null);
```

scripts/apps/authoring-react/data-layer.ts
thecalcc: Add 'graphic'

scripts/apps/authoring-react/data-layer.ts
```diff
@@ -69,6 +69,8 @@ function getArticleContentProfile<T>(item: IArticle, fieldsAdapter: IFieldsAdapt
 
                     return result;
                 })
+                .filter((t) => t.editorItem != null)
```
tomaskikutis, line 72: is that your local fix :upside_down_face:

scripts/apps/authoring-react/data-layer.ts
thecalcc: After looking at the code a bit I found that we consider 'media' type to be only `picture`, `video` and `audio` so do you think I should still add `graphic` type in the condition?
image

tomaskikutis (approved):

PR #4197, opened 2023-02-02
integrate full width toggling UI from menu into extensions

tomaskikutis:
SDESK-6800

image

### DIFF
--- a/scripts/apps/workspace/directives/WorkspaceSidenavDirective.ts
+++ b/scripts/apps/workspace/directives/WorkspaceSidenavDirective.ts
@@ -45,7 +45,7 @@ export function WorkspaceSidenavDirective(superdeskFlags, Keys,
 
                 if (fullWidthConfig.enabled) {
                     if (fullWidthConfig.allowed) {
-                        fullWidthConfig.onToggle();
+                        fullWidthConfig.onToggle(!scope.fullWidthEnabled);
                     }
                 } else {
                     // eslint-disable-next-line no-lonely-if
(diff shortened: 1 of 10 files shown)


scripts/core/superdesk-api.d.ts
```diff
@@ -790,10 +790,15 @@ declare module 'superdesk-api' {
 
     // PAGE
 
+    /**
+     * When not enabled, existing monitoring hiding functionality has to work.
+     */
+    export type IFullWidthPageCapabilityConfiguration = {enabled: false} | {enabled: true; allowed: false} | {enabled: true; allowed: true; onToggle: () => void};
```
petrjasek, line 796: what's the difference between enabled and allowed? might be also useful to get the full width state in the `onToggle` callback so the component knows how it changed

tomaskikutis: I've clarified the difference in the interface comments

scripts/apps/workspace/directives/WorkspaceSidenavDirective.ts
```diff
@@ -41,11 +41,20 @@ export function WorkspaceSidenavDirective(superdeskFlags, Keys,
...
+                    }
                 } else {
-                    superdeskFlags.flags.hideMonitoring = false;
+                    // eslint-disable-next-line no-lonely-if
+                    if (superdeskFlags.flags.authoring && state) {
+                        e.preventDefault();
+                        superdeskFlags.flags.hideMonitoring = !superdeskFlags.flags.hideMonitoring;
+                    } else {
+                        superdeskFlags.flags.hideMonitoring = false;
+                    }
```
thecalcc, line 57: Asignment on line 54 always evaluates to `false`.

tomaskikutis: How come? It would, if I was checking `superdeskFlags.flags.hideMonitoring` in the condition on line 52, but I'm checking for a different flag

thecalcc (approved):

PR #4198, opened 2023-02-06
Show modal if no highlights are configured & small improvements

thecalcc:


scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -185,16 +184,20 @@ const getExportModal = (
...
-                        article={getItem()}
-                        closeModal={closeModal}
-                    />
-                );
-            })
-        ),
+        onTrigger: () => {
+            sdApi.highlights.fetchHighlights().then((res) => {
+                if ((res._items.length ?? 0) === 0) {
+                    ui.alert(gettext('Highlights aren\'t configured on this instance.'));
```
tomaskikutis, line 190: Keep in mind that `an instance` doesn't mean anything for end users. For them it's THE Superdesk. They have only one, thus it'd be better to avoid such phrasing. I wouldn't also call it an instance configuration even technically same as you wouldn't call it an instance configuration if no articles have been created yet.

I'd use `No highlights have been created yet.`

tomaskikutis (approved):

scripts/core/ui-utils.tsx
```diff
@@ -2,11 +2,23 @@ import {showModal} from '@superdesk/common';
...
+| 'bottom-left'
+| 'bottom-right';
+
 export const ui = {
-    alert: (message: string) => (
+    alert: (message: string, position: modalPosition = 'top') => (
         showModal(({closeModal}) => {
             return (
                 <Modal
+                    position={position}
```
tomaskikutis, line 21: it'd be better to simply hardcode this to top. An alert is supposed to be one/two sentences, so no other positioning would be desired

tomaskikutis (approved):

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -185,16 +184,20 @@ const getExportModal = (
...
-                    <HighlightsModal
-                        article={getItem()}
-                        closeModal={closeModal}
-                    />
-                );
-            })
-        ),
+        onTrigger: () => {
+            sdApi.highlights.fetchHighlights().then((res) => {
+                if ((res._items.length ?? 0) === 0) {
```
tomaskikutis, line 189: are items `undefined` if there are no highlihgts? I think it should still be an empty array, thus applying a default isn't needed

tomaskikutis (approved):

PR #4203, opened 2023-02-14
Proofreading themes

dzonidoo:
SDESK-6253


tomaskikutis:
undo package-lock.json changes
page refresh shouldn't be needed to theme update to take effect
slugline font size is affected when it should not be
make `textColor` a standalone function (move out of the class)
uiThemeFontSize - move the function up the component tree, only pass REM value to editor3component
remove all usages of `any` types

TODO: check proofreading-theme-modal file again when lint issues are fixed

use spacer instead of custom CSS in proofreading-modal.scss

tomaskikutis:
don't forget: check if global themes work in broadcasting module

tomaskikutis (commented):
revert package lock
scripts/apps/authoring-react/toolbar/proofreading-modal.scss -> /proofreading.scss
import {showModal} from '@superdesk/common/dist/src/ui/show-modal'; -> from '@superdesk/common'
proofreadEnable -> proofreadingEnabled
uiThemeFontSize -> getUiThemeFontSize
else if (value === 'medium') -> add if ('large')
themeEnabled -> themingEnabled

const uiThemeFontSize = (value: 'small' | 'medium' | 'large') => {
    if (value === 'small') {
        return '1.4rem';
    } else if (value === 'medium') {
        return '1.6rem';
    } else if (value === 'large') {
        return '1.8rem';
    } else {
        assertNever(value);
    }
};

showThemeConfigModal -> private showThemeConfigModal

themeShadow -> themeColorSecondary

revert whitespace - scripts/apps/authoring/views/authoring.html
revert whitespace - scripts/apps/authoring/views/theme-select.html

'Configure Editor themes' -> use gettext

ProofreadingPanel -> to separate file
ThemeSelector -> to separate file

Fix class names - only use classnames for themes that are defined in a separate themes.scss file

uiTheme is sometimes optional - go over all usages and double check if we account for `undefined` - also correct interfaces and make it optional if it's possible that it will be undefined or null. Make it mandatory and possibly undefined when needed: uiTheme: IAuthoringSectionTheme | undefined;

backgroundColor: themeApplies && this.props.uiTheme.backgroundColor <-- use a ternary

Add a comment:
/**
* There is global theming and article-edit specific.
* When inside article-edit we always pass a theme,
* otherwise we want to use `Editor3-single-line-style` for global theme styling.
*/
'Editor3-single-line-style': this.props.singleLine === true || this.props.uiTheme == null,

StyleButton -> remove propTypes add an interface

tomaskikutis:
## left from previous review
revert package lock
revert whitespace - scripts/apps/authoring/views/authoring.html
revert whitespace - scripts/apps/authoring/views/theme-select.html

## newly found issues

remove dynamic classname
badgeColor: 'gray' | 'green';

availableThemes -> add type for the array
fontOptions -> add type for the array
ThemeSelector -> rename to BackgroundColorSelector
IFontSizeOptions -> rename to singular
onToggle: any; -> remove "any"
ThemeSelector -> type of value has to be the same as type of item inside onChange

tomaskikutis:
Trusting Nikola that all is working

PR #4204, opened 2023-02-17
Fix : The related item in the item preview is not updated [SDESK-6830]

devketanpro:


tomaskikutis:
It's a bit dangerous to fire a network request on any property change. Can you throttle(use lodash) it? First one should start immediately and subsequent ones should wait 1 second.

petrjasek (approved):

PR #4207, opened 2023-02-17
Link the related articles showing up in the preview of the main articles to the corresponding articles in the Belga archive [SDBELGA-711]

devketanpro:


tomaskikutis (changes requested):
Customer-specific changes should go to a customer repository, not to core. I'll look into the requirements next week and will propose a cleaner approach.

tomaskikutis:
@devketanpro I've talked to Petr and we decided to do it in 2 steps. First we provide a quick fix to avoid it crashing - for that we need to disable opening of related items that are from belga archive. It would be good to have a more generic condition. We have `externalsource` in `IArticle['_type']` maybe that would work? Could you open a PR for that?

The second step we'll do for another release sometime later.

PR #4208, opened 2023-02-21
Search across translations

thecalcc:
SDESK-6727

### DIFF
--- a/scripts/apps/vocabularies/controllers/VocabularyConfigController.ts
+++ b/scripts/apps/vocabularies/controllers/VocabularyConfigController.ts
@@ -17,7 +17,7 @@ export interface IScope extends IDirectiveScope<void> {
     tags: IVocabulary['tags'];
     loading: boolean;
     mediaTypes: object;
-    localQuery: string;
+    searchString: string;
     openVocabulary(vocabulary: IVocabulary): void;
     downloadVocabulary(vocabulary: IVocabulary): void;
     uploadConfig(): void;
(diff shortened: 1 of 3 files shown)


scripts/apps/vocabularies/controllers/VocabularyConfigController.ts
```diff
@@ -11,11 +11,13 @@ function getOther() {
 }
 
 export interface IScope extends IDirectiveScope<void> {
+    filterVocab(vocabulary: IVocabulary): boolean;
     vocabularies: Array<IVocabulary>;
     vocabulary: any;
     tags: IVocabulary['tags'];
     loading: boolean;
     mediaTypes: object;
+    localQuery: string;
```
tomaskikutis, line 20: `localQuery` was a bad suggestion, since it's no longer limited to a template but is now also used in a directive. Let's call it something that's easy to grep, how about `vocabularySearchString` or something similar

scripts/apps/vocabularies/controllers/VocabularyConfigController.ts
```diff
@@ -140,6 +142,22 @@ export function VocabularyConfigController($scope: IScope, $route, $routeParams,
     $scope.existsVocabulariesForTag = (currentTag: IVocabularyTag, tab: string) =>
         ($scope.vocabularies || []).some((vocabulary) => checkTag(vocabulary, currentTag, tab));
 
+    /**
+     * Checks if a vocabulary's display_name matches the search
+     * query or if it matches a translation of that vocabulary.
+     */
+    $scope.filterVocab = function(vocabulary: IVocabulary): boolean {
+        const translationMach = Object.values(vocabulary?.translations?.display_name ?? [])
+            .find((tr: string) => tr.toLowerCase().indexOf($scope.localQuery) != -1);
```
tomaskikutis, line 151: avoid abbreviations, `tr` > `translation`

`!= -1`
should use double equals

tomaskikutis (approved):

PR #4215, opened 2023-03-01
add embeds as external in case association for ID is missing

tomaskikutis:
SDESK-6834

The issue was caused by incorrect data migration script (EWTN)


thecalcc (approved):
Looks good

PR #4217, opened 2023-03-03
Fix : Can’t use 3 dot menu from list view to send a locked article to other desk/stage [SDESK-6809]

devketanpro:


scripts/apps/ingest/services/SendService.ts
```diff
@@ -212,10 +212,13 @@ export function SendService(
...
                 });
-        } else if (!item.lock_user) {
+        } else {
             return api.save('move', {}, {task: data, allPackageItems: config.sendAllPackageItems}, item)
                 .then((_item) => {
                     $rootScope.$broadcast('item:update', {item: _item});
+                    if (item.lock_user) {
+                        getAuthoringWorkspace().close();
+                    }
                     if (config.open) {
```
tomaskikutis, line 222: what does config.open mean? When is it true? Would it be compatible with you closing the item in the line before?

scripts/apps/ingest/services/SendService.ts
```diff
@@ -212,10 +212,13 @@ export function SendService(
                     $rootScope.$broadcast('item:update', {item: _item});
                     return _item;
                 });
-        } else if (!item.lock_user) {
+        } else {
             return api.save('move', {}, {task: data, allPackageItems: config.sendAllPackageItems}, item)
                 .then((_item) => {
                     $rootScope.$broadcast('item:update', {item: _item});
+                    if (item.lock_user) {
```
tomaskikutis, line 219: you need to also check if the same item is open in authoring. Otherwise it would close authoring with article A open when you send article B to another stage when it is locked, but not open in authoring.

scripts/apps/ingest/services/SendService.ts
devketanpro: `config.open` means , if user select option send and open, so in this case first we move that item then open it.

Superdesk - Monitoring - desk testing

scripts/apps/ingest/services/SendService.ts
```diff
@@ -216,7 +216,7 @@ export function SendService(
             return api.save('move', {}, {task: data, allPackageItems: config.sendAllPackageItems}, item)
                 .then((_item) => {
                     $rootScope.$broadcast('item:update', {item: _item});
-                    if (item.lock_user) {
+                    if (item?.lock_user && getAuthoringWorkspace().getItem()?._id === item?._id) {
```
tomaskikutis, line 219: Don't use optional chaining where it's not needed. Is is possible that `item` can be `null`? I see optional chaining wasn't used before so I'd guess that not. In that case you also don't need it in ` item?._id`.

Also, don't use type casting in `if` conditions. Do convert it to boolean explicitly - `if (item.lock_user != null)`

tomaskikutis (approved):

PR #4219, opened 2023-03-06
When sending an item to another desk/stage the current stage of the item should be disabled in the send panel [SDESK-6814]

devketanpro:


tomaskikutis:
Can you describe the fix? Why wasn't `stage._id === item.task.stage` working? Looks logical to me - `item.task.stage` is the stage an article is located at any given moment.

devketanpro:
> Can you describe the fix? Why wasn't `stage._id === item.task.stage` working? Looks logical to me - `item.task.stage` is the stage an article is located at any given moment.

because we couldn't find item here, that's why this condition is not working.

PR #4221, opened 2023-03-10
Create a UI toggle for switching between old/new authoring

dzonidoo:
SDESK-6845


tomaskikutis:
closing for now because it requires a different approach

PR #4222, opened 2023-03-14
Template editing authoring react

thecalcc:


tomaskikutis:
* improve naming and comments in superdesk.api.d.ts
* getInlineToolbarActions - make optional
* AuthoringIntegrationWrapper - IProps - keep id mandatory
* hideToolbar - which one???
* remove hideSidebarAndHeader
* template editor modal - authoring react bridge might work if we add a scope.fn outside of _.defer call - directly inside `link` - if that works, move `isDirty` and other properties back to script TemplateEditorModal.ts

thecalcc:
The last commit has a mistake in it's description. It's supposed to be `v2` not `v1`

tomaskikutis:
* improve toolbar naming - primary/secondary
* use inline style instead of .sd-authoring-bridge

PR #4224, opened 2023-03-17
Body footer

thecalcc:
SDESK-6259


tomaskikutis:
Does it now support "unsetting" a vocabulary for predefined snippets in case someone wanted to disable it?

PR #4230, opened 2023-04-04
DST fix date calculations wrong

marwoodandrew:


tomaskikutis:
Why are you droping typescript types? Could you install a matching version instead?

MarkLark86:
@tomaskikutis In newer version of Moment lib, they have incorporated the types into the library. Therefore adding the types package is no longer required

MarkLark86 (approved):

PR #4241, opened 2023-05-04
Read only and archive from

thecalcc:


scripts/apps/authoring-react/authoring-angular-integration.tsx
```diff
@@ -144,11 +144,46 @@ function getInlineToolbarActions(options: IExposedFromAuthoring<IArticle>): IAut
...
 
+    const getReadOnlyAndArchivedFrom = (): Array<ITopBarWidget<IArticle>> => {
+        const actions: Array<ITopBarWidget<IArticle>> = [];
+
+        if (item._type === 'archived') {
+            actions.push({
+                group: 'start',
+                availableOffline: false,
+                component: () => (
+                    <span>
```
tomaskikutis, line 155: it is possible for this string to get very long. Ellipsis should appear in case it no longer fits in the toolbar - but not sooner.  It should work in `DeskAndStage` component - check it out how it works there.

scripts/apps/authoring-react/authoring-angular-integration.tsx
```diff
@@ -144,11 +144,46 @@ function getInlineToolbarActions(options: IExposedFromAuthoring<IArticle>): IAut
         availableOffline: true,
     });
 
+    const getReadOnlyAndArchivedFrom = (): Array<ITopBarWidget<IArticle>> => {
```
tomaskikutis, line 147: what about states at lines 327-370? It should have read-only archived indicators too. It would probably be good to have some `commonActions` that would be everywhere.

tomaskikutis:
fix this one and don't forget to reply to my comments

scripts/apps/authoring-react/authoring-angular-integration.tsx
thecalcc: Yes, it does work just fine for that so we can reuse it there.

scripts/apps/authoring-react/authoring-angular-integration.tsx
thecalcc: I plant to address those lines in the next bulltet point of the ticket which is about all of these other cases.

tomaskikutis:
after changing base, it seems there are more changes that are coming than read-only/archived stuff. If I merge latest stuff  to post big merge it should resolve I believe. I'll try doing that

PR #4245, opened 2023-05-11
Fix : Publish button not activating [SDESK-6902]

devketanpro:
`otherDeskSelected` is works according to send_to functionality.

### DIFF
--- a/scripts/core/interactive-article-actions-panel/utils/get-initial-destination.ts
+++ b/scripts/core/interactive-article-actions-panel/utils/get-initial-destination.ts
@@ -18,11 +18,11 @@ export function getInitialDestination(
     const currentDeskId = sdApi.desks.getCurrentDeskId();
 
     let destinationDesk: string = (() => {
-        if (lastDestination?.type === 'desk' && lastDestination.desk != null) {
-            return lastDestination.desk;
-        } else if (currentDeskId != null) {
+        if (currentDeskId != null) {
             return currentDeskId;
-        } else if (items.length === 1 && items[0].task?.desk != null) {
+        }else if (lastDestination?.type === 'desk' && lastDestination.desk != null) {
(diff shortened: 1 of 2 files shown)


tomaskikutis:
@thecalcc could you review this in depth when you have time?  It doesn't look right to me. Compare it to angular logic, figure out why wasn't it activating and evaluate if this is the right solution.

scripts/core/interactive-article-actions-panel/utils/get-initial-destination.ts
```diff
@@ -18,11 +18,11 @@ export function getInitialDestination(
...
     let destinationDesk: string = (() => {
-        if (lastDestination?.type === 'desk' && lastDestination.desk != null) {
-            return lastDestination.desk;
-        } else if (currentDeskId != null) {
+        if (currentDeskId != null) {
             return currentDeskId;
-        } else if (items.length === 1 && items[0].task?.desk != null) {
+        }else if (lastDestination?.type === 'desk' && lastDestination.desk != null) {
+            return lastDestination.desk;
+        }else if (items.length === 1 && items[0].task?.desk != null) {
```
thecalcc, line 25: Did you test the other use cases and make sure that these changes don't break anything else?

devketanpro: we've updated the functionality based on angular logic, thus when we open a **sendto and publish tab**, we receive the `currentDesk` instead of the `lastDestinationDesk`.

devketanpro: our `otherDeskSelected `logic fails due to this behaviour. now its works fine as expected.

thecalcc (approved):
All good

PR #4250, opened 2023-05-22
fix non wrapping category config in user preferences

petrjasek:
SDESK-6923


petrjasek:
before:
categories-before
after:
categories-after

fritzSF (approved):
OK, looks good.

tomaskikutis (approved):

PR #4252, opened 2023-05-24
improve table copy pasting into editor3

petrjasek:
when there is no editor instance, like from different instance or different window.

SDNTB-814


tomaskikutis (approved):

tomaskikutis (approved):

tomaskikutis:
> When copy&pasting between windows without editor instance

I forgot when is that the case and why wouldn't we be able to get editor3 instance?

petrjasek:
> I forgot when is that the case and why wouldn't we be able to get editor3 instance?

it's when you open the item for editing in a new tab, copy something there, then you got back to the other tab and paste it

petrjasek:
yep that's right

PR #4258, opened 2023-05-31
fixed filtering inside of more templates

dzonidoo:


tomaskikutis:
> (from ticket:) Also looks like scrolling doesn’t work inside that selection dropdown.

@dzonidoo did you resolve this too?

tomaskikutis:
@dzonidoo how are notifications now?

PR #4260, opened 2023-06-01
fixed panel footer UI

dzonidoo:
SDESK-6942


tomaskikutis (approved):
I suspect the issue might have been markup of `PanelFooter` changing on ui-framework side

tomaskikutis (approved):
I suspect the issue might have been markup of `PanelFooter` changing on ui-framework side

dzonidoo:
I do not think like that... only the absence of the ButtonGroup component or style of buttons is a problem in this case. I chose to wrap button

tomaskikutis:
I checked an older commit and the issue was still there, so you might be correct. One more thing - check usages for when `markupV2` is `true` and see if your change still looks good in that case.

dzonidoo:
I checked and it works

tomaskikutis:
ok thanks

PR #4265, opened 2023-06-15
Improve publish tab so you have the ability to save values to the article

thecalcc:
SDESK-6961

### DIFF
--- a/scripts/core/interactive-article-actions-panel/actions/publish-tab.tsx
+++ b/scripts/core/interactive-article-actions-panel/actions/publish-tab.tsx
@@ -35,6 +35,7 @@ interface IProps {
     handleUnsavedChanges(): Promise<IArticle>;
     markupV2: boolean;
     onError: (error: IPanelError) => void;
+    setSave: (item: IArticle) => void;
 }
 
 interface IState {
@@ -53,9 +54,9 @@ export class PublishTab extends React.PureComponent<IProps, IState> {
             selectedDestination: getInitialDestination([this.props.item], false),
             publishingDateOptions: getInitialPublishingDateOptions([props.item]),
             publishingTarget: {
-                target_subscribers: [],
-                target_regions: [],
-                target_types: [],
+                target_subscribers: this.props.item.target_subscribers,
+                target_regions: this.props.item.target_regions,
+                target_types: this.props.item.target_regions,
             },
             subscribers: null,
         };
@@ -159,13 +160,26 @@ export class PublishTab extends React.PureComponent<IProps, IState> {
                     <div style={style}>
                         <div>
                             {
-                                publishFromEnabled && (
+                                true && (
                                     <ToggleBox title={gettext('From')} initiallyOpen>
                                         <DestinationSelect
                                             value={this.state.selectedDestination}
                                             onChange={(value) => {
                                                 this.setState({
                                                     selectedDestination: value,
+                                                }, () => {
+                                                    const dest = this.state.selectedDestination;
+
+                                                    if (dest.type === 'desk') {
+                                                        this.props.setSave({
+                                                            ...this.props.item,
+                                                            task: {
+                                                                ...(this.props.item.task ?? {}),
+                                                                desk: dest.desk,
+                                                                stage: dest.stage,
+                                                            },
+                                                        });
+                                                    }
                                                 });
                                             }}
                                             includePersonalSpace={false}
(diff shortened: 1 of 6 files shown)


scripts/core/interactive-article-actions-panel/actions/publish-tab.tsx
```diff
@@ -53,9 +54,9 @@ export class PublishTab extends React.PureComponent<IProps, IState> {
             selectedDestination: getInitialDestination([this.props.item], false),
             publishingDateOptions: getInitialPublishingDateOptions([props.item]),
             publishingTarget: {
-                target_subscribers: [],
-                target_regions: [],
-                target_types: [],
+                target_subscribers: this.props.item.target_subscribers,
```
tomaskikutis, line 57: you need to consider that `target_subscribers` and others might be `null` and use initial value in that case

scripts/core/interactive-article-actions-panel/actions/publish-tab.tsx
```diff
@@ -159,13 +160,26 @@ export class PublishTab extends React.PureComponent<IProps, IState> {
                     <div style={style}>
                         <div>
                             {
-                                publishFromEnabled && (
+                                true && (
```
tomaskikutis, line 163: did you review your code yourself before opening the PR? :upside_down_face:

scripts/core/interactive-article-actions-panel/actions/publish-tab.tsx
```diff
@@ -159,13 +160,26 @@ export class PublishTab extends React.PureComponent<IProps, IState> {
...
                                     <ToggleBox title={gettext('From')} initiallyOpen>
                                         <DestinationSelect
                                             value={this.state.selectedDestination}
                                             onChange={(value) => {
                                                 this.setState({
                                                     selectedDestination: value,
+                                                }, () => {
```
tomaskikutis, line 170: It would be better to decouple UI from this new saving functionality. For example, could you do the saving in `componentDidUpdate`?

scripts/core/interactive-article-actions-panel/subcomponents/destination-select.tsx
```diff
@@ -59,11 +59,11 @@ export class DestinationSelect extends React.PureComponent<IProps> {
                         })()}
                         items={destinations}
                         onChange={(val) => {
-                            if (val.id === PERSONAL_SPACE) {
+                            if (val != null && val.id === PERSONAL_SPACE) {
```
tomaskikutis, line 62: how about you do `if (val == null) { return; }` at the top and keep the remaining code unchanged? Antoher issue is that if there was a value, and user removed the value and closed the panel - it would be expected the value NOT to be there if he reopens the pane.

scripts/core/interactive-article-actions-panel/actions/publish-tab.tsx
```diff
@@ -35,6 +35,7 @@ interface IProps {
     handleUnsavedChanges(): Promise<IArticle>;
     markupV2: boolean;
     onError: (error: IPanelError) => void;
+    setSave: (item: IArticle) => void;
```
tomaskikutis, line 38: The name doesn't look good to me. How about we call it `onChange` or `onDataChange`. What happens after the change is not known by this component so it's better not to use naming that assumes it. Even if the assumption is correct at the moment.

scripts/core/interactive-article-actions-panel/actions/publish-tab.tsx
thecalcc: Yes, missed this one

scripts/core/interactive-article-actions-panel/subcomponents/destination-select.tsx
thecalcc: Yes good point on one hand about the `if` handling. Another thing is that originally in sd-develop you can't have an empty value there. Check out how it works on that instance and you'll see.

tomaskikutis: if UI doesn't allow empty value then all good.

scripts/core/interactive-article-actions-panel/actions/publish-tab.tsx
thecalcc: That would be unnecessary updates. If we do it in the componentDidUpdate we'd have to go through all values no matter which one is changed. If we don't change it and handle it as I've done that wouldn't be the case. What I mean is that over all it seems like it would bring unnecessary complexity.

tomaskikutis: > That would bring unnecessary updates

if you add an `if` to check for it, there won't be unnecessary updates.

It's computationally cheap to run a few extra loops, so if we can trade it for better code structure(decoupling UI from saving logic) I'd do that. See if it's possible to do it and keep code clean and readable. If not, I'm open to leaving it as is.

tomaskikutis (approved):

tomaskikutis:
CI has to pass before merging

PR #4267, opened 2023-06-19
Make translations load synchronously

thecalcc:
SDESK-6971


scripts/translations.js
```diff
@@ -0,0 +1,52 @@
...
+    window.translations = translations;
+    window.pluralForms = pluralForms;
+}
+
+const language = getUserInterfaceLanguage();
+
+if (language === 'en') {
+    const translations = {'': {'language': 'en', 'plural-forms': 'nplurals=2; plural=(n != 1);'}};
+
+    window.language = translations['']['language'];
```
tomaskikutis, line 42: I was commenting in another place that it would be better to have 1 variable on window that would have more translations related stuff inside, but now I'm looking at it again and am thinking that since `language` and `pluralForms` are computed from `translations`, what's even the point to have 3 of them instead only having translations on window, and accessing the other 2 in another bundle?

tomaskikutis (approved):
all good, fix CI and merge it

PR #4269, opened 2023-06-22
add missing translation to transifex

petrjasek:
SDCP-704

### DIFF
--- a/scripts/apps/search/components/fields/translations.tsx
+++ b/scripts/apps/search/components/fields/translations.tsx
@@ -56,7 +56,7 @@ export class Translations extends React.PureComponent<IProps> {
                             this.renderOriginalArticle(event.target as HTMLElement);
                         }}
                     >
-                        {gettext('translation')}
+                        {gettext('Translation')}
                     </button>
                 )}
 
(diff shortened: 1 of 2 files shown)


scripts/apps/search/components/fields/translations.tsx
```diff
@@ -70,7 +70,7 @@ export class Translations extends React.PureComponent<IProps> {
                     >
                         {'('}<b>{this.props.item.translations.length}</b>{')'}
                         {' '}
-                        {gettextPlural(this.props.item.translations.length, 'translation', 'translations')}
+                        {this.props.item.translations.length === 1 ? gettext('Translation') : gettext('Translations')}
```
tomaskikutis, line 73: you should keep `gettextPlural` - other languages have more plural forms than one that depend on item count (in this case `translations.length`) and when the script is doing string extraction, it will mark it as plural. As a result translators would have to set 3-4 or how many plural forms there are in their language.

petrjasek: yep problem is that it doesn't work atm, those parts don't show up on transifex

tomaskikutis: what about .pot file?

tomaskikutis (approved):

scripts/apps/search/components/fields/translations.tsx
petrjasek: they are in the `.pot`, that part seems ok. will investigate it further this is rather temporary fix

PR #4278, opened 2023-06-28
Z index property

dzonidoo:
SDESK-6952


tomaskikutis (approved):
don't merge it yet

tomaskikutis:
Let's wait until 2.7 is released and only merge it afterwards

petrjasek:
do we need this one still @dzonidoo ?

tomaskikutis:
We do. At the time we were planning to do a release so I didn't want to push code that is likely to break things. Do you think Petr now is a good time? Do we have any releases planned soon?

petrjasek:
now it should be fine, there is already a branch for 2.8 so there is time before there will be any release from develop

tomaskikutis:
@dzonidoo then update the branch and merge it

PR #4280, opened 2023-06-29
fixed overlap

dzonidoo:
SDESK-6940

### DIFF
--- a/scripts/apps/authoring/attachments/AttachmentsWidget.tsx
+++ b/scripts/apps/authoring/attachments/AttachmentsWidget.tsx
@@ -13,7 +13,6 @@ import {withAttachments} from './AttachmentsWrapper';
 interface IProps extends IAttachmentsWrapperProps {
     updateItem(updates: Partial<IArticle>): void;
     readOnly: boolean;
-    isWidget: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


tomaskikutis:
since `is-widget` will now be always `true` can you remove the prop entirely and any code that handled the situation when it is `false`?

dzonidoo:
Yep... I'll try to do that...

MarkLark86:
> since `is-widget` will now be always `true` can you remove the prop entirely and any code that handled the situation when it is `false`?

Might need to check SAMS extension, as it looks like this is used there: https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/extensions/sams/src/apps/samsAttachmentsWidget.tsx#L158

tomaskikutis:
[image](https://github.com/superdesk/superdesk-client-core/assets/2076953/e651bbca-ac95-4b74-854b-dc546508c64c)
[image](https://github.com/superdesk/superdesk-client-core/assets/2076953/dfccd71a-2990-44cc-a264-4b97db336c33)


We missed some usages @dzonidoo. It is used in inline mode in authoring so we can't remove that property. We'll need to close this PR and fix overlay issues in a different way. `isWidget` is `false` for some reason when the widget is pinned. We need to address this instead. Either to make it `true`, or modify the condition in the HTML.

PR #4282, opened 2023-06-30
Fix masterdesk users view

thecalcc:


scripts/apps/master-desk/components/UsersComponent.tsx
```diff
@@ -61,7 +63,8 @@ class UsersComponent extends React.Component<IProps, IState> {
 
             this.setState({
                 roles: roles._items,
-                usersByRole: users._items,
+                usersByRole: users._items.filter((item) => item.role != null),
```
tomaskikutis, line 66: This is a good use case for lodash's `partition` method.

User count will be < 100 in most cases so iterating the same thing twice doesn't hurt in terms of performance, but it's a good habit to do it more efficiently when it doesn't require significantly more effort.

---

Another thing - it'd better to call it `usersWithRole`(especially when you have a opposite value called `usersWithoutRole`). When naming convention of `{entity}By{Field}` is used, I expect an object/map. In case of `usersByRole`, it'd expect `{[roleId: string]: IUser}`

scripts/apps/master-desk/components/UsersComponent.tsx
```diff
@@ -82,38 +85,74 @@ class UsersComponent extends React.Component<IProps, IState> {
...
+                        const usersWithoutRole = this.getUsersWithoutRole(desk);
+
+                        return (
+                            <div className="sd-board" key={desk._id}>
+                                <div className="sd-board__header">
+                                    <h3 className="sd-board__header-title">{desk.name}</h3>
+                                </div>
+                                <div className="sd-board__content sd-padding-t--1">
+                                    {
+                                        usersWithoutRole.length && (
```
tomaskikutis, line 118: `usersWithoutRole.length` is not a boolean ;)

tomaskikutis (approved):

PR #4284, opened 2023-07-04
Article type

thecalcc:


tomaskikutis:
`article.type` can not be missing, so instead we should solve the issue that causes it to be missing

tomaskikutis (approved):

PR #4298, opened 2023-07-20
import ResizeObserverComponent from ui-framework

dzonidoo:


tomaskikutis (approved):
test it before merging

PR #4302, opened 2023-07-31
Fix datetime required stars

thecalcc:
SDESK-6904

### DIFF
--- a/scripts/apps/authoring-react/authoring-section/get-field-container.tsx
+++ b/scripts/apps/authoring-react/authoring-section/get-field-container.tsx
@@ -45,7 +45,10 @@ export function getFieldContainer(
                             style={{minHeight: 'auto'}}
                         >
                             <Spacer h gap="8" noGrow>
-                                <span>{field.name}</span>
+                                <span>
+                                    {field.fieldConfig.required && <span style={{color: '#e41b21', fontSize: '1.2rem'}}>* </span>}
+                                    {field.name}
+                                </span>
                                 <span>{toggle}</span>
                             </Spacer>
                         </span>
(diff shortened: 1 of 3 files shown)


scripts/apps/authoring-react/authoring-section/get-field-container.tsx
```diff
@@ -45,7 +45,12 @@ export function getFieldContainer(
                             style={{minHeight: 'auto'}}
                         >
                             <Spacer h gap="8" noGrow>
-                                <span>{field.name}</span>
+                                <span>
+                                    {field.fieldConfig.required && (
+                                        <span style={{color: '#e41b21', fontSize: '1.2rem'}}>* </span>
```
fritzSF, line 50: Please don't use any hardcoded values for colors and font-sizes. CSS variables should be used instead.

thecalcc: Which variable should be used for this color?

tomaskikutis (approved):

scripts/apps/authoring-react/authoring-section/get-field-container.tsx
tomaskikutis: @fritzSF did you find the variable yet or you want us to use the hardcoded one? :laughing:

Actually we did try to find the variable in the beginning as well, but couldn't. I think we need some cleanup on colors so we have less of them and they're more descriptive like `color-text-error` vs `red-600`

tomaskikutis: never mind, @thecalcc says you replied to him personally

tomaskikutis (approved):

PR #4303, opened 2023-08-01
fixed mark for highlight data

dzonidoo:
SDESK-6948

### DIFF
--- a/e2e/server/requirements.in
+++ b/e2e/server/requirements.in
@@ -1,3 +1,3 @@
 honcho
 gunicorn
-git+https://github.com/superdesk/superdesk-core.git@hotfix/2.6.5#egg=Superdesk-Core
+git+https://github.com/superdesk/superdesk-core.git@develop#egg=Superdesk-Core
--- a/e2e/server/requirements.txt
+++ b/e2e/server/requirements.txt
@@ -10,7 +10,7 @@ arrow==0.13.0
     # via
     #   eve-elastic
     #   superdesk-core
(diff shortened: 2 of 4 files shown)


tomaskikutis:
what's wrong with CI?

tomaskikutis:
I will take a look at e2e issue

tomaskikutis:
```
INFO: pip is looking at multiple versions of superdesk-core to determine which version is compatible with other requirements. This could take a while.
  ERROR: Could not find a version that satisfies the requirement celery[redis]<5.3,>=5.2.7 (from superdesk-core)
```

@petrjasek can you update e2e to work against develop for e2e to pass? It's failing because old server is still expecting highlights to be a string. I tried updating it myself using `pip-compile` but it didn't work

tomaskikutis:
@dzonidoo update your branch from develop and see if e2e is good now

tomaskikutis:
@petrjasek I was running e2e against `hotfix/2.6.5` branch as you specified, but the server still complains about the array and expects a string. Are you sure it's a correct version to use?

```
# GET /api/marked_for_highlights
{
    "_status": "ERR",
    "_issues": {
        "highlights": "must be of string type"
    },
    "_error": {
        "code": 400,
        "message": "Insertion failure: 1 document(s) contain(s) error(s)"
    }
}
```

petrjasek:
@tomaskikutis that wouldn't be atm in 2.6, so either develop or release/2.7

tomaskikutis (approved):

PR #4309, opened 2023-08-16
make sure description text is plain text without html

petrjasek:
SDCP-710


tomaskikutis:
at the moment, we have `singleLine` field responsible for both - single line and plain text. If it's set as single line - it will always be plain text. Do you need description to be multi-line and still plain text?

petrjasek:
yep it needs to be multi line

tomaskikutis (approved):

PR #4312, opened 2023-08-22
remove toggle button theme from header

dzonidoo:
SDESK-7037


scripts/apps/users/directives/UserPreferencesDirective.ts
```diff
@@ -44,7 +44,7 @@ export function UserPreferencesDirective(
             const body = angular.element('body');
 
             scope.activeNavigation = null;
-            scope.activeTheme = localStorage.getItem('theme');
+            scope.activeTheme = localStorage.getItem('theme') !== '' ? localStorage.getItem('theme') : 'light-ui';
```
tomaskikutis, line 47: It's not ideal to read local storage twice. I'd be nicer to do it like this:

```
scope.activeTheme = localStorage.getItem('theme');

if (scope.activeTheme === '') {
    scope.activeTheme = 'light-ui';
}
```

It would also be better to use a constant instead of inlining `light-ui`.

tomaskikutis:
can you use a constant? :)

scripts/apps/users/directives/UserPreferencesDirective.ts
```diff
@@ -4,6 +4,8 @@ import {gettext} from 'core/utils';
 import {appConfig, getUserInterfaceLanguage} from 'appConfig';
 import {applyDefault} from 'core/helpers/typescript-helpers';
 
+const theme = 'light-ui';
```
tomaskikutis, line 7: use capital letters for a variable name

tomaskikutis (changes requested):
check the comment about variable naming

tomaskikutis (approved):

PR #4313, opened 2023-08-23
Switch to authoring react in broadcasting

thecalcc:


scripts/core/menu/authoring-switch.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    if (action === 'register') {
+        setAuthoringReact(true);
+
+        registerAuthoringReactWidgets();
+        registerAuthoringReactFields();
+    } else {
+        setAuthoringReact(false);
+        unregisterInternalExtension(authoringReactWidgetsExtension);
+        unregisterInternalExtension('authoring-react--fields');
```
tomaskikutis, line 39: can you extract this string into a constant too?

scripts/core/superdesk-api.d.ts
```diff
@@ -5,18 +5,18 @@ declare module 'superdesk-api' {
 
     type OrderedMap<K, V> = import('immutable').OrderedMap<K, V>;
```
tomaskikutis, line 7: looks like there are no changes here apart from whitespace, can you revert it then?

scripts/core/menu/authoring-switch.ts
```diff
@@ -0,0 +1,45 @@
+import {authoringReactEnabledUserSelection, extensions, setAuthoringReact} from 'appConfig';
+import {registerAuthoringReactFields} from 'apps/authoring-react/fields/register-fields';
+import {registerAuthoringReactWidgets, authoringReactWidgetsExtension} from 'apps/authoring-react/manage-widget-registration';
+import {unregisterInternalExtension} from 'core/helpers/register-internal-extension';
+import {trimStartExact} from 'core/helpers/utils';
+import {flatMap} from 'lodash';
+import ng from 'core/services/ng';
+
+export const switchAuthoring = (url: string) => {
```
tomaskikutis, line 9: `maybeEnableAuthoringReact` would be a better name. Or add a comment saying that in certain conditions we're enabling react based authoring

scripts/core/menu/menu.ts
```diff
@@ -286,15 +289,26 @@ angular.module('superdesk.core.menu', [
...
                         if (route && route.href) {
                             scope.activeMenuItemPath = getActiveMenuItemPath(route.href);
                         }
                     }
 
-                    scope.$on('$locationChangeStart', () => {
+                    switchAuthoring($location.absUrl());
+
+                    /**
+                     * Gets triggered after the location has been changed for the final time.
```
tomaskikutis, line 303: `final time` is a bit confusing because the URL only changes one time to begin with. That other function that is being called multiple times is about the intention to change the URL, but can as well cancel the change. I think it might even be more clear without the comment because we have two listeners for different things which are named pretty well.

scripts/core/menu/menu.ts
```diff
@@ -286,15 +289,26 @@ angular.module('superdesk.core.menu', [
...
+                     * Using this is consistent and assures you that `nextUrl` is the final url
+                     * you'll receive for the location change.
+                     */
+                    scope.$on('$locationChangeSuccess', (_, newUrl, oldRoute) => {
+                        switchAuthoring(newUrl);
+                    });
+
+                    scope.$on('$locationChangeStart', (_, a, b) => {
```
tomaskikutis, line 311: revert changes on this line that you ended up not using

scripts/core/datetime/datetime.ts
```diff
@@ -8,11 +8,11 @@ const ISO_DATE_FORMAT = 'YYYY-MM-DD';
 const ISO_WEEK_FORMAT = 'YYYY-W';
 const ISO_YEAR_FORMAT = 'YYYY';
 
-const LONG_FORMAT = appConfig.longDateFormat || 'LLL';
-const TIME_FORMAT = appConfig.shortTimeFormat || 'hh:mm';
-const DATE_FORMAT = appConfig.shortDateFormat || 'MM/DD';
-const WEEK_FORMAT = appConfig.shortWeekFormat || 'dddd, ' + TIME_FORMAT;
-const ARCHIVE_FORMAT = appConfig.ArchivedDateFormat || DATE_FORMAT;
+const LONG_FORMAT = appConfig?.longDateFormat || 'LLL';
```
tomaskikutis, line 11: that is a bit weird, appConfig should never be null. Try reproducing when it is and let's look at it together.

tomaskikutis:
also fix CI :)

tomaskikutis (approved):

PR #4314, opened 2023-08-23
Introuduce private custom field option

thecalcc:
SDESK-7043


scripts/core/superdesk-api.d.ts
```diff
@@ -3359,6 +3359,7 @@ declare module 'superdesk-api' {
         editorComponent: React.ComponentClass<IEditorComponentProps<IValueOperational, IConfig, IUserPreferences>>;
         previewComponent: React.ComponentType<IPreviewComponentProps<IValueOperational, IConfig>>;
 
+        private?: boolean;
```
tomaskikutis, line 3362: can you add a comment describing the use case?

tomaskikutis (approved):

PR #4317, opened 2023-08-30
fix e2e server requirements

petrjasek:


dzonidoo (approved):

petrjasek:
can't reproduce those failing e2e tests locally, any idea what might be wrong here @tomaskikutis ?

tomaskikutis:
No idea. Maybe chrome driver updated again? :laughing: There was one test that was randomly failing from time to time, but it's not it this time. I'll check it out.

petrjasek:
I don't know, have the same chromedriver version locally, but an updated chrome, will probably let it rest a bit

tomaskikutis:
image
I've found a clue in screenshots, list fails to load for some reason. I couldn't reproduce it, but will try to match versions on my local machine to what github uses

tomaskikutis:
I ran e2e with chrome 116, launched the server with docker, ran `docker-compose build --no-cache` to ensure latest server code, built client for production and still can't reproduce it. I'll added outputting of browser console errors, hopefully we can see something there.

PR #4326, opened 2023-09-13
Link shortcut and formatting options zindex fix

thecalcc:
SDNTB-832

### DIFF
--- a/scripts/apps/workspace/content/views/FormattingOptionsMultiSelect.tsx
+++ b/scripts/apps/workspace/content/views/FormattingOptionsMultiSelect.tsx
@@ -50,6 +50,7 @@ export class FormattingOptionsTreeSelect extends React.Component<IProps> {
                 label={gettext('Formatting options')}
                 labelHidden
                 inlineLabel
+                zIndex={2000}
             />
         );
     }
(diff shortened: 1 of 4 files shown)


scripts/apps/workspace/content/views/FormattingOptionsMultiSelect.tsx
```diff
@@ -50,6 +50,7 @@ export class FormattingOptionsTreeSelect extends React.Component<IProps> {
                 label={gettext('Formatting options')}
                 labelHidden
                 inlineLabel
+                zIndex={2000}
```
tomaskikutis, line 53: do 1050 or next one that works instead of jumping a thousand

tomaskikutis (approved):
verify also in the UI that after keybinding gets hit - the keybinding focus goes to link input

thecalcc:
> verify also in the UI that after keybinding gets hit - the keybinding focus goes to link input
@tomaskikutis 
That's not the case even if you click on it with your mouse. This is different because the style of this formatting option is different. Instead it's indicated that this is the formatting option which is active by the popup being displayed.

thecalcc:
@tomaskikutis you'll notice I added the `&& selectionState.isCollapsed() !== true` check. That's because we have to "disable" the possibility to activate the link option if nothing is selected. This is basically mimicking the normal functionality (when clicking on it with your mouse). If there's no text selected we disable the option to add a link.

PR #4327, opened 2023-09-14
Fix : missing translate in h3 attribute

devketanpro:


scripts/apps/users/views/user-preferences.html
```diff
@@ -121,7 +121,7 @@ <h3 id="article-defaults" class="sd-heading sd-text-align--left sd-text--sans sd
                 </li>
 
                 <li class="simple-list__item simple-list__item--stacked simple-list__item--justify-flex-start" ng-hide="profileConfig.category === false">
-                    <h3 id="prefered-categories" class="sd-heading sd-text-align--left sd-text--sans sd-heading--h3 sd-txt-transform--capitalize">{{ preferences['categories:preferred'].category }}</h3>
+                    <h3 id="prefered-categories" class="sd-heading sd-text-align--left sd-text--sans sd-heading--h3 sd-txt-transform--capitalize" translate>{{ preferences['categories:preferred'].category }}</h3>
```
petrjasek, line 124: here we shouldn't use it

scripts/apps/users/views/user-preferences.html
```diff
@@ -163,7 +163,7 @@ <h4 class="sd-heading sd-text-align--left sd-text--sans sd-heading--h4 sd-margin
                 </li>
 
                 <li class="simple-list__item simple-list__item--stacked simple-list__item--justify-flex-start" ng-hide="profileConfig.desks === false">
-                    <h3 id="prefered-desks" class="sd-heading sd-text-align--left sd-text--sans sd-heading--h3 sd-txt-transform--capitalize">{{ :: preferences['desks:preferred'].category }}</h3>
+                    <h3 id="prefered-desks" class="sd-heading sd-text-align--left sd-text--sans sd-heading--h3 sd-txt-transform--capitalize" translate>{{ :: preferences['desks:preferred'].category }}</h3>
```
petrjasek, line 166: and here

scripts/apps/users/views/user-preferences.html
tomaskikutis: yep, @devketanpro translate attributes can only be used on inline strings. If it's a variable, we can't add a `translate` attribute and have to ensure that string is translated before it goes to a variable.

petrjasek (approved):

PR #4334, opened 2023-10-03
Add classes on item dragging, fix local storage upload

thecalcc:
SDESK-7063


tomaskikutis (approved):
Was this a regression? Did you find the reason?

Check whether the same issue isn't happening on post broadcasting with authoring react

PR #4335, opened 2023-10-05
Fix dragging UI for media gallery [REACT]

thecalcc:
SDESK-6946


scripts/core/ui/components/drop-zone-3.tsx
```diff
@@ -102,8 +89,17 @@ export class DropZone3 extends React.PureComponent<IDropZoneComponentProps, ISta
 
         return (
             <div
+                onDragOver={(e) => {
+                    e.preventDefault();
+                    this.setState({dragging: true});
```
tomaskikutis, line 94: you should check if it's already true before calling `setState`, `dragOver` will be triggered dozens of times as you drag an element

PR #4338, opened 2023-10-06
merging 2.5.3 and 2.6.3 into develop

petrjasek:


petrjasek:
seems like same test is failing on develop atm, so not an issue with those changes

PR #4339, opened 2023-10-10
TGA sign off event listener

thecalcc:
TGA-59


scripts/apps/authoring/authoring/directives/AuthoringContainerDirective.ts
```diff
@@ -32,6 +32,12 @@ export function AuthoringContainerDirective(authoringWorkspace: AuthoringWorkspa
         scope: {},
         require: 'sdAuthoringContainer',
         link: function(scope, elem, attrs, ctrl) {
+            scope.$on('author_approval:updated', (event) => {
```
tomaskikutis, line 35: can you add a comment for the record that it's handled through `resource:updated` event in authoring-react and this one is only needed for angular-based authoring? You should also test that it actually works on authoring react too even if it's not used yet.

tomaskikutis: > // Needed for authoring Angular.

"**Only** needed for authoring Angular." would be better phrasing.

tomaskikutis (approved):

tomaskikutis (approved):

PR #4340, opened 2023-10-12
[TGA-59] fix(author_approval): incorrect watch function signature

MarkLark86:
@thecalcc I have fixed the function signature so now it works processing the websocket notification. What is still not working, and would need front-end teams expertise, to fix the editor not updating with the changes


thecalcc:
> @thecalcc I have fixed the function signature so now it works processing the websocket notification. What is still not working, and would need front-end teams expertise, to fix the editor not updating with the changes

Ok thanks, I’ll fix that

thecalcc (approved):

PR #4342, opened 2023-10-18
Improve toolbar styles for small screens

tomaskikutis:
SDESK-7058

## before

image

## after
image

### DIFF
--- a/e2e/client/specs/authoring_spec.ts
+++ b/e2e/client/specs/authoring_spec.ts
@@ -663,7 +663,12 @@ describe('authoring', () => {
         workspace.selectDesk('XEditor3 Desk'); // has media gallery in content profile
 
         el(['content-create']).click();
-        el(['content-create-dropdown']).element(by.buttonText('editor3 template')).click();
+
+        const templateBtn = el(['content-create-dropdown']).element(by.buttonText('editor3 template'));
+
+        browser.wait(ECE.elementToBeClickable(templateBtn));
+
+        templateBtn.click();
 
         browser.wait(ECE.visibilityOf(el(['authoring-field--media-gallery', 'media-gallery--upload-placeholder'])));
         expect(ECE.hasElementCount(els(['authoring-field--media-gallery', 'media-gallery-image']), 0)()).toBe(true);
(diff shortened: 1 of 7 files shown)


e2e/client/specs/helpers/monitoring.ts
```diff
@@ -396,7 +396,7 @@ class Monitoring {
         this.tabAction = function(tab) {
             const btn = element(by.css('[ng-click="vm.current_tab = \'' + tab + '\'"]'));
 
-            browser.wait(ECE.elementToBeClickable(btn), 2000);
+            browser.wait(ECE.elementToBeClickable(btn));
```
petrjasek, line 399: this will make it wait for minutes before failing, not sure that's useful in general

e2e/client/specs/authoring_spec.ts
```diff
@@ -663,7 +667,12 @@ describe('authoring', () => {
         workspace.selectDesk('XEditor3 Desk'); // has media gallery in content profile
 
         el(['content-create']).click();
-        el(['content-create-dropdown']).element(by.buttonText('editor3 template')).click();
+
+        const templateBtn = el(['content-create-dropdown']).element(by.buttonText('editor3 template'));
+
+        browser.wait(ECE.elementToBeClickable(templateBtn));
```
petrjasek, line 673: would be good also to add explicit time to wait

tomaskikutis:
I think we should control it via a default timeout so we don't have to pass it every time. 5 seconds would probably be fine in most cases. If I were to set it tests would probably start timing out on the instance and not locally as it is already happening to an extent and I can't understand the problem. Whether github virtual machine has resource limited or protractor is buggy or a combination of the 2.

My idea is that we keep it on life support for now and when you finish back-end for DB snapshots we start moving tests to another framework which is more stable and where we can set ~5 second default timeout. Do you have other ideas?

petrjasek:
problem is that default timeout is also used for the whole spec. switching to a different framework makes sense, protractor is EOLed

tomaskikutis:
> problem is that default timeout is also used for the whole spec.

Interesting, I didn't know that. I'd be reasonable then to make a constant for a single action and always pass it, but given that we'll be moving to another framework I don't think it's worth investing the time now to go over all `browser.wait` calls to update it and deal with all new false failures that would result from it. What do you think?

petrjasek:
yep I wouldn't go over all, just would pass something to new ones and not remove it from existing

PR #4346, opened 2023-10-23
MacOS shortcut for links in editor3

thecalcc:
SDNTB-851
- [x] Don't merge before Nikola tests it on Mac.

### DIFF
--- a/scripts/core/editor3/components/Editor3Component.tsx
+++ b/scripts/core/editor3/components/Editor3Component.tsx
@@ -290,11 +290,11 @@ export class Editor3Component extends React.Component<IProps, IState> {
     keyBindingFn(e) {
         const {key, shiftKey, ctrlKey, metaKey} = e;
         const selectionState = this.props.editorState.getSelection();
-        const osBasedKeyRule = this.isMacOS() ? metaKey : ctrlKey;
+        const modifierKey = this.isMacOS() ? metaKey : ctrlKey;
 
         if (
             key === 'k'
-            && osBasedKeyRule
+            && modifierKey
             && this.props.editorFormat.includes('link')
             && selectionState.isCollapsed() !== true
         ) {
--- a/scripts/core/editor3/components/Editor3Component.tsx
+++ b/scripts/core/editor3/components/Editor3Component.tsx
@@ -280,7 +280,7 @@ export class Editor3Component extends React.Component<IProps, IState> {
     }
 
     isMacOS() {
-        if (navigator.userAgent.indexOf('Mac') != -1) {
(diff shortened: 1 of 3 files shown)


scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -278,13 +278,22 @@ export class Editor3Component extends React.Component<IProps, IState> {
         }
     }
 
+    isMacOS() {
+        if (navigator.userAgent.indexOf('Mac') != -1) {
```
tomaskikutis, line 282: Mac is a very short word and could trigger a false positive when found as a part of some other string https://www.thefreedictionary.com/words-containing-mac

You have it capitalized so it's a bit protected, but I'd rather convert to lower case and check if it includes "macintosh" or "mac os"

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -278,13 +278,22 @@ export class Editor3Component extends React.Component<IProps, IState> {
...
+        }
+
+        return false;
+    }
+
     keyBindingFn(e) {
-        const {key, shiftKey, ctrlKey} = e;
+        const {key, shiftKey, ctrlKey, metaKey} = e;
         const selectionState = this.props.editorState.getSelection();
+        const osBasedKeyRule = this.isMacOS() ? metaKey : ctrlKey;
```
tomaskikutis, line 292: keyboardModifier, mainModifier, modifierKey or modifier would be better choices for naming

tomaskikutis (approved):

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -278,13 +278,22 @@ export class Editor3Component extends React.Component<IProps, IState> {
         }
     }
 
+    isMacOS() {
```
tomaskikutis, line 281: also you need to bind it

scripts/core/editor3/components/Editor3Component.tsx
thecalcc: I agree on your point, but userAgent isn't limitless in regards of what words it might contain. Most answers on the internet say you should check for the string "Mac". Why should I change it if that's the case?

scripts/core/editor3/components/Editor3Component.tsx
thecalcc: But this isn't some random naming, I'm taking `metaKey` from the event. If the platform is macOS and `metaKey` (cmd on Mac) is pressed then I have to use that key. I'm not sure whether we can use just any name here

tomaskikutis (approved):

thecalcc:
Don't merge before Nikola tests it on Mac.

tomaskikutis:
also fix CI

PR #4347, opened 2023-10-25
focus first input when opening authoring

tomaskikutis:
SDAAP-99


thecalcc (approved):

marwoodandrew:
I've pulled this down into my dev environment and I'm not convinced it works, querySelectorAll doesn't find anything , The test instance doesn't seem to be working, I wonder if it should have been rasied against the hotfix/2.5.3 branch?

thecalcc (approved):

tomaskikutis:
@marwoodandrew I checked it again and in fact it didn't work when simulating a slower network. I've added a interval to work around this. We'll be throwing out angular-based authoring code so I'm not putting time in a cleaner solution.

marwoodandrew:
@tomaskikutis, that seems to work, thank you!

marwoodandrew (approved):

PR #4351, opened 2023-10-31
Add support for legacy extensions which have "modals" to be displayed properly

thecalcc:
SWP-2253


pavlovicnemanja (approved):

tomaskikutis:
What was this about @thecalcc? Why would changing side panel be relevant to modals?

thecalcc:
> What was this about @thecalcc? Why would changing side panel be relevant to modals?

Had nothing to do with modals, I had to allow extensions to take their full height in order to fix the publisher extension view

PR #4353, opened 2023-10-31
Template filtering

thecalcc:
SDESK-7084
SDESK-7085

### DIFF
--- a/scripts/api/user.ts
+++ b/scripts/api/user.ts
@@ -6,6 +6,12 @@ function hasPrivilege(privilege: string): boolean {
     return privileges.userHasPrivileges({[privilege]: 1});
 }
 
+function getCurrentUserId(): string {
+    const session = ng.get('session');
+
+    return session?.identity?._id;
+}
+
 function isLoggedIn() {
     const session = ng.get('session');
 
@@ -15,4 +21,5 @@ function isLoggedIn() {
 export const user = {
     hasPrivilege,
     isLoggedIn,
+    getCurrentUserId,
 };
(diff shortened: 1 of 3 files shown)


scripts/api/user.ts
```diff
@@ -6,6 +6,12 @@ function hasPrivilege(privilege: string): boolean {
     return privileges.userHasPrivileges({[privilege]: 1});
 }
 
+function getCurrentUserId(): string {
```
tomaskikutis, line 9: reuse this one https://github.com/superdesk/superdesk-client-core/blob/authoring-react-post-broadcasting/scripts/core/superdesk-api.d.ts#L2861

in the implementation it's inlined, so extract into a separate function

scripts/core/ui/components/content-create-dropdown/more-templates.tsx
```diff
@@ -28,18 +29,34 @@ export class MoreTemplates extends React.PureComponent<IProps, IState> {
...
             method: 'GET',
             path: '/content_templates',
             urlParams: {
                 max_results: pageSize,
                 page: pageToFetch,
-                sort: nameof<ITemplate>('template_name'),
-                where: this.state.searchString.length < 1 ? undefined : {
-                    [nameof<ITemplate>('template_name')]: {
+                sort: templateName,
+                where: this.state.searchString.length < 1 ? {$and: [criteria]} : {
```
tomaskikutis, line 54:
```
const where = {$and: [criteria]};

if (this.state.searchString.length < 1) {
    where[templateName] = {
        $regex: this.state.searchString,
        $options: '-i',
    },
}

httpRequestJsonLocal(where, ...)
```

Avoid repeating the same logic twice ( `{$and: [criteria]}` ). In my pseudo-code it's explicit that `{$and: [criteria]}` is always applied and search is applied conditionally. In your version it wouldn't be that clear for someone adding more filters. Should he add it on both places or only one for example? It is clear at the moment, but as code grows it becomes an issue.

scripts/api/user.ts
thecalcc: How should I import this?

tomaskikutis: https://github.com/superdesk/superdesk-client-core/blob/authoring-react-post-broadcasting/scripts/core/get-superdesk-api-implementation.tsx#L337

look at implementation of this. First you extract it into a function and then use that function anywhere in the code AND also expose via extensions API

tomaskikutis (approved):

PR #4355, opened 2023-11-01
fix publish schedule time picker

petrjasek:
avoid time conversion from local time to utc when saving.

SDESK-7080


scripts/core/utils.tsx
```diff
@@ -364,7 +365,14 @@ export function toQueryString(
  * Output example: "1970-01-19T22:57:38"
  */
 export function toServerDateFormat(date: Date): string {
-    return date.toJSON().slice(0, 19);
+    return moment(date).toISOString(true).slice(0, 19);
```
tomaskikutis, line 368: Why do you need moment here? Could you implement the same thing without it? We're creating a black box here by using moment that supports a thousand features and it's not clear which will be invoked here. You could also use a function from `date-fns` - those are much more explicit.

tomaskikutis: Looks like `formatISO` from `date-fns` does the job. It'd be good to add a comment also that this is your intention so someone who looks at it in the future understands why aren't you using the built-in ".toISOString"

tomaskikutis (approved):

PR #4356, opened 2023-11-02
marked 'not for publication' invisible publish tab

dzonidoo:
SDESK-7079


scripts/core/interactive-article-actions-panel/index-ui.tsx
```diff
@@ -73,14 +73,16 @@ export class InteractiveArticleActionsPanel
         const {activeTab} = this.state;
         const markupV2 = authoringReactViewEnabled && this.props.markupV2 === true;
         const handleUnsavedChanges = this.props.handleUnsavedChanges ?? handleUnsavedChangesDefault;
+        const item = items[0];
```
tomaskikutis: Even though `items[0]` is used in multiple places it's not correct to declare it at the top of the scope, because you would confuse other developers on why are you checking only first item when this component supports multiple.

tomaskikutis (approved):

PR #4366, opened 2023-11-06
Add item action to superdesk api

thecalcc:
Necessary because a planning extension has to use it.

### DIFF
--- a/scripts/api/article.ts
+++ b/scripts/api/article.ts
@@ -378,6 +378,12 @@ function get(id: IArticle['_id']): Promise<IArticle> {
     return dataApi.findOne<IArticle>('archive', id);
 }
 
+function itemAction(article: IArticle): any {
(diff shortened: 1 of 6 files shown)


scripts/api/article.ts
```diff
@@ -430,7 +437,7 @@ interface IArticleApi {
      */
     isPublished(article: IArticle, includeScheduled?: boolean): boolean;
 
-    itemAction(article: IArticle): {[key: string]: boolean};
+    itemAction(article: IArticle): {[key in IAuthoringActionType]: boolean};
```
tomaskikutis, line 440: To avoid repeating the type you could use `ReturnType<typeof itemAction>`. If it was repeated more heavily I'd suggest to create another type like `type IItemActionsResult = {[key in IAuthoringActionType]: boolean}`

tomaskikutis (approved):

PR #4370, opened 2023-11-09
publish tab fix

dzonidoo:
SDESK-7079

### DIFF
--- a/scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
+++ b/scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
@@ -48,7 +48,7 @@ export function AuthoringTopbarDirective(
                 dispatchInternalEvent('interactiveArticleActionStart', {
                     items: [scope.item],
                     tabs: ['send_to', 'publish'],
-                    activeTab: scope.item.flags.marked_for_not_publication ? 'send_to' : 'publish',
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -48,7 +48,7 @@ export function AuthoringTopbarDirective(
                 dispatchInternalEvent('interactiveArticleActionStart', {
                     items: [scope.item],
                     tabs: ['send_to', 'publish'],
-                    activeTab: 'publish',
+                    activeTab: scope.item.flags.marked_for_not_publication ? 'send_to' : 'publish',
```
petrjasek, line 51: would use probably `item.flags?.marked_...` not sure if we have those flags set always

petrjasek (approved):

tomaskikutis:
I was reviewing #4390 from @devketanpro and some questions about intended behavior came up which brought me to this PR - which seems incorrect @petrjasek. I was checking the JIRA ticket where it says that publish tab shouldn't be available in certain conditions. Then the primary thing we need to change is which tabs are available - set via 
`tabs: ['send_to', 'publish']`. It was correct to adjust `activeTab`, but available tabs should have been adjusted as well.

I'll ask Ketan to improve it, posting here just to bring it to your attention Petr since you were reviewing it.

PR #4371, opened 2023-11-09
Content Profile - > Width field although required can be saved without initialization

dzonidoo:
SDESK-7099

### DIFF
--- a/scripts/core/ui/components/generic-form/input-types/select_single_value.tsx
+++ b/scripts/core/ui/components/generic-form/input-types/select_single_value.tsx
@@ -100,7 +100,7 @@ export function getSelectSingleValue(
                     <label className="sd-line-input__label">{this.props.formField.label}</label>
                     <select
                         disabled={this.props.disabled || this.state.items == null || this.state.items.length < 1}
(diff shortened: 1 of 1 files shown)


scripts/core/ui/components/generic-form/input-types/select_single_value.tsx
```diff
@@ -100,7 +100,7 @@ export function getSelectSingleValue(
                     <label className="sd-line-input__label">{this.props.formField.label}</label>
                     <select
                         disabled={this.props.disabled || this.state.items == null || this.state.items.length < 1}
-                        value={this.props.value || ''}
+                        value={this.props.value || this.state.items[0].id}
```
tomaskikutis, line 103: It doesn't look like a good solution to me. If the problem is that an empty value is being saved, we can solve it by either adding validation or not allowing empty value. If we're going with an approach of not allowing empty values, we'd need to add a prop to this component and in that case don't render the `<option>` tag with empty string a a value. The display of value shouldn't be messed with. The value should be displayed as it is.

tomaskikutis:
will solve on the back-end

PR #4372, opened 2023-11-10
fix annotations list when updated item is no longer in the list

petrjasek:
SDESK-7091

### DIFF
--- a/scripts/core/helpers/crud-manager-http.tsx
+++ b/scripts/core/helpers/crud-manager-http.tsx
@@ -34,6 +34,7 @@ export function connectCrudManagerHttp<Props, Entity extends IBaseRestApiRespons
                 _meta: null,
                 activeSortOption: defaultSortOption ?? null,
                 activeFilters: {},
+                itemsById: {},
             };
 
             this.create = this.create.bind(this);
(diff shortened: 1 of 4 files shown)


tomaskikutis:
I understand it was throwing an exception and crashing. What was the cause?

petrjasek:
it was trying to use the item from the list which wasn't there anymore if user changed the value used for filtering, so this part would return `null`:
https://github.com/superdesk/superdesk-client-core/pull/4372/files#diff-50e5a74c6f2cbb85d07847eb75eb1ec5829dcbd0e2608a7edb477501b93bd659L693-L695
but I'm not sure actually this fix is needed, in theory would be enough to unset `editItemId` after saving, I thought it was showing the updated item after saving but that's probably not the case.

tomaskikutis:
I've looked into it. A simple option as you say is to close the item after you edit it. An improved version would be to only close if it's no longer available - the same issue might happen even due to pagination/sorting and not filters.

The harder option is to fetch the individual item and keep the preview anyway. Both would work for me, but I'd want to maintain clean code.

What I don't like in this implementation is that `crudManager._items` is maintained and `itemsById` is added on top. One problem with this is data duplication - I would rather compute the map on demand + use memo for performance. This wouldn't be straight-forward, because unlike it would intuitivelly seem - the way you use `itemsById` it is not simply a map made from the list but contains different data - preview item, which is a messy approach IMO.

I think it would be cleanest to simply have a preview specific field inside the [state](https://github.com/superdesk/superdesk-client-core/blob/authoring-react-post-broadcasting/scripts/core/ui/components/ListPage/generic-list-page.tsx#L44) that would hold the entire object and not only ID as it is now.

petrjasek:
I'm not sure if it would be just for preview or it would also need the edited item to be there.

PR #4373, opened 2023-11-10
use storage records api in e2e

petrjasek:


tomaskikutis:
Progress at last :tada: . Are there instructions somewhere on how do I try it out?

petrjasek:
added a readme @tomaskikutis

tomaskikutis:
I see you removed the existing prepopulate thing entirely and probably created a full dump from it. The problem is that prepopulate script produced a corrupted DB in a sense that schema wasn't respected and you would find all kinds of states that are not possible in production. My idea was to continue using prepopulate for old tests, but make a new full dump for new tests. That still can be done in a way - I would rename existing full dump to `legacy_dump` and would only run old tests against it. And make a new clean one for new tests that we'd run with a new framework.

In summary, for this PR it'd be enough if you could rename current full dumps to `legacy_dump` or something similar and create a clean full dump using the same process as you'd use for production.

tomaskikutis:
by the way, I'm wondering now, is there a way to modify the full dump instead of creating a new record? I'm thinking that for 90% of the cases we'd want one big snapshot/dump to be able to speed up and parallelize test runs, and for the edge cases we'd use records where a specific state needs to be simulated that can't co-exists in the main dump.

petrjasek:
not sure what data should be in this full dump, we don't have initial dataset.

> by the way, I'm wondering now, is there a way to modify the full dump instead of creating a new record? I'm thinking that for 90% of the cases we'd want one big snapshot/dump to be able to speed up and parallelize test runs, and for the edge cases we'd use records where a specific state needs to be simulated that can't co-exists in the main dump.

I think that's the usecase for records, it's just a diff from the full dump.

tomaskikutis:
> not sure what data should be in this full dump, we don't have any initial dataset.

I'd initialize the DB, add mandatory metadata and then 1 desk, 1 stage, 1 article and that'd be good for a start. Later we could add things as we go along.

> I think that's the usecase for records, it's just a diff from the full dump.

I know what records are useful for. What I was asking is whether we can also append stuff to the main dump and not create a record?

petrjasek:
> > I think that's the usecase for records, it's just a diff from the full dump.
> 
> I know what records are useful for. What I was asking is whether we can also append stuff to the main dump and not create a record?

what's the usecase? it still sounds to me as a new record but I'm probably missing something

tomaskikutis:
> what's the usecase? it still sounds to me as a new record but I'm probably missing something

Imagine we write 100 tests per year. Would it mean then that after that everytime before e2e runs, 100 records need to be applied to the base dump to generate the latest database state? If it's fast I don't mind, but I suspect it might be a bit inefficient.

tomaskikutis (approved):

PR #4376, opened 2023-11-13
Fix desks template filter

thecalcc:
SDESK-7085

### DIFF
--- a/scripts/core/query-formatting.ts
+++ b/scripts/core/query-formatting.ts
@@ -39,9 +39,9 @@ function toElasticFilter(q: ILogicalOperator | IComparison) {
             case '$in':
                 return {terms: {[field]: value}};
             case '$exists':
-                return {terms: {exists: {[field]: value}}};
+                return {query: {exists: {field: field}}};
             case '$notExists':
-                return {terms: {must_not: {exists: {[field]: value}}}};
+                return {query: {bool: {must_not: [{exists: {field: field}}]}}};
             case '$stringContains':
                 return {query: {simple_query_string: {query: value.val, fields: [field]}}};
             }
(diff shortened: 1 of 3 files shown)


scripts/core/ui/components/content-create-dropdown/more-templates.tsx
```diff
@@ -41,7 +41,12 @@ export class MoreTemplates extends React.PureComponent<IProps, IState> {
             deskCriteria.push({[templateDesks]: {$in: [currentDeskId]}});
         }
 
-        const criteria = {$or: [{$or: deskCriteria}, {user: sdApi.user.getCurrentUserId()}]};
+        const criteria = {
```
tomaskikutis, line 44: can you use `ISuperdeskQuery` as [here](https://github.com/superdesk/superdesk-client-core/blob/authoring-react-post-broadcasting/scripts/api/article-send.tsx#L20)?

the logic also seems off inside `criteria`. I'd expect it to be
```$or: [deskCriteria, {user: sdApi.user.getCurrentUserId()}]```
why double $ors?

Comment (thecalcc) [reply scripts/core/ui/components/content-create-dropdown/more-templates.tsx:44]: I can't use `ISuperdeskQuery`, since we're working on the `where` filter which we don't do anything for in the `ISuperdeskQuery` or `toElasticQuery`. About the `or` - is because `deskCriteria` is an array of options.

File: scripts/core/query-formatting.ts [hunk 2]
```diff
@@ -38,6 +38,12 @@ function toElasticFilter(q: ILogicalOperator | IComparison) {
                 return {range: {[field]: {'lte': value}}};
             case '$in':
                 return {terms: {[field]: value}};
+            case '$exists':
+                return {terms: {exists: {[field]: value}}};
+            case '$notExists':
+                return {terms: {must_not: {exists: {[field]: value}}}};
```
tomaskikutis, line 44: let's remove `notExists` if we can do the same via `$exists: false`

tomaskikutis (approved):

tomaskikutis (approved):

PR #4378, opened 2023-11-14
show error message on delete

petrjasek:


scripts/core/helpers/crud-manager-http.tsx
```diff
@@ -97,6 +97,14 @@ export function connectCrudManagerHttp<Props, Entity extends IBaseRestApiRespons
                 .then(() => this.refresh())
                 .then(() => {
                     notify.success(gettext('The item has been deleted.'));
+                })
+                .catch((reason) => {
+                    if (reason != null && reason.message != null) {
```
tomaskikutis, line 102: you can also use a shorthand `reason?.message != null` instead

tomaskikutis (approved):

PR #4380, opened 2023-11-15
Fix table width

thecalcc:
SDESK-7116

I didn't find the ticket specific to this issue but we can consider it a part of the general fixes for multi-line quote

### DIFF
--- a/scripts/core/editor3/components/tests/tables.spec.tsx
+++ b/scripts/core/editor3/components/tests/tables.spec.tsx
@@ -53,6 +53,7 @@ describe('editor3.component.table-cell', () => {
     it('should render', () => {
         const wrapper = shallow(
             <TableCell
+                fullWidth={{}}
                 editorState={EditorState.createWithContent(ContentState.createFromText('abc'))}
                 onChange={() => { /* no-op */ }}
                 readOnly={false}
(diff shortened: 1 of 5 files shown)


tomaskikutis:
actual tables shouldn't be set to 100% by default. Add a config option so it's only applied to pull quote

thecalcc:
> actual tables shouldn't be set to 100% by default. Add a config option so it's only applied to pull quote

Actually we don't need a config option since we already have className which we can check against whether it's a multi line quote or not

tomaskikutis:
Classnames are for styling and you shouldn't be using them for anything else. If I could, I would write a lint rule not allowing to compare a classname using `===` :D

Use a separate prop like `fullWidth: boolean` and keep classname decoupled.

scripts/core/editor3/components/tests/tables.spec.tsx
```diff
@@ -53,6 +53,7 @@ describe('editor3.component.table-cell', () => {
     it('should render', () => {
         const wrapper = shallow(
             <TableCell
+                fullWidth={{}}
```
tomaskikutis, line 56: how are you passing an empty object where a boolean is required?

tomaskikutis (approved):

PR #4381, opened 2023-11-16
bundle css from @superdesk/common

tomaskikutis:
NHUB-432


thecalcc (approved):
Did you test what happens if you delete a file?

tomaskikutis:
no, but since it watches all events - it should triger on edition/deletion too

PR #4382, opened 2023-11-16
Fix positioning of create new article bottom toolbar button

tomaskikutis:
SDESK-7075

### DIFF
--- a/scripts/core/popper-utils.tsx
+++ b/scripts/core/popper-utils.tsx
@@ -1,0 +1,68 @@
+import {Modifier, Instance as PopperInstance} from '@popperjs/core';
+
+export const sameWidth: Modifier<any, any> = {
+    name: 'sameWidth',
+    enabled: true,
+    phase: 'beforeWrite',
+    requires: ['computeStyles'],
+    fn: ({state}) => {
+        state.styles.popper.width = `${state.rects.reference.width}px`;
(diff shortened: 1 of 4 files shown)


scripts/core/popper-utils.tsx
```diff
@@ -0,0 +1,68 @@
...
+export const applyMaxSize: Modifier<any, any> = {
+    name: 'applyMaxSize',
+    enabled: true,
+    phase: 'beforeWrite',
+    requires: ['maxSize'],
+    fn: ({state}) => {
+        const {height} = state.modifiersData.maxSize;
+
+        // subtracting 10 in order to make a gap between the edge of the viewport
+        state.styles.popper.maxHeight = `${height - 10}px`;
```
thecalcc, line 29: What if someone is using a different than the standard screen size? Would `10` still be enough?

thecalcc (approved):

scripts/core/popper-utils.tsx
tomaskikutis: Why would screen size matter? As you see from the comment, the goal is to make a gap - i.e. to avoid popover appearing "glued" to the bottom/top. The purpose is to make it clear visually where is the end of the popover so user doesn't try scrolling around to find the end.

PR #4386, opened 2023-11-21
Carousel implementation

thecalcc:
NHUB-432


thecalcc:
Example looks:
image
image
image

MarkLark86:
@thecalcc @tomaskikutis Will this work across both React 16 & 17? I see the package.json has React 16 as a dependency, where as Newshub's [package.json](https://github.com/superdesk/newsroom-core/blob/develop/package.json#L76) has React 17.

thecalcc:
> @thecalcc @tomaskikutis Will this work across both React 16 & 17? I see the package.json has React 16 as a dependency, where as Newshub's [package.json](https://github.com/superdesk/newsroom-core/blob/develop/package.json#L76) has React 17.

Should work, at least it did locally

tomaskikutis:
image
a good future improvement would be good to improve spacing of button. Instead of the button leaving image area and going on top of margin, it'd be better if it remained inside image area and even was a moved from the edge the same amount as we have margin on the side.

PR #4390, opened 2023-11-22
FIX : for correction items enable send-correction tab instead of publish tab [SDESK-7111]

devketanpro:


scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -45,13 +46,33 @@ export function AuthoringTopbarDirective(
...
                 dispatchInternalEvent('interactiveArticleActionStart', {
                     items: [scope.item],
-                    tabs: ['send_to', 'publish'],
-                    activeTab: scope.item.flags?.marked_for_not_publication ? 'send_to' : 'publish',
+                    tabs: getTabsArray(activeTab),
+                    activeTab: activeTab,
                 });
             };
 
+            function determineActiveTab(item) {
```
tomaskikutis, line 58: This is making implicit assumptions about which tabs are available. First check if a tab exists in the list of available tabs and add a fallback to 1st tab.

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -45,13 +46,33 @@ export function AuthoringTopbarDirective(
...
+                if (item.flags?.marked_for_not_publication) {
+                    return 'send_to';
```
tomaskikutis, line 60: is there a reason to set tab to `send_to` here, rather than to a first tab?

tomaskikutis (commented):
- improve naming: `getTabsArray` / `determineActiveTab` -> `getAvailableTabs` / `getActiveTab`
- ensure it also works correctly in refactored authoring (by setting `authoringReactEnabledUserSelection` to `true` [here](https://github.com/superdesk/superdesk-client-core/blob/authoring-react-post-broadcasting/scripts/appConfig.ts#L49))

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
devketanpro: this is something I'm not sure of.
Previously, @dzonidoo made the following changes: https://github.com/superdesk/superdesk-client-core/pull/4370, he may be able to explain this better.

tomaskikutis: That change he made is unrelated. It was about publishing tab appearing when it wasn't supposed to appear. If you chose first available tab here all would be fine.

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
devketanpro: available tabs are decided based on active tab:
if `activeTab = correct` so in this case `availableTabs = ['send_to', 'correct']` and if `activeTab = publish/send_to` in this case, `availableTabs = ['send_to', 'publish']`
So first, we have to check activeTab based on the item. then send both activeTab and availableTabs

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
devketanpro: got it

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -45,13 +46,34 @@ export function AuthoringTopbarDirective(
...
+            function getAvailableTabs(): Array<IArticleActionInteractive> {
+                if (scope.isCorrection(scope.item)) {
+                    return ['send_to', 'correct'];
+                } else {
+                    return ['send_to', 'publish'];
+                }
+            }
+
+            function getActiveTab(availableTabs: Array<IArticleActionInteractive>): IArticleActionInteractive {
+                if (availableTabs.includes('correct') && scope.isCorrection(scope.item)) {
```
tomaskikutis, line 68: If `scope.isCorrection(scope.item)` is `false`, tab should not be **available**. And if tab is not available it can't be active. You should move that check to `getAvailableTabs` function.

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -45,13 +46,34 @@ export function AuthoringTopbarDirective(
...
                 dispatchInternalEvent('interactiveArticleActionStart', {
                     items: [scope.item],
-                    tabs: ['send_to', 'publish'],
-                    activeTab: scope.item.flags?.marked_for_not_publication ? 'send_to' : 'publish',
+                    tabs: availableTabs,
+                    activeTab: activeTab,
                 });
             };
 
+            function getAvailableTabs(): Array<IArticleActionInteractive> {
```
tomaskikutis, line 59: Integrate requirements from SDESK-7079  (#4370) to this function, because now you removed it completely.

tomaskikutis (approved):

PR #4393, opened 2023-11-24
Revert "Carousel implementation"

thecalcc:
Reverts superdesk/superdesk-client-core#4386


tomaskikutis:
why did you revert it?

PR #4397, opened 2023-11-28
Annotations library fixes

thecalcc:
SDESK-7091

### DIFF
--- a/scripts/core/ui/components/ListPage/generic-list-page.tsx
+++ b/scripts/core/ui/components/ListPage/generic-list-page.tsx
@@ -444,7 +444,11 @@ export class GenericListPageComponent<T, P>
                                         key={this.props.getId(item)}
                                         item={item}
                                         page={page}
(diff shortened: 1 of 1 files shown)


scripts/core/ui/components/ListPage/generic-list-page-item-view-edit.tsx
```diff
@@ -159,7 +166,7 @@ export class GenericListPageItemViewEdit<T> extends React.Component<IProps<T>, I
                 this.setState({
                     issues: {},
                 }, () => {
-                    this.props.onEditModeChange(false);
+                    this.props.onEditModeChange(true);
```
tomaskikutis, line 169: are you sure this change is needed? It seems to me it was correct - after you're done editing - you should go into preview mode - thus editing mode should be set to `false`

tomaskikutis (commented):
looks good otherwise

PR #4399, opened 2023-11-29
Incorrect height of search bar in Annotations Library

dzonidoo:
SDESK-7135


tomaskikutis:
did you go over all other usages of search bar to check if your addition isn't breaking it anywhere else?

dzonidoo:
yep... and Search bar on monitoring also have height: 100%...

PR #4406, opened 2023-12-13
add email field to desk config

petrjasek:
CPCN-502

### DIFF
--- a/scripts/apps/desks/views/desk-config-modal.html
+++ b/scripts/apps/desks/views/desk-config-modal.html
@@ -62,6 +62,15 @@
                             </div>
                         </div>
                     </div>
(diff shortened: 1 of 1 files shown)


tomaskikutis:
is it really for `CPCN-502`? I don't see anything about emails there

petrjasek:
you're right, it's rather CPCN-503

tomaskikutis (approved):

tomaskikutis:
I'm thinking maybe we should do it conditional, because it doesn't have any use in superdesk

tomaskikutis:
Ideal scenario would be being able to set it up from extensions API

petrjasek:
there is a specific ticket for the hiding later

PR #4410, opened 2023-12-20
Editor3 performance improvements

tomaskikutis:
SDCP-727

https://github.com/superdesk/superdesk-client-core/assets/2076953/7861f043-23fc-40e5-b986-a45973757111

https://github.com/superdesk/superdesk-client-core/assets/2076953/ab6bbd4a-2d9f-4f4c-8090-67a2702a19c4

https://github.com/superdesk/superdesk-client-core/assets/2076953/cd8ed3ca-a570-4c68-8464-b1f8e06bd25e

PR #4419, opened 2024-01-22
Editor3 performance improvements

tomaskikutis:
reopened version of https://github.com/superdesk/superdesk-client-core/pull/4410


petrjasek (approved):

e2e/client/specs/helpers/authoring.ts
```diff
@@ -450,7 +450,9 @@ class Authoring {
         this.save = function() {
             browser.wait(() => this.save_button.isEnabled(), 2000);
             this.save_button.click();
-            return browser.wait(() => this.save_button.getAttribute('disabled'));
+            return browser.wait(
+                ECE.stalenessOf(this.save_button.element(by.css('[data-test-id="loading-indicator"]'))),
+            );
```
petrjasek, line 455: it would be good to add explicit timeout, otherwise it uses the default 200s

tomaskikutis: I'd say explicit timeout only makes sense when the test is exceptional and needs more time. Otherwise it'd go with a default timeout - and lower the default if it's too long.

petrjasek: problem is it uses the default timeout also for the whole spec, so we can't make it lower

tomaskikutis: By whole `spec` you mean `describe` block? Ok, then we can add an additional limit to browser.wait

petrjasek: I think it's for `it` block

PR #4427, opened 2024-02-07
Implement onFieldChange extension point, content profile switching in authoring react,

thecalcc:


scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -452,22 +452,28 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
...
+    handleFieldChange(fieldId: string, data: unknown, exposed: IExposedFromAuthoring<T>) {
         const {state} = this;
 
         if (state.initialized !== true) {
             throw new Error('can not change field value when authoring is not initialized');
         }
 
-        const {onFieldChange} = this.props;
         const fieldsDataUpdated = state.fieldsDataWithChanges.set(fieldId, data);
+        const resultFromOnFieldChange = this.props.onFieldChange?.(
```
tomaskikutis, line 463: Avoid optional chaining for function calls, especially when you expect something returned. Here for example `resultFromOnFieldChange` can evaluate to `undefined` and on next line you access its properties as it were an object.

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -560,13 +584,21 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
...
             }
+
+            return initialState;
+        });
+    }
+
+    componentDidMount() {
+        const {authoringStorage} = this.props;
+
+        this.switchProfile(() => authoringStorage.getEntity(this.props.itemId)).then((initialState) => {
```
tomaskikutis, line 600: since `switchProfile` profile already provides an ID of an item, you should pass that to `getEntity` instead of the one from props.

scripts/core/superdesk-api.d.ts
```diff
@@ -160,6 +161,7 @@ declare module 'superdesk-api' {
         initiateClosing(): void;
         keepChangesAndClose(): void;
         stealLock(): void;
+        loadContentProfile(contentProfileId?: string): void
```
tomaskikutis, line 164: why `contentProfileId` is optional? How would you load a profile without ID?

tomaskikutis (commented):
fix tests

scripts/apps/authoring-react/authoring-react.tsx
thecalcc: I agree, but in this case I've handled it on the next lines, so it's ok

tomaskikutis: image

Doesn't seem handled. If you want to keep the conditional function call at least add the type:
`const resultFromOnFieldChange: undefined | MyType = ...`

tomaskikutis (commented):
```
commit 5f43e7b403be1c827a0b72d8a0f27b3c95243c00
Author: Konstantin Markov <maga.ot.oz@protonmail.com>
Date:   Tue Apr 9 12:13:55 2024 +0300

    Changes after review
```

You did something wrong with this commit, probably merging. You're dropping my implementation of `preferIncomplete` in `authoring-react.tsx`. Adjust your PR not to touch it.

tomaskikutis (changes requested):
>  Files changed 160

Seems like there is still something wrong with the branch. There are even more things that this PR is removing that it shouldn't.

PR #4428, opened 2024-02-07
Broadcasting: (Inline) comments panes not opening for new items

dzonidoo:
SDESK-7174


tomaskikutis (approved):

dzonidoo:
@tomaskikutis but we don't have this issue in monitoring

tomaskikutis:
Why don't we have the same issue in authoring if it's the same code? Are you sure you have authoring-react enabled?

dzonidoo:
@tomaskikutis yes authoring-react was enabled... and we don't have issue on monitoring... its working different... when u click on plus button to create article - your article have 'id' already.. but its not case with rundown.. u need to save it first and then your rundown get 'id'

petrjasek (approved):

PR #4429, opened 2024-02-07
Broadcasting: Wrong fields width in authoring

dzonidoo:
SDESK-7173


tomaskikutis (changes requested):
fix tests

petrjasek:
I wonder if we might rather set the width in https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/extensions/broadcasting/src/rundown-items/content-profile.tsx
but not sure exactly where is this `null` coming from

PR #4430, opened 2024-02-07
e2e test duplication in monitoring

tomaskikutis:
SDESK-7164


scripts/apps/search/components/SmoothLoader.tsx
```diff
@@ -55,7 +55,9 @@ export class SmoothLoader extends React.Component<IProps> {
                 {
                     loading && (
                         <div style={{height: '100%', position: 'relative'}}>
-                            <ListItemLoader />
+                            <div data-test-id="loading-placeholder">
+                                <ListItemLoader />
+                            </div>
```
thecalcc, line 60: Since you wrap this component in a div with the same data-test-id wouldn't it be better if we put the wrapper inside the component itself? Note: the component is in superdesk-ui-framework.

scripts/core/itemList/LazyLoader.tsx
```diff
@@ -207,7 +207,9 @@ export class LazyLoader<T> extends React.Component<IProps<T>, IState<T>> {
                     {(() => {
                         if (loading === true) {
                             return (
-                                <ListItemLoader />
+                                <div data-test-id="loading-placeholder">
+                                    <ListItemLoader />
+                                </div>
```
thecalcc, line 212: This is the second wrapping.

petrjasek (approved):

PR #4436, opened 2024-02-13
fix extension field type in superdesk-api

petrjasek:


thecalcc (approved):

tomaskikutis:
`ICustomFieldType<unknown, unknown, unknown, unknown>` was correct. Otherwise it wouldn't compile to use a type that takes arguments without providing them. Compiler errors don't show up for d.ts files for some reason.

petrjasek:
it didn't compile belga coverage extension with that, complaining that
```ts
ICustomFieldType<string, string, ICommonFieldConfig, never>
```
is not compatible with it

tomaskikutis:
It isn't, but I usually cast it when providing to superdesk core:

```
const field: ICustomFieldType<string, string, ICommonFieldConfig, never> = {...};

return {
    customFieldTypes?: [
        field as ICustomFieldType<unknown, unknown, unknown, unknown>,
    ],
}
```

It achieves the main goal to cover the types inside the extension.

PR #4439, opened 2024-02-16
Fix styles when dragging embed

thecalcc:
SDANSA-549


tomaskikutis (approved):
make tests pass before merging

thecalcc:
Other tests that are broken are also broken on develop, and I'm making a fix for those right now, so will be merging this one.

PR #4440, opened 2024-02-16
Port embed test to playwright

thecalcc:


e2e/client/playwright/authoring.spec.ts
```diff
@@ -38,3 +38,28 @@ test('can apply "populate abstract" macro', async ({page}) => {
         page.locator(s('authoring', 'authoring-field=abstract')).getByRole('textbox'),
     ).toHaveText('test sport story body');
 });
+
+test('can embed if url is valid', async ({page}) => {
```
tomaskikutis, line 42: I'd call it "can add embeds". If you want to test invalid urls, I would add another test "editor3 doesn't crash when attempting to add invalid embeds"

Also, it's better to move this test to `editor3.spec.ts` so we separate editor stuff from the rest of authoring tests.

e2e/client/playwright/authoring.spec.ts
```diff
@@ -38,3 +38,28 @@ test('can apply "populate abstract" macro', async ({page}) => {
...
+
+    await page.locator(
+        s('monitoring-group=Sports / Working Stage', 'article-item=test sports story'),
+    ).dblclick();
+
+    await expect(
+        page.locator(s('authoring', 'authoring-field=body_html')).getByRole('textbox'),
+    ).toHaveText('test sport story body');
+
+    await page.locator(s('Embed')).getByRole('button').click();
```
tomaskikutis, line 59: avoid global selectors like `s('embed')`. If someone modifies the content profile in the future and enables embeds in other fields it might crash your test. Since there may possible be multiple fields and you want to work only with body HTML, use `(s('authoring', 'authoring-field=body_html', 'embed'))`

e2e/client/playwright/authoring.spec.ts
```diff
@@ -38,3 +38,28 @@ test('can apply "populate abstract" macro', async ({page}) => {
...
+
+    await page.getByPlaceholder('Enter URL or code to embed').type('https://google.com');
```
tomaskikutis, line 61: How about using some fake url to avoid mentioning other companies/brands? Or let's at least adversite ourselves  `.type('https://superdesk.org')`

e2e/client/playwright/authoring.spec.ts
```diff
@@ -38,3 +38,28 @@ test('can apply "populate abstract" macro', async ({page}) => {
...
+
+    await page.locator(s('embed-controls', 'submit')).click();
+    await expect(page.getByText('https://google.com')).toBeDefined();
```
tomaskikutis, line 64: This doesn't seem to be testing much except that you enter the URL and later have the URL present in the page. I'd think embedding does more than that?

e2e/client/playwright/editor3.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    await page.goto('/#/workspace/monitoring');
+
+    await monitoring.selectDesk('Sports');
+
+    await page.locator(
+        s('monitoring-group=Sports / Working Stage', 'article-item=test sports story'),
+    ).dblclick();
+
+    await expect(
```
tomaskikutis, line 18: this assert isn't relevant for this test

e2e/client/playwright/editor3.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+        page.locator(s('authoring', 'authoring-field=body_html')).getByRole('textbox'),
+    ).toHaveText('test sport story body');
+
+    await page.getByRole('button', {name: 'Embed'}).click();
```
tomaskikutis, line 22: use a more specific selector

e2e/client/playwright/editor3.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    await page.getByPlaceholder('Enter URL or code to embed').type('https://sourcefabric.org');
```
tomaskikutis, line 24: use a more specific selector

e2e/client/playwright/editor3.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    await page.locator(s('embed-controls', 'submit')).click();
+    await expect(page.getByText('https://sourcefabric.org')).toBeDefined();
```
tomaskikutis, line 27: image

If we intercept the request and return a hardcoded response for testing we could make the test better by at least checking that title and description were fetched correctly.

e2e/client/playwright/editor3.spec.ts
```diff
@@ -1,12 +1,31 @@
...
+    await page.route(
+        `https://iframe.ly/api/oembed?callback=?&url=
+        ${requestRoute}
+        &api_key=${appConfig.iframely.key}
+        &omit_script=true&iframe=true`,
+        (route) => {
+            route.fulfill({
+                body: JSON.stringify([{
+                    title: 'Open Source Software for Journalism',
+                    // eslint-disable-next-line max-len
```
tomaskikutis, line 23: when a line is too long, split in 2 and concatenate

```
description: 'Sourcefabric is Europe\'s largest developer of '
 + 'open source tools for news media, powering news and media organisations around the world.',
 ```

**Review by tomaskikutis — COMMENTED** [review]
Let's agree that comments are only "resolved" on github by the person reviewing it. I did unresolve a few that weren't addressed.

**Review by tomaskikutis — COMMENTED** [review]
do handle conflicts

File: e2e/client/playwright/editor3.spec.ts [hunk 10]
```diff
@@ -3,6 +3,48 @@ import {test, expect} from '@playwright/test';
...
```
tomaskikutis: if you did concatenate it - you long longer need to disable eslint for that line

scripts/core/editor3/components/toolbar/IconButton.tsx
```diff
@@ -20,12 +20,14 @@ interface IProps {
 export const IconButton: React.FunctionComponent<IProps> = ({onClick, iconName, tooltip, uiTheme}) => (
     <div
         data-flow={'down'}
+        data-test-id="icon-button"
+        data-test-value={tooltip}
```
tomaskikutis: Since values are dynamic anyway here, there's no benefit to use data-test-id.

Use `getByRole('button', {name: 'abc'}`. In case of an icon button, playwright will match `name` against aria-label

tomaskikutis (commented):
if you remove `data-test-id`, remove `data-test-value` as well

thecalcc:
@tomaskikutis can you resolve all comments that are addressed in the code

tomaskikutis (approved):
All good now, let's get tests passing before merging. Unit tests you need to look at and e2e tests I'll take a look at because I saw they fail on develop.

tomaskikutis (approved):

PR #4442, opened 2024-02-21
Rundowns: Change the button labels for editing/saving Rundowns and Rundown items

dzonidoo:
https://sofab.atlassian.net/browse/SDESK-7186

### DIFF
--- a/scripts/core/ui/components/ActionButton.tsx
+++ b/scripts/core/ui/components/ActionButton.tsx
@@ -1,0 +1,22 @@
+import * as React from 'react';
+
+interface IProps {
+    ariaValue: string;
+    refValue?(event): void;
+    onClick(event: React.MouseEvent): void;
+}
+
+export class ActionButton extends React.PureComponent<IProps> {
+    render() {
+        return (
+            <button
+                className="sd-navbtn"
+                aria-label={this.props.ariaValue}
+                ref={this.props.refValue}
+                onClick={this.props.onClick}
+            >
+                <i className="icon-dots-vertical" />
+            </button>
+        );
+    }
+}
--- a/scripts/extensions/broadcasting/src/rundowns/rundown-view-edit.tsx
+++ b/scripts/extensions/broadcasting/src/rundowns/rundown-view-edit.tsx
@@ -376,7 +376,7 @@ export class RundownViewEditComponent extends React.PureComponent<IProps, IState
                                     gap="16"
                                     justifyContent="space-between"
                                     noWrap
-                                    style={{paddingLeft: 16, paddingRight: 16}}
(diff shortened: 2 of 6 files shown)


scripts/extensions/broadcasting/src/rundowns/rundown-view-edit.tsx
```diff
@@ -469,12 +467,13 @@ export class RundownViewEditComponent extends React.PureComponent<IProps, IState
                                                 },
                                             ]}
                                         >
-
-                                            <IconButton
-                                                ariaValue={gettext('Actions')}
-                                                icon="dots-vertical"
+                                            <button
```
petrjasek, line 470: I thought we want to use components for buttons everywhere

dzonidoo: I know.... but I needed to edit the styles and add a class... but the IconBUtton component doesn't have a prop for className.. so I had to find a specific solution

petrjasek: I get the requirements but not sure about the fix, we should then probably make those components if they are missing in the framework, that's pretty common usecase we have in other places as well

dzonidoo: thats one options.. another one is to add className prop on Button conponent in ui-framework side

scripts/core/ui/components/ActionButton.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+
+interface IProps {
+    ariaValue: string;
+    refValue?(event): void;
+    onClick(event: React.MouseEvent): void;
+}
+
+export class ActionButton extends React.PureComponent<IProps> {
```
petrjasek, line 9: I would probably call it `MoreActionsButton` so it's more obvious

scripts/core/ui/components/ActionButton.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+
+interface IProps {
+    ariaValue: string;
```
petrjasek, line 4: `ariaLabel` would be better imo if it's used for `aria-label`

scripts/core/ui/components/ActionButton.tsx
```diff
@@ -0,0 +1,22 @@
...
+    refValue?(event): void;
```
petrjasek, line 5: it's not a value if it's a callback, but calling it just `ref` would be probably ok

petrjasek (approved):

PR #4443, opened 2024-02-21
Show selcted Rundown item

dzonidoo:
SDESK-7188

link: https://github.com/superdesk/superdesk-ui-framework/pull/824

### DIFF
--- a/scripts/extensions/broadcasting/src/interfaces.ts
+++ b/scripts/extensions/broadcasting/src/interfaces.ts
@@ -61,6 +61,7 @@ export interface IRundownItemBase {
         }
     };
     subitems?: Array<ISubitem>;
+    _id?: string;
 }
 
 export type IRundownItem = IRundownItemBase & IBaseRestApiResponse & ILockInfo;
--- a/scripts/extensions/broadcasting/src/interfaces.ts
+++ b/scripts/extensions/broadcasting/src/interfaces.ts
@@ -61,7 +61,6 @@ export interface IRundownItemBase {
         }
     };
     subitems?: Array<ISubitem>;
-    _id?: string;
 }
 
 export type IRundownItem = IRundownItemBase & IBaseRestApiResponse & ILockInfo;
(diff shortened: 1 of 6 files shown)


dzonidoo:
first merge ui-framework side

scripts/extensions/broadcasting/src/interfaces.ts
```diff
@@ -61,6 +61,7 @@ export interface IRundownItemBase {
         }
     };
     subitems?: Array<ISubitem>;
+    _id?: string;
```
petrjasek, line 64: that's defined in the `IBaseRestApiResponse` so should be available on `IRundownItem` type

scripts/extensions/broadcasting/src/rundowns/rundown-view-edit.tsx
```diff
@@ -549,6 +549,11 @@ export class RundownViewEditComponent extends React.PureComponent<IProps, IState
                                                             ),
                                                         });
                                                     }}
+                                                    itemId={
```
petrjasek, line 552: not sure about the name here, maybe `selectedItem` would be better

scripts/extensions/broadcasting/src/interfaces.ts
dzonidoo: i know that... but i have error in TS if i dont add in `IRundownItem` too

scripts/extensions/broadcasting/src/rundowns/components/rundown-items.tsx
```diff
@@ -30,6 +30,7 @@ interface IPropsReadOnly<T extends IRundownItem | IRundownItemBase> {
     getActions(item: T): JSX.Element;
     preview(item: T): void;
     edit(item: T): void;
+    itemId?: string | null;
```
petrjasek, line 33: I would change the name here as well

petrjasek (approved):

PR #4459, opened 2024-03-22
Code restructuring for better usability from other projects

thecalcc:
SDBELGA-768


petrjasek (approved):

tomaskikutis:
@thecalcc @petrjasek @MarkLark86 I stumbled upon [changes](https://github.com/user-attachments/assets/93185eb0-1085-4b19-8db1-5cf4f2535f77)  in `superdesk-api.d.ts` and traced it back to this PR. We are breaking the principle of not exposing implementation details there and I would like to drop those methods from the API.

It looks to me that it was done in order to [render a preview in related articles in planning](https://github.com/user-attachments/assets/05c94012-0b1c-407c-8494-898a6a06c52b). What I'm not understanding is why didn't we expose the entire preview component, instead of internal superdesk implementation details needed for building it on the planning side?

We have a [print preview component](https://github.com/user-attachments/assets/a82c4ce2-0ea1-47e4-8eb2-826fa0dfd948) which would work great here. We might do some design improvements, but we would probably want those in print preview too.

thecalcc:
> @thecalcc @petrjasek @MarkLark86 I stumbled upon [changes](https://github.com/user-attachments/assets/93185eb0-1085-4b19-8db1-5cf4f2535f77) in `superdesk-api.d.ts` and traced it back to this PR. We are breaking the principle of not exposing implementation details there and I would like to drop those methods from the API.
> 
> It looks to me that it was done in order to [render a preview in related articles in planning](https://github.com/user-attachments/assets/05c94012-0b1c-407c-8494-898a6a06c52b). What I'm not understanding is why didn't we expose the entire preview component, instead of internal superdesk implementation details needed for building it on the planning side?
> 
> We have a [print preview component](https://github.com/user-attachments/assets/a82c4ce2-0ea1-47e4-8eb2-826fa0dfd948) which would work great here. We might do some design improvements, but we would probably want those in print preview too.

To me it makes sense that in planning we receive the article in the usual format, so it should be all good to export the component itself. 

From what I checked, these functions are the most crucial to be hidden: 
`getProjectedFieldsArticle`, `getSortedFields`, `getSortedFieldsFiltered`
  
 I don't know why we would not receive a normal article in planning and simply reuse the component from client-core for the preview functionality. I think since it's client-core, then it only makes sense we have core-components and functions only inside it, and that we should not expose the internal workings of it to the rest of the projects.

PR #4465, opened 2024-04-03
templates e2e test

dzonidoo:
SDESK-7164


e2e/client/playwright/templates.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    await expect(page.locator(s('template-content', 'content-template--template 1'))).toBeVisible();
+});
+
+test('editing template', async ({page}) => {
+    await restoreDatabaseSnapshot();
+    await page.goto('/#/settings/templates');
+
+    await page.locator(s('template-content', 'content-template--story', 'template-actions')).click();
+    await page.locator(s('template-actions--options')).getByRole('button', {name: 'Edit'}).click();
```
tomaskikutis, line 23: When you are testing something that has dynamic test IDs, take your time and convert it to data-test-id/data-test-value. Especially when it's easy to do so - there are only 2 lines in old tests that you'd need to update.

e2e/client/playwright/templates.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+test('Assign template to a desk', async ({page}) => {
+    const monitoring = new Monitoring(page);
+
+    await restoreDatabaseSnapshot();
+    await page.goto('/#/workspace/monitoring');
+    await monitoring.selectDeskOrWorkspace('Finances');
+
+    await page.locator(s('content-create')).click();
+    await page.locator(s('content-create-dropdown')).getByRole('button', {name: 'More Templates...'}).click();
+    await expect(page.locator(s('content-create-dropdown')).getByRole('button', {name: 'Story 2'})).not.toBeVisible();
```
tomaskikutis, line 49: Do a search before asserting it not being visible. Otherwise simply adding more templates to the database would break your test.

e2e/client/playwright/templates.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    // assign template to the desk
+    await page.goto('/#/settings/templates');
+    await page.locator(s('template-content', 'content-template--story 2', 'template-actions')).click();
+    await page.locator(s('template-actions--options')).getByRole('button', {name: 'Edit'}).click();
+    await page.locator(s('template-edit-view', 'desks', 'desk--Finances')).first().click();
```
tomaskikutis, line 55: It would be ideal to use the `.check()` method for checkboxes. Example: `expect(page.getByLabel('abc')).check();`

This doesn't look like a checkbox, but we could use role="checkbox". If it doesn't work out easily - at least add a comment that you expect that checkbox is off and by clicking it - you are toggling it on

e2e/client/playwright/templates.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    await page.goto('/#/workspace/monitoring');
+    await monitoring.selectDeskOrWorkspace('Sports');
```
tomaskikutis, line 58: shouldn't you be checking finances desk here? because you switched template visibility for finances desk

e2e/client/playwright/templates.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    await page.locator(s('desk-config-modal', 'field--default-content-template')).selectOption({label: 'story 2'});
+    await page.locator(s('desk-config-modal')).getByRole('button', {name: 'done'}).click();
+
+    await page.goto('/#/workspace/monitoring');
+    await monitoring.selectDeskOrWorkspace('Sports');
+    await page.locator(s('content-create')).click();
+    await expect(page.locator(s('content-create-dropdown', 'default-desk-template'))).toHaveText('story 2');
+});
+
+test('Prefill template', async ({page}) => {
```
tomaskikutis, line 82: Review naming of all newly added tests in this PR. The name should start with a lowercase letter and it should read well when `test` function is read as a part of the test name. Also make test names more descriptive so it can be understood better by people who are not too familiar with the feature. For example this could be more clearly phrased `test new article prefilling with content set in template`

scripts/apps/templates/views/create-template.html
```diff
@@ -1,62 +1,64 @@
...
+                    <div class="field">
+                        <div class="control">
+                            <sd-check ng-model="template.is_public">Desk template</sd-check>
+                        </div>
+                    </div>
+
+                    <div class="field" ng-if="template.is_public !== false">
+                        <label for="template-desk" translate>Desk</label>
+                        <div class="control">
+                            <select ng-model="template.desk" ng-options="desk._id as desk.name for desk in template.desks" data-test-id="select-desk"></select>
```
tomaskikutis, line 41: Avoid data-test-id tags for inputs. Instead get inputs by label. In this case id/for isn't setup correctly, so do that and we'll improve accessibility at the same time

tomaskikutis (commented):
Good stuff Nikola :tada: I still had some comments, but tests are pretty good already.

e2e/client/playwright/page-object-models/monitoring.ts
```diff
@@ -50,4 +50,18 @@ export class Monitoring {
             }
         }
     }
+
+    async executeActionInEditor(...actionPath: Array<string>): Promise<void> {
```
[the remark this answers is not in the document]
tomaskikutis: I see you moved the helper, but I expected you would also respond to the second part of the question :)

e2e/client/playwright/monitoring.personal-space.spec.ts
```diff
@@ -60,6 +60,8 @@ test('sending an item from personal space', async ({page}) => {
         page.locator(s('article-item=personal space article 1')),
         'Send to',
     );
+    await page.locator(s('interactive-actions-panel', 'open-popover')).click();
```
tomaskikutis, line 63: how was this working before? did anything change?

e2e/client/playwright/templates.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    await page.locator(s('content-create-dropdown')).getByRole('button', {name: 'More Templates...'}).click();
+    await page.locator(s('content-create-dropdown', 'search')).fill('Story 2');
+    await expect(page.locator(s('content-create-dropdown')).getByRole('button', {name: 'Story 2'})).not.toBeVisible();
+
+    // assign template to the desk
+    await page.goto('/#/settings/templates');
+    await page.locator(s('template-content', 'content-template=story 2', 'template-actions')).click();
+    await page.locator(s('template-actions--options')).getByRole('button', {name: 'Edit'}).click();
+    await page.locator(s('template-edit-view', 'desks', 'desk--Finances')).click();
+    await page.locator(s('template-edit-view')).getByRole('button', {name: 'Save'}).click();
```
tomaskikutis, line 58: could we somehow wait to ensure that saving finished before navigating away? maybe we could wait until modal closes?

tomaskikutis (commented):
also see remaining comments that I didn't resolve

e2e/client/playwright/monitoring.personal-space.spec.ts
dzonidoo: before we had only one Desk ('Sports').. now we have plus one (two in total)... so i need to select on which one i want to send item to avoid error in some cases

e2e/client/playwright/page-object-models/monitoring.ts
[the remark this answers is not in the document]
dzonidoo: checked... works

tomaskikutis:
See one unresolved comment. There's still an input using `data-test-id="name-input"`

tomaskikutis:
all good now with code. Do update the branch by merging develop and resolve conflicts

tomaskikutis (approved):

PR #4466, opened 2024-04-04
avoid using docker-compose on github actions

petrjasek:


tomaskikutis (commented):
did docker implement its own `compose` method making docker-compose obsolete?

tomaskikutis (approved):

PR #4469, opened 2024-04-04
Broadcasting: Notifications pane going under part of the Broadcasting UI

dzonidoo:
SDESK-7221


dzonidoo:
@tomaskikutis On broadcasting is an exception because we use a react component from UI Framework there and that component SubNav contains a z-index of 1000.

tomaskikutis (approved):

PR #4501, opened 2024-04-18
Fix for editor3 performance improvements

tomaskikutis:
SDESK-7248

- [x] TODO: check authoring-react


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -542,36 +546,54 @@ export function AuthoringDirective(
...
-                            'Article cannot be published. Please accept or reject all suggestions first.',
-                        ),
+                $scope.$applyAsync(() => {
+                    $scope.loading = true;
+                });
+
+                return $q((resolve) => {
+                    // delay required for loading state to render
+                    // before possibly long operation (with huge articles)
+                    setTimeout(() => {
```
petrjasek, line 556: if you use `$timeout` protractor will know it should wait for it, not sure if that's causing any issues but just in case

PR #4508, opened 2024-04-23
Ai assistant widget

thecalcc:
Can you suggest me a good way to handle the `id` of the extension? I'm using that to track a bunch of things and I need it to be accessible from belga's repo and also client-core. One option is to put it in the config and having a constant with the id at the top. Then this config property would be accessible in clien-core as well as belga's repo, and for the future clients that want to use the ai widget setup would also be pretty easy. 

- [x] verified to be working with both versions of authoring


petrjasek:
you mean this id @thecalcc ? https://github.com/superdesk/superdesk-belga/pull/448/files#diff-2ae9b1d748ad93663c8ec84a69971adf5d1bcfcfc412d12859d6fa484cb0c805R58

scripts/core/superdesk-api.d.ts
```diff
@@ -757,6 +757,10 @@ declare module 'superdesk-api' {
                     ruleHandlers?: {[key: string]: IIngestRuleHandlerExtension};
                 };
             };
+            aiAssistant?: {
```
tomaskikutis, line 760: superdesk-api should be limited to core functionalities. AI widget is an add-on so it shouldn't go here.

Instead you should make it a regular extension and expose customization options from the extension itself. You can check how it's done with [broadcasting](https://github.com/superdesk/superdesk-insajder/blob/0b95804d6af160608b4156d0f092af579dc9e2ab/client/index.ts#L9)

tomaskikutis (commented):
Can you add a video demonstrating how the widget works?

I've reviewed the integration and after I check the video I'll review the widget code.

scripts/core/superdesk-api.d.ts
tomaskikutis: > Can you suggest me a good way to handle the id of the extension? I'm using that to track a bunch of things and I need it to be accessible from belga's repo and also client-core.

after you make it a regular extension, you won't have this problem either.

styles/sass/sf-icons.scss
```diff
@@ -166,8 +166,10 @@
 
 [class^="svg-icon-"], [class*="svg-icon-"] {
     display: inline-block;
-    width:20px;
-    height: 20px;
+    :is(.svg-icon-illustration) {
```
MarkLark86, line 169: Why the change to limit width/height to just `svg-icon-illustration`?
Looking at the history of this file, the width and height for `svg-icon` has been `20px` for 9+ years. Does this break anything?

thecalcc: Yes it does, we have new components for SVG illustrations that fritz introduced. Because of that if I don't have this condition the alignment of them breaks.

thecalcc:
@tomaskikutis here's a demo of the functionality

https://github.com/superdesk/superdesk-client-core/assets/45574768/bafa9a1f-f4ff-477b-b02d-398feac565e7

scripts/extensions/ai-widget/src/ai-assistant.tsx
```diff
@@ -0,0 +1,224 @@
...
+     */
+    loadingHeadlines: boolean;
+    loadingSummary: boolean;
+
+    headlines: Array<string>;
+    error: boolean;
+    summary: string;
+}
+
+export default function getAiAssistantWidget(superdesk: ISuperdesk, label: string) {
```
tomaskikutis, line 35: Don't use factory function. It was only used before because it wasn't possible to access
`(superdesk: ISuperdesk)` at the top level, but now it is - see hello world extension.

scripts/extensions/ai-widget/src/requests.ts
```diff
@@ -0,0 +1,12 @@
+import {IArticle, ISuperdesk} from 'superdesk-api';
+
+export interface IRequests {
+    generateHeadlines?: (article: IArticle, superdesk: ISuperdesk) => Promise<Array<string>>;
```
tomaskikutis, line 4: It should be mandatory. In `activate` method of the extension, check if it is set, and if not - display an error alert about wrong configuration and don't return anything from activate method so it doesn't get registered.

scripts/extensions/ai-widget/src/requests.ts
```diff
@@ -0,0 +1,12 @@
...
+    generateSummary?: (article: IArticle, superdesk: ISuperdesk) => Promise<string>;
+}
+
+export const requests: IRequests = {};
+
+export function setRequests(_requests: IRequests) {
```
tomaskikutis, line 10: `configure` or `setup` would be a better name. Then `IRequests` would be `IConfigurationOptions`

styles/sass/sf-icons.scss
tomaskikutis: It's not clear to me too. You should either add CSS or if existing CSS conflicts with your markup, you should use `:not` selector to make existing CSS not apply to your new markup. It seems to be that you're instead changing behavior for all existing usages that are not yours. Or am I not correct?

scripts/apps/authoring-bridge/widgets/ai-widget/ai-assistant.tsx
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    setError() {
+        this.setState({
+            error: true,
+        });
+    }
+
+    generateHeadlines() {
+        if (this.aiAssistant?.generateHeadlines == null) {
```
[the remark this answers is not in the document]
thecalcc: fyi, you commented on the previous implementation, but the new one is similar. Well, that can't be done with the design you suggested. Using this design it's possible to register the extension, but you can skip registering the custom API requests. So that's why if someone wants, they can register the extension. And actually for the future I see a benefit since since some clients might want to have e.g. summary, translations but not headlines generation.

scripts/apps/authoring-bridge/widgets/ai-widget/ai-assistant.tsx
```diff
@@ -0,0 +1,239 @@
...
+            <AuthoringWidgetLayout
+                header={(
+                    <Spacer v gap="0" alignItems="center">
+                        <AuthoringWidgetHeading
+                            widgetName={getLabel()}
+                            editMode={false}
+                        />
+                        {(this.state.summaryOpen || this.state.headlinesOpen) && (
+                            <>
+                                <div className="p-1">
```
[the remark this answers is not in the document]
thecalcc: It's a helper class from UIF. I'm not sure how you should be aware of that, I don't want to leave a comment on every class I use. Since we're aiming for reusability and control we have to use helper classes and no inline styles as much as possible.

scripts/apps/authoring-bridge/widgets/ai-widget/headlines.tsx
```diff
@@ -0,0 +1,97 @@
...
+                                    ariaValue={gettext('Copy')}
+                                    icon="copy"
+                                    onClick={() => {
+                                        navigator.clipboard.writeText(headline);
+                                    }}
+                                />
+                            </ButtonGroup>
+                        </Container>
+
+                        <ContentDivider type="dashed" margin="small" />
```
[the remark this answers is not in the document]
thecalcc: I have to put a dividing line as this is in the design file

scripts/apps/authoring-bridge/widgets/ai-widget/ai-assistant.tsx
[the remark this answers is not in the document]
tomaskikutis: In another comment I asked to make `generateHeadlines/generateSummary` mandatory. If you want to keep it optional, update the UI so if headlines/summary is not passed - buttons to generate shouldn't appear either.

And if buttons appear only when it's registered - there's no use in this function to again check if it's registered.

scripts/apps/authoring-bridge/widgets/ai-widget/ai-assistant.tsx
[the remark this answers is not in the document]
tomaskikutis: I was wondering if you left if from testing since name isn't much different from `a-1 b-2 c-3` :smile:

scripts/apps/authoring-bridge/widgets/ai-widget/headlines.tsx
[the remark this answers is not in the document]
tomaskikutis: I'm not against putting a dividing line, I'm just saying you should put it into a correct place.

```typescript
// BAD

const component1 = () => (
    <>
        <p>paragraph 1</p>
        <div className="divider" />
    </>
);

const component2 = () => (
    <p>paragraph 2</p>
);

const main = () => (
    <>
        {component1}
        {component2}
    </>
);

// GOOD

const component1 = () => (
    <p>paragraph 1</p>
);

const component2 = () => (
    <p>paragraph 2</p>
);

const main = () => (
    <>
        {component1}
        <div className="divider" />
        {component2}
    </>
);
```

scripts/extensions/ai-widget/src/ai-assistant.tsx
```diff
@@ -0,0 +1,221 @@
...
+                                                activeSection: 'summary',
+                                            });
+                                        }}
+                                    >
+                                        <SvgIconIllustration illustration="summary" />
+                                    </IllustrationButton>
+                                )}
+                            </div>
+                        );
+                    } else if (this.state.activeSection === 'headlines') {
```
tomaskikutis, line 171: Imagine a few more tabs got added, it would be a textbook definition of spaghetti code.

```
<header>
    {(() => {
        if (section === 'aaa') {
            // return ...
        } else if (section === 'bbb') {
            // return ...
        } else if (section === 'ccc') {
            // return ...
        } else if (section === 'ddd') {
            // return ...
        }
    })()}
</header>

<body>
    {(() => {
        if (section === 'aaa') {
            // return ...
        } else if (section === 'bbb') {
            // return ...
        } else if (section === 'ccc') {
            // return ...
        } else if (section === 'ddd') {
            // return ...
        }
    })()}
</body>

<footer>
    {(() => {
        if (section === 'aaa') {
            // return ...
        } else if (section === 'bbb') {
            // return ...
        } else if (section === 'ccc') {
            // return ...
        } else if (section === 'ddd') {
            // return ...
        }
    })()}
</footer>
```

Instead separate each tab into it's own component. A similar thing is implemented here https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/interactive-article-actions-panel/index-ui.tsx#L202

One difference in your case would be to persist state of closed tabs, but that's not too hard either - simply store state of all tabs in a root component.

scripts/apps/authoring-bridge/widgets/ai-widget/headlines.tsx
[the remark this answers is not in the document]
tomaskikutis: I've looked more into this and you're probably right. header/content/footer abstraction prevents you from adding a separator the way I was suggesting

tomaskikutis (commented):
I left some comments unresolved that you didn't address. Let's talk tomorrow and go through them.

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -452,9 +452,9 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
                                             />
                                         );
                                     } else if (sideWidget != null) {
-                                        return getWidgetsFromExtensions(item).find(
-                                            ({_id}) => sideWidget === _id,
-                                        ).component;
+                                        return getWidgetsFromExtensions(item).find((x) => {
```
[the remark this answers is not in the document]
thecalcc: This is a workaround which I unintentionally left, the root fix is probably different, but the issue was with opening widgets.

scripts/extensions/ai-widget/src/ai-assistant.tsx
```diff
@@ -0,0 +1,221 @@
...
+                                    >
+                                        <IconButton
+                                            size="small"
+                                            icon="arrow-left"
+                                            onClick={() => {
+                                                this.setState({
+                                                    activeSection: null,
+                                                });
+                                            }}
+                                            ariaValue={gettext('Close') + this.state.activeSection === 'headlines'
```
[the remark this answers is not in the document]
thecalcc: Makes sense, I changed it so it's proper now

scripts/extensions/ai-widget/src/ai-assistant.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+    setError() {
+        this.setState({
+            error: true,
+        });
+    }
+
+    generateHeadlines() {
+        if (configuration.generateHeadlines == null) {
+            this.setError();
```
tomaskikutis, line 52: Headlines are only generated from headlines widget. And we don't show headlines widget if we don't have the configuration. Then it means this line would never execute - so let's drop it. Same with `generateSummary`

tomaskikutis (approved):

tomaskikutis:
> verified to be working with both versions of authoring

also check that checkbox if you verified it already

tomaskikutis (approved):

PR #4516, opened 2024-05-07
Custom blocks tab

thecalcc:
BOR-12
Add custom blocks tab, fix adding and deleting vocabularies with no reload bug

### DIFF
--- a/scripts/apps/vocabularies/controllers/VocabularyConfigController.ts
+++ b/scripts/apps/vocabularies/controllers/VocabularyConfigController.ts
@@ -110,6 +110,7 @@ export function VocabularyConfigController($scope: IScope, $route, $routeParams,
     $scope.matchFieldTypeToTab = (tab, fieldType) =>
         tab === 'vocabularies' && !fieldType || fieldType &&
         (tab === 'text-fields' && fieldType === 'text' ||
+            tab === 'custom-editor-blocks' && fieldType === 'block' ||
             tab === 'date-fields' && fieldType === 'date' ||
             tab === 'urls-fields' && fieldType === 'urls' ||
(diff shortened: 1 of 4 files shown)


scripts/apps/vocabularies/controllers/VocabularyConfigController.ts
```diff
@@ -110,6 +110,7 @@ export function VocabularyConfigController($scope: IScope, $route, $routeParams,
     $scope.matchFieldTypeToTab = (tab, fieldType) =>
         tab === 'vocabularies' && !fieldType || fieldType &&
         (tab === 'text-fields' && fieldType === 'text' ||
+            tab === 'custom-editor-blocks' && fieldType === 'block' ||
```
tomaskikutis, line 113: can you rename `block` to `custom-editor-block` or `editor-block`?

scripts/apps/vocabularies/services/VocabularyService.ts
```diff
@@ -46,10 +46,11 @@ export function VocabularyService(api, $q, $filter, $rootScope) {
      * @name VocabularyService#getVocabularies
      * @public
      * @description Returns the manageable vocabularies.
+     * @param {String} vocabularyDeleted if true, fetch new vocabularies values again
      * @return {Promise} {Object} vocabularies
      */
-    this.getVocabularies = function() {
-        if (_.isNil(self.vocabularies)) {
+    this.getVocabularies = function(vocabularyDeleted) {
```
tomaskikutis, line 52: `noCache` would be more clear as an argument name. It'd even better to use an options object so usages are more clear too e.g. `getVocabularies({noCache: true})`

tomaskikutis (approved):

PR #4518, opened 2024-05-09
Make spellchecker work for multi line quotes

tomaskikutis:
BOR-13

Since multi line quotes and upcoming custom blocks are based on tables, spellchecking techninally works on tables too, except that it would spellcheck each cell separatelly which would probably degrade performance so I've disabled it.


tomaskikutis:
I finally handled the issues with unit tests. @petrjasek do review this one when you have time

petrjasek (approved):

PR #4519, opened 2024-05-14
Custom blocks tab

thecalcc:
BOR-12
BOR-5
BOR-6
BOR-7


thecalcc:
Demo:
Uploading Screen Recording 2024-05-14 at 15.16.32.mov…

thecalcc:
Sample config:

[my custom editor block.json](https://github.com/superdesk/superdesk-client-core/files/15308222/my.custom.editor.block.json)

scripts/core/editor3/directive.tsx
```diff
@@ -342,10 +349,12 @@ class Editor3Directive {
                     }
                 })();
 
-                const renderEditor3 = () => {
+                const renderEditor3 = (options?: {unmount: boolean}) => {
                     const element = $element.get(0);
 
-                    ReactDOM.unmountComponentAtNode(element);
+                    if (options?.unmount) {
```
tomaskikutis, line 355: You're funny :D Before, it was unmounting unconditionally. After your change it never unmounts, but you're passing arguments trying to trick me that you're preserving current behavior.

scripts/core/editor3/directive.tsx
```diff
@@ -519,6 +529,20 @@ class Editor3Directive {
                     renderEditor3();
                 });
 
+                $scope.$on('formattingOptions-update', (_e, data) => {
```
tomaskikutis, line 532: what about adding a watcher instead that would look at `editorFormat` changes? Then it would apply universally - also for authoring which is not the case now. It would also be safer to dispatch an action vs reinitialize the entire store.

scripts/core/editor3/directive.tsx
```diff
@@ -380,6 +389,7 @@ class Editor3Directive {
 
                     const editor3 = (
                         <Editor3
+                            uiTheme={{}}
```
tomaskikutis, line 392: do you need this or it's some code left from testing?

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
```diff
@@ -120,6 +189,17 @@ export function VocabularyEditController(
...
         delete $scope.vocabulary['_deleted'];
 
+        if ($scope.vocabulary.field_type === EDITOR_BLOCK_FIELD_TYPE) {
+            $scope.requestEditor3DirectivesToGenerateHtml.forEach((fn) => {
+                fn();
+            });
+            $scope.vocabulary.field_options = $scope.vocabulary.field_options ?? {};
+            $scope.vocabulary.field_options = {
+                ...$scope.vocabulary.field_options,
+                template: $scope.item.fields_meta?.[editorBlockFieldId].draftjsState,
```
tomaskikutis, line 199: shouldn't you also save selected formatting options?

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
```diff
@@ -175,7 +255,7 @@ export function VocabularyEditController(
             return false;
         }
 
-        if ($scope.vocabulary.field_options == null || $scope.vocabulary.field_options.allowed_types == null) {
+        if ($scope.vocabulary.field_type !== 'related_content') {
```
tomaskikutis, line 258: Are you sure this change is correct? It might be - just looks suspicious to me.

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
```diff
@@ -111,6 +126,60 @@ export function VocabularyEditController(
...
+    if ($scope.vocabulary.field_type === EDITOR_BLOCK_FIELD_TYPE) {
+        const vocabulary = $scope.vocabulary;
+
+        $scope.item = {
+            fields_meta: {
+
+                /**
+                 * Fake field, needed for compatibility with sdEditor3 directive
+                 */
+                [editorBlockFieldId]: {draftjsState: vocabulary.field_options?.template},
```
tomaskikutis, line 138: either set fields meta conditionally(when `vocabulary.field_options.template` is present) or do a fallback - empty content state converted to raw. It should never be
`fields_meta: {"editor_block_field": undefined}`

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
```diff
@@ -111,6 +126,60 @@ export function VocabularyEditController(
...
+                'comments',
+                'annotation',
+                'suggestions',
+                'table',
+                'media',
+            ]);
+
+            const formattingOptions = Object.entries(getEditor3RichTextFormattingOptions())
+                .map(([notTranslatedOption, translatedOption]) => ({value: [notTranslatedOption, translatedOption]}))
+                .filter(({value}: {value: [RICH_FORMATTING_OPTION, string]}) =>
```
tomaskikutis, line 154: When possible you should first filter and only then map, not to waste CPU cycles mapping what you're gonna drop anyway. There wouldn't be performance impact here, but there might be reputational impact if someone checks git blame :D Also it's a good habit to do it the right way so when you write back-end code eventually where it will matter, you wouldn't do such mistakes.

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
```diff
@@ -85,9 +100,9 @@ export function VocabularyEditController(
         if (angular.isDefined(response.data._issues)) {
             if (angular.isDefined(response.data._issues['validator exception'])) {
                 notify.error(gettext('Error: ' +
-                                     response.data._issues['validator exception']));
+                    response.data._issues['validator exception']));
```
tomaskikutis, line 103: if you're touching this already - let's correct it
`gettext('Error: {{message}}', {message: response.data._issues['validator exception']})`

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
```diff
@@ -32,13 +35,18 @@ interface IScope extends IScopeConfigController {
     _errorUniqueness: boolean;
     errorMessage: string;
     save: () => void;
+    formattingOptionsOnChange?: (options: Array<RICH_FORMATTING_OPTION>) => void;
```
tomaskikutis, line 38: separate custom block related props here.

```
// custom block props START
requestEditor3DirectivesToGenerateHtml: Array<() => void>;
// custom block props END
```

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
```diff
@@ -21,6 +23,7 @@ VocabularyEditController.$inject = [
 ];
 
 interface IScope extends IScopeConfigController {
+    item?: Partial<IArticle>;
```
tomaskikutis, line 26: better call it `fakeItem` and comment that it's `needed for editor3 directive(custom blocks)`

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
thecalcc: It's correct because the type that was used for checking has these fieldOptions

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
thecalcc: It can possibly only be `{fields_meta: {"editor_block_field": {draftjsState: undefined}}`, that's supported by types. Everything is properly typed here and linter doesnt fail

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
thecalcc: No, these are saved on change in `$scope.formattingOptionsOnChange`, because I need to reinitialize the editor through createEditorStore.

scripts/apps/workspace/content/views/FormattingOptionsMultiSelect.tsx
```diff
@@ -1,36 +1,20 @@
...
-    getEditor3PlainTextFormattingOptions,
-    HAS_RICH_FORMATTING_OPTIONS,
-} from '../components/get-content-profiles-form-config';
 
 interface IProps {
     value: Array<string> | null;
     fieldId: string;
     fields: Dictionary<string, IArticleField>;
     onChange(value: Array<string>, fieldId: string): void;
+    formattingOptions: Array<{value: [RICH_FORMATTING_OPTION, string]}>;
```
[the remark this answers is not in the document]
thecalcc: I agree, not only because of the one usage but because the available formattingOptions calculation doesn't happen inside the component and also my usage passes onChange with a different signature

scripts/apps/vocabularies/controllers/VocabularyEditController.tsx
tomaskikutis: I'm not suggesting the linter fails. I'm saying it's a bad idea to have such object possible.

e2e/client/playwright/utils/tree-select-driver.ts
```diff
@@ -1,44 +1,45 @@
...
-     * Get all available options.
-     */
-    async getValues(page: Page): Promise<Array<string>> {
-        await page.locator(s('open-popover')).click();
+export const treeSelectDriver = (dataTestId: string) => {
+    return {
+        /**
+         * Get all available options.
+         */
+        async getValues(page: Page): Promise<Array<string>> {
```
tomaskikutis, line 9: > 2. The method is called get getValues and your implementation is for getOptions(which doesn't exist here)

Did you see this one in the [previous comment](https://github.com/superdesk/superdesk-client-core/pull/4519#discussion_r1604661415)?

tomaskikutis (approved):
All good with code. The only thing left is to fix the tests which I'll do in my branch.

PR #4532, opened 2024-05-29
Custom editor3 blocks

tomaskikutis:
BOR-1

This PR contains code already reviewed in https://github.com/superdesk/superdesk-client-core/pull/4519.

To see only changes that are unique to this PR, merge this branch to #4519 with `--squash --no-commit` options.


scripts/apps/workspace/content/views/FormattingOptionsMultiSelect.tsx
```diff
@@ -1,48 +1,30 @@
...
-        )
-            .map(([notTranslatedOption, translatedOption]) => ({value: [notTranslatedOption, translatedOption]}));
-
-        const values = this.props.value != null
-            ? formattingOptions
-                .filter((option) => this.props.value.includes(option.value[0]))
-                .map((option) => option.value)
-            : [];
+        const lookup = new Map(
+            this.props.options.map(({value}) => value) as unknown as Array<[string, RICH_FORMATTING_OPTION]>,
```
thecalcc, line 15: You've reversed the types in `Array<[string, RICH_FORMATTING_OPTION]`, according to props it's `[RICH_FORMATTING_OPTION, string]`

scripts/apps/workspace/content/views/FormattingOptionsMultiSelect.tsx
```diff
@@ -1,48 +1,30 @@
...
+        );
+        const values: Array<[string, string]> = (this.props.value ?? []).map((id) => [id, lookup.get(id)]);
```
thecalcc, line 17: Are you sure type of `values` is correct here? Might be that it's `Array<[RICH_FORMATTING_OPTION, string]>`

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -581,8 +581,9 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
          * add much performance overhead if it was replaced unconditionally, but I don't want to break it
          * nor spend time on testing so I'm keeping it as is for now.
          */
-        const dropAreaEnabled =
-            this.props.editorFormat.includes('media') || this.props.editorFormat.includes('embed articles');
+        const dropAreaEnabled = this.props.editorFormat.some(
+            (option) => formattingOptionsThatRequireDragAndDrop.has(option),
+        );
```
thecalcc, line 586: Might be good to extract this as a boolean at this point and reuse in `scripts/core/editor3/components/BaseUnstyledComponent.tsx:54`

scripts/core/editor3/components/custom-block.scss
```diff
@@ -0,0 +1,46 @@
...
+    border-start-end-radius: 2px;
+
+    background: var(--sd-editor-colour__controls-border);
+
+    &:hover {
+        background: gray;
+    }
+
+    &:active {
+        background:var(--sd-colour-primary)
```
thecalcc, line 44: add `;` at the end of this line

scripts/core/editor3/components/media/MediaBlock.tsx
```diff
@@ -22,6 +22,25 @@ function getTranslationForAssignRights(value) {
...
+
+interface IReduxStateProps {
+    readOnly: boolean;
+    showTitle: boolean;
+}
+
+interface IDispatchProps {
+    cropImage(entityKey, entityData, options): void;
+    removeBlock(blockKey): void;
+    changeCaption(entityKey, newCaption, field): void;
```
thecalcc, line 38: can you add functions' prop types here?

scripts/core/editor3/components/toolbar/index.tsx
```diff
@@ -177,15 +183,49 @@ class ToolbarComponent extends React.Component<IProps, IState> {
...
+                                (option) => availableOptions.has(option as RICH_FORMATTING_OPTION),
+                            );
+                        }
+                        default:
+                            assertNever(activeCell.additional);
+                        }
+                    })()}
+                />
+            );
+        } else {
```
thecalcc, line 228: Can you improve nesting for this whole block and inside it? Looks like we can at least avoid using else and get one level up. I think that's fine here since we only have one condition, so it doesn't sacrifice readability

scripts/core/editor3/html/to-html/AtomicBlockParser.ts
```diff
@@ -218,4 +225,37 @@ export class AtomicBlockParser {
...
         return html;
     }
+
+    parseCustomBlock(data: IEditor3CustomBlockData): string {
+        if (this.disabled.indexOf('table') > -1) {
+            return '';
+        }
+
+        function getHighestHeadingText(el: HTMLElement): string | null {
+            for (const tag of ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']) {
```
thecalcc, line 235: can you type these `h1`, `h2`, `h3`,...

scripts/core/editor3/reducers/table.tsx
```diff
@@ -56,19 +73,28 @@ const addTable = (state, data) => {
...
+                nextCells.push([]);
             }
         });
 
-        return {data: {cells, numRows, numCols, withHeader}};
+        return {
+            data: {
+                ...tableData,
+                cells: nextCells,
+                numRows: numRows + 1,
```
thecalcc, line 95: Don't leave magic numbers, I know it's just adding an additional role, but would be easier to read if it's something like + `ONE_ADDITIONAL_ROW`

scripts/core/editor3/reducers/table.tsx
```diff
@@ -94,41 +128,47 @@ const removeRow = (state) =>
...
-                        }
+    processCells(state, (tableData, activeCell) => {
+        const {cells, numCols} = tableData;
+        const {j} = activeCell;
 
-                        let orig = jj;
+        return {
+            data: {
+                ...tableData,
+                numCols: tableData.numCols + 1,
```
thecalcc, line 138: No magic numbers

scripts/core/editor3/reducers/table.tsx
```diff
@@ -94,41 +128,47 @@ const removeRow = (state) =>
...
+                cells: cells.map((_, ii) =>
```
thecalcc, line 139: this `ii` looks a bit odd. Maybe do firstIndex and for `jj` do `secondIndex`

scripts/core/helpers/utils.tsx
```diff
@@ -31,6 +31,10 @@ export function pick<T extends IOnlyStringKeys, K extends keyof T>(obj: T, ...ke
     return picked;
 }
 
+export function getObjectEntriesGeneric<KEY extends string, VALUE>(obj: any): Array<[KEY, VALUE]> {
+    return Object.entries(obj) as Array<[KEY, VALUE]>;
+}
+
```
thecalcc, line 37: Unused code

thecalcc (commented):
All good, small changes only

scripts/core/editor3/components/Editor3Component.tsx
tomaskikutis: It requires an argument so it can't be a boolean. It could be a function, but other usage is a bit different, so I think having a constant for formatting options is fine.

scripts/core/editor3/components/media/MediaBlock.tsx
tomaskikutis: Functions have arguments, not props ;) I've added props from the data in proptypes and it wasn't that detailed. I didn't run that feature, just improved props so I don't know the types of arguments. I did add a few of the ones that were easy to understand from usage.

scripts/core/editor3/reducers/table.tsx
tomaskikutis: It's not magic here, because the purpose is clear and it couldn't work with any other number. It makes sense to use constants when it's arbitrary numbers e.g. items per page in paginated content.

scripts/core/editor3/reducers/table.tsx
tomaskikutis: I didn't write this code, just refactored arguments. Another thing is that inside active cell there's a convention that `i` is row and `j` is column. When working with matrices you want short index names so ii/jj is quite a good choice.

scripts/core/helpers/utils.tsx
tomaskikutis: There's an exception to the unused code rule for utilities. I also wrote some e2e helpers(addFormattingOptionToContentProfile) that I'm not using at the moment, but will probably need later.

thecalcc:
Fix CI

thecalcc (approved):

PR #4538, opened 2024-06-11
Fix saving when item closed with changes

thecalcc:
SDESK-7295

Tested on authoring react and angular.


thecalcc:
@tomaskikutis I'm not sure why the changes for `requestEditor3DirectivesToGenerateHtml` appear in the AuthoringToolbarDirective. They should be there since you did the performance improvements, no?

thecalcc:
Just double checked, you reverted it in `1cd183e7c36796fe96a4c69c1b2ae6a6d7ac388d` commit. If that's the case, should I keep them or revert? If it's the second, then implementation would totally change I think.

thecalcc:
I've reverted my changes, leaving your latest in the AuthoringTopbarDirective. The fix is different, and looks better. It looks like you've achieved what I wanted to do in the first place. I simply made the call to compute html before passing scope to other functions.

tomaskikutis (approved):

PR #4539, opened 2024-06-11
do not show publishing tab for articles where desk_type is authoring and global noPublishOnAuthoringDesk config is set

tomaskikutis:
SDESK-7294


petrjasek (approved):

tomaskikutis:
I've checked locally and unit tests are passing so I'll ignore CI result here.

PR #4540, opened 2024-06-12
generate html from editor3 on autosave

petrjasek:
SDESK-7273


scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -850,9 +850,8 @@ export function AuthoringDirective(
             // default to true
             $scope.firstLineConfig.wordCount = $scope.firstLineConfig.wordCount ?? true;
 
-            $scope.autosave = function(item, timeout) {
-                $scope.dirty = true;
-                angular.extend($scope.item, item); // make sure all changes are available
+            const _autosave = debounce((timeout) => {
+                $scope.requestEditor3DirectivesToGenerateHtml?.forEach((fn) => fn());
```
tomaskikutis, line 854: Since `requestEditor3DirectivesToGenerateHtml` not existing would break the feature, we shouldn't even check if it's defined and let it crash if it's not set.

petrjasek: yep it was crashing a test but will check why, it was easier to add it :sweat_smile:

petrjasek: fixed it

tomaskikutis (approved):

PR #4541, opened 2024-06-12
Add new 'getCustomVacabulary' function in superdesk-api [SDESK-7268]

devketanpro:
Add new `getCustomVacabulary` func in superdesk-api which we used in supedesk-analytics.
https://github.com/superdesk/superdesk-analytics/pull/151


tomaskikutis:
Do we have a clear definition of what exactly is a "custom" vocabulary? I'd rather expose a method to get all vocabularies from [here](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/api/vocabularies.ts#L6)

devketanpro:
Custom vocabularies means which vocab we can add via metadata settings. 
In the backend we fetch this type of vocab by query :- "field_type": None, "service": {"$exists": True}

tomaskikutis:
Since I was looking at vocabulary code I also pushed some improvements.

Regarding your code, one improvement I made is to keep API slimmer is to only expose `isCustomVocabulary` method. I also exposed a method to get all vocabularies so you can later filter them using `isCustomVocabulary`.

Let me know if you have other improvements in mind.

tomaskikutis (approved):

PR #4551, opened 2024-06-18
Fix widget pinning state getting reset

thecalcc:
SDESK-7303

### DIFF
--- a/scripts/apps/authoring/widgets/views/authoring-widgets.html
+++ b/scripts/apps/authoring/widgets/views/authoring-widgets.html
@@ -5,7 +5,7 @@
         'right-extended': active.extended
     }"
 >
-    <!-- RIGHT SIDEBAR START -->
+    <!-- AVAILABLE WIDGETS LIST START -->
     <div class="sd-sidetab-menu sd-sidetab-menu--right" ng-if="item">
         <ul class="navigation-tabs" data-test-id="navigation-tabs">
             <li ng-repeat="widget in widgets | orderBy:'order'"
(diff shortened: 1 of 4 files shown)


scripts/apps/authoring/widgets/widgets.ts
```diff
@@ -377,7 +378,11 @@ function WidgetsManagerCtrl(
             $scope.active.afterClose($scope);
         }
 
-        $scope.active = null;
+        if ($scope.pinnedWidget != null) {
```
tomaskikutis, line 381: shouldn't it close in either case - whether pinned or not?

scripts/apps/authoring/widgets/views/authoring-widgets.html
```diff
@@ -1,44 +1,11 @@
...
-<div class="widget-wrapper"
+<div
+    class="widget-wrapper"
     ng-class="{
         'active-right': active,
-        'right-extended': active.extended}">
+        'right-extended': active.extended
+    }"
+>
+    <!-- RIGHT SIDEBAR START -->
```
tomaskikutis, line 8: Widget itself is also a part of the sidebar, so the name isn't that good. I'd suggest "available widgets list start/end"

scripts/apps/authoring/widgets/widgets.ts
thecalcc: No logic is correct, such that if you have a pinned widget but then open another one on top of it when you close this "unpinned"/"overlayed" one, the pinned is back to active. If that's not the case and you're just closing a widget then it'd get closed as normally.

tomaskikutis (approved):

PR #4552, opened 2024-06-18
use spec reporter for karma

tomaskikutis:


thecalcc (approved):

tasks/options/karma.js
```diff
@@ -3,15 +3,15 @@ module.exports = {
...
     },
     single: {
-        reporters: 'dots',
+        reporters: 'spec',
     },
     watch: {
         singleRun: false,
         autoWatch: true,
-        reporters: 'dots',
+        reporters: ['progress'],
```
thecalcc, line 14: Is this correct, previously this was a string and now an array?

tomaskikutis: I changed it for testing, will set it back to a string.

petrjasek (approved):

PR #4558, opened 2024-06-20
SDBELGA-830

thecalcc:
SDBELGA-830 

Related improvements

### DIFF
--- a/scripts/core/ui/components/SelectUser.tsx
+++ b/scripts/core/ui/components/SelectUser.tsx
@@ -115,7 +115,7 @@ export class SelectUser extends SuperdeskReactComponent<IPropsSelectUser, IState
                 label={gettext('Select a user')}
                 inlineLabel={true}
                 labelHidden={true}
-                getItems={(searchString) => {
+                getItems={this.props.getItems != null ? this.props.getItems : (searchString) => {
                     this.abortController?.abort();
                     this.abortController = new AbortController();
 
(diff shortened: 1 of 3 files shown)


thecalcc:
https://github.com/superdesk/superdesk-client-core/assets/45574768/263e53be-cbb8-4a59-bf87-17ab3c59611a

scripts/core/ui/components/SelectUser.tsx
```diff
@@ -119,11 +121,16 @@ export class SelectUser extends SuperdeskReactComponent<IPropsSelectUser, IState
...
-                        searchString != null && searchString.length > 0
-                            ? getUserSearchMongoQuery(searchString)
-                            : {},
-                    );
+                    const desk = sdApi.desks.getDeskById(this.props.deskId);
+                    const deskMemberIds = (desk?.members ?? []).map((member) => member.user);
+                    const query = {
+                        $and: [
+                            this.props.deskId != null ? {_id: {$in: deskMemberIds}} : {},
+                            searchString != null && searchString.length > 0
```
tomaskikutis: if `searchString` is empty, return early and don't query at all.

Also adjust how query is made so you never end up with empty objects inside the query. Use `Array.push` to add things conditionally.

thecalcc: I've tested it and it doesn't make any difference, do you think it has some other impact?

thecalcc: Also if you want to do it this way can you give me a better example? If I push `{"$and":[]}` server responds with a 500. How do you suggest I implement this so that I don't need to check for `query.$and.length` and send the query conditionally?

tomaskikutis:
```
const desk = sdApi.desks.getDeskById(this.props.deskId);
const deskMemberIds = (desk?.members ?? []).map((member) => member.user);

if ((searchString ?? '').length < 1) {
    return Promise.resolve([]);
}

const query = {
    $and: [
        getUserSearchMongoQuery(searchString),
    ],
};

if (this.props.deskId != null) {
    query.$and.push({_id: {$in: deskMemberIds}});
}
```

Also try using `prepareSuperdeskQuery` to have it type safe.

tomaskikutis (approved):

tomaskikutis (approved):

PR #4571, opened 2024-07-11
improve notifications toggle for parent-child relationship [SDBELGA-818]

devketanpro:
SDBELGA-818

- [x] Check default settings before merging, should all be set to true by default. Check if planning-ext assignments registration works as well.
- [x] Before merging, refactor existing notification handling in the panel/modal such that it matches the new type signatures

Note: having `scope.preferences` and `scope.extensionsNotifications` both is necessary since we have to update both separately and together while having them in completely separate components (first variable is responsible for the whole email notifications group updating, second only for notifications from extensions, being passed to a react component).

Also `preferencesService.registerUserPreference(key, 1);` is the core for how notifications from extensions should be done. This prevents any strings hardcoding and allows for modularity.

**Demo (a notification registered as an extension coming from client-core and one coming from the planning extension)**

https://github.com/user-attachments/assets/e3ad8a28-1e3d-4ae6-8b89-53fe71a480ec

### DIFF
--- a/scripts/apps/search/index.ts
+++ b/scripts/apps/search/index.ts
@@ -10,6 +10,7 @@ import {MultiActionBarReact} from 'apps/monitoring/MultiActionBarReact';
 import {reactToAngular1} from 'superdesk-ui-framework';
 import {SearchPanelWidgets} from './components/search-panel-widgets';
 import {PreviewSubject} from './components/preview-subject';
+import EmailNotificationPreferences from 'apps/users/components/EmailNotificationPreferences';
 
 angular.module('superdesk.apps.search.react', [
     'superdesk.apps.highlights',
@@ -71,6 +72,14 @@ angular.module('superdesk.apps.search', [
             MultiActionBarReact,
             ['articles', 'hideMultiActionBar', 'getCoreActions', 'compact'],
         ),
+    )
+
+    .component(
+        'sdNotificationsList',
+        reactToAngular1(
+            EmailNotificationPreferences,
+            ['toggleEmailNotification'],
+        )
     )
 
     .component(
--- a/scripts/apps/search/index.ts
+++ b/scripts/apps/search/index.ts
@@ -78,7 +78,7 @@ angular.module('superdesk.apps.search', [
         'sdNotificationsList',
         reactToAngular1(
             EmailNotificationPreferences,
-            ['toggleEmailNotification'],
+            ['toggleEmailNotification', 'preferences'],
         ),
     )
 
--- a/scripts/apps/search/index.ts
+++ b/scripts/apps/search/index.ts
@@ -79,7 +79,7 @@ angular.module('superdesk.apps.search', [
         reactToAngular1(
             EmailNotificationPreferences,
             ['toggleEmailNotification'],
-        )
+        ),
     )
 
     .component(
--- a/scripts/apps/search/index.ts
+++ b/scripts/apps/search/index.ts
@@ -10,7 +10,7 @@ import {MultiActionBarReact} from 'apps/monitoring/MultiActionBarReact';
 import {reactToAngular1} from 'superdesk-ui-framework';
 import {SearchPanelWidgets} from './components/search-panel-widgets';
 import {PreviewSubject} from './components/preview-subject';
-import EmailNotificationPreferences from 'apps/users/components/EmailNotificationPreferences';
+import {EmailNotificationPreferences} from 'apps/users/components/EmailNotificationPreferences';
 
 angular.module('superdesk.apps.search.react', [
     'superdesk.apps.highlights',
@@ -75,7 +75,7 @@ angular.module('superdesk.apps.search', [
     )
 
     .component(
-        'sdNotificationsList',
+        'sdEmailNotificationsList',
         reactToAngular1(
             EmailNotificationPreferences,
             ['toggleEmailNotification', 'preferences'],
--- a/scripts/apps/users/components/EmailNotificationPreferences.tsx
+++ b/scripts/apps/users/components/EmailNotificationPreferences.tsx
@@ -6,7 +6,7 @@ interface IProps {
     preferences?: {[key: string]: any};
 }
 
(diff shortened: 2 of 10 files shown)


tomaskikutis (commented):
"Mark for user" feature is built as an extension, thus all it's code should live inside the extension folder. If we want to allow configuration of extension notifications in settings page - we need to implement an extension point and make the extension use it.

We already have `IExtensionActivationResult['contributions']['notifications']` extension point which allows handling of notifications. That extension point now doesn't have a label, only ID and handler function. If we adjust the API to add a label:

```
notifications?: {
    [id: string]: {
        label: string;
        handler: (notification) => {
            body: string;
            actions: Array<{label: string; onClick(): void;}>;
        };
    };
};
```

then we could get all configured notifications from extensions and also allow disabling them from the settings page.

The same could then be applied to a planning extension too which we have inside a planning repo.

devketanpro:
> "Mark for user" feature is built as an extension, thus all it's code should live inside the extension folder. If we want to allow configuration of extension notifications in settings page - we need to implement an extension point and make the extension use it.
......

Hey @tomaskikutis , Thank you for your feedback.

Regarding the recent changes I made, they were specifically aimed at ensuring correct toggle behavior for user preferences, such as enabling/disabling email notifications. This functionality is directly related to user preferences and their settings.
The recent updates were not related to "Mark for user" feature managed by extensions functionalities. Instead, they were solely focused on enhancing the user experience with regards to their preference settings.

tomaskikutis (changes requested):
I understand perfectly the purpose of the PR.

`"mark_for_user:notification"` - this string is a part of mark for user feature and should only be used inside `scripts/extensions/markForUser` folder. Same with `'assignment:notification'` - it should only be mentioned inside planning repo.

devketanpro:
> I understand perfectly the purpose of the PR.
> 
> `"mark_for_user:notification"` - this string is a part of mark for user feature and should only be used inside `the'scripts/extensions/markForUser` folder. Same with `'assignment:notification'` - it should only be mentioned inside planning repo.

alright, I will make the changes. Thanks!

thecalcc:
Before merging, refactor existing notification handling in the panel/modal such that it matches the new type signatures

thecalcc:
Check default settings before merging, should all be set to true by default. Check if planning-ext assignments registration works as well.

thecalcc:
Demo:

https://github.com/user-attachments/assets/e3ad8a28-1e3d-4ae6-8b89-53fe71a480ec

thecalcc:
Note: having `scope.preferences` and `scope.extensionsNotifications` both is necessary since we have to update both separately and together (first variable is responsible for the whole email notifications group updating, second only for notifications from extensions, being passed to a react component)

scripts/apps/search/index.ts
```diff
@@ -73,6 +74,14 @@ angular.module('superdesk.apps.search', [
         ),
     )
 
+    .component(
+        'sdNotificationsList',
```
tomaskikutis: `sdEmailNotificationsList` would be a better name. Or is it not limited to emails?

scripts/apps/users/components/EmailNotificationPreferences.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import {CheckGroup, Checkbox} from 'superdesk-ui-framework/react';
+
+interface IProps {
+    toggleEmailNotification: (notificationId: string) => void;
+    preferences: {[key: string]: any};
+}
+
+export default class EmailNotificationPreferences extends React.PureComponent<IProps> {
```
tomaskikutis: don't do default exports. Without them, it's easier to track usages and it also enforces consistent naming.

scripts/apps/users/directives/UserPreferencesDirective.ts
```diff
@@ -56,7 +54,53 @@ export function UserPreferencesDirective(
...
+                    ...(scope.preferences[notificationId] ?? {}),
+                    enabled: !(scope.preferences[notificationId]?.enabled ?? false),
+                };
+
+                scope.extensionsNotifications[notificationId] = {
+                    ...scope.extensionsNotifications[notificationId],
+                    enabled: scope.preferences[notificationId].enabled,
+                };
+
+                const notificationsForGroupAreOff = Object.keys(scope.extensionsNotifications)
```
tomaskikutis: `.some` and `.every` are functionally the same, just reverse. In terms of performance it is the same. Which to use depends on which term more accuratelly describes what you're doing. In this case it seems you're inverting the logic - `const allFalse = arr.some(x => x === true)`. While technically correct, the following matches naming better: `const allFalse = arr.every(x => x === false)`

scripts/apps/users/directives/UserPreferencesDirective.ts
```diff
@@ -56,7 +54,53 @@ export function UserPreferencesDirective(
              */
 
             scope.preferencesLoaded = false;
-            var orig; // original preferences, before any changes
+            var orig: {[key: string]: any}; // original preferences, before any changes
+
+            scope.extensionsNotifications = {};
```
tomaskikutis: What about calling this `emailNotificationsFromExtensions`?. I saw there are other kinds of notifications too. We could say that there are multiple notification types supported in extensions, but since you are interating this object in other methods and don't use any conditionals, then it makes sense to limit this object only to email notifications. If we decide in the future to modify it, it will be clear this is only used for email type and it would be easier to refactor thanks to self-documenting naming.

scripts/apps/search/index.ts
thecalcc: Good point, in theory this component `EmailNotificationPreferences` can be used for any sort of notifications list, but there's no other one besides email ones now.

scripts/apps/users/directives/UserPreferencesDirective.ts
thecalcc: I had it that way, just renamed it with its current name so it's shorter, I use it in lots of places and thought it's a good compromise

tomaskikutis (approved):

PR #4572, opened 2024-07-11
fix media metadata editor changes not being saved

petrjasek:
it was calling authoring editor instances to generate html which then overwrote any changes done in modal.

SDESK-7304


scripts/apps/archive/views/upload.html
```diff
@@ -35,7 +35,7 @@
                 <div class="upload__info-icon"></div>
                 <h3 class="upload__info-heading" translate>Drag Your Files Here</h3>
                 <div class="upload__info-label" translate>or</div>
-                <button ng-click="invokeImagesInput()" class="btn btn--hollow upload__info-button" translate>Select them from folder</button>
+                <button ng-click="invokeImagesInput()" data-id="select-file-btn" class="btn btn--hollow upload__info-button" translate>Select them from folder</button>
```
tomaskikutis: should be `data-test-id`

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -882,6 +875,10 @@ export function AuthoringDirective(
                 _autosave(timeout);
             };
 
+            $scope.syncEditor = () => {
```
tomaskikutis: `generateHtml` would be a better name.

scripts/apps/authoring/authoring/directives/AuthoringDirective.ts
```diff
@@ -270,11 +267,11 @@ export function AuthoringDirective(
             /**
              * Create a new version
              */
-            $scope.save = function() {
+            $scope.save = function({reloadEditor3 = false} = {}) {
```
tomaskikutis: `generateHtml = true` would be better naming. Reloading usually means resetting state and executing the same code again, which isn't the case here.

e2e/client/test-files/iptc-photo.jpg
tomaskikutis, line 1: Why not use one of existing pictures in the same folder? Not sure about licensing of this one. It says copyright iptc.org :)

scripts/apps/authoring/authoring/directives/ArticleEditDirective.ts
```diff
@@ -342,6 +343,8 @@ export function ArticleEditDirective(
                  * @description Opens the Change Image Controller to modify the image metadata.
                  */
                 scope.editMedia = (defaultTab = 'view') => {
+                    scope.syncEditor();
```
tomaskikutis: why do we generate HTML before editing media? Could you add a comment explaining it?

e2e/client/playwright/page-object-models/media-editor.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export class MediaEditor {
+    private page: Page;
+
+    constructor(page: Page) {
+        this.page = page;
+    }
+
+    field(field: string) {
+        return this.page.locator(s('media-metadata-editor', 'field--' + field)).getByRole('textbox');
```
tomaskikutis: Don't use concatentation inside the value of `data-test-id`. That way it is difficult to find where the tags are in the code. And the other way around - if one sees tags, it's diffucult to find where they are used. I've been experimenting with using `data-test-value` attributes in combination with `data-test-id` and it works great so far.

```
// markup
<div class="field body"
     data-test-id="authoring-field"
     data-test-value="body_html"
>
    ...
</div>

// selector
s('authoring', 'authoring-field=body_html')
```

You can check other examples if you grep for `s\('.+=` in the codebase.

e2e/client/playwright/page-object-models/monitoring.ts
```diff
@@ -60,4 +60,13 @@ export class Monitoring {
             }
         }
     }
+
+    async uploadMedia(): Promise<void> {
```
tomaskikutis: `openMediaUploadView` would be a better name. Upload media implies that after calling it, it would start uploading.

e2e/client/playwright/page-object-models/media-editor.ts
```diff
@@ -0,0 +1,29 @@
+import {Page} from '@playwright/test';
+import {s} from '../utils';
+import path from 'path';
+
+export class MediaEditor {
```
tomaskikutis, line 4: Can you add return types to methods? Typescript can imply it, but it's useful to have it typed to document intent.

Also scoping would be useful to add to all methods. Wrapping the entire modal with
`data-test-id="media-editor"` would do.

e2e/client/playwright/page-object-models/monitoring.ts
```diff
@@ -60,4 +60,13 @@ export class Monitoring {
...
+        await this.page.locator(s('content-create')).click();
+        await this.page.locator(s('content-create-dropdown')).getByRole('button', {name: 'Upload media'}).click();
+    }
+
+    listArticle(headline: string) {
```
tomaskikutis: I'm wondering whether having page object model methods which only serve to get a locator is too much abstraction for tests. I used to hate the equivalent of page object models in protractor, because protractor setup was very tedious to debug and page object models would add a separate layer. Since this one is easy to debug I don't mind I guess. The only complaint remaining is naming. Usually functions are called `{verb}{noun}` and you're using `{noun}{noun}` or only `{noun}` which is confusing. A better naming scheme would be `get{noun}Locator()`. e.g. `mediaEditor.getFieldLocator('headline')`, `monitoring.getArticleLocator('test123')`

e2e/client/playwright/authoring.picture.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    await monitoring.selectDeskOrWorkspace('Sports');
+
+    await monitoring.uploadMedia();
+
+    await mediaEditor.selectUploadFile('iptc-photo.jpg');
+
+    await expect(mediaEditor.field('headline')).toContainText('The Headline');
+
+    await mediaEditor.field('headline').fill('');
```
tomaskikutis: doesn't `.clear()` work?

e2e/client/playwright/page-object-models/authoring.ts
```diff
@@ -21,4 +21,18 @@ export class Authoring {
             .getByRole('button', {name: actionPath[actionPath.length - 1]})
             .click();
     }
+
+
+    async openMediaMetadataEditor() {
+        await this.page.locator(s('image-overlay')).hover();
+        await this.page.locator(s('edit-metadata')).click();
```
tomaskikutis: We should add some scoping - both to ensure uniqueness of selectors and to make it easier for developer reading the test to understand which part of the UI your selector is located in.

e.g. `s('authoring', 'edit-metadata')` would already be much better.

Another important thing is that there can be multiple media fields so we should pass a field value:

```
async openMediaMetadataEditor({field}: {field: string}) {
    await this.page.locator(s(`authoring-field=${field}`, 'image-overlay')).hover();
    await this.page.locator(s(`authoring-field=${field}`, 'edit-metadata')).click();
}
    ```

File: e2e/client/playwright/authoring.picture.spec.ts [hunk 11]
```diff
@@ -0,0 +1,50 @@
...
+import {Authoring} from './page-object-models/authoring';
+
+test.setTimeout(30000);
+
+/**
+ * upload a picture
+ * edit metadata
+ * test metadata changes from modal are visible in the editor
+ */
+test('edit picture metadata in modal', async ({page}) => {
```
tomaskikutis: All other playwright tests use naming convention where `test` function is read as a part of the name
e.g. `test('editing picture metadata in modal)`, `test('correcting with unsaved changes'`

I'd say let's keep it. I'd also not use the word "modal" in the name since it's not a detail that impacts anything that's being tested. How about `test('media metadata editor')`.

e2e/client/playwright/page-object-models/authoring.ts
```diff
@@ -21,4 +21,18 @@ export class Authoring {
...
+    }
+
+    field(field: string) {
+        return this.page.locator(s('authoring', 'field--' + field)).getByRole('textbox');
+    }
+
+    async waitForAutosave() {
```
tomaskikutis: I see this is not used in the end. Can we drop it? It's a hack anyway, in 90% of cases tests should work solely with the DOM. API mocking can be useful to test integration with 3rd party services, but other than that we should avoid interaction with HTTP requests.

.eslintrc.js
```diff
@@ -44,6 +44,8 @@ module.exports = Object.assign({}, sharedConfigs, {
         '@typescript-eslint/ban-types': 'off',
         '@typescript-eslint/no-inferrable-types': 'off',
         '@typescript-eslint/no-this-alias': 'off',
+        '@typescript-eslint/no-explicit-any': 'off',
```
tomaskikutis, line 47: what's this for? I've removed those lines you added and lint passed without issues.

e2e/client/test-files/iptc-photo.jpg
petrjasek: those have no metadata embedded, I wanted to test also how is that form populated automatically

.eslintrc.js
petrjasek: it passes but generates 600+ warnings which makes it pretty hard to find anything in the output

e2e/client/test-files/iptc-photo.jpg
tomaskikutis: I see. We could also modify one of the existing images to put metadata on it.

.eslintrc.js
tomaskikutis: FYI eslint has the `--quiet` flag that hides all warnings so only errors are displayed.

e2e/client/playwright/page-object-models/authoring.ts
```diff
@@ -21,4 +21,13 @@ export class Authoring {
             .getByRole('button', {name: actionPath[actionPath.length - 1]})
             .click();
     }
+
+    async openMediaMetadataEditor(field: string = 'media') {
```
tomaskikutis: Could you remove the default? It's a looks messy and might imply there's some default behavior too in the code being tested when there is none. Also it'd be better to have named arguments so one doesn't need to go to definition to see what the string means.

tomaskikutis (commented):
Check out unresolved comments. I did resolve the ones you addressed.

e2e/client/playwright/page-object-models/authoring.ts
petrjasek: it's actually not media field withing text item but just media preview in picture item, so there will be only 1 and it has no specific name

.eslintrc.js
petrjasek: also wanted to avoid it on github actions, I don't see those warnings useful anyhow if that passes the CI

tomaskikutis: yeah, I agree with disabling it.

e2e/client/playwright/page-object-models/authoring.ts
tomaskikutis: I see. Isn't it the same component as it would be in a text item having media? If it is, then we should be thinking from a perspective that this is a authoring page object model -> authoring view has a content profile -> content profiles can have multiple media fields. Image content profile can only have one media field, but since this is page object model for authoring, not for image profile, we can't make an assumption here.

petrjasek: we have different component for associated media and for media within picture/audio/video item, they have some actions in common but in text item you can drag and drop there other picture items, you can remove the associated picture, that's not available here

tomaskikutis: I see. If they are different components, the question then is how to communicate this using naming/structure/comments of page object models so it's obvious by reading a function signature.

I think having a separate page object model for authoring view with non-text content profiles would be best. Then we wouldn't allow arguments to `openMediaMetadataEditor` and similar functions and it would be clear that only one item is allowed.

tomaskikutis (commented):
Can you check the remaining unresolved comments?

e2e/client/playwright/page-object-models/media-editor.ts
petrjasek: we already have those ids there and it's used by other tests, we can change that all over but wouldn't do it in the 2.7.1 bugfix but rather separately

e2e/client/playwright/page-object-models/monitoring.ts
petrjasek: for locators I don't mind such names, it's easier to read imo when it's like `monitoring.article.click()`, but don't mind changing it if that will be the rule

e2e/client/playwright/page-object-models/media-editor.ts
tomaskikutis: I see. I thought you were adding new ones. Then yes, let's not refactor other tests now.

e2e/client/playwright/page-object-models/monitoring.ts
tomaskikutis: We can leave this up to developers writing tests for now and in the future if it still annoys us we could decide on a rule.

To me it looks like if we have `data-test-id` attributes, tests are pretty stable and flexible already so I'd not enforce on developers a rule to extract locators to page object models. I'd rather go for speed, use that playwright test generator in most cases and only use page object models for multi-step actions like selecting from a tree menu.

e2e/client/playwright/page-object-models/media-editor.ts
```diff
@@ -0,0 +1,19 @@
...
+        this.page = page;
+    }
+
+    field(field: string): Locator {
+        return this.page.locator(s('media-metadata-editor', field)).getByRole('textbox');
+    }
+
+    async saveMetadata(): Promise<void> {
+        await this.page.locator(s('media-editor', 'apply-metadata-button')).click();
+        await this.page.locator(s('change-image', 'done')).click();
```
tomaskikutis, line 17: is `change-image` image selector still inside "media-editor"? If it is, it'd be good to scope it:
```
await this.page.locator(s('media-editor', 'change-image', 'done')).click();
```

The reason being not uniqueness, but documentation - ability to quickly understand the hierarchy without having to look it up inside dev tools.

tomaskikutis (approved):

e2e/client/playwright/page-object-models/media-editor.ts
petrjasek: those are on the same level

PR #4573, opened 2024-07-11
[SDESK-7332] and [SDESK-7333]

fritzSF:
Plus additional layout fixes in some settings modals


tomaskikutis (commented):
You have removed `superdesk-ui-framework` from `package.json` so all tests were failing. I reverted changes to `package.json` and `package-lock.json`

tomaskikutis (approved):

PR #4574, opened 2024-07-16
Widget state persistance

thecalcc:
SDBELGA-847

TODO:
- [x] Look into possible different solutions for `closedThroughAction`
- [x] Add comment for the functionality
- [x] Look into state getting reset, when:
         a) widget is pinned and you click to close it
         b) widget is pinned or not, you switch back and forth between views, and switch widgets
- [x] Final check against develop
- [x] Update UIF so `disabled` take effect ([version](https://github.com/superdesk/superdesk-ui-framework/pull/872))
- [ ] Add a playwright test
- [x] Verify persistence works across all widgets in React Authoring

**What this PR aims to address:** When a user switches back and forth between full view width and normal width, if there's a pinned widget that has state changes, the changes should persist between the two view modes. In this PR that's being handled using a HOC (`WidgetStatePersistanceHOC`) and a variable (`closedThroughAction`). The HOC handles the logic for state persistence using `widgetState` variable and `ref` (to get child component state - current widget). Inside authoring when a widget gets loaded we handle the `initialState` of the widget using `closedThroughAction` variable - If a widget has been closed by the user on purpose, once the same widget is reopened, state does get reset. If that's not the case (switching between the two article view modes) we count that as the widget not being closed through a direct user interaction, thus retrieving the old state of the widget and loading it with the widget rendering. 
 
**Side issues this PR fixes:** Widget `pinned` state in the different scenarios, and widget persistence using `preferencesService`. 

**Additional solutions:** Another solution for `setTimeout` for the initialState in authoring-react - Add a variable `renderCount` and based on the count delete the localStorage value, then reset `renderCount`. Problems with that - adds more complexity, if logic changes we'd need to adjust that on top of everything.


scripts/appConfig.ts
```diff
@@ -50,7 +50,7 @@ export const debugInfo = {
     translationsLoaded: false,
 };
 
-export let authoringReactEnabledUserSelection = (JSON.parse(localStorage.getItem('auth-react') ?? 'false') as boolean);
+export let authoringReactEnabledUserSelection = true;
```
tomaskikutis: this should not be commited

scripts/apps/authoring-react/article-widgets/demo-widget.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import {IArticleSideWidget, IArticle, IExtensionActivationResult} from 'superdesk-api';
+import {IArticleSideWidget, IArticle, IExtensionActivationResult, IArticleSideWidgetComponentType} from 'superdesk-api';
```
tomaskikutis: drop unused imports

scripts/core/superdesk-api.d.ts
```diff
@@ -635,7 +635,7 @@ declare module 'superdesk-api' {
         label: string;
         order: number; // Integer. // NICE-TO-HAVE: manage order in the UI instead of here
         icon: string;
-        component: React.ComponentType<IArticleSideWidgetComponentType>;
+        component: React.ComponentClass<IArticleSideWidgetComponentType, any>;
```
tomaskikutis: state type is optional. There's no use to pass `any`.

scripts/extensions/ai-widget/src/ai-assistant.tsx
```diff
@@ -95,6 +95,10 @@ export class AiAssistantWidget extends React.PureComponent<IArticleSideWidgetCom
             return {
                 activeSection: null,
             };
+        case undefined:
```
tomaskikutis, line 96:
```
switch (section) {
case null:
case undefined:
    return {
        activeSection: null,
    };
// ...
```
You can handle multiple cases with one code block.

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -328,9 +332,17 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
         };
 
         widgetReactIntegration.pinWidget = () => {
+            const pinWidget = !(this.props.sideWidget?.pinned ?? false);
```
tomaskikutis: `<verb><noun>` naming is used for functions. `widgetPinned` should be used for a boolean value.

scripts/apps/authoring/widgets/widgets.ts
```diff
@@ -140,6 +140,8 @@ export const widgetReactIntegration: IWidgetIntegration = {
     disableWidgetPinning: false,
 };
 
+export let closedThroughAction = {closed: true};
```
tomaskikutis: From what I understood by reading a comment in the HOC component, you want to preserve the state if the component re-mounts. But not if unmounting was expected due to component being closed by user action.

I'm wondering if this could handled a bit cleaner, with less usages of this special variable. Now the variable has to be set when genuine user actions occur - to prevent persisting the state when it must not be persisted. What about if we never persist it, except during special actions. Then we'd only need to add special behavior only to special circumstances - not to all regular user actions.

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -327,10 +350,26 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
...
                     activeTab={this.state.sideWidget?.id}
                     onActiveTabChange={(val) => {
+                        if (val == null && closedThroughAction.closed == false) {
+                            closedThroughAction.closed = true;
+                        }
+
+                        const widgetFromPreferences = sdApi.preferences.get(PINNED_WIDGET_USER_PREFERENCE_SETTINGS);
+
+                        const isWidgetPinned = (() => {
+                            if (widgetFromPreferences._id == val) {
```
tomaskikutis: shouldn't we only take preferences into account to come up with a initial value in the constructor?

scripts/apps/authoring-react/widget-persistance-hoc.tsx
```diff
@@ -0,0 +1,50 @@
...
+import {IArticleSideWidgetComponentType} from 'superdesk-api';
+
+export let widgetState = {};
+
+interface IProps {
+    children: (
+        widgetRef: RefObject<React.PureComponent<IArticleSideWidgetComponentType>>,
+    ) => JSX.Element | Array<JSX.Element>;
+    sideWidgetId: string;
+    pinned: boolean;
```
tomaskikutis: unused property

scripts/apps/dashboard/widget-react.tsx
```diff
@@ -25,11 +49,20 @@ export class WidgetReact extends React.PureComponent<IProps, IState> {
 
         return (
             <Component
+                ref={this.widgetRef}
                 key={key}
                 article={this.props.article}
                 initialState={(() => {
                     const localStorageWidgetState = JSON.parse(localStorage.getItem('SIDE_WIDGET') ?? 'null');
 
+                    if (localStorageWidgetState == null && closedThroughAction.closed === false) {
```
tomaskikutis: I think you should adjust the order a bit so it's easier to understand. First you should check `closedThroughAction`, and if state needs to be restored you do that first thing and return. Reading local storage should only happen in the "else branch" (or after a `return`)

scripts/apps/authoring-react/widget-persistance-hoc.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+        // Reset widgetState if widget was closed through a function, or
+        // if it wasn't pinned and got closed from re-rendering
+        if (closedThroughAction.closed === true) {
+            delete widgetState[this.props.sideWidgetId];
+        }
+    }
+
+    render(): React.ReactNode {
+        return this.state.mounted ? this.props.children(this.widgetRef) : <></>;
```
tomaskikutis: use `null` if your intention is not to render anything. Use a fragment only if you need to return multiple sibling elements without wrapping it.

tomaskikutis (commented):
* Add a playwright test
* Add documentation as a comment somewhere describing the issue and your solution

scripts/apps/authoring/widgets/widgets.ts
thecalcc: It can probably be flipped yes, I'll look into it

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -327,10 +350,22 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
                 <Nav.SideBarTabs
                     activeTab={this.state.sideWidget?.id}
                     onActiveTabChange={(val) => {
+                        if (val == null && closedThroughAction.closed == false) {
+                            closedThroughAction.closed = true;
+                        }
+
+                        const isWidgetPinned = (() => {
+                            if (this.state.sideWidget?.id == val && this.state.sideWidget.id != null) {
```
tomaskikutis: `this.state.sideWidget?.id == val` is double equals instead triple intentional?

thecalcc: Not intentional

scripts/apps/authoring/widgets/widgets.ts
thecalcc: Because of the nature (specific cases where this variable has to be set)  of the fix, I'm not so sure now this can be improved, while using this approach with the variable. And currently I can't think of another way besides using a variable, and handling logic like this in the authoring itself.

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
tomaskikutis: can you use triple equals then?

tomaskikutis (commented):
take a look at 2 remaining unresolved comments

scripts/apps/dashboard/widget-react.tsx
thecalcc: This is depending on what we want to prioritise. In theory there can't be a state where we have both `SIDE_WIDGET` and `widgetState` for the same widget, but in practice invalid states can occur. That's why I'm checking if the local storage `SIDE_WIDGET` is empty while also checking `closedThroughAction`. That way I prioritise `SIDE_WIDGET` value over everything.

thecalcc:
"Versions & History" and "Comments" widget are not the usual implementations of widgets. 

Nested components with multiple levels of state management, and more. I'd say time wasted is not worth it now, we can make a ticket to check it when working on authoring-react.

thecalcc:
Lint failing because UIF isn't yet updated

scripts/core/superdesk-api.d.ts
```diff
@@ -216,7 +216,7 @@ declare module 'superdesk-api' {
 
         sideWidget: null | {
             id: string;
-            pinned: boolean;
+            pinned?: boolean;
```
tomaskikutis, line 219: why are you making it optional?

scripts/core/superdesk-api.d.ts
```diff
@@ -2084,6 +2085,7 @@ declare module 'superdesk-api' {
     export interface IPropsWidgetHeading {
         widgetName: string;
         editMode: boolean;
+        widgetId: string;
```
tomaskikutis, line 2098: Why do you need widget id? How about using label is unique identifier? It's shouldn't be supported to have multiple widgets using the same name anyway.

scripts/apps/authoring/widgets/widgets.ts
```diff
@@ -140,6 +140,11 @@ export const widgetReactIntegration: IWidgetIntegration = {
     disableWidgetPinning: false,
 };
 
+/**
+ * Used for controlling when a widget is closed intentionally by a user action
+ */
+export let closedOnRender = {closed: false};
```
tomaskikutis: Since the object is only neccessary for mutation to work, it would be better to call the object property "value", because the real name is already used in the constant.

I don't think `closedOnRender` is a good name either. What does it have to do with rendering? `closedIntentionally` sounds better. Also comment on how it may get closed unitentionally.

scripts/apps/authoring-react/widget-persistance-hoc.tsx
```diff
@@ -0,0 +1,51 @@
...
+        widgetRef: RefObject<React.PureComponent<IArticleSideWidgetComponentType>>,
+    ) => JSX.Element | Array<JSX.Element>;
+    sideWidgetId: string;
+}
+
+interface IState {
+    mounted: boolean;
+}
+
+// Functionality to persist widget state between re-renders. Solves issues related to widget state being lost when switching between different authoring views.
```
tomaskikutis: > Solves issues related to widget state being lost

You've spent a lot of time fixing well and know it well at the moment. Could you list which issues exactly used to occur, and switching between which views?

I think it's only full-width switching? Or also widgets?

scripts/extensions/ai-widget/src/ai-assistant.tsx
```diff
@@ -84,14 +86,24 @@ export class AiAssistantWidget extends React.PureComponent<IArticleSideWidgetCom
         this.inactiveTabState = {};
         this.getDefaultState = this.getDefaultState.bind(this);
         this.setSection = this.setSection.bind(this);
+
+        const computeInitialState = () => {
+            if (this.props.initialState.currentTab != null) {
```
tomaskikutis: why is conditional needed? I'd expect it would work the same regardless of state structure

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -323,14 +346,29 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
...
-                    activeTab={this.state.sideWidget?.id}
+                    disabled={sideWidget?.pinned}
+                    activeTab={sideWidget?.id}
                     onActiveTabChange={(val) => {
+                        if (val == null && closedOnRender.closed == true) {
+                            closedOnRender.closed = false;
+                        }
+
+                        const isWidgetPinned = (() => {
+                            if (sideWidget?.id === val && sideWidget.id != null) {
```
tomaskikutis: can `sideWidget` be `null`? You are accessing it twice and once you use optional chaining and second time not.

scripts/apps/dashboard/widget-react.tsx
tomaskikutis: Makes sense, but it would be better to improve the code, so logical order follows code order. If local storage is prioritized, it should be in the code first as well. i.e.

```

if (localStorage) {
    // ...
} else if (restoreState) {
    // ...
} else {
    return undefined;
}
```

scripts/core/superdesk-api.d.ts
thecalcc: Because it can be `null`. Type was wrong

scripts/core/superdesk-api.d.ts
thecalcc: I need it, in order to calculate properly the `pinned` state in another component. Using labels as unique IDs doesn't sound right to me, that's why I added `widgetId`

scripts/extensions/ai-widget/src/ai-assistant.tsx
thecalcc: It's needed because of how state is saved through places. Sometimes through "SIDE_WIDGET" the currentTab is only saved. And other times when saved through `widget-react.tsx` the full state is saved. We could make the "SIDE_WIDGET" save the whole state, or we can also make a custom parser for each widget, to process its state accordingly. You might quickly say option 1, for now I think it'd be best too, but for the future we should consider option 2 as well because of nesting state inside the widgets.

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
thecalcc: Yes it can be `null` of course. Yes, I am, I had some reason behind that. `val` can also be null, that's why I added the second condition. If `val` is `null` another branch needs to execute.

scripts/core/superdesk-api.d.ts
tomaskikutis: I see.

scripts/core/superdesk-api.d.ts
tomaskikutis: > doesn't sound right

Doesn't sound like an argument to me :smile:

I won't ask you to undo this, but let's not do this in the future. If we have a value that is already unique it doesn't make sense to add another. It might be said that labels might not be unique - but they should in this and all similar cases. To enforce it, I would rather add a check on appStart, throw an error and refuse to start the app if some values that are supposed to be unique - are not. If we don't do this, the app technically works, but users see 2 options to click that have the same name, but do different things - which also must not happen.

scripts/extensions/ai-widget/src/ai-assistant.tsx
tomaskikutis: > It's needed because of how state is saved through places. Sometimes through "SIDE_WIDGET" the currentTab is only saved.

Could you save the entire state there? It's best if we can use whatever state structure we need in each widget and it doesn't impact code outside of the widget. Would that work? In that case I don't think we need any custom parsers for anything. Or am I missing something?

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
tomaskikutis: If it can be `null` shouldn't you use optional chaining too in the second part of the condition where you are reading the `id`?

Also if you use `&&` operator, order doesn't matter. And if one part of the condition is checking against `null` it makes more sense to do it first- use optional chaining - and for the second part you don't need optional chaining then `sideWidget?.id != null && sideWidget.id === val`
 
Another issue is that when `val` is `null` your `setState` call should be `setState({sideWidget: null})`. If you set `sideWidget` to something other than `null`, `id` is mandatory.

```
this.setState({
    sideWidget: {
        id: val, // <-- val might be null, `id` is mandatory. You can't use it here without a runtime check
        pinned: isWidgetPinned,
    },
});
```

You should also call it `nextWidgetId`, `val` is confusing.

scripts/extensions/ai-widget/src/ai-assistant.tsx
thecalcc: Would work totally, I already did that, custom parsers would be for future proofing I meant to say, for now of course we don't need them

scripts/core/superdesk-api.d.ts
thecalcc: Ok, I can remove the ID and use name. Also if I do that I'll need to do extra modifications so that `getPinnedWidget` in `authoring-react.tsx` can return the widget name. Currently it only can access `id` since that's what's stored in state.

scripts/apps/authoring/widgets/widgets.ts
tomaskikutis: > Also comment on how it may get closed unitentionally.

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
tomaskikutis: the issue I pointed out to `id` being set to `null` is still not addressed.

scripts/apps/dashboard/widget-react.tsx
```diff
@@ -37,9 +63,15 @@ export class WidgetReact extends React.PureComponent<IProps, IState> {
                         localStorage.removeItem('SIDE_WIDGET');
 
                         return initialState;
-                    } else {
-                        return undefined;
+                    } else if (closedIntentionally.value === true) {
+                        const prevWidgetState = widgetState[this.props.widget.active._id];
+
+                        if (prevWidgetState != null) {
```
tomaskikutis, line 69: not all code paths return a value. add an `else` with `return undefined` just to be explicit

tomaskikutis (approved):

PR #4576, opened 2024-07-22
Fix notifications on cancel

thecalcc:
SDBELGA-818

Reverts changes when a user clicks cancel, instead of save inside user preferences view.


tomaskikutis:
> Fixes on cancel not working correctly.

On which cancel? Could you describe more clearly what wasn't working?

tomaskikutis (approved):

PR #4577, opened 2024-07-23
update mark_for_user notification perefrence name [SDBELGA-818]

devketanpro:


thecalcc (approved):

tomaskikutis (changes requested):
let me think a bit before we merge this

thecalcc:
@tomaskikutis backend related changes

tomaskikutis:
The way I think it should be, is that we have 1 notification. Let's say "mark for user" and the extension exposes only that. On client-core side, in settings we could manage whether we want to receive only email, only in-app. The extension should only expose that a notification exists and client core would take care of via which channels it should be sent.

Could we register it as a single notification instead of 3?

What do you think @petrjasek

petrjasek:
yep that makes sense, maybe it should be more like:
```ts
{
    name: 'mark_for_user',
    label: '...',
    email: true,
    desktop: true,
}
```

tomaskikutis:
What would `email, desktop` options mean? Whether a particular channel is available to be chosen in settings? When wouldn't it be possible?

petrjasek:
yep, but not sure actually if there would be any where only 1 makes sense

tomaskikutis:
I'd say let's go without them and we will expand the API if we ever need it in the future.

thecalcc:
> I'd say let's go without them and we will expand the API if we ever need it in the future.

Don't we need it now in order for disable all desk notifications to work?

tomaskikutis:
No. Extension will provide the id/label of a notification and then it's up for the core code to display notification delivery options.

tomaskikutis:
@petrjasek could you guide @devketanpro on how to adjust the back-end so we support registering a single notification and are able to deliver it using different channels? Or do you think we can't do it that easily and would need to still use 2 notifications on the back-end?

petrjasek:
so how would we store it in preferences, under single `notifications` object like:
```ts
{"notifications": {"mark_for_user": {"email": false, "desktop" true}}}
```
?

tomaskikutis:
Yep, the one you proposed would do.

petrjasek:
ok will do the backend changes

petrjasek:
btw there is this websocket notification for the extension, that should be sent no matter what's the user preference right?

tomaskikutis:
Websocket message about a new notification? No point to send it if desktop notification is disabled. If technically difficult to not send it, we can continue sending and simply not show it on the client.

petrjasek:
seems like there is `item:marked` and `item:unmarked` sent, so I'm wondering if that's needed for some UI refreshing or it's used only for desktop notifications

tomaskikutis:
`'item:marked'` is used to check if re-fetching needs to happen so we'll have to keep it for now as we don't have any alternative way to get notified. websockets come close, but they don't provide values, only field names.

tomaskikutis:
> websockets come close, but they don't provide values, only field names.

I meant the generic `resouce:update` message

petrjasek:
btw would the UI handle undefined values as enabled?

tomaskikutis:
At the moment I see it doesn't check message payload at all - since we now have 2 different notifications for marked/unmarked. If we change it, we'll have to adjust the code a bit and we'd be able to apply a default too.

petrjasek:
did the backend changes in https://github.com/superdesk/superdesk-core/pull/2652, can you handle the frontend @tomaskikutis @thecalcc ?

PR #4582, opened 2024-07-31
Update datetime formatting

dzonidoo:
SDESK-7350

### DIFF
--- a/scripts/appConfig.ts
+++ b/scripts/appConfig.ts
@@ -12,6 +12,14 @@ if (appConfig.shortTimeFormat == null) {
     appConfig.shortTimeFormat = 'HH:mm'; // 24h format
 }
 
+if (appConfig.view.dateformat == null) {
+    appConfig.view.dateformat = 'MM/DD'; // 24h format
+}
+
+if (appConfig.view.timeformat == null) {
+    appConfig.view.timeformat = 'hh:mm'; // 24h format
+}
+
 if (appConfig.ui == null) {
     appConfig.ui = {};
 
--- a/scripts/appConfig.ts
+++ b/scripts/appConfig.ts
@@ -13,11 +13,11 @@ if (appConfig.shortTimeFormat == null) {
 }
 
 if (appConfig.view.dateformat == null) {
-    appConfig.view.dateformat = 'MM/DD'; // 24h format
+    appConfig.view.dateformat = 'MM/DD';
 }
 
 if (appConfig.view.timeformat == null) {
-    appConfig.view.timeformat = 'hh:mm'; // 24h format
+    appConfig.view.timeformat = 'hh:mm';
 }
 
 if (appConfig.longDateFormat == null) {
--- a/scripts/appConfig.ts
+++ b/scripts/appConfig.ts
@@ -20,6 +20,10 @@ if (appConfig.view.timeformat == null) {
     appConfig.view.timeformat = 'hh:mm'; // 24h format
 }
 
+if (appConfig.longDateFormat == null) {
+    appConfig.longDateFormat = 'LLL';
+}
+
 if (appConfig.ui == null) {
     appConfig.ui = {};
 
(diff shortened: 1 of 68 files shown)


tomaskikutis (commented):
fix tests

scripts/appConfig.ts
```diff
@@ -12,6 +12,14 @@ if (appConfig.shortTimeFormat == null) {
     appConfig.shortTimeFormat = 'HH:mm'; // 24h format
 }
 
+if (appConfig.view.dateformat == null) {
+    appConfig.view.dateformat = 'MM/DD'; // 24h format
```
tomaskikutis: your comment doesn't make sense since days of the month don't come in 24h format

scripts/appConfig.ts
```diff
@@ -12,6 +12,14 @@ if (appConfig.shortTimeFormat == null) {
...
+}
+
+if (appConfig.view.timeformat == null) {
+    appConfig.view.timeformat = 'hh:mm'; // 24h format
```
tomaskikutis: Previously `HH:mm` was commented as being "24h format". Are you sure this is 24h format too?

scripts/apps/archive/services/ArchiveService.ts
```diff
@@ -81,8 +81,7 @@ export function ArchiveService(desks, session, api, $q, search, $location) {
      *  @return {Object} the list of archive items
      */
     this.getRelatedItems = function(item, fromDateTime) {
-        var beforeDateTime = fromDateTime || moment().subtract(1, 'days')
-            .format(appConfig.view.dateformat);
+        var beforeDateTime = formatDate(fromDateTime || moment().subtract(1, 'days'));
```
tomaskikutis: as you can see this function returns `api.query`. Do not change this usage since it is being used for a query, not outputting to user interface.

scripts/apps/authoring/authoring/directives/AuthoringHeaderDirective.ts
```diff
@@ -194,8 +195,7 @@ export function AuthoringHeaderDirective(
                 scope.missing_link = false;
                 if (scope.item.slugline && scope.item.type === 'text') {
                     // get the midnight based on the default timezone not the user timezone.
-                    var fromDateTime = moment().tz(appConfig.default_timezone)
-                        .format(appConfig.view.dateformat);
+                    var fromDateTime = formatDate(moment());
```
tomaskikutis: revert this too, date is being used for an API call not UI output

scripts/core/datetime/datetime.ts
```diff
@@ -106,23 +107,36 @@ export function scheduledFormat(__item: IArticle): {short: string, long: string}
...
+    const _date = datetime.str,
+        _time = datetime.moment.format(appConfig.view.timeformat);
 
-    let short = isSameDay(datetime, now) ? '@ '.concat(_time) : _date.concat(' @ ', _time);
+    let short = isSameDay(datetime.moment, now) ? '@ '.concat(_time) : _date.concat(' @ ', _time);
 
     return {
         short: short,
-        long: longFormat(datetime),
+        long: longFormatDate(datetime.moment),
```
tomaskikutis: You're replacing a usage of `longFormat` here, which is fine. Also see if you can replace all usages of that function and remove the function completelly.

scripts/appConfig.ts
dzonidoo: no is not... i just miss the comment

tomaskikutis:
You were saying @dzonidoo you are waiting review for this. I don't see any code changes after my last review

dzonidoo:
@tomaskikutis I didnt say that i was waiting for a review, but that we are in the review phase

tomaskikutis (commented):
add a ticket number
fix e2e

tomaskikutis (approved):

PR #4589, opened 2024-08-05
Refactor all notifications from extensions and UI toggling such that …

thecalcc:
SDBELGA-818

- [x] Registered notifications are displayed with value `true` even if a user doesn't have them saved in preferences
- [x] Email notification group toggling works and does toggle all sub notifications no matter if they haven't been set before by the user, or if they have been
- [x] Saving works in both cases
- [x] Canceling works in both cases
- [x] Desktop notifications aren't shown in the notifications sidebar if a user has toggled them off
- [ ] Email notifications aren't received if a user has toggled them off

### DIFF
--- a/scripts/apps/users/components/EmailNotificationPreferences.tsx
+++ b/scripts/apps/users/components/EmailNotificationPreferences.tsx
@@ -3,7 +3,7 @@ import {CheckGroup, Checkbox} from 'superdesk-ui-framework/react';
 
 interface IProps {
     toggleEmailNotification: (notificationId: string) => void;
-    preferences?: {[key: string]: any};
+    preferences?: {[key: string]: {label: string; email?: boolean; default?: boolean;}};
 }
 
 export class EmailNotificationPreferences extends React.PureComponent<IProps> {
@@ -17,7 +17,7 @@ export class EmailNotificationPreferences extends React.PureComponent<IProps> {
                         onChange={() => {
                             this.props.toggleEmailNotification(key);
                         }}
-                        checked={value?.enabled ?? value?.default ?? false}
+                        checked={value?.email ?? value?.default ?? false}
                     />
                 ))}
             </CheckGroup>
--- a/scripts/apps/users/components/EmailNotificationPreferences.tsx
+++ b/scripts/apps/users/components/EmailNotificationPreferences.tsx
@@ -3,7 +3,7 @@ import {CheckGroup, Checkbox} from 'superdesk-ui-framework/react';
 
 interface IProps {
     toggleEmailNotification: (notificationId: string) => void;
-    preferences?: {[key: string]: {label: string; email?: boolean; default?: boolean;}};
+    preferences?: {[key: string]: {label: string; email: boolean;}};
 }
 
 export class EmailNotificationPreferences extends React.PureComponent<IProps> {
@@ -17,7 +17,7 @@ export class EmailNotificationPreferences extends React.PureComponent<IProps> {
                         onChange={() => {
                             this.props.toggleEmailNotification(key);
                         }}
-                        checked={value?.email ?? value?.default ?? false}
+                        checked={value.email}
                     />
                 ))}
             </CheckGroup>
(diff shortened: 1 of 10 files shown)


scripts/core/superdesk-api.d.ts
```diff
@@ -706,16 +706,7 @@ declare module 'superdesk-api' {
     }
 
     interface IEmailNotification {
```
tomaskikutis: interface is empty and unused, remove it

scripts/extensions/markForUser/src/extension.tsx
```diff
@@ -48,8 +48,7 @@ const extension: IExtension = {
                     },
                     notifications: {
                         'item:marked': {
-                            type: 'desktop',
-                            label: gettext('open item'),
+                            name: gettext('Mark for User'),
```
tomaskikutis, line 51: does this unified notification notify both of marking and unmarking? Check with if API exposes this information and coordinate with Petr if not

scripts/apps/users/components/EmailNotificationPreferences.tsx
```diff
@@ -3,7 +3,7 @@ import {CheckGroup, Checkbox} from 'superdesk-ui-framework/react';
 
 interface IProps {
     toggleEmailNotification: (notificationId: string) => void;
-    preferences?: {[key: string]: any};
+    preferences?: {[key: string]: {label: string; email?: boolean; default?: boolean;}};
```
tomaskikutis: shouldn't this be `desktop` insead of `default`?
```
{label: string; email?: boolean; desktop?: boolean;}
                                 ^
```

scripts/apps/users/directives/UserPreferencesDirective.ts
```diff
@@ -58,27 +59,31 @@ export function UserPreferencesDirective(
 
             scope.emailNotificationsFromExtensions = {};
 
-            scope.buildNotificationsFromExtensions = function() {
+            scope.buildNotificationsFromExtensions = function(rebuild: boolean) {
```
tomaskikutis: `rebuild` argument not used

scripts/apps/users/directives/UserPreferencesDirective.ts
```diff
@@ -58,27 +59,31 @@ export function UserPreferencesDirective(
...
+                                default: true,
+                                label: gettext('Send {{ name }} notifications', {name: value.name}),
+                            };
                     }
                 }
             };
 
             scope.buildNotificationsFromExtensions();
 
+            scope.buildNotificationsFromExtensions(false);
```
tomaskikutis: there is the same call above

scripts/apps/users/directives/UserPreferencesDirective.ts
```diff
@@ -87,19 +92,15 @@ export function UserPreferencesDirective(
             };
 
             scope.toggleEmailNotification = function(notificationId: string) {
```
tomaskikutis, line 84: Shouldn't there also be a "toggleDesktopNotification" function?

scripts/apps/users/components/EmailNotificationPreferences.tsx
thecalcc: `default` is a different value, which I take as an `email` default, if email wasn't set. I agree things can be improved in a lot of ways here. As for `desktop` I haven't included that since we practically don't have granular desktop notifications control for now.

scripts/apps/users/directives/UserPreferencesDirective.ts
thecalcc: In the future when we need to use it yes. This is used purely for rendering any "sub" notifications that need the individual control. For desktop we only have the main toggle, which shouldn't be touched.

tomaskikutis (commented):
let's verify that unified notification works (mark/unmark) including enabling/disabling

scripts/extensions/markForUser/src/extension.tsx
thecalcc: Yes, works for marking and unmarking. Only difference between the two previously was the body message, handler was indeed the same one.

tomaskikutis (approved):

PR #4597, opened 2024-08-13
fix protractor e2e

petrjasek:
SDESK-7363

with recent webdriver


tomaskikutis (approved):
Which driver version did you reproduce it with?

petrjasek:
`127.0.6533.99`

PR #4599, opened 2024-08-13
accessibility fix for widget pinning

petrjasek:


scripts/core/superdesk-api.d.ts
```diff
@@ -3717,5 +3717,9 @@ declare module 'superdesk-api' {
         /** original source of the data, eg. wikidata */
         original_source?: string;
         parent?: string;
+
+        // AI generated subjects
+        creator?: string;
```
tomaskikutis: let's overwrite types in the extension instead

image

thecalcc (approved):

PR #4607, opened 2024-08-20
fix duplicated items in saved search on desk

petrjasek:
it was not using the config to filter out other versions in monitoring saved searches, only output stage.

SDESK-7369

### DIFF
--- a/e2e/client/playwright.config.ts
+++ b/e2e/client/playwright.config.ts
@@ -13,7 +13,7 @@ import path from 'path';
 export default defineConfig({
     testDir: './playwright',
     /* Run tests in files in parallel */
-    fullyParallel: true,
+    fullyParallel: false,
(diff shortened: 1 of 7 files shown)


e2e/client/playwright.config.ts
```diff
@@ -13,7 +13,7 @@ import path from 'path';
 export default defineConfig({
     testDir: './playwright',
     /* Run tests in files in parallel */
-    fullyParallel: true,
+    fullyParallel: false,
```
thecalcc, line 16: why are we no longer gonna do parallel?

thecalcc (approved):

e2e/client/playwright.config.ts
petrjasek: because that's causing tests to fail when there are 2 tests modifying data

PR #4608, opened 2024-08-21
Fix logic for shortcut buttons (#4606)

thecalcc:
SDCP-809

### DIFF
--- a/scripts/api/article.ts
+++ b/scripts/api/article.ts
@@ -74,9 +74,7 @@ function canPublish(item: IArticle): boolean {
         return false;
     }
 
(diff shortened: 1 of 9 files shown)


petrjasek:
it's better to cherry-pick it and push it there so when we merge the branch later back to develop there won't be conflicts

petrjasek (approved):

thecalcc:
@petrjasek CI failing here, you did make some fixes yesterday to that right? i wonder if backporting them here as well would do the job

petrjasek:
let me check, but those errors are different

PR #4609, opened 2024-08-21
Publish and continue var name

thecalcc:
SDCP-809

PR #4610, opened 2024-08-22
Fix logic for shortcut buttons (#4606)

petrjasek:
cherry picking from develop

replaces https://github.com/superdesk/superdesk-client-core/pull/4608


thecalcc (approved):

thecalcc:
we need the next one as well i think https://github.com/superdesk/superdesk-client-core/pull/4609

PR #4617, opened 2024-09-03
Fix prop access

thecalcc:
SDBELGA-872

### DIFF
--- a/e2e/client/playwright/internal-destinations.spec.ts
+++ b/e2e/client/playwright/internal-destinations.spec.ts
@@ -1,0 +1,24 @@
+import {test, expect} from '@playwright/test';
+import {restoreDatabaseSnapshot, s} from './utils';
+import {login} from 'end-to-end-testing-helpers';
+
+test.describe('internal destinations', async () => {
+    test('display and removal of active filters', async ({page}) => {
+        await login('admin', 'admin');
+        await restoreDatabaseSnapshot();
+        await page.goto('/#/settings/internal-destinations');
+
+        await expect(page.locator(s('list-page--filters-active', 'tag-label'))).toHaveCount(0);
+        await page.locator(s('toggle-filters')).click();
+
+        await page.locator(s('list-page--filters-form', 'gform-input--desk')).click();
+        await page.locator(s('list-page--filters-form', 'gform-input--desk')).getByRole('button', {name: 'Sports'}).click();
+        await page.locator(s('list-page--filters-form', 'filters-submit')).click();
+        await expect(page.locator(s('list-page--filters-active', 'tag-label'))).toHaveCount(1);
+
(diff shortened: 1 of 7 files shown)


dev-devlegion:
yep thats it

dzonidoo:
yep thats it

dzonidoo (approved):

tomaskikutis (approved):

e2e/client/playwright/internal-destinations.spec.ts
```diff
@@ -0,0 +1,24 @@
+import {test, expect} from '@playwright/test';
+import {restoreDatabaseSnapshot, s} from './utils';
+import {login} from 'end-to-end-testing-helpers';
+
+test.describe('internal destinations', async () => {
+    test('display and removal of active filters', async ({page}) => {
+        await login('admin', 'admin');
```
tomaskikutis: You don't need to login. Once database snapshot is restored, user is already logged in.

e2e/client/playwright/internal-destinations.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+        await page.locator(s('toggle-filters')).click();
+
+        await page.locator(s('list-page--filters-form', 'gform-input--desk')).click();
+        await page.locator(s('list-page--filters-form', 'gform-input--desk')).getByRole('button', {name: 'Sports'}).click();
+        await page.locator(s('list-page--filters-form', 'filters-submit')).click();
+        await expect(page.locator(s('list-page--filters-active', 'tag-label'))).toHaveCount(1);
+
+        const activeFilter = page.locator(s('list-page--filters-active', 'tag-label')).first();
+
+        await page.getByText(await activeFilter.textContent()).click();
```
tomaskikutis: Would it be "Sports"? No need to make it dynamic in tests. Might even be more clear when it's hardcoded.

e2e/client/specs/internal_destinations_spec.ts
```diff
@@ -155,28 +155,26 @@ describe('internal destinations & generic-page-list', () => {
         expect(el(['gform-output--name'], null, items.get(1)).getText()).toBe('charlie');
     });
 
-    it('can display and remove active filters', () => {
-        var items = els(['list-page--items', 'internal-destinations-item']);
-
-        browser.wait(ECE.hasElementCount(items, 3));
+    fit('can display and remove active filters', () => {
```
tomaskikutis: Was this the test that was failing? This is the one that's quite good compared to other protractor tests. Did you find the root cause of why was it failing?

thecalcc: This was the test yes. The root cause I suspect is in the autocomplete component re-rendering multiple times, so after you get it with a selector in the test the reference can't be used after that same component re-renders. This is why we were getting a stale element exception

PR #4620, opened 2024-09-04
Event and Planning profile editor doesn't display Groups in Fields

dzonidoo:
STT-74

### DIFF
--- a/scripts/core/ui/components/style.scss
+++ b/scripts/core/ui/components/style.scss
@@ -32,9 +32,16 @@ time.Datetime {
 }
 
 .sd-list-item-group {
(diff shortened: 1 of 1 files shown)


tomaskikutis (approved):

tomaskikutis (changes requested):
I take back my approval :D Why is planning-related code added to client-core rather than planning?

Could you explain what is causing the issue?

dzonidoo:
i can show u...

dzonidoo:
We are closing this pull request because the CSS code is specific to Planning and should be added to planning repo...

PR #4622, opened 2024-09-05
remove media_description field

tomaskikutis:
SDESK-7362

@petrjasek we already started the conversation about it a while ago 
 - https://sourcefabric.slack.com/archives/D9191P9FD/p1651505039824909
 
 I did try to add it to a picture item in mongo, but couldn't get it to display - which makes sense given it's not used anywhere but these 3 places I'm removing it from. It's be best if you could write an upgrade script to drop it from existing content profiles so I don't have to exclude it manually there.
 
Or do you think I'm missing something and it is used in some way?

### DIFF
--- a/scripts/apps/authoring-react/data-layer.ts
+++ b/scripts/apps/authoring-react/data-layer.ts
@@ -64,12 +64,20 @@ export function getArticleContentProfile<T>(
         const {editor, fields, schema} = fakeScope;
         const fieldExists = (fieldId) => fakeScope.editor[fieldId] != null;
 
-        // Avoid having unnecessary adapters for fields
-        // to which we do not write data e.g. 'footer'.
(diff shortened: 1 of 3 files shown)


petrjasek:
it seems like this should be a description for featuredmedia stored on the parent text item where the image is embedded

tomaskikutis:
It doesn't work though, does it?

One can always access feature media description by `article.associations.featuremedia.description_text`. We could keep it for compatibility, but if it doesn't even work now that's not relevent. Or am I missing something?

petrjasek:
yep I was just checking in the code why we have it, looks like it's an old one which we probably don't need anymore

tomaskikutis:
can you approve it so I can merge it?

petrjasek (approved):

PR #4623, opened 2024-09-06
Authoring react required field indicators and validation errors

tomaskikutis:
SDESK-7258

[screenshot](https://github.com/user-attachments/assets/015cbae6-b74b-49d0-860a-35f86cd459d6)

### DIFF
--- a/scripts/apps/authoring-react/authoring-section/authoring-section.tsx
+++ b/scripts/apps/authoring-react/authoring-section/authoring-section.tsx
@@ -95,13 +95,13 @@ export class AuthoringSection<T> extends React.PureComponent<IPropsAuthoringSect
                     backgroundColor: themeApplies ? this.props.uiTheme.backgroundColor : undefined,
                     display: 'flex',
                     flexDirection: 'column',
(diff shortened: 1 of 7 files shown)


scripts/apps/authoring-react/authoring-section/authoring-section.tsx
```diff
@@ -90,7 +95,7 @@ export class AuthoringSection<T> extends React.PureComponent<IPropsAuthoringSect
                     backgroundColor: themeApplies ? this.props.uiTheme.backgroundColor : undefined,
                     display: 'flex',
                     flexDirection: 'column',
-                    gap: '12px',
+                    gap: this.props.useHeaderLayout === true ? '12px' : '32px',
```
petrjasek: this would be better handled via `gap-` classes

petrjasek (approved):

scripts/apps/authoring-react/authoring-section/authoring-section.tsx
tomaskikutis: I've replaced the hardcoded values with css same variables that are used in those CSS classes.

PR #4637, opened 2024-09-12
Originating desk not selected in the publishing dialog

dzonidoo:
SDESK-7398

### DIFF
--- a/scripts/core/interactive-article-actions-panel/utils/get-initial-destination.ts
+++ b/scripts/core/interactive-article-actions-panel/utils/get-initial-destination.ts
@@ -86,11 +86,9 @@ export function getInitialDestination(
 }
 
 export function getCurrentDeskDestination(id?: string): ISendToDestination {
-    const destinationDeskId: string = sdApi.desks.getCurrentDeskId();
-
     return {
         type: 'desk',
-        desk: id ?? destinationDeskId,
+        desk: id ?? sdApi.desks.getCurrentDeskId(),
         stage: null,
     };
 }
--- a/scripts/core/interactive-article-actions-panel/utils/get-initial-destination.ts
+++ b/scripts/core/interactive-article-actions-panel/utils/get-initial-destination.ts
@@ -85,12 +85,13 @@ export function getInitialDestination(
     return destination;
 }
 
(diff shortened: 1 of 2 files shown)


scripts/core/interactive-article-actions-panel/utils/get-initial-destination.ts
```diff
@@ -85,12 +85,13 @@ export function getInitialDestination(
     return destination;
 }
 
-export function getCurrentDeskDestination(): ISendToDestination {
-    let destinationDesk: string = sdApi.desks.getCurrentDeskId();
+export function getCurrentDeskDestination(id?: string): ISendToDestination {
+    const destinationDeskId: string = sdApi.desks.getCurrentDeskId();
+    const itemDeskId: string = sdApi.desks.getAllDesks().get(id)._id;
```
tomaskikutis: what is the purpose to call `getAllDesks`?

dzonidoo: change it to:
desk: id ?? destinationDeskId,

tomaskikutis: More than changing I'm interested why did you go with that approach to begin with. But yeah, I don't make sense get the entire object if you're not going to use anything from it. Unless you wanted to verify that desk ID exists, but then you should provide a fallback.

scripts/core/interactive-article-actions-panel/utils/get-initial-destination.ts
```diff
@@ -85,12 +85,12 @@ export function getInitialDestination(
     return destination;
 }
 
-export function getCurrentDeskDestination(): ISendToDestination {
-    let destinationDesk: string = sdApi.desks.getCurrentDeskId();
+export function getCurrentDeskDestination(id?: string): ISendToDestination {
+    const destinationDeskId: string = sdApi.desks.getCurrentDeskId();
```
tomaskikutis: Can you remove this variable and inline this code?

dzonidoo: okay

tomaskikutis (approved):

PR #4646, opened 2024-09-25
[SDCP-848] Open translated item from api result instead of original, on translate action

thecalcc:
SDCP-848


petrjasek (approved):

thecalcc:
needs backporting to release/2.8 right @petrjasek ?

petrjasek:
yep but I will cherry-pick it there, it can be tested on sd-develop meanwhile

PR #4647, opened 2024-09-25
use prefetched user list in select user component

petrjasek:
there was an issue when fetching desk members
for a desk with many users, it would fail on
too big request when fetching 50+ users by id.

SDCP-849

### DIFF
--- a/scripts/api/index.ts
+++ b/scripts/api/index.ts
@@ -11,6 +11,7 @@ import {time} from './time';
 import {user} from './user';
 import {vocabularies} from './vocabularies';
 import {contentProfiles} from './content-profiles';
+import * as utils from './utils';
 
 /**
  * This is core API, not extensions API.
@@ -31,4 +32,5 @@ export const sdApi = {
     vocabularies,
     highlights,
     contentProfiles,
+    utils,
 };
--- a/scripts/api/index.ts
+++ b/scripts/api/index.ts
@@ -11,7 +11,6 @@ import {time} from './time';
 import {user} from './user';
 import {vocabularies} from './vocabularies';
 import {contentProfiles} from './content-profiles';
-import * as utils from './utils';
 
 /**
  * This is core API, not extensions API.
@@ -32,5 +31,4 @@ export const sdApi = {
     vocabularies,
     highlights,
     contentProfiles,
-    utils,
 };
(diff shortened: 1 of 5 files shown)


thecalcc (approved):

scripts/api/index.ts
```diff
@@ -11,6 +11,7 @@ import {time} from './time';
 import {user} from './user';
 import {vocabularies} from './vocabularies';
 import {contentProfiles} from './content-profiles';
+import * as utils from './utils';
```
tomaskikutis: why not import from lodash directly?

scripts/core/ui/components/SelectUser.tsx
```diff
@@ -118,47 +117,23 @@ export class SelectUser extends SuperdeskReactComponent<IPropsSelectUser, IState
...
-
-                    if (this.props.deskId != null && this.props.deskId != '') {
-                        query.$and.push({_id: {$in: deskMemberIds}});
-                    }
+                    const deskMembers = sdApi.desks.getDeskMembers(this.props.deskId);
 
                     if (searchString != null && searchString.length > 0) {
-                        query.$and.push(getUserSearchMongoQuery(searchString));
-                    }
+                        const query = getUserSearchMongoQuery(searchString);
```
tomaskikutis: Your filtering here only supports a subset of mongo query, ($or operator). How about next to `getUserSearchMongoQuery` you put `matchUserByMongoQuery(user: IUser): boolean` so if someone changes the query they would also see client side matching code and would update it too.

tomaskikutis:
> too big request when fetching 50+ users by id

You mean GET request URL would get too long?

tomaskikutis:
By the way, that playwright test is unstable. You can either try restarting e2e or disable that test. (the problem is actually the feature that is unstable, not the test)

petrjasek:
> > too big request when fetching 50+ users by id
> 
> You mean GET request URL would get too long?

yes exactly

scripts/api/index.ts
petrjasek: will drop it, seems like we have another helper for it

scripts/core/ui/components/SelectUser.tsx
petrjasek: done

tomaskikutis (approved):

tomaskikutis:
I was looking around the code and realized this change is breaking `SelectUser` interface and probably usages. According to the interface `deskId` is optional and according to your code it looks like it wouldn't work without it?

petrjasek:
yep you're right, will fix it

PR #4648, opened 2024-09-25
warn about incompatible extensions

tomaskikutis:
SDESK-7399

Display invalid instance configuration modal if unsupported extension point `contributions.authoringHeaderComponents` is being used. It is now deprecated, but still supported - until authoring-react is released. The message will only show up if authoring-react is running.


petrjasek:
would be good to limit this to administrators only

tomaskikutis:
Not sure if this is the best idea. If someone tests on UAT with a normal user they won't see the issues.

petrjasek (approved):

PR #4650, opened 2024-09-27
Add article fetch to api, adjust itemaction name

thecalcc:
SDESK-7015

### DIFF
--- a/scripts/api/article.ts
+++ b/scripts/api/article.ts
@@ -479,7 +479,7 @@ function get(id: IArticle['_id']): Promise<IArticle> {
 function itemAction(_article: IArticle): {[key in IAuthoringActionType]: boolean} {
     const authoring = ng.get('authoring');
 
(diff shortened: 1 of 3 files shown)


scripts/api/article.ts
```diff
@@ -479,7 +479,7 @@ function get(id: IArticle['_id']): Promise<IArticle> {
 function itemAction(_article: IArticle): {[key in IAuthoringActionType]: boolean} {
     const authoring = ng.get('authoring');
 
-    return authoring.itemAction(_article);
+    return authoring.itemActions(_article);
```
tomaskikutis, line 482: are the 2 methods with similar name? what's the difference?

thecalcc: They're not, there's no `itemAction` method, it's called `itemActions`

tomaskikutis (approved):

PR #4653, opened 2024-10-04
add cache param to dataApi.findOne

petrjasek:
need to force refresh a computed resource in planning

### DIFF
--- a/scripts/core/helpers/CrudManager.tsx
+++ b/scripts/core/helpers/CrudManager.tsx
@@ -121,21 +121,22 @@ export function generatePatchIArticle(a: IArticle, b: IArticle) {
     return patch;
 }
 
-const cache = {};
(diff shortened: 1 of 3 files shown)


tomaskikutis (approved):
I think we should drop that cache thing in the future completelly. Cache that caches only for the duration of 1 request is pretty useless. We should probably rather fix the usages and reduce the number of calls. Do you remember why did you add that cache option to begin with?

petrjasek:
yep I think it was there for angular where multiple requests could fire at the same time due to digest handling and then this would avoid certain duplicated queries, like when fetching the same vocabulary

tomaskikutis:
Do you remember if it was authoring or another part of the app?

petrjasek:
I don't remember where exactly

PR #4654, opened 2024-10-04
Pass article through translate endpoint on create article 

thecalcc:
SDBELGA-889

### DIFF
--- a/scripts/core/superdesk-api.d.ts
+++ b/scripts/core/superdesk-api.d.ts
@@ -2891,6 +2891,7 @@ declare module 'superdesk-api' {
         };
         entities: {
             article: {
+                translate(article: IArticle, language: string): Promise<IArticle>;
                 get(articleId: IArticle['_id']): Promise<IArticle>;
                 isLocked(article: IArticle): boolean; // returns true if locked by anyone, including the current user
                 isLockedInCurrentSession(article: IArticle): boolean;
--- a/scripts/core/superdesk-api.d.ts
+++ b/scripts/core/superdesk-api.d.ts
@@ -2891,8 +2891,7 @@ declare module 'superdesk-api' {
         };
         entities: {
             article: {
-                createTranslation(article: IArticle, language: string): Promise<IArticle>;
-                translateAndPatch(article: IArticle, language: string, contentPatch: string): Promise<IArticle>;
+                translate(article: IArticle, language: string): Promise<IArticle>;
                 get(articleId: IArticle['_id']): Promise<IArticle>;
                 isLocked(article: IArticle): boolean; // returns true if locked by anyone, including the current user
                 isLockedInCurrentSession(article: IArticle): boolean;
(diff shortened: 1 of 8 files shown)


tomaskikutis:
What was the issue here? I though there was already an integration with translation service?

Could you double check if your fix runs well on authoring-react?

thecalcc:
> What was the issue here? I though there was already an integration with translation service?
> 
> Could you double check if your fix runs well on authoring-react?

The issue is that we're not running the article through the `/archive/translate` endpoint, which processes the article and adds some data to it, which is then rendered on the UI. That was missing because I was simply creating a new article

thecalcc:
> What was the issue here? I though there was already an integration with translation service?
> 
> Could you double check if your fix runs well on authoring-react?

For authoring-react, currently there are problems with applying the patch after the translation. Either when the dangerous options are enabled etags don't match, and in the case when this is not enabled, the patch doesn't get applied. 

Do you think this is an ok way to do it? If yes, this means that the changes would not be saved and the save button should be in active state, thus would require the user to click to persist the translated content. 

```
this.props.onFieldsDataChange?.(
  this.props.fieldsData?.set(
  'body_html',
  superdesk.helpers.editor3ToOperationalFormat(
  {rawContentState: rawState},
  this.props.article.language,
  ),
));
```

scripts/core/superdesk-api.d.ts
```diff
@@ -2891,7 +2891,8 @@ declare module 'superdesk-api' {
         };
         entities: {
             article: {
-                translate(article: IArticle, language: string): Promise<IArticle>;
+                createTranslation(article: IArticle, language: string): Promise<IArticle>;
```
tomaskikutis: `translate` was a better name - it'd be good to go back to it. Since this is exposed to API - we should keep it single purpose and not trigger opening of an article. There's already a separate method for this in the API.

Regarding `translateAndPatch` - why is that needed at all? We already have a "translate" method. Patching should be available already via an independent method.

scripts/extensions/ai-widget/src/translations/translations-body.tsx
```diff
@@ -124,26 +124,21 @@ export default class TranslationsBody extends React.Component<IProps, IState> {
...
-                                superdesk.entities.article.createNewWithData({
-                                    body_html: translation,
-                                    task: taskData,
-                                    language: this.props.activeLanguageId,
-                                }, article.profile);
+                                translate(
+                                    article,
+                                    this.props.activeLanguageId,
+                                ).then((translatedItem) =>
+                                    get(translatedItem._id).then((fullTranslatedItem) => {
```
tomaskikutis: Reduce the destructuring a bit so it's not `get(translatedItem._id)`,
but `article.get(translatedItem._id)` or `articleApi.get(translatedItem._id)`

It's getting hard to follow where these short generic methods(get,patch) are coming from.

tomaskikutis (approved):

PR #4657, opened 2024-10-08
fix SelectUser component when there is no deskId

petrjasek:
SDESK-7411

### DIFF
--- a/scripts/api/user.ts
+++ b/scripts/api/user.ts
@@ -1,4 +1,5 @@
 import ng from 'core/services/ng';
+import {IUser} from 'superdesk-api';
 
 function hasPrivilege(privilege: string): boolean {
     const privileges = ng.get('privileges');
@@ -18,8 +19,13 @@ function isLoggedIn() {
     return session?.identity?._id != null;
 }
 
+function getAll(): Array<IUser> {
+    return ng.get('desks').users._items;
+}
+
 export const user = {
     hasPrivilege,
     isLoggedIn,
     getCurrentUserId,
+    getAll,
 };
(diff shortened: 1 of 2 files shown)


scripts/api/user.ts
```diff
@@ -18,8 +19,13 @@ function isLoggedIn() {
...
 
+function getAll(): Array<IUser> {
+    return ng.get('desks').users._items;
+}
+
 export const user = {
     hasPrivilege,
     isLoggedIn,
     getCurrentUserId,
+    getAll,
```
tomaskikutis, line 31: Can you export an `OrderedMap<IUser['_id'], IUser]>` from the API? So we also have a capability to get a user by ID without having to iterate.

tomaskikutis: It works quite well with vocabularies which I did a while ago https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/api/vocabularies.ts#L97

When thinking more about it, I'm not sure whether it's better to expose an ordered map or simply have 2 methods - getAll that would return an array and getById that would return a single item - without involving immutable.

petrjasek: I think it makes sense to get the map, imo when you get all users it's usually rather for a lookup and not for listing them, this is rather an edge case.

tomaskikutis (approved):

scripts/api/user.ts
tomaskikutis: Functionally it works great both - lookup and iteration. The only thing I'm wary about is developers being confused that getting all items when they only need one would be expensive in terms of performance when in reality it wouldn't.

PR #4659, opened 2024-10-11
Authoring actions rework

thecalcc:
SDESK-6267


thecalcc:
Note: 

**React version:** 
image

**Angular version:**
image

I suggest to take things piece by piece. The strategy of unifying the list & authoring actions might not be good for now. But a next good step for me seems like unifying authoring actions from angular & react into a single React component. From there we can think about list actions and what to do with them. Let's merge this PR so develop is not broken and from there we can pickup this part of the refactoring any time.

thecalcc:
DO NOT merge until https://github.com/superdesk/superdesk-planning/pull/1907 has been merged

tomaskikutis:
> next good step for me seems like unifying authoring actions from angular & react into a single React component

It's one of the things that I would like to get done, but the question is when is it best to do it. I'd say when we have extra time or when it becomes a roadblock. I'd like to prioritize the release of authoring-react and if those list item actions don't block it - I'd delay that work for sometime later.

thecalcc:
> > next good step for me seems like unifying authoring actions from angular & react into a single React component
> 
> It's one of the things that I would like to get done, but the question is when is it best to do it. I'd say when we have extra time or when it becomes a roadblock. I'd like to prioritize the release of authoring-react and if those list item actions don't block it - I'd delay that work for sometime later.

Yep exactly, if we do it how I suggest we can delay until whenever

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -410,7 +393,7 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
                                     getExportModal(getLatestItem, handleUnsavedChanges, hasUnsavedChanges),
                                     getTranslateModal(getLatestItem),
                                     ...authoringActionsFromExtensions,
-                                    ...articleActionsFromExtensions,
+                                    // ...articleActionsFromExtensions,
```
tomaskikutis: this should probably not be commented out?

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
```diff
@@ -31,7 +32,10 @@ export function AuthoringTopbarDirective(
         templateUrl: 'scripts/apps/authoring/views/authoring-topbar.html',
         link: function(scope) {
             function setActionsFromExtensions() {
-                scope.articleActionsFromExtensions = getArticleActionsFromExtensions(scope.item);
+                scope.articleActionsFromExtensions = partition(
```
tomaskikutis: I would ask you to make this work with any number of groups, not only planning/non-planning. But since this is angular-authoring that we'll soon drop, it's fine to keep it.

What would be good to do which is easy is to not use a tuple here but 2 separate objects like `planningActions` / `nonPlanningActions`. Accessing the tuple using index is very confusing.

thecalcc: I know, but given that it's in angular I don't think it's worth spending time on

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
thecalcc: Yes, this should be completely removes, since if we leave it, both `list item three dots menu` actions & `authoring three dots menu` actions show up in `authoring three dots menu`, thus causing duplication of actions. Regardless of duplication this shouldn't be there, since we agreed that getActions is only for the list item

scripts/apps/authoring/authoring/directives/AuthoringTopbarDirective.ts
tomaskikutis: > What would be good to do which is easy is to not use a tuple here but 2 separate objects like planningActions / nonPlanningActions. Accessing the tuple using index is very confusing.

This one do anyway since it's quick. I don't want to mess up angular code even more if it's easy to avoid.

tomaskikutis (approved):

PR #4667, opened 2024-10-18
Cosmetic metadata widget changes, persist changes on full width toggl…

thecalcc:
SDESK-7341
SDESK-7417
SDESK-7340

### DIFF
--- a/scripts/apps/authoring-react/authoring-react.tsx
+++ b/scripts/apps/authoring-react/authoring-react.tsx
@@ -1146,7 +1146,11 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
      * and when outside code sends updated `{item:T }` there is not other way for authoring-react
      * to apply it to `fieldsData`, but to re-initialize.
      */
-    reinitialize(state: IStateLoaded<T>, itemWithUpdates: T) {
+    reinitialize(
+        state: IStateLoaded<T>,
+        itemWithUpdates: T,
+        newProfile?: IContentProfileV2,
+    ) {
         const item: {
             saved: T;
             autosaved: T;
@@ -1157,7 +1161,7 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
 
         this.setState(getInitialState(
             item,
-            state.profile,
+            newProfile ?? state.profile,
             state.userPreferencesForFields,
             state.spellcheckerEnabled,
             this.props.fieldsAdapter,
@@ -1411,6 +1415,13 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
             },
         };
 
+        const onChangeSideWidget = (item: T) => {
+            authoringStorage.getContentProfile(item, this.props.fieldsAdapter)
+                .then((res) => {
+                    this.reinitialize(state, item, res);
+                });
+        };
+
         return (
             <div style={{display: 'contents'}} ref={this.setRef}>
                 {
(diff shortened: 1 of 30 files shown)


scripts/apps/authoring-react/fields/dropdown/index.ts
```diff
@@ -26,6 +26,7 @@ export function getDropdownField(): ICustomFieldType<IDropdownValue, IDropdownVa
         getEmptyValue: (config) => config.multiple ? [] : null,
         differenceComponent: Difference,
         configComponent: Config,
+        toOperationalFormat: (valueStorage) => valueStorage,
```
tomaskikutis: if operational and storage formats are the same then you don't need to specify it

tomaskikutis (approved):

tomaskikutis:
Could you post a screenshot with profile selection?

thecalcc:
> Could you post a screenshot with profile selection?

Yes, here it is.

Unopened:
image

---
Opened:
image

scripts/core/superdesk-api.d.ts
```diff
@@ -210,7 +210,10 @@ declare module 'superdesk-api' {
         // used for side widgets
         getSidePanel?(options: IExposedFromAuthoring<T>, readOnly: boolean): React.ReactNode;
 
-        secondaryToolbarWidgets: Array<React.ComponentType<{item: T}>>;
+        secondaryToolbarWidgets: Array<React.ComponentType<{
+            item: T;
+            onChange(itemWithChanges: T): void;
```
tomaskikutis: Would this "onChange" actually reinitialize? If so I think we should call it "reinitialize" so it is clear that it will be performance intensive and can't be called on every key press.

tomaskikutis:
Can you revert the latest commit and prepare the branch to be merged? Let's do widget persistance in another PR

tomaskikutis (approved):

thecalcc:
Fix for toolbar width - showcasing how it looks with the different authoring widths. Height for it should also be changed on UIF so that the changes look proper.

image

image

Note: authoring becomes scrollable here. 
image

thecalcc:
Showcase of how the reworked toolbar style looks

image
image
image

Authoring itself becomes scrollable, toolbar included.
image

thecalcc:
What needs to be decided is if we should have a third toolbar. In authoring angular this is how the view currently looks (and it is as a part of the heather fields):
image

tomaskikutis:
only use `&&` for booleans or renderng JSX directly. In case you have a variable like `const a = b && <div>b</div>`
 use a ternary `const a = b == null ? null : <div>b</div>`
 
Replace classname soup(className="flex flex-row flex-wrap align-center px-2 gap-1 py-1") with inline styles or a single classname.

Regarding the third toolbar - let's not have it. It doesn't show anything that is not already displayed elsewhere. In the long run we'll probably make it more customizable so it would be possible to set up any number of toolbars from root repo.

thecalcc:
> Replace classname soup(className="flex flex-row flex-wrap align-center px-2 gap-1 py-1") with inline styles or a single classname.

What's the difference? Only style of how we write CSS? And if we change our spacing at some point, that wouldn't be reflected.

tomaskikutis:
> What's the difference?

Easier to work with. Why spend time on coming up and remembering class names when we can use standard CSS rules directly. For spacing and other things - use variables. `<div style={{padding: 'var(--padding-small)'}} />`

thecalcc:
I've made it look the same as in Angular, after talking to fritz:
image

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -1491,26 +1503,43 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
                                                 </ButtonGroup>
                                             </div>
                                         )}
+                                        headerStyles='authoring-header--padding-medium pt-1'
```
tomaskikutis: do not use CSS classnames coming outside of this project. Address other occurences of the same in this PR too.

scripts/apps/authoring-react/subcomponents/content-profile-dropdown.tsx
```diff
@@ -1,60 +1,64 @@
...
-                    value={this.state.selectedProfileId}
-                    onChange={(profileId) => {
-                        this.setState({
-                            selectedProfileId: profileId,
-                        }, () => {
-                            this.props.reinitialize({
-                                ...this.props.article,
-                                profile: profileId,
+            <Spacer gap='4' h noGrow noWrap>
+                <span className='authoring-header__label'>{gettext('PROFILE')}</span>
```
tomaskikutis: Keep the string identical. Use CSS to adjust case.

scripts/apps/authoring-react/subcomponents/content-profile-dropdown.tsx
```diff
@@ -0,0 +1,62 @@
...
+import {Spacer} from 'superdesk-ui-framework/react';
+import {sdApi} from 'api';
+
+interface IProps<T> {
+    article: T;
+    reinitialize: (itemWithChanges: T) => void;
+}
+
+interface IState {
+    profilesList: Array<IContentProfile>;
```
tomaskikutis: if a variable is typed - do not add type related info to the variable name. `profiles` is fine here, it'd be typed as an array so we'd know what it is

scripts/apps/authoring-react/subcomponents/content-profile-dropdown.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import {IArticle, IContentProfile} from 'superdesk-api';
+import {gettext} from 'core/utils';
+import {Spacer} from 'superdesk-ui-framework/react';
+import {sdApi} from 'api';
+
+interface IProps<T> {
+    article: T;
```
tomaskikutis: if it's article - it's not `T`, but `IArticle` and if it's `T` it's not an article :D `item` is probably the best name here.

thecalcc: It's `T`, because if it's not the usage doesn't compile from authoring-react.

tomaskikutis: yeah, I thought it's `T`, but then it means it's not an article - thus we should name the variable `item` and not `article`

scripts/apps/authoring-react/subcomponents/content-profile-dropdown.tsx
thecalcc: String is identical:
image

tomaskikutis: ok, maybe I looked at intermediary diff where it was being reverted.

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -1502,7 +1502,7 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
                                                 </ButtonGroup>
                                             </div>
                                         )}
-                                        headerStyles="authoring-header--padding-medium pt-1"
+                                        headerPadding={{top: 8}}
```
tomaskikutis, line 1505: should have been paddingBlockStart

tomaskikutis:
Hey, I missed an important issue here. I though you were doing a profile selector via angular wrapper thing. You can't put it inside authoring-react.ts, because authoring-react is generic. You don't know if it's an article or some other entity you're dealing with. [The compiler was warning you](https://github.com/user-attachments/assets/9b447171-39a6-4668-ae4e-ab27123e9d1b)  :smile:

thecalcc:
Move the usage from authoring-react and pass it through the api or `defaultToolbarItems` in angular wrapper

PR #4669, opened 2024-10-18
make SelectUser dynamic again

petrjasek:
uses new api for desk members

SDBELGA-894

### DIFF
--- a/scripts/core/ui/components/SelectUser.tsx
+++ b/scripts/core/ui/components/SelectUser.tsx
@@ -1,11 +1,11 @@
 /* eslint-disable react/no-multi-comp */
 import React from 'react';
-import {IPropsSelectUser, IUser} from 'superdesk-api';
-import {gettext, searchUsers} from 'core/utils';
+import {IPropsSelectUser, IUser, IRestApiResponse} from 'superdesk-api';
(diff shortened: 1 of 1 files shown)


tomaskikutis (approved):
I've created a [data store](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/data-store.ts) recently to define often used resources and always keep them up to date via websocket notifications so they can be accessed synchronously at any time. I think it would be good to have users there since it's a frequently accessed resource and UI would be smoother if we had synchronous access.

We need to consider performance implications to avoid hammering the server, but I don't think it would be an issue. Worst case scenario, we could keep in sync a subset of `IUser` data like display name, avatar, online status.

What do you think?

petrjasek:
with users it's tricky as they update quite often. if we would have some subset of the data which we would keep up to date via websockets it would be ok, but if we would need to make a request for that it would be too much

tomaskikutis:
> with users it's tricky as they update quite often

What exactly is changing often there? Preferences?

petrjasek:
preferences and also `last_activity_at` timestamp which we use for online indication

tomaskikutis:
yep, `last_activity_at` is the offender. We'd need `active:boolean` so it changes once in 10 min not every second.

PR #4673, opened 2024-10-23
Context menu not visible for nested events in narrow view

dzonidoo:
STT-94


thecalcc (approved):

tomaskikutis:
What was causing the issue?

dzonidoo:
@tomaskikutis 'div - wrappers' make a issue for grid layout

tomaskikutis:
check if it still works well in other usages, those wrappers were probably added for a reason

dzonidoo:
I dont see any issue with removing the extra div on browser. Would you like to have it put back? But need to be display contents...
My opinion is that we dont need 4 divs in a row in any case...

tomaskikutis:
The same way adding a div can cause style issues, it might cause them if you remove it. I'm not saying not to remove it, but check other usages to ensure that removing doesn't break anything.

PR #4676, opened 2024-10-28
Resizing authoring on widget pin, allow for a secondary widget

thecalcc:
SDESK-7343
SDESK-7344

### DIFF
--- a/scripts/apps/authoring-react/article-widgets/find-and-replace.tsx
+++ b/scripts/apps/authoring-react/article-widgets/find-and-replace.tsx
@@ -134,7 +134,7 @@ class FindAndReplaceWidget extends React.PureComponent<IArticleSideWidgetCompone
                                             this.highlightMatches();
                                         });
                                     }}
-                                    disabled={this.state.replaceValue?.trim().length < 1}
+                                    disabled={(this.state.replaceValue ?? '').trim().length < 1}
                                 />
 
                                 <Button
--- a/scripts/apps/authoring-react/article-widgets/find-and-replace.tsx
+++ b/scripts/apps/authoring-react/article-widgets/find-and-replace.tsx
@@ -134,7 +134,7 @@ class FindAndReplaceWidget extends React.PureComponent<IArticleSideWidgetCompone
                                             this.highlightMatches();
                                         });
                                     }}
-                                    disabled={this.state.replaceValue.trim().length < 1}
+                                    disabled={this.state.replaceValue?.trim().length < 1}
                                 />
 
                                 <Button
@@ -150,7 +150,7 @@ class FindAndReplaceWidget extends React.PureComponent<IArticleSideWidgetCompone
                                             this.highlightMatches();
                                         });
                                     }}
-                                    disabled={this.state.replaceValue.trim().length < 1}
+                                    disabled={this.state.replaceValue?.trim().length < 1}
                                 />
                             </Spacer>
                         </div>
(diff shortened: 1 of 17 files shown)


thecalcc:
@tomaskikutis could you check the code

tomaskikutis:
yep, it's on my list

scripts/apps/authoring-react/generic-widgets/comments/CommentsWidget.tsx
```diff
@@ -216,7 +216,7 @@ class CommentsWidget<T> extends React.PureComponent<IProps<T>, IState> {
                         text="post"
                         type="primary"
                         onClick={this.save}
-                        disabled={this.state.newCommentMessage.length < 1}
+                        disabled={this.state.newCommentMessage?.length < 1}
```
tomaskikutis: Don't compare undefined to a number. Stricter compiler options wouldn't allow it.

scripts/apps/authoring-react/article-widgets/find-and-replace.tsx
```diff
@@ -134,7 +134,7 @@ class FindAndReplaceWidget extends React.PureComponent<IArticleSideWidgetCompone
                                             this.highlightMatches();
                                         });
                                     }}
-                                    disabled={this.state.replaceValue.trim().length < 1}
+                                    disabled={this.state.replaceValue?.trim().length < 1}
```
tomaskikutis: Don't compare undefined to a number. Stricter compiler options wouldn't allow it.

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -241,6 +241,7 @@ interface IPropsWrapper extends IProps {
 interface IState {
     sidebarMode: boolean | 'hidden';
     sideWidget: null | ISideWidget;
+    sideWidgetSecondary: null | ISideWidget;
```
tomaskikutis: According the the interface it's possible for both widgets to be pinned at once. I'm not requesting to fix it - rather keep in mind in the future to avoid similar situations when possible.

tomaskikutis (commented):
Not sure if we talked about before - I'm thinking whether I didn't previously request the opposite. I think it'd be better to have this behavior as a part of authoring-react component so if it's used somewhere else it would work out of the box without having to re-implement it via a wrapper.

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
tomaskikutis: image

this would be better for example

scripts/apps/authoring-react/article-widgets/versions-and-item-history/versions-tab.tsx
```diff
@@ -114,11 +114,11 @@ export class VersionsTab extends React.PureComponent<IArticleSideWidgetComponent
 
         compareAuthoringEntities({
             item1: {
-                label: gettext('version {{n}}', {n: from._current_version}),
+                label: gettext('version {{n}}', {n: from?._current_version}),
```
tomaskikutis: You can't output a string with pieces missing to the UI - which would happen since you use optional chaining without a fallback.

Current version is mandatory, not sure why are you applying optional chaining all over the place here.

tomaskikutis (approved):

PR #4679, opened 2024-10-30
Add UI feedback

thecalcc:
SDCP-854

### DIFF
--- a/scripts/apps/highlights/services/HighlightsService.ts
+++ b/scripts/apps/highlights/services/HighlightsService.ts
@@ -29,10 +29,6 @@ export function HighlightsService(api, $q, $cacheFactory, packages: IPackagesSer
     var cache = $cacheFactory('highlightList');
 
     /**
-     * Allows for state control when a request is made.
-     */
-    service.requestLoading = false;
-    /**
      * Get cached value for given key
      *
      * @param {string} key
@@ -59,11 +55,13 @@ export function HighlightsService(api, $q, $cacheFactory, packages: IPackagesSer
         var criteria = {};
 
         if (desk) {
-            criteria = {where: {$or: [
-                {desks: desk},
-                {desks: {$size: 0}},
-            ],
-            },
+            criteria = {
+                where: {
+                    $or: [
+                        {desks: desk},
+                        {desks: {$size: 0}},
+                    ],
+                },
             };
         }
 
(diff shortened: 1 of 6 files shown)


scripts/apps/highlights/services/HighlightsService.ts
```diff
@@ -111,7 +114,25 @@ export function HighlightsService(api, $q, $cacheFactory, packages: IPackagesSer
      * Mark an item for a highlight
      */
     service.markItem = function(highlight, markedItem) {
-        return api.save('marked_for_highlights', {highlights: [highlight], marked_item: markedItem._id});
+        dispatchEvent(new CustomEvent('highlights-loading', {
```
tomaskikutis: Can you abstract it into a function so if you want to use it for some other action you would only need to call a function, but wouldn't need to know about events?

```
function trackArticleActionProgress<T>(p: () => Promise<T>, successMessage: string, errorMessage: string): Promise<T> {
    //
}

function markItem(highlight, markedItem) {
    return trackArticleActionProgress(
        () => api.save(
            'marked_for_highlights',
            {highlights: [highlight], marked_item: markedItem._id},
        ),
        gettext('Item marked'),
        gettext('Couldn\'t mark item'),
    );
}

```

scripts/apps/search/components/Item.tsx
```diff
@@ -119,13 +120,21 @@ export class Item extends React.Component<IProps, IState> {
...
     }
 
+    handleHighlightsLoading(e: CustomEvent) {
+        if (e.detail.itemId === this.props.item._id) {
+            this.setState({loading: e.detail.loading});
+        }
+    }
+
     componentDidMount() {
+        addEventListener('highlights-loading', this.handleHighlightsLoading);
```
tomaskikutis: Let's not couple it to highlights here. Just call it loading indicator. That way we'll be able to reuse it for other actions and will not be lying via naming.

scripts/core/helpers/network.tsx
```diff
@@ -173,3 +174,30 @@ export function uploadFileWithProgress<T>(
             });
         });
 }
+
+export function trackArticleActionProgress<T>(
+    p: () => Promise<T>,
```
tomaskikutis: `getPromise` would be a better name. `p` was a placeholder in my initial snippet.

scripts/core/helpers/network.tsx
```diff
@@ -173,3 +174,30 @@ export function uploadFileWithProgress<T>(
...
+    itemId: string,
+    successMessage: string,
+    errorMessage: string,
+): Promise<T> {
+    dispatchEvent(new CustomEvent('action-loading', {
```
tomaskikutis: would be good to call it `article-action-loading` so if anyone is logging events they would discover quicker from which part is it coming from.

tomaskikutis (approved):

thecalcc:
cherry pick to release/2.8 and update cp dev

PR #4684, opened 2024-11-11
Fix article content profile switch in authoring

thecalcc:
SDESK-7417

### DIFF
--- a/scripts/apps/authoring-react/authoring-integration-wrapper.tsx
+++ b/scripts/apps/authoring-react/authoring-integration-wrapper.tsx
@@ -56,8 +56,8 @@ const defaultToolbarItems: Array<React.ComponentType<{
     article: IArticle;
     reinitialize: (itemWithChanges: IArticle) => void;
 }>> = [
-        CreatedModifiedInfo,
-    ];
+    CreatedModifiedInfo,
+];
 
 interface IProps {
     itemId: IArticle['_id'];
@@ -378,7 +378,7 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
                             ),
                             availableOffline: false,
                             group: 'start',
-                            priority: 1
+                            priority: 1,
                         }])}
                         getLanguage={(article) => article.language ?? 'en'}
                         onEditingStart={(article) => {
--- a/scripts/apps/authoring-react/authoring-integration-wrapper.tsx
+++ b/scripts/apps/authoring-react/authoring-integration-wrapper.tsx
@@ -56,8 +56,8 @@ const defaultToolbarItems: Array<React.ComponentType<{
     article: IArticle;
     reinitialize: (itemWithChanges: IArticle) => void;
 }>> = [
-    CreatedModifiedInfo,
-];
+        CreatedModifiedInfo,
+    ];
 
 interface IProps {
     itemId: IArticle['_id'];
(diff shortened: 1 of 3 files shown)


scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -359,6 +359,22 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
                             },
                             retrieveStoredValue: (item: IArticle, fieldId) => item.extra?.[fieldId] ?? null,
                         }}
+                        customHeader={((options) => (
```
tomaskikutis: use `exposed` to keep naming consistent

scripts/core/superdesk-api.d.ts
```diff
@@ -215,6 +216,8 @@ declare module 'superdesk-api' {
             reinitialize(itemWithChanges: T): void;
         }>>;
 
+        customHeader?: (options: IExposedFromAuthoring<T>) => JSX.Element;
```
tomaskikutis: Return `Array<ITopBarWidget<T>>` here. We already have a nice interface that supports ordering buttons, adding keybindings. This is just another toolbar and it'd be best if it was configurable the same way.

scripts/core/superdesk-api.d.ts
```diff
@@ -161,6 +161,7 @@ declare module 'superdesk-api' {
         initiateClosing(): void;
         keepChangesAndClose(): void;
         stealLock(): void;
+        onChangeToolbarWidget(item: T): void;
```
tomaskikutis: Naming doesn't make sense for me here. What is a toolbar widget? What does it mean for a toolbar widget to "change"? from widget A to widget B? or how? Why would then the article be passed if it was only the toolbar widget that changed?

I think you should simply expose `reinitialize: (itemWithChanges: IArticle) => void;` that is already exposed in multiple places in the API, just add a `profile` argument.

thecalcc: I can't because `state` needs to be passed, so some wrapper would always have to be there

tomaskikutis:
fix ci

scripts/core/superdesk-api.d.ts
tomaskikutis: remove `onChangeToolbarWidget` completely since there are no usages anymore

scripts/core/superdesk-api.d.ts
```diff
@@ -215,6 +216,8 @@ declare module 'superdesk-api' {
             reinitialize(itemWithChanges: T): void;
         }>>;
 
+        customHeader?: (options: IExposedFromAuthoring<T>) => Array<ITopBarWidget<T>>;
```
tomaskikutis: headerToolbar would be a better name.

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -359,6 +359,37 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
...
+                                    exposed.reinitialize(item, profile);
+                                });
+
+                            return [{
+                                component: ({entity}) => (
+                                    <div className="authoring-header__general-info">
+                                        <ContentProfileDropdown
+                                            item={entity}
+                                            reinitialize={(item) => {
+                                                if (exposed.hasUnsavedChanges()) {
```
tomaskikutis: It'd be cleaner if `getProfileAndReinitialize` would only be used once. This way it is also clear that it will get called regardless of the condition.

```
const handledChanges = exposed.hasUnsavedChanges()
    ? exposed.handleUnsavedChanges()
    : Promise.resolve();

handledChanges.then(() => {
    getProfileAndReinitialize(item);
});
```

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -1220,6 +1219,7 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
                     pinnedId: id === activeWidgetId ? activeWidgetId : this.props.sideWidget?.pinnedId,
                 });
             },
+            reinitialize: (item, profile) => this.reinitialize(state, item, profile),
```
tomaskikutis, line 1222: you didn't add `reinitialize` to the interface.

tomaskikutis (approved):
fix ci

PR #4685, opened 2024-11-12
remove zIndex property

dzonidoo:
SDESK-7429


scripts/apps/authoring-react/authoring-angular-integration.tsx
```diff
@@ -450,9 +457,8 @@ function getInlineToolbarActions(
             component: () => (
                 <>
                     <Popover
-                        zIndex={1050}
                         triggerSelector="#marked-for-desks"
-                        title={gettext('Marked for')}
+                        title={gettext('Marked for 123123123')}
```
tomaskikutis: restore the old string

tomaskikutis (approved):

tomaskikutis:
did you expose a function from ui-framework to get latest z-index when not using ui-framework components?

dzonidoo:
but i dont need it for this pr...

tomaskikutis:
How likely it is that someone will need it later? Probably 100% :D Do expose it now(from ui-framework side) so others don't need to later.

dzonidoo:
but this is ui framework side... i just need to export it.. why is that in relationship with this pr?

dzonidoo:
do i need to bump version of ui framework again after exporting?

tomaskikutis:
fix ci

tomaskikutis:
@dzonidoo I've restarted unit tests, they do pass now, but there are issues with extensions - do take a look

dzonidoo:
@tomaskikutis  so do i need to reset the failed test?

tomaskikutis:
no, you need to fix the issues with extensions - see logs

petrjasek:
there are also some z-indexes in angular templates like [here](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/activity/views/activity-modal.html#L1), should those be removed too? seems like it's causing some issues currently

tomaskikutis:
The reason why z-indexes were removed from react components is because we are handling z-indexes on ui-framework side. That only happens for react components though.

I think there are 2 options here @dzonidoo, either to initialize z-index on ui-framework side to a lower value so it doesn't conflict with angular z-indexes, or export that function from ui-framework to generate a next z-index and use it for angular components.

PR #4689, opened 2024-11-21
DateTime picker implementation

thecalcc:
STT-109


thecalcc:
Note: z-index fixes aren't applied here so lint will fail

scripts/apps/authoring-react/fields/time copy/editor.tsx
```diff
@@ -0,0 +1,37 @@
...
+type IProps = IEditorComponentProps<ITimeValueOperational, ITimeFieldConfig, ITimeUserPreferences>;
+
+export class Editor extends React.PureComponent<IProps> {
+    render() {
+        const Container = this.props.container;
+
+        return (
+            <Container>
+                <DateTimePicker
+                    dateFormat='MM/DD/YYYY'
```
tomaskikutis: do not hardcode format - use the one from appConfig

tomaskikutis (commented):
Do not use interfaces owned by time field(like ITimeFieldConfig). Add separate interface for a new field type. Even if it will be identical at the moment.

Rename "time copy" to something better

scripts/apps/authoring-react/fields/datetime/editor.tsx
```diff
@@ -17,7 +18,7 @@ export class Editor extends React.PureComponent<IProps> {
         return (
             <Container>
                 <DateTimePicker
-                    dateFormat='MM/DD/YYYY'
+                    dateFormat={appConfig.shortDateFormat}
```
tomaskikutis: use `appConfig.view.dateformat`

tomaskikutis (approved):

PR #4693, opened 2024-11-28
test extensions build on github actions

petrjasek:


tomaskikutis (commented):
We're already compiling extensions as a part of the main CI task.

```
// package.json scripts
"test": "npm run lint && npm run unit && node tasks/verify-client-api-changes.js",
                                         ^
```

petrjasek:
interesting, because we have it failing here
https://github.com/superdesk/superdesk/actions/runs/12028429253/job/33531537951?pr=3508

tomaskikutis:
It's most likely because of cherry-picking one thing which depended on another and that another wasn't cherry picked.

In this case I see that it's crashing due to changes in ui-framework, not due to API changes.

petrjasek:
yep looks like it

PR #4694, opened 2024-12-02
lock ui framework version

petrjasek:
to avoid getting updates without having them tested

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13587,9 +13587,9 @@
       }
     },
     "superdesk-ui-framework": {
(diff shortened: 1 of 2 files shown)


tomaskikutis (approved):

tomaskikutis:
since it's a release branch, `npm ci` will be used and no library versions should change. Have we updated some by accident in the past?

petrjasek:
we got version updated in superdesk repo but not sure if that was done manually or via dependabot

PR #4696, opened 2024-12-06
Setting "Appearance" preferences in the user profile is not kept upon re-login

dzonidoo:
SDESK-7436


scripts/apps/users/directives/UserPreferencesDirective.ts
```diff
@@ -281,6 +285,7 @@ export function UserPreferencesDirective(
                     }
                 });
 
+                scope.activeTheme = data['application:theme']['theme'];;
```
tomaskikutis: will this not crash if no theme is set in preferences?

tomaskikutis (commented):
fix lint

tomaskikutis (approved):

PR #4697, opened 2024-12-09
Fix authoring and field

thecalcc:
STT-109

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "sass-loader": "6.0.6",
     "shortid": "2.2.8",
     "style-loader": "0.20.2",
-    "superdesk-ui-framework": "^4.0.3",
+    "superdesk-ui-framework": "^4.0.4",
     "ts-loader": "3.5.0",
     "typescript": "4.9.5",
     "uuid": "8.3.1",
--- a/scripts/core/menu/authoring-switch.ts
+++ b/scripts/core/menu/authoring-switch.ts
@@ -23,8 +23,10 @@ export const setupAuthoringReact = (url: string) => {
         (extensionUrl) => extensionUrl.startsWith(trimStartExact(parsedPath.hash, '#')),
     ) != null;
 
+    const isNavToPlan = parsedPath.hash.startsWith('#/planning');
+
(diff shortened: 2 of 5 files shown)


scripts/core/menu/authoring-switch.ts
```diff
@@ -23,8 +23,10 @@ export const setupAuthoringReact = (url: string) => {
         (extensionUrl) => extensionUrl.startsWith(trimStartExact(parsedPath.hash, '#')),
     ) != null;
 
+    const isNavToPlan = parsedPath.hash.startsWith('#/planning');
```
tomaskikutis: `isNavigatingToPlanning` would fit on the line too, wouldn't it? :)

tomaskikutis (approved):

package.json
```diff
@@ -123,7 +123,7 @@
     "sass-loader": "6.0.6",
     "shortid": "2.2.8",
     "style-loader": "0.20.2",
-    "superdesk-ui-framework": "4.0.3",
+    "superdesk-ui-framework": "^4.0.4",
```
petrjasek, line 126: imo would be better to avoid `^` there, we had changes in hotfix version which broke builds, the idea to remove that was that we get the updated framework version tested before it's used by clients
cc @tomaskikutis

tomaskikutis: You might have replied about it before, but I forgot - why package-lock isn't enough for hotfixes?

petrjasek: it will start breaking builds in superdesk repo but not in superdesk-client-core when there is a change that requires changes in core

tomaskikutis: why would it break builds in main superdesk repo? It also has package-lock

tomaskikutis: or you say that when client-core is being updated from root repo it would update the version then?

petrjasek: dependabot will update ui framework and the build will fail, but it won't be visible on client-core repo but rather on some superdesk-cp where we would be using an older version, but which none of the ui team is following

tomaskikutis: I've added a grep-lint rule for this https://github.com/superdesk/superdesk-client-core/pull/4710

PR #4698, opened 2024-12-11
implement prompt

tomaskikutis:
STT-63

A styled alternative to `window.prompt`.

### DIFF
--- a/scripts/core/prompt-modal.tsx
+++ b/scripts/core/prompt-modal.tsx
@@ -4,8 +4,7 @@ import {gettext} from './utils';
 
 interface IProps {
     label: string;
-    closeModal(): void;
-    onDone(value: string): void;
+    closeModal(result: {value: string | null}): void;
     okButtonText?: string;
     cancelButtonText?: string;
 }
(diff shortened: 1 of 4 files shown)


scripts/core/prompt-modal.tsx
```diff
@@ -0,0 +1,63 @@
...
+                                type="default"
+                                text={this.props.cancelButtonText ?? gettext('Cancel')}
+                                onClick={this.props.closeModal}
+                            />
+                            <Button
+                                type="primary"
+                                text={this.props.okButtonText ?? gettext('Ok')}
+                                onClick={() => {
+                                    this.props.onDone(this.state.value);
+                                    this.props.closeModal();
```
thecalcc: Why don't you pass the state to `closeModal` and just have one function? So in the usage it would be: 

`closeModal={(value) => {
 ...processing
 closeModal(); <-- coming from `showModal`
})`

scripts/core/prompt-modal.tsx
```diff
@@ -0,0 +1,63 @@
...
+            value: '',
+        };
+    }
+    render() {
+        return (
+            <Modal
+                position="top"
+                visible
+                onHide={this.props.closeModal}
+                headerTemplate=""
```
thecalcc: Is this necessary? From what I checked in UIF this is the interface for the prop `headerTemplate?: JSX.Element | string;`

thecalcc (approved):

scripts/core/prompt-modal.tsx
thecalcc: Since both functions are not exposed to the options interface it makes sense to me

scripts/core/prompt-modal.tsx
thecalcc: I meant to select only `headerTemplate` as the row to comment on, my bad

tomaskikutis: that's how I understood it and only dropped headerTemplate

PR #4699, opened 2024-12-11
"Send to" tab in the publish widget: Do not show embargo field per default

dzonidoo:
SDESK-7434


tomaskikutis:
fix tests

tomaskikutis:
I don't see you changing any defaults here. Should app config be updated?

dzonidoo:
we have default value for embargo in send to plane... its false in appConfing file

tomaskikutis (approved):

PR #4700, opened 2024-12-12
sams use helper from sourcefabric common

tomaskikutis:
STT-63

### DIFF
--- a/scripts/extensions/sams/package.json
+++ b/scripts/extensions/sams/package.json
@@ -7,7 +7,7 @@
     "watch": "tsc -watch -p src"
   },
   "dependencies": {
-    "@superdesk/common": "^0.0.13",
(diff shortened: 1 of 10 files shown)


thecalcc (approved):

scripts/extensions/sams/package.json
```diff
@@ -7,7 +7,7 @@
     "watch": "tsc -watch -p src"
   },
   "dependencies": {
-    "@superdesk/common": "^0.0.13",
+    "@sourcefabric/common": "0.0.33",
```
MarkLark86, line 10: @tomaskikutis Where is the code for v0.0.33, I don't see it in https://github.com/superdesk/sourcefabric-common

tomaskikutis: my bad forgot to push it - it's there now

PR #4701, opened 2024-12-12
use authoring-react in planning module

tomaskikutis:
STT-63

### DIFF
--- a/scripts/core/superdesk-api.d.ts
+++ b/scripts/core/superdesk-api.d.ts
@@ -133,6 +133,8 @@ declare module 'superdesk-api' {
 
     export type IFieldsData = import('immutable').Map<string, unknown>;
 
+    export type IAuthoringValidationErrors = {[fieldId: string]: string};
+
     /**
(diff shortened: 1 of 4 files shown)


tomaskikutis:
FYI after this one is merged @thecalcc you can use client-core@develop with embedded plannings PR on planning side

scripts/core/superdesk-api.d.ts
```diff
@@ -133,6 +133,8 @@ declare module 'superdesk-api' {
 
     export type IFieldsData = import('immutable').Map<string, unknown>;
 
+    export type IAuthoringValidationErrors = {[fieldId: string]: string};
```
thecalcc, line 136: Didn't we have a global Dictionary type defined? If so, maybe it can be used here

thecalcc (approved):

scripts/core/superdesk-api.d.ts
tomaskikutis: This is even nicer sometimes allowing to name a key

PR #4705, opened 2024-12-13
Enable Parent Item Selection In Dropdown

thecalcc:
STT-63

### DIFF
--- a/scripts/apps/authoring-react/fields/dropdown/editor-using-manual-source-or-vocabulary.tsx
+++ b/scripts/apps/authoring-react/fields/dropdown/editor-using-manual-source-or-vocabulary.tsx
@@ -80,6 +80,7 @@ export class EditorUsingManualSourceOrVocabulary extends React.PureComponent<IPr
                     kind="synchronous"
                     getOptions={() => options}
                     values={selected}
+                    canSelectBranchWithChildren={config.canSelectBranchWithChildren}
                     onChange={(_values) => {
                         const ids = _values.map((val) => val.id);
 
--- a/scripts/core/superdesk-api.d.ts
+++ b/scripts/core/superdesk-api.d.ts
@@ -355,6 +355,7 @@ declare module 'superdesk-api' {
         vocabularyId: IVocabulary['_id'];
         multiple: boolean;
         filter?(vocabulary: IVocabularyItem): boolean;
+        canSelectBranchWithChildren?: boolean;
     }
 
     export interface IDropdownConfigRemoteSource extends ICommonFieldConfig {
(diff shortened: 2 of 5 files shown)


scripts/apps/authoring-react/fields/dropdown/editor-using-manual-source-or-vocabulary.tsx
```diff
@@ -80,6 +80,7 @@ export class EditorUsingManualSourceOrVocabulary extends React.PureComponent<IPr
                     kind="synchronous"
                     getOptions={() => options}
                     values={selected}
+                    canSelectBranchWithChildren={config.canSelectBranchWithChildren}
```
tomaskikutis, line 83: keep capability to allow/disallow based on a tree node

```
canSelectBranchWithChildren?(branch: ITreeNode<unknown>): boolean;
```

scripts/core/superdesk-api.d.ts
```diff
@@ -366,18 +367,21 @@ declare module 'superdesk-api' {
...
         getId(item: unknown): string;
-        canSelectBranchWithChildren?(branch: ITreeNode<unknown>): boolean;
+        canSelectBranchWithChildren?: boolean;
         optionTemplate?: React.ComponentType<{item: unknown}>;
         valueTemplate?: React.ComponentType<{item: unknown}>;
         multiple: boolean;
     }
 
+    /**
+     * @deprecated Use IDropdownConfigManualSource/IDropdownConfigVocabulary/IDropdownConfigRemoteSource
```
tomaskikutis: can it be removed instead of deprecating?

tomaskikutis: Not sure if we need to deprecate it. Maybe it's enough to warn not to use this one if use case can be satisfied using other dropdown sources.

thecalcc: No need, but I thought it's the better direction

scripts/apps/authoring-react/fields/dropdown/editor-using-manual-source-or-vocabulary.tsx
thecalcc: not supported in uif component, we talked about it and decided to leave as is

scripts/core/superdesk-api.d.ts
thecalcc: done

thecalcc: done

tomaskikutis (approved):

PR #4706, opened 2024-12-17
Move boolean field to core project

thecalcc:
STT-63


scripts/apps/authoring-react/fields/boolean/preview.tsx
tomaskikutis, line 1: move interfaces to api-d.ts next to `IDateValueOperational` so it could be imported from extensions outside of this repo

tomaskikutis (approved):

PR #4718, opened 2025-01-09
Add option to hide contact header

thecalcc:
STT-31


scripts/apps/contacts/components/ContactInfo.tsx
```diff
@@ -9,51 +9,54 @@ import {ContentDivider, Icon, Spacer} from 'superdesk-ui-framework/react';
 interface IProps {
     item: IContact;
     labelInactive?: string;
+    isActive?: boolean;
```
tomaskikutis: what is active? contact info? What does it mean for contact info to be active? Better naming would help here. That labelInactive I'd rename too

thecalcc: Done

tomaskikutis (approved):

PR #4720, opened 2025-01-13
Get authoring ref

thecalcc:
STT-31

### DIFF
--- a/scripts/apps/authoring-react/authoring-react.tsx
+++ b/scripts/apps/authoring-react/authoring-react.tsx
@@ -1187,24 +1187,13 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
         ));
     }
 
-    render() {
-        const state = this.state;
-        const {authoringStorage, fieldsAdapter, storageAdapter, getLanguage, getSidePanel} = this.props;
-
-        if (state.initialized !== true) {
-            return null;
-        }
-
-        // TODO: remove test code
-        if (uiFrameworkAuthoringPanelTest) {
-            return (
-                <div>
-                    <EditorTest />
-                </div>
-            );
-        }
-
-        const exposed: IExposedFromAuthoring<T> = {
+    getExposed(
+        state: IStateLoaded<T>,
+        authoringStorage: IAuthoringStorage<T>,
+        storageAdapter: IStorageAdapter<T>,
+        fieldsAdapter: IFieldsAdapter<T>,
+    ): IExposedFromAuthoring<T> {
+        return {
             item: state.itemWithChanges,
             contentProfile: state.profile,
             getLatestItem: this.computeLatestEntity,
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -1187,24 +1187,13 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
...
-        if (uiFrameworkAuthoringPanelTest) {
-            return (
-                <div>
-                    <EditorTest />
-                </div>
-            );
-        }
-
-        const exposed: IExposedFromAuthoring<T> = {
+    getExposed(
```
tomaskikutis: This one can't be public because it needs internal details as arguments. Since all those details come from props anyway, you can drop arguments and then the method could be public.

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -1187,24 +1182,16 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
...
     }
 
-    render() {
-        const state = this.state;
-        const {authoringStorage, fieldsAdapter, storageAdapter, getLanguage, getSidePanel} = this.props;
+    getExposed(): IExposedFromAuthoring<T> | null {
+        const {state} = this;
 
-        if (state.initialized !== true) {
+        if (state.initialized === false) {
```
tomaskikutis, line 1188: Let's better throw an error, so our return type is not nullable. Returning `null` isn't useful anyway, if I get `null` what I'm supposed to do? call again hoping it would be initialized? :D

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -1187,24 +1182,16 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
         ));
     }
 
-    render() {
-        const state = this.state;
-        const {authoringStorage, fieldsAdapter, storageAdapter, getLanguage, getSidePanel} = this.props;
+    getExposed(): IExposedFromAuthoring<T> | null {
```
tomaskikutis: mark as public

scripts/core/get-superdesk-api-implementation.tsx
```diff
@@ -121,6 +121,8 @@ import {editor3ToOperationalFormat} from 'apps/authoring-react/fields/editor3';
 import {prepareSuperdeskQuery} from './helpers/universal-query';
 import {showPopup} from 'superdesk-ui-framework/react';
 import {ui} from './ui-utils';
+import {forwardRef, RefObject} from 'react';
+import React from 'react';
 
 export function openArticle(
```
tomaskikutis: no changes in this file?

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -1256,7 +1243,26 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
...
+    render() {
+        const state = this.state;
+        const {getLanguage, getSidePanel} = this.props;
+
+        if (state.initialized !== true) {
+            return null;
+        }
+
+        // TODO: remove test code
+        if (uiFrameworkAuthoringPanelTest) {
```
tomaskikutis: drop this block

tomaskikutis (approved):

PR #4725, opened 2025-01-15
Expose validation errors and improve unsaved changes modal

thecalcc:
STT-31

### DIFF
--- a/scripts/apps/authoring-react/authoring-react.tsx
+++ b/scripts/apps/authoring-react/authoring-react.tsx
@@ -1441,14 +1441,14 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
                             <Layout.AuthoringFrame
                                 header={primaryToolbarWidgets.length < 1
                                     && extraPrimaryToolbarWidgets?.length < 1 ? null : (
-                                    <SubNav>
-                                        <AuthoringToolbar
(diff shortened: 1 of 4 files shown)


scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -933,6 +932,11 @@ export class AuthoringReact<T extends IBaseRestApiResponse> extends React.PureCo
                         if (this.state.initialized) {
                             resolve(this.state.itemOriginal);
                         }
+                    }).catch((e) => {
+                        // Since we don't give modal control to the developer using authoring react
+                        // we close the prompt and return an error
+                        closePromptFn();
+                        reject(e);
```
tomaskikutis: I'd be better to reject without arguments. Otherwise it's very unpredictable what the rejection type is going to be.

Why does saving reject for you in the first place?

thecalcc: Because of validation errors in the fields

tomaskikutis: I see. It should be good then to reject without arguments since you're exposing validation errors separatelly anyway

tomaskikutis (approved):

tomaskikutis (approved):

PR #4730, opened 2025-01-22
replace usages of MultiSelect with TreeSelect

dzonidoo:


scripts/apps/authoring-react/toolbar/multi-edit-toolbar-action.tsx
```diff
@@ -27,6 +27,10 @@ export class MultiEditToolbarAction extends React.Component<IProps, IState> {
     }
 
     render(): ReactNode {
+        const options = sdApi.article.getWorkQueueItems().filter((article) =>
```
tomaskikutis: Either inline it or move inside `getOptions` function so it only executes when `getOptions` is called.

tomaskikutis (approved):

PR #4731, opened 2025-01-22
Datetime improvements

thecalcc:
STT-31


scripts/apps/authoring-react/fields/datetime/editor.tsx
```diff
@@ -23,15 +24,11 @@ export class Editor extends React.PureComponent<IProps> {
                         text: gettext('Date time'),
                         hidden: true,
                     }}
-                    onChange={(value) => {
-                        this.props.onChange(value);
+                    onChange={(val) => {
+                        // if value passed to onChange has the same reference, storageAdapter won't be triggered
```
tomaskikutis, line 28: what would be the purpose of triggering it if nothing changed?

PR #4738, opened 2025-01-29
Authoring react field template

tomaskikutis:
STT-63


scripts/apps/authoring-react/authoring-section/authoring-section-field.tsx
```diff
@@ -32,21 +34,22 @@ export class AuthoringSectionField<T> extends React.PureComponent<IProps<T>> {
     constructor(props: IProps<T>) {
         super(props);
 
-        this.getFieldContainer = memoize(getFieldContainer, 1);
+        this.getFieldContainer = memoizeOne(getFieldContainer, ([a], [b]) => shallowEqualObjects(a, b));
```
thecalcc, line 37: I wonder if we can't tweak the internal `memoize` function to handle this use case here, so we don't pile up dependencies.

thecalcc (approved):

scripts/apps/authoring-react/authoring-section/authoring-section-field.tsx
tomaskikutis: You end up with a dependency in any case - the only question is whether you maintain it or someone else. As long as they do a good job I choose "someone else" :D

thecalcc (approved):

PR #4739, opened 2025-02-03
FIX Planning preferences in user profile: do not save and deselect not possible [SDESK-6444]

devketanpro:


tomaskikutis (commented):
Do not add empty values to vocabularies. Those vocabularies are used in other places of the application where empty values would cause bugs.

It should work to implement this by only changing user profile page.

devketanpro:
Thanks, @tomaskikutis. Can you please suggest me some another approach to fix this?

tomaskikutis:
Sure. Empty item is fine UI wise. The important part is to add an empty item only to that input - not to the vocabulary itself.

Here's an [angularjs example](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/users/views/edit-form.html#L220). Same principle would be applied in react - to render an extra option in the input field itself.

scripts/apps/authoring/metadata/views/MetaDataDropdownSingleSelectReact.tsx
```diff
@@ -35,7 +35,7 @@ export class MetaDataDropdownSingleSelectReact extends React.PureComponent<IProp
                     tabIndex={tabIndex}
                     className="sd-line-input__select"
                 >
-                    <option value="" disabled hidden>{gettext('-- Choose --')}</option>
+                    <option value="" >{""}</option>
```
tomaskikutis: `<option value="" />` is enough, you don't need the empty string

scripts/apps/users/directives/UserPreferencesDirective.ts
```diff
@@ -474,7 +474,7 @@ export function UserPreferencesDirective(
                         });
                     }
 
-                    patchObject[key] = Object.assign(val, scope.preferences[key]);
+                    patchObject[key] = _.merge(val, scope.preferences[key]);
```
tomaskikutis, line 477: What was the issue with `Object.assign`?

devketanpro: Actually, `Object.assign` only performs a shallow copy, so nested objects get overwritten instead of merged.  
That's why I used `_.merge`, which properly merges deeply nested values.  

We faced a similar issue previously with `_.extend`:  
https://github.com/superdesk/superdesk-client-core/pull/4664/files#diff-f4b362667df302f286707982cb3cbb90a0ebda8f40ff78683e9d74c523d0d5ebR423

tomaskikutis: I see. All good then.

tomaskikutis (approved):

PR #4740, opened 2025-02-03
Vocabulary empty state

thecalcc:
STT-31


tomaskikutis:
I suspect the reason I did this was to handle [dependent fields](https://github.com/user-attachments/assets/b2864727-2913-4686-a36d-3d503e50fa58). See if dropping this code does not break it. I suspect it might. See if same can be achieved while maintaining support for dependent fields.

PR #4744, opened 2025-02-04
update karma webpack libraries

petrjasek:
and fix tests which were failing randomly due to
relying on state set by other tests.

### DIFF
--- a/scripts/apps/authoring/authoring/authoring-custom-field.tsx
+++ b/scripts/apps/authoring/authoring/authoring-custom-field.tsx
@@ -45,7 +45,7 @@ export class AuthoringCustomField extends React.PureComponent<IProps, IState> {
 
         this.state = {
             value: getValue(props),
-            preferences: ng.get('preferencesService').getSync(),
+            preferences: ng.get('preferencesService').getSync() || {},
         };
 
         this.lastPropsValue = this.state.value;
--- a/scripts/apps/authoring/authoring/authoring-custom-field.tsx
+++ b/scripts/apps/authoring/authoring/authoring-custom-field.tsx
@@ -45,7 +45,7 @@ export class AuthoringCustomField extends React.PureComponent<IProps, IState> {
 
         this.state = {
             value: getValue(props),
-            preferences: ng.get('preferencesService').getSync() || {},
+            preferences: ng.get('preferencesService').getSync() ?? {},
         };
 
         this.lastPropsValue = this.state.value;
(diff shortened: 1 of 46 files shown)


scripts/apps/authoring/authoring/authoring-custom-field.tsx
```diff
@@ -45,7 +45,7 @@ export class AuthoringCustomField extends React.PureComponent<IProps, IState> {
 
         this.state = {
             value: getValue(props),
-            preferences: ng.get('preferencesService').getSync(),
+            preferences: ng.get('preferencesService').getSync() || {},
```
tomaskikutis: prefer `??` operator over boolean `OR` for setting defaults

scripts/apps/search/tests/MultiActionController.spec.ts
```diff
@@ -1,7 +1,5 @@
 import {registerExtensions} from 'core/register-extensions';
 import {AuthoringWorkspaceService} from 'apps/authoring/authoring/services/AuthoringWorkspaceService';
-import {ISuperdeskGlobalConfig} from 'superdesk-api';
-import {appConfig} from 'appConfig';
 import {noop} from 'lodash';
 import {getMultiActions} from '../controllers/get-multi-actions';
```
tomaskikutis: drop unused imports

scripts/apps/workspace/content/tests/content.spec.ts
```diff
@@ -188,6 +188,7 @@ describe('superdesk.apps.workspace.content', () => {
 
         it('should notify appropriate error when created profile is not unique', inject((
             notify, $controller, content, $q, $rootScope) => {
+            return;
```
tomaskikutis: That's probably some debug code you left here

scripts/core/auth/auth.ts
```diff
@@ -21,6 +21,13 @@ window.fetch = (...args) => {
         return originalFetch(resource, config);
     }
 
+    if (appConfig.testing) {
```
tomaskikutis: It'd be good to avoid `if(testing)` code. I've commented it out and tests still pass so let's drop it.

If dropping it wasn't an option we could look into overwriting fetch in test code rather than here. [reference](https://github.com/facebook/create-react-app/issues/967#issuecomment-256358738)

tomaskikutis (approved):

scripts/core/superdesk-api.d.ts
```diff
@@ -3330,6 +3330,8 @@ declare module 'superdesk-api' {
 
 
     export interface ISuperdeskGlobalConfig {
+        testing?: boolean;
```
tomaskikutis: one more thing - this is probably not used anymore

PR #4745, opened 2025-02-05
Authoring UI improvements

thecalcc:
STT-31

Improve UI so it's less coupled to actions, and embedded editors in planning look better.

### DIFF
--- a/scripts/apps/authoring-react/authoring-integration-wrapper.tsx
+++ b/scripts/apps/authoring-react/authoring-integration-wrapper.tsx
@@ -14,7 +14,7 @@ import {
     IRestApiResponse,
     IFieldsData,
 } from 'superdesk-api';
-import {AuthoringReact, SPELLCHECKER_PREFERENCE} from './authoring-react';
+import {AuthoringReact} from './authoring-react';
 import {getFieldsAdapter} from './field-adapters';
 import {dispatchCustomEvent} from 'core/get-superdesk-api-implementation';
 import {appConfig, extensions} from 'appConfig';
@@ -46,8 +46,6 @@ import {AuthoringIntegrationWrapperSidebar} from './authoring-integration-wrappe
 import {assertNever} from 'core/helpers/typescript-helpers';
 import {ContentProfileDropdown} from './subcomponents/content-profile-dropdown';
 import {IconButton} from 'superdesk-ui-framework';
-import {preferences} from 'api/preferences';
-import {dispatchEditorEvent} from './authoring-react-editor-events';
 
 export function getWidgetsFromExtensions(article: IArticle): Array<IArticleSideWidget> {
     return Object.values(extensions)
--- a/scripts/apps/authoring-react/authoring-integration-wrapper.tsx
+++ b/scripts/apps/authoring-react/authoring-integration-wrapper.tsx
@@ -459,28 +459,7 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
                                 fieldsData,
                             );
 
-                            const getSpellcheckerAction = (): IAuthoringAction => {
-                                if (appConfig.features.useTansaProofing !== true) {
-                                    return {
-                                        label: spellchecker.enabled
-                                            ? gettext('Disable spellchecker')
-                                            : gettext('Enable spellchecker'),
-                                        onTrigger: () => {
-                                            spellchecker.toggleSpellchecker(!spellchecker.enabled);
-                                        },
-                                        keyBindings: {
-                                            'ctrl+shift+y': () => {
-                                                spellchecker.toggleSpellchecker(!spellchecker.enabled);
-                                            },
-                                        },
-                                    } as IAuthoringAction;
-                                }
-
-                                return {} as IAuthoringAction;
-                            };
-
-                            return [
-                                getSpellcheckerAction(),
+                            const actions = [
                                 getSaveAsTemplate(getLatestItem),
                                 getCompareVersionsModal(
                                     getLatestItem,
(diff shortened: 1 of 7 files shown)


tomaskikutis:
is this not ready for review? could you describe the PR?

thecalcc:
> is this not ready for review? could you describe the PR?

yes it's not, that's why i put it to draft. Purpose is to address all UI issues fritz mentioned

scripts/core/superdesk-api.d.ts
```diff
@@ -212,6 +212,7 @@ declare module 'superdesk-api' {
         authoringStorage: IAuthoringStorage<T>;
         storageAdapter: IStorageAdapter<T>;
         fieldsAdapter: IFieldsAdapter<T>;
+        extraActions?: Array<IAuthoringAction>;
```
tomaskikutis: is there a reason why an additional prop is needed vs just passing everything via `getActions`?

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -378,6 +382,90 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
...
+                            const actions = [];
+                            const {state} = this;
+
+                            if (appConfig.features.useTansaProofing !== true) {
+                                if (state.spellcheckerEnabled) {
+                                    const nextValue = false;
+
+                                    actions.push({
+                                        label: gettext('Disable spellchecker'),
+                                        onTrigger: () => {
```
tomaskikutis: Spellchecking implementation needs to be either inside our outside authoring react. We won't be easily able to move it out, not sure even if it would make sense. Thus let's keep it inside. Then do not expose **implementation details** outside authoring react. If you need to interact with it - use `exposed` to expose the interface like

```
spellchecker: {
    enabled: boolean;
    enable(): void;
    disable(): void;
},
```

scripts/core/superdesk-api.d.ts
```diff
@@ -175,6 +175,10 @@ declare module 'superdesk-api' {
         reinitialize(item: T, profile?: IContentProfileV2): void;
         getValidationErrors(): IAuthoringValidationErrors;
         setValidationErrors(validationErrors: IAuthoringValidationErrors): void;
+        spellchecker: {
+            enabled: boolean;
+            toggleSpellchecker: (nextValue: boolean) => void;
```
tomaskikutis: When a function is called `toggleSomething` it should not accept arguments since toggling means changing to opposite state.

This would be better
`setSpellcheckerStatus: (enabled: boolean) => void;`

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -1317,92 +1335,7 @@ export class AuthoringReact<T extends IBaseRestApiResponse>
...
-                            },
-                        },
-                    });
-                }
-            }
-
-            return [...coreActions, ...actions];
-        })();
-
+        const authoringActions = (() => this.props.getActions?.(exposed) ?? [])();
```
tomaskikutis: why wrap into IIFE?

`const authoringActions = this.props.getActions?.(exposed) ?? [];` looks fine

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -447,14 +451,36 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
...
+                                            : gettext('Enable spellchecker'),
+                                        onTrigger: () => {
+                                            spellchecker.toggleSpellchecker(!spellchecker.enabled);
+                                        },
+                                        keyBindings: {
+                                            'ctrl+shift+y': () => {
+                                                spellchecker.toggleSpellchecker(!spellchecker.enabled);
+                                            },
+                                        },
+                                    } as IAuthoringAction;
```
tomaskikutis: Do not cast when it can be avoided because you are disabling type checking for this segment of code

image

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -447,14 +451,36 @@ export class AuthoringIntegrationWrapper extends React.PureComponent<IPropsWrapp
...
+                                }
+
+                                return {} as IAuthoringAction;
```
tomaskikutis: If you don't have an item just don't put it into the array instead of putting a fake item and casting to make the error go away

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -14,10 +14,10 @@ import {
     IRestApiResponse,
     IFieldsData,
 } from 'superdesk-api';
-import {AuthoringReact} from './authoring-react';
+import {AuthoringReact, SPELLCHECKER_PREFERENCE} from './authoring-react';
```
tomaskikutis: remove unused imports

scripts/apps/authoring-react/authoring-react.tsx
thecalcc: didn't finish the refactoring from previous code

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
thecalcc: it's even easier and can be done inline without unnecessary cost declarations
image

scripts/apps/authoring-react/authoring-integration-wrapper.tsx
```diff
@@ -291,6 +291,7 @@ interface IPropsWrapper extends IProps {
 interface IState {
     sidebarMode: boolean | 'hidden';
     sideWidget: ISideWidget;
+    spellcheckerEnabled: boolean,
```
tomaskikutis: drop this one, it's not used

thecalcc:
Wait before merging: Update UI framework to `v4.0.6`

scripts/apps/authoring-react/fields/datetime/index.tsx
```diff
@@ -10,7 +10,7 @@ import {Editor} from './editor';
 import {Preview} from './preview';
 import {Difference} from './difference';
 
-export const DATETIME_FIELD_ID = 'datetime';
+export const DATETIME_FIELD_ID = 'datetime-v2';
```
tomaskikutis, line 13: why v2? We only have one implementation of a datime in authoring-react, no?

thecalcc: We have some other one in extensions with the same id

tomaskikutis: Good catch. What about we stick to a single implementation? And the one from the extension is already used elsewhere so maybe it would work here too?

tomaskikutis (approved):
fix ci

PR #4751, opened 2025-02-11
add custom validation capability to field types in authoring-react

tomaskikutis:
STT-31

### DIFF
--- a/scripts/apps/authoring-react/authoring-react.tsx
+++ b/scripts/apps/authoring-react/authoring-react.tsx
@@ -994,42 +994,38 @@ export class AuthoringReact<T extends IBaseRestApiResponse>
             const {profile} = state;
             const allFields = profile.header.merge(profile.content);
 
-            const validationErrorsArray: Array<{field: IAuthoringFieldV2; error: string}> =
-                allFields
-                    .toArray()
-                    .flatMap((field) => {
-                        const fieldType = getField(field.fieldType);
-                        const fieldValue = state.fieldsDataWithChanges.get(field.id);
+            let validationErrors = Map<string, string>(); // fieldId, errorMessage
 
-                        const error = ((): string | null => {
-                            if (!fieldType.hasValue(fieldValue)) {
-                                return field.fieldConfig.required === true ? gettext('Field is required') : null;
-                            } else if (fieldType.validate != null) {
-                                return fieldType.validate(fieldValue, field.fieldConfig);
-                            } else {
-                                return null;
-                            }
-                        })();
+            allFields.forEach((field) => {
+                const fieldType = getField(field.fieldType);
+                const fieldValue = state.fieldsDataWithChanges.get(field.id);
 
-                        return error == null ? [] : {field, error};
-                    });
+                const error = ((): string | null => {
+                    if (!fieldType.hasValue(fieldValue)) {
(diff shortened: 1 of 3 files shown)


scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -993,25 +994,43 @@ export class AuthoringReact<T extends IBaseRestApiResponse>
...
+                    .toArray()
+                    .flatMap((field) => {
+                        const fieldType = getField(field.fieldType);
+                        const fieldValue = state.fieldsDataWithChanges.get(field.id);
+
+                        const error = ((): string | null => {
+                            if (!fieldType.hasValue(fieldValue)) {
+                                return field.fieldConfig.required === true ? gettext('Field is required') : null;
+                            } else if (fieldType.validate != null) {
+                                return fieldType.validate(fieldValue, field.fieldConfig);
```
thecalcc: You can do 
```
else {
  return fieldType.validate?.(fieldValue, field.fieldConfig);
}
```
or even without the `else` just `return`

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -993,25 +994,43 @@ export class AuthoringReact<T extends IBaseRestApiResponse>
...
             const allFields = profile.header.merge(profile.content);
 
-            const validationErrors: IAuthoringValidationErrors = allFields.toArray()
-                .filter((field) => {
-                    if (field.fieldConfig.required === true) {
-                        const FieldEditorConfig = getField(field.fieldType);
+            const validationErrorsArray: Array<{field: IAuthoringFieldV2; error: string}> =
+                allFields
+                    .toArray()
+                    .flatMap((field) => {
```
thecalcc: Might be better to keep it an `object` and reduce to `validationErrors: IAuthoringValidationErrors` which also would be an object. If I recall correctly `allFields` is a dictionary as well, so shouldn't be a problem. You then won't need to convert it back to an `object` on line `1070`

thecalcc: You can make an argument that `object` doesn't hold key order, but you can also use Map which guarantees that. In that regard we might improve the `IAuthoringValidationErrors` interface to use Map instead of an object. I'm talking about native JS Map

thecalcc (approved):

scripts/apps/authoring-react/authoring-react.tsx
```diff
@@ -58,6 +58,7 @@ import ng from 'core/services/ng';
 import {focusFirstChildInput} from 'utils/focus-first-child-input';
 import {EDITOR_3_FIELD_TYPE} from './fields/editor3';
 import memoizeOne from 'memoize-one';
+import {arrayToDictionary} from 'core/helpers/array-to-dictionary';
```
thecalcc: remove unused import

scripts/core/helpers/array-to-dictionary.tsx
```diff
@@ -0,0 +1,14 @@
+export function arrayToDictionary<T, V>(
```
thecalcc, line 1: remove unused function

tomaskikutis: I'll leave the helper, we might use it later

PR #4756, opened 2025-02-17
SDAAP-101: Fix issue with find pattern matching replace

rbi-aap:
- Resolved problem where find pattern incorrectly matched the replace
- Improved regex handling to ensure correct replacements
- Added test cases to verify expected behavior

### DIFF
--- a/scripts/core/editor3/helpers/table.ts
+++ b/scripts/core/editor3/helpers/table.ts
@@ -96,7 +96,29 @@ export function setCell(data, row, col, cellEditorState: EditorState) {
 
     return {data, needUpdate, forceUpdate};
 }
+export function getData(contentState: ContentState, blockKey: string) {
+    const block = contentState.getBlockForKey(blockKey);
+    const entityKey = block.getEntityAt(0);
 
+    if (!entityKey) {
+        return {};
+    }
+
+    const entity = contentState.getEntity(entityKey);
+    const entityData = entity.getData();
+
+    const blockData = block.getData()?.get('data');
+
+    if (!blockData) {
+        return entityData || {};
+    }
+
+    try {
+        return JSON.parse(blockData);
+    } catch (error) {
+        return {};
+    }
+}
 /**
  * @ngdoc method
  * @name getData
@@ -105,7 +127,7 @@ export function setCell(data, row, col, cellEditorState: EditorState) {
  * @description Returns the data contained in the entity of this atomic block.
  * @return {Object}
  */
-export function getData(contentState: ContentState, blockKey: string) {
+export function getDataOld(contentState: ContentState, blockKey: string) {
     const block = contentState.getBlockForKey(blockKey);
     const entityKey = block.getEntityAt(0);
     const {data} = contentState.getEntity(entityKey).getData();
@@ -118,7 +140,6 @@ export function getData(contentState: ContentState, blockKey: string) {
 
     return JSON.parse(blockData);
 }
-
 /**
  * @ngdoc method
  * @name setData
--- a/scripts/core/editor3/reducers/find-replace.tsx
+++ b/scripts/core/editor3/reducers/find-replace.tsx
@@ -1,6 +1,4 @@
-// eslint complains about imported types not being used
-// eslint-disable-next-line no-unused-vars
-import {Modifier, EditorState} from 'draft-js';
+import {Modifier, EditorState, SelectionState} from 'draft-js';
 import {clearHighlights, quietPush, forEachMatch} from '../helpers/find-replace';
 import {onChange} from './editor3';
 import {escapeRegExp} from 'core/utils';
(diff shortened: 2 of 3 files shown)


rbi-aap:
Hi Andrew and Team, please review, thanks

MarkLark86:
@tomaskikutis Could you please review this PR, and see if it is also needed in v2.8, 2.9 & develop branches please.
Cheers

tomaskikutis:
@rbi-aap if the browser was freezing it was probably due to infinite loop. Why was it causing an infinite loop?

Is your code only handling the infinite loop issue? If not, what other issues is it handling?

I see after your changes, 2 different functions exist - one to replace at index and another to replace all. Could you stick to a single implementation in a similar fashion how it was previously? e.g. when multiple replacements were needed, the function doing a single replacement would be called multiple times.

marwoodandrew (approved):

rbi-aap:
@tomaskikutis, 
Thanks for the review. I refactored the replacement logic to improve stability, prevent unintended behavior, and ensure correct text updates. Unit tests were added to validate edge cases. Appreciate your time.

tomaskikutis (changes requested):
Could you answer to each of my 3 questions separatelly?

rbi-aap:
@tomaskikutis ,Thanks for the review. The infinite loop was caused by unintended offset shifts leading to continuous execution and in turn, possible issues like callback hell, which have been fixed for stability. The update also prevents duplicate replacements and ensures correct text updates. The functions were split to avoid instability, as a single implementation didn’t handle multiple replacements reliably. thanks

rbi-aap:
Hi Team, can you review more test case for QA with its function, thanks

tomaskikutis:
Can you provide a sample input text, pattern and replacement text which was causing the infinite loop?

> The update also prevents duplicate replacements and ensures correct text updates.

Can you specify samples where it was causing duplicate replacements or incorrect updates?

<hr>

Please rework the tests so they are simpler
- limit sample text to a single line.
- Clean up test names so it is clear which use case is being tested. For example - this it not clear at all - "HIGHLIGHTS_FIND_REPLACE_SINGLE_AND_REPLACE_ALL"

<hr>

`Replace all $ with $AUD ,but no change with $AUD` - this one is asserting incorrect behavior. It should replace `$AUD` too with the pattern that you specified.

```javascript
'I have $100 in my wallet. The total cost is $50, but sometimes it is $AUD60'.replace(/\$/g, '$AUD')
// > 'I have $AUD100 in my wallet. The total cost is $AUD50, but sometimes it is $AUDAUD60'
```

rbi-aap:
Hi @tomaskikutis, 

Thanks for your review!

- Infinite loop sample: "$100" with pattern "$" and replacement "$AUD" looped endlessly.
- Update resolves duplicates and ensures correct replacements.
- Fixed duplicate sample: "$AUD60" no longer becomes "$AUDAUD60" with "$" → "$AUD".
- Tests are simpler, using AAA pattern—see new commit.
- Sample text now single-line, e.g., "Total: $100, cost: $50".
- Test names improved, e.g., "CombinedReplacementScenarios".
- Added QA test for edge cases.

Really appreciate your support!

tomaskikutis:
I've looked into the issue and indeed there is a bug with an infinite loop. The cause was that the function was performing replacement on text that was already replaced.

For example if text was `air` and we wanted to replace "air" with "airplane", it would do it correctly the first iteration, but then would search for "air" in "airplane" and would replace to "airplaneplane", then to "airplaneplaneplane" - and would never stop.

I've fixed the infinite loop in https://github.com/superdesk/superdesk-client-core/pull/4758 by adjusting code so it does not attempt to replace what was already replaced once.

Besides that - I don't see any other issues.

Your comment about duplicates is describing correct behavior. I've added a test that given text "dev dev dev" and trying to replace "dev" with "developers" it should output "developers developers developers". It's not a bug - that's how "replace all" is supposed to work.

Regarding currency conversion where you want to change "$" to "$AUD" - but keep existing mentions of "$AUD" intact - that can't be done with "find and replace" feature without regex support. I don't suggest to expose regex support directly to end-users either - that'd be an overkill for superdesk users.

The best solution I think would be to have pre-made custom replacement strategies. It could either be integrated into "find and replace" widget or you could use macros. There is one already for [USD to CAD](https://github.com/user-attachments/assets/c4f879e2-2da2-4a3a-9b18-47147bd31476) conversion.

rbi-aap:
Hi Team, 
Please review the updated PR
### Fix infinite loop, shrinking bug & improve find-replace logic
### Summary
This PR addresses key issues in find-replace functionality:
- Fixed shrinking replacement issue ($AUD → $) to ensure correct replacements.
- Prevented infinite loop by correctly updating `lastIndex` in regex.
- Improved&FIX  handling of $ → $AUD likewise pattern
- Enhanced table entity detection for find-replace inside tables.
### 
QA & Testing
- Unit tests cover shrinking & expanding replacements.
- Tested `$ → $AUD` and `$AUD → $` transformations for consistency.
- able replacement logic verified.

#Review Notes
- [ ]  Please check if all edge cases are correctly handled.
- [ ] Let me know if you find any further improvements.
Thanks for reviewing.

tomaskikutis:
I've merged the infinite loop fix (https://github.com/superdesk/superdesk-client-core/pull/4758) and cherry-picked to [hotfix/2.5.3](https://github.com/superdesk/superdesk-client-core/tree/hotfix/2.5.3).

Besides that, I'm not aware of any other issues. As I've mentioned, your comments about "duplicates" are describing correct behavior.

Let me know if you see any legitimate bugs remaining.

PR #4760, opened 2025-02-25
support node 20+, update webpack

petrjasek:


package.json
```diff
@@ -162,13 +163,15 @@
...
-    "test": "npm run lint && npm run unit && node tasks/verify-client-api-changes.js",
+    "test": "npm run lint && npm run unit && npm run verify-client-api-changes",
     "debug-unit-tests": "karma start --reporters=progress --browsers=Chrome",
     "unit": "karma start --single-run",
     "lint": "python3 ./grep-lint.py && tsc -p scripts --noEmit && eslint --quiet --parser=@typescript-eslint/parser --ext .js --ext .jsx --ext .ts --ext .tsx scripts e2e/client tasks",
     "lint-fix": "eslint --fix --parser=@typescript-eslint/parser --ext .js --ext .jsx --ext .ts --ext .tsx scripts e2e/client tasks",
     "server": "grunt server",
-    "dev": "npm run server"
+    "dev": "npm run server",
+    "build": "grunt build",
```
tomaskikutis: running client-core without main repo isn't supported so we shouldn't add `dev` and `build` script here to avoid confusion.

.github/workflows/tests.yml
```diff
@@ -1,51 +1,76 @@
...
-    # avoid file watch limit error
-    - run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
-    - run: npm ci
-    - run: npm run test || npm run test || npm run test
+
+    - run: npm install # needed while we support v14
+
+    - run: npm run lint
+    - run: npm run unit
+    - run: npm run verify-client-api-changes
```
tomaskikutis: why not only run `test`? It includes all 3 tasks

build-tools/src/po-to-json/index.js
```diff
@@ -90,7 +89,7 @@ function compileTranslationsPoToJson(translationsPoDir, translationsJsonDir) {
             return;
         }
 
-        if (_.endsWith(filename, '.po') !== true) {
+        if (/\.po$/.test() !== true) {
```
tomaskikutis: `endsWith` is now in standard library

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

end-to-end-testing-helpers/package.json
```diff
@@ -14,11 +14,11 @@
...
   },
   "devDependencies": {
-    "protractor": "^7.0.0",
+    "protractor": "7.0.0",
     "superdesk-code-style": "1.3.0",
     "typescript": "^4.5.2"
   },
   "peerDependencies": {
-    "protractor": "5.4.2"
+    "protractor": "7.0.0"
```
tomaskikutis: if you have it in dev dependencies there is probably no reasons to also put it in `peerDependencies`

scripts/core/services/asset.ts
```diff
@@ -14,13 +14,14 @@ export default angular.module('superdesk.core.services.asset', [])
         this.templateUrl = function(path) {
             var ret = path;
 
-            if (/^(apps|core)/.test(path)) {
+            if (/^(core|apps)/.test(path)) {
```
tomaskikutis: doe this make a difference?

scripts/core/editor3/helpers/draftInsertEntity.spec.ts
```diff
@@ -3,23 +3,19 @@
 import {insertEntity} from './draftInsertEntity';
 import {convertFromRaw, SelectionState, EditorState, RawDraftContentState, convertToRaw} from 'draft-js';
 
-// removes keys from blocks and their entity ranges so resulting objects can be compared by value as JSON strings
+// set static keys for blocks and their entity ranges so resulting objects can be compared by value as JSON strings
```
tomaskikutis, line 6: what wasn't working with simply removing them?

scripts/core/auth/auth.ts
```diff
@@ -17,7 +17,7 @@ window.fetch = (...args) => {
     /**
      * Prevent requests not coming from our app get intercepted.
      */
-    if (resource.toString().includes(appConfig.server.url) === false) {
+    if (resource.toString().includes(appConfig.server.url) === false || resource.toString().includes('client_config')) {
```
tomaskikutis: wouldn't `includes(appConfig.server.url)` condition also match `client_config`? Base server URL must be the same?

scripts/core/utils.tsx
```diff
@@ -53,11 +53,16 @@ function requestListener() {
 if (language === 'en') {
     applyTranslations(DEFAULT_ENGLISH_TRANSLATIONS);
 } else {
-    const req = new XMLHttpRequest();
-
-    req.addEventListener('load', requestListener);
-    req.open('GET', filename, false);
-    req.send();
+    fetch(filename).then((response) => {
```
tomaskikutis: By allowing it to be async, some code might execute sooner and call `gettext` before translations are loaded.

See if you can keep the old blocking version.

In the long run we should extract translation loading out of the main bundle in order to use gettext synchronously at the top level in the main bundle.

scripts/apps/vocabularies/tests/vocabularies.spec.tsx
```diff
@@ -61,7 +61,10 @@ describe('vocabularies', () => {
         const fakeEvent = {target: {value: '2'}};
 
         wrapper.find('input[type="number"]').simulate('change', fakeEvent);
-        expect(instance.getItemsForSaving()[0].qcode).toBe(2);
+
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        expect(instance.getItemsForSaving()[0].priority).toBe(2);
```
tomaskikutis: `(x as unknown as any).abc` allows to avoid `@ts-ignore`

tasks/options/webpack.js
```diff
@@ -1,31 +1,9 @@
...
-
-    config.progress = !grunt.option('webpack-no-progress');
-    config.devtool = grunt.option('webpack-devtool') || 'source-map';
-
-    config.module.rules = config.module.rules.map((rule) => {
-        if (rule.loader === 'ts-loader') {
-            // skipping typechecking here to make build faster
-            // types are checked in `npm run lint` using
-            // typescript compiler directly which is faster
-            rule.options.transpileOnly = true;
```
tomaskikutis, line 16: dropping `transpileOnly` might make webpack dev server reloads slower.

.github/workflows/tests.yml
petrjasek: it was failing on some and this makes it easier to see where exactly

scripts/core/services/asset.ts
petrjasek: no

.github/workflows/tests.yml
petrjasek: reverted to unit

scripts/core/auth/auth.ts
petrjasek: yep that shouldn't be needed

tasks/options/webpack.js
petrjasek: it's set in webpack.config.js

tomaskikutis:
I've resolved most comments. Could you respond to remaining 2?

scripts/core/editor3/helpers/draftInsertEntity.spec.ts
petrjasek: I was getting errors like:
```
Property 'key' is missing in type '{ text: string; type: string; depth: number; inlineStyleRanges: undefined[]; entityRanges: undefined[]; data: {}; }' but required in type '{ key: Any | ExpectedRecursive<string>; type: Any | ExpectedRecursive<string>; text: Any | ExpectedRecursive<string>; depth: Any | ExpectedRecursive<...>; inlineStyleRanges: Any | ExpectedRecursive<...>; entityRanges: Any | ExpectedRecursive<...>; data?: Any | ExpectedRecursive<...>; }'
```

tomaskikutis: I see

tomaskikutis (approved):

PR #4772, opened 2025-03-21
improve ui-framework linking

tomaskikutis:
Will ensure that the same version of ui-framework is used in development in all apps(planning, analytics) and extensions.

@thecalcc you can use it even before it's merged to test radio group in planning

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14254,9 +14254,9 @@
       }
     },
     "superdesk-ui-framework": {
-      "version": "4.0.24",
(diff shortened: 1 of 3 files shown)


petrjasek:
I wonder how do you know it's only used during development

tomaskikutis:
Good point. The comment is not accurate after all. If client-core is the only user of ui-framework and there are no other apps like planning or analytics - it will get used too in production

petrjasek (approved):

PR #4777, opened 2025-03-31
new line on enter, support expandable config option

thecalcc:
STT-156

### DIFF
--- a/scripts/core/editor3/components/Editor3Component.tsx
+++ b/scripts/core/editor3/components/Editor3Component.tsx
@@ -152,6 +152,7 @@ export interface IPropsEditor3Component {
 interface IState {
     draggingInProgress: boolean;
     expanded?: boolean;
+    showExpandButton?: boolean;
 
     /**
      * Only content changes are tracked (addition/removal of chars, whitespace).
@@ -187,7 +188,6 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
     div: HTMLDivElement;
     editor: IEditor3;
     onDragEnd: () => void;
-    showExpandButton: boolean;
     collapsedHeight: number;
 
     private removeListeners: Array<() => void> = [];
@@ -208,12 +208,11 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
         this.keyBindingFn = this.keyBindingFn.bind(this);
         this.handleDropOnEditor = this.handleDropOnEditor.bind(this);
         this.spellcheck = this.spellcheck.bind(this);
-        this.shouldShowCollapse = this.shouldShowCollapse.bind(this);
+        this.shouldShowFieldToggle = this.shouldShowFieldToggle.bind(this);
         this.scheduleSpellchecking = debounce(this.spellcheck.bind(this), 1000);
 
         this.spellcheckAbortController = new AbortController();
 
-        this.showExpandButton = true;
         this.collapsedHeight = LINE_HEIGHT * (this.props.expandable?.numberOfRowsWhenCollapsed ?? 1);
 
         this.onDragEnd = () => {
@@ -223,6 +222,7 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
         };
 
         this.state = {
+            showExpandButton: true,
             expanded: this.props.expandable?.defaultValue ?? false,
             draggingInProgress: false,
             contentChangesAfterLastFocus: 0,
@@ -231,13 +231,14 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
         this.removeListeners = [];
     }
 
-    shouldShowCollapse() {
+    shouldShowFieldToggle() {
         const textInputArea = this.div?.querySelector?.('.focus-screen');
+        const shouldShowExpand = textInputArea?.scrollHeight > this?.collapsedHeight;
 
-        if (textInputArea?.scrollHeight > this?.collapsedHeight) {
-            this.showExpandButton = true;
-        } else {
-            this.showExpandButton = false;
+        if (this.state.showExpandButton != shouldShowExpand) {
+            this.setState({
+                showExpandButton: shouldShowExpand,
+            });
         }
     }
 
@@ -528,7 +529,7 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
     componentDidMount() {
         $(this.div).on('dragover', this.onDragOver);
 
-        this.shouldShowCollapse();
+        this.shouldShowFieldToggle();
 
         if (!window[EDITOR_GLOBAL_REFS]) {
             window[EDITOR_GLOBAL_REFS] = {};
@@ -572,12 +573,19 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
         }
     }
 
-    componentDidUpdate(prevProps) {
+    componentDidUpdate(prevProps: IPropsEditor3Component) {
         if (window.hasOwnProperty('instgrm')) {
             window.instgrm.Embeds.process();
         }
 
-        this.shouldShowCollapse();
+        if (this.props.expandable?.defaultValue != prevProps.expandable?.defaultValue) {
+            // eslint-disable-next-line react/no-did-update-set-state
+            this.setState({
+                expanded: this.props.expandable?.defaultValue,
+            });
+        }
+
+        this.shouldShowFieldToggle();
 
         if (
             this.props.spellchecking.enabled &&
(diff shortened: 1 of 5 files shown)


tomaskikutis:
What's the purpose of the PR?

tomaskikutis:
The title gives a hint, but maybe you could post a demo video if this is a new feature. New lines I thought we already have and expandable config I don't know how it is supposed to work.

thecalcc:
Demo: 

https://github.com/user-attachments/assets/83b27f88-85ac-476a-baab-c099e084249d

thecalcc:
> The title gives a hint, but maybe you could post a demo video if this is a new feature. New lines I thought we already have and expandable config I don't know how it is supposed to work.

I posed a demo. new lines we do have, I've only changed the shortcut from `Shift + Enter` to just `Enter` to make it consistent with the old editor

thecalcc:
Updated demo with an auto hiding button when content isn't pass the collapse height threshold:

https://github.com/user-attachments/assets/d81c9181-513d-43df-87ea-ec5dc5540248

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -645,85 +676,98 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
...
-                        )
-                    }
+                        )}
 
-                    {this.props.loading && <div className="loading-overlay active" />}
+                        {this.props.loading && <div className="loading-overlay active" />}
+                    </div>
                 </div>
-            </div>
+                {this.showExpandButton && this.props.expandable?.enabled && (
```
tomaskikutis: If you are doing conditional rendering based on a value - you probably want things to re-render if that value changes - thus you should store it in the state rather than as a component property.

scripts/core/superdesk-api.d.ts
```diff
@@ -469,6 +469,12 @@ declare module 'superdesk-api' {
         cleanPastedHtml?: boolean;
         disallowedCharacters?: Array<string>;
 
+        expandable?: {
+            enabled: true;
+            defaultValue: boolean;
+            numberOfRowsWhenCollapsed: boolean;
```
tomaskikutis: if prop changes on subsequent renders - it will be ignored. It shouldn't be.

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -62,6 +63,8 @@ const VALID_MEDIA_TYPES = [
 
 export const EDITOR_GLOBAL_REFS = 'editor3-refs';
 const editor3AutocompleteClassName = 'editor3-autocomplete';
+const LINE_HEIGHT = 30;
```
tomaskikutis: This will not work if font size changes. User zooming might break it too. See if you can setup a `em` unit based height, like `3em`

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -139,10 +142,16 @@ export interface IPropsEditor3Component {
     canAddArticleEmbed?: (srcId: string) => Promise<typeof canAddArticleEmbed>;
     uiTheme?: IEditorComponentProps<unknown, unknown, unknown>['uiTheme'];
     showPopup?(type: any, data: any): void;
+    expandable?: {
```
tomaskikutis, line 143: does it work correctly if authoring panel is resized? And with body_html style fields?

thecalcc: yes

scripts/core/editor3/components/Editor3Component.tsx
thecalcc: zooming doesn't actually change the dimensions of DOM elements. but the other case is valid

scripts/core/editor3/components/Editor3Component.tsx
tomaskikutis: I didn't see resize listeners. How does it detect when line count has changed and "expand/collapse" button needs to be shown/hidden?

thecalcc: I thought you mean resizing through the content profile, not resizing the panel by dragging

scripts/core/editor3/components/Editor3Component.tsx
thecalcc: So are you saying if font-size is e.g. 40px we should have a higher threshold for when to show the expand collapse? Another PoV is that no matter the font size for text inside the input box, we should keep the height of the input box the same.

tomaskikutis: Your prop name is `numberOfRowsWhenCollapsed`. Which is good. When limiting height - you set maxHeight in pixels, hardcoding line height. You can't hardcode line height in pixels, because both font family and font size impacts it.

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -572,12 +573,19 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
...
     }
 
-    componentDidUpdate(prevProps) {
+    componentDidUpdate(prevProps: IPropsEditor3Component) {
         if (window.hasOwnProperty('instgrm')) {
             window.instgrm.Embeds.process();
         }
 
-        this.shouldShowCollapse();
+        if (this.props.expandable?.defaultValue != prevProps.expandable?.defaultValue) {
```
tomaskikutis: would it work to use get derived state from props here to avoid this?

thecalcc:
This is a demo with dynamic row height 


https://github.com/user-attachments/assets/cc04d5cf-6e2e-45d1-b653-29289eaab22b

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -541,11 +584,23 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
         }
     }
 
-    componentDidUpdate(prevProps) {
+    static getDerivedStateFromProps(props: IPropsEditor3Component, state: IState): Partial<IState> | null {
```
tomaskikutis: move static methods to the top of the class

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -603,39 +658,71 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
...
+            <Spacer v gap="0" justifyContent="center" alignItems="center" noWrap>
+                <div
+                    className={cx}
+                    ref={(div) => {
+                        this.div = div;
+
+                        // if element gets resized from somewhere we should
+                        // check if the expand button should still show
+                        if (div) {
+                            const resizeObserver = new ResizeObserver(() => {
```
tomaskikutis: ref function may run multiple times - ensure you do not end up with more than one observer.

I was about to suggest using lodash.once, but that would be incorrect too, because on every run, the element reference might be a different one and if you cache the first one, you would end up with a stale reference.

You'll need to disconnect old observer here and connect a new one.

scripts/core/editor3/components/Editor3Component.tsx
thecalcc: actually this doesn't work, I re-ran the code and when state changes this also runs and overrides the value I wanted to actually see. I'm going to keep it as it was - conditionally running in `componentDidUpdate`.

tomaskikutis: ok. might be that it is only supported if you don't have manually managed state

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -584,21 +585,18 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
...
-            return null;
-        }
-    }
-
     componentDidUpdate(prevProps: IPropsEditor3Component) {
         if (window.hasOwnProperty('instgrm')) {
             window.instgrm.Embeds.process();
         }
 
+        if (prevProps.expandable?.defaultValue !== this.props.expandable?.defaultValue) {
```
tomaskikutis: This is not correct I think. `defaultValue` should be only a **default**. Here it looks like you would use the default to override user selection which shouldn't happen.

thecalcc: I thought I included this in the first place, because you said value isn't getting changed if the `defaultValue` changes so I made it. I can remove this and have it set once in the constructor and that's it

thecalcc: I found the comment - https://github.com/superdesk/superdesk-client-core/pull/4777#discussion_r2024432464

tomaskikutis: yep, let's drop it.

It should respect the props until first usage, but fixing it in constructur will be fine too

tomaskikutis (approved):

PR #4780, opened 2025-04-07
Availability manager settings

tomaskikutis:
SDBELGA-924


scripts/extensions/availability-manager/src/settings/edit-workday-modal.tsx
```diff
@@ -0,0 +1,257 @@
...
+                            text={gettext('Cancel')}
+                            onClick={() => this.handleClose()}
+                            noMargin
+                            disabled={this.state.savingInProgress}
+                        />
+
+                        <Button
+                            text={gettext('Save')}
+                            type="primary"
+                            onClick={() => {
```
petrjasek: would prefer such a callback in its own function, mixing the rendering with too much logic is hard to read for me

scripts/extensions/availability-manager/src/settings/edit-workday-modal.tsx
```diff
@@ -0,0 +1,257 @@
...
+                                tagsWhitelist={this.props.tagsWhitelist}
+                            >
+                                {({inputs, labels}) => {
+                                    const columnCount = labels.length;
+
+                                    return (
+                                        <div
+                                            style={{
+                                                display: 'grid',
+                                                gap: '8px',
```
petrjasek: do we have any gap variable we could use here?

petrjasek (approved):

scripts/validate-instance-configuration.tsx
```diff
@@ -1,60 +1,117 @@
...
-                        Unsupported extension point is being used: `contributions.authoringHeaderComponents`
-                    </p>
+        issues.push((
+            <>
+                <Heading
+                    text={'Unsupported extension point is being used: `contributions.authoringHeaderComponents`'}
+                />
+
+                <p>
+                    You are likely running an outdated version of auto tagging extension.
```
petrjasek: seems like it's mixing translated strings with english ones, I wouldn't mind english only here, or having all translated, but wouldn't mix it

e2e/client/playwright/extensions/availability-manager/settings.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+    });
+
+    test('day preview', async ({page}) => {
+        await openAvailabilitySettings(page);
+
+        page.locator(s('month=mar')).getByRole('button', {name: '10'}).click();
+
+        await expect(page.locator(s('working-day-view', 'tags', 'tag'))).toHaveText(['Austria']);
+        await expect(page.locator(s('working-day-view'))).toContainText('Tuesday, March 10, 1970');
+        await expect(page.locator(s('working-day-view'))).toHaveScreenshot();
```
petrjasek, line 27: I wonder how reliable are those screenshots, I guess we will soon find out :smile:

petrjasek (approved):

e2e/client/playwright/extensions/availability-manager/settings.spec.ts
tomaskikutis: Yep, well have to see how well it works. I'm optimistic it would work well. We'd need to regenerate the screenshots when doing UI changes which might get time consuming, but we could automate it. We'd review the changes and approved ones would get regenerated.

Playwright has a very good screenshot comparison tool. There's also a config how much difference is tolerated, but I'll see how far we go without it.

https://github.com/user-attachments/assets/1f4d1a30-d7a3-416f-8210-9343d727ad4a


I had an issue with it failing due to font rendering. Loading the font in a blocking way might help to avoid fallback font displayed. I saw some screenshots where it seemed like the font did load, but was rendering in a buggy way in headless mode - letters being squashed too close.

I've set font to Arial for tests and it appears to work well now.

PR #4789, opened 2025-04-15
create 'item appearing in publish queue after publishing with subscriber' test

dzonidoo:
SDESK-7558

### DIFF
--- a/e2e/client/playwright/publish-queue.spec.ts
+++ b/e2e/client/playwright/publish-queue.spec.ts
@@ -1,7 +1,7 @@
 import {test, expect} from '@playwright/test';
 import {Authoring} from './page-object-models/authoring';
 import {Monitoring} from './page-object-models/monitoring';
-import {restoreDatabaseSnapshot, s} from './utils';
+import {getCellValueByColumTitle, restoreDatabaseSnapshot, s} from './utils';
 
 test.setTimeout(50000);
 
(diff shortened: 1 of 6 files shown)


e2e/client/playwright/publish-queue.spec.ts
```diff
@@ -26,3 +27,31 @@ test.skip('item appearing in publish queue after publishing', async ({page}) =>
 
     await expect(page.locator(s('publish-queue-item=test sports story'))).toBeAttached({timeout: 10000});
 });
+
+test('item appearing in publish queue after publishing with subscriber', async ({page}) => {
```
tomaskikutis, line 31: What are you attempting to test here? This test looks the same as the one above. If subscribers dropdown was broken it would probably still pass? If so, we're not testing anything that's not being tested already.

e2e/client/playwright/utils/index.ts
```diff
@@ -50,3 +50,26 @@ export function sleep(ms: number): Promise<void> {
...
+    for (let i = 0; i < count; i++) {
+        const text = await headers.nth(i).innerText();
+
+        if (text === tableHeading) {
+            columnIndex = i;
+            break;
+        }
+    }
+
+    const item = row.locator(`td:nth-child(${columnIndex + 1})`);
```
tomaskikutis, line 76: If `tableHeading` can't be found, it would return first column. Better throw an error.

e2e/client/playwright/publish-queue.spec.ts
```diff
@@ -47,11 +47,19 @@ test('item appearing in publish queue after publishing with subscriber', async (
...
+    await expect(page.locator(s('publish-queue-item=test sports story'))).toBeVisible({timeout: 10000});
 
     await page.locator(s('publish-queue-item=test sports story')).click();
 
     await expect(page.locator(
         s('authoring-preview', 'field--headline'),
     )).toHaveText('test sports story');
+
+    const value = await getCellValueByColumTitle(
+        page.locator('[data-test-id="publish-queue-table"]'),
```
tomaskikutis: use `s` function

tomaskikutis (approved):

PR #4791, opened 2025-04-22
Fix user editing

thecalcc:
SDESK-7573

### DIFF
--- a/scripts/apps/users/directives/UserEditDirective.ts
+++ b/scripts/apps/users/directives/UserEditDirective.ts
@@ -54,6 +54,10 @@ interface IScope {
     $watchCollection: (...args: any) => any;
 }
 
+const notAllowedToChangeYourself = ['is_active', 'is_enabled'] satisfies Array<keyof IUser>;
+
+export const readOnlyUserFields = ['dateline_source', 'last_activity_at'] satisfies Array<keyof IUser>;
+
 UserEditDirective.$inject = ['api', 'notify', 'usersService', 'userList', 'session', 'lodash',
     'langmap', '$location', '$route', 'superdesk', 'features', 'asset', 'privileges',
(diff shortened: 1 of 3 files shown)


scripts/apps/users/directives/UserEditDirective.ts
```diff
@@ -169,15 +217,29 @@ export function UserEditDirective(api, notify, usersService, userList, session,
                          * which will be used my extension point)
                          * usersService.save is only intended to be used when creating a new user
                          */
-                        const save = scope.$parent.$parent.onSave ?? usersService.save;
+                        const notAllowedToChangeYourself = ['is_active', 'is_enabled'];
+                        const fieldsToOmit = ['dateline_source', 'last_activity_at'];
```
tomaskikutis: why are they being omitted?

tomaskikutis (commented):
Use `satisfies Array<keyof IUser>` when omitting keys so field names are typed.

Why wasn't it saving?

scripts/apps/users/directives/UserEditDirective.ts
thecalcc: because `dateline_source` is a readOnly field that's not even in the backend schema and  `last_activity_at` can't be changed through users PATCH.

thecalcc: To answer why they weren't getting saved - these fields were getting passed, and then also the wrong input was getting passed to `usersService.save`. It needs two arguments - original item and patch containing only changed fields.

tomaskikutis: > because dateline_source is a readOnly field that's not even in the backend schema and last_activity_at can't be changed through users PATCH

would be nice to have `const readOnlyFields = [...]` so it's clear to everyone reading your code

tomaskikutis (approved):

thecalcc:
Don't merge before updating feature embedded plannings instance

PR #4802, opened 2025-04-30
Update ui framework

thecalcc:
SDBELGA-939

### DIFF
--- a/e2e/client/playwright/authoring.spec.ts
+++ b/e2e/client/playwright/authoring.spec.ts
@@ -52,7 +52,7 @@ test('cancel and ignore buttons from unsaved changes modal', async ({page}) => {
 
     // check unsaved changes modal is visible
     await page.locator(s('authoring-topbar', 'close')).click();
-    await expect(page.locator(s('unsaved-changes-dialog')).getByRole('dialog')).toBeVisible();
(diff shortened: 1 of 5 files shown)


thecalcc:
E2E test changes explained: 

After the change for `data-test-id` for the `Modal` component [here](https://github.com/superdesk/superdesk-ui-framework/commit/4fa4cf2b5838d2321409fdf5b49702ab95ee77a3#diff-88dcddc5384daa1f7d26a031620bb902618fa2cc6cb679375381b74413758308R65) searching by `role="dialog"` html for this modal looks like `<div role="dialog" data-test-id="unsaved-changes-dialog">`.

And trying to find this modal by chaining multiple selectors like this:

`await expect(page.locator(s('unsaved-changes-dialog').getByRole('dialog'))).toBeVisible();`

No longer works because this selector would only get a match if the there was a `div` with `role="dialog"` and inside it another `div` with `data-test-id="unsaved-changes-dialog"`. 


If you wish to have a combined selector by `role` and `data-test-id` we can do:

```
await expect(page.locator(s('unsaved-changes-dialog')).and(page.getByRole('dialog'))).toBeVisible();
```

In my opinion it's not necessary so I removed it.

tomaskikutis (approved):

PR #4806, opened 2025-05-05
Correct item e2e test

dzonidoo:
SDESK-7602

### DIFF
--- a/e2e/client/playwright/correct-item-corrections-workflow.spec.ts
+++ b/e2e/client/playwright/correct-item-corrections-workflow.spec.ts
@@ -32,7 +32,7 @@ test('can correct published item using corrections workflow', async ({page}) => 
         page.locator(s('monitoring-group=Sports / Working Stage', 'article-item=Story 5')),
     ).toBeVisible();
     await expect(
-        page.locator(s('monitoring-group=Sports / Working Stage', 'article-item=Story 5')).getByTitle('Correction'),
+        page.locator(s('monitoring-group=Sports / Working Stage', 'article-item=Story 5')).getByTitle('Correction', {exact: true}),
     ).toBeVisible();
 
     // edit item
@@ -53,6 +53,9 @@ test('can correct published item using corrections workflow', async ({page}) => 
     await expect(
         page.locator(s('monitoring-group=Sports desk output', 'article-item=Story 5.1')).getByTitle('Corrected'),
     ).toBeVisible();
+    await expect(
+        page.locator(s('monitoring-group=Sports desk output', 'article-item=Story 5.1')).getByTitle('Correction', {exact: true}),
+    ).not.toBeVisible();
 
     await page.goto('/#/publish_queue');
     await expect(page.locator(s('publish-queue-item=Story 5.1'))).toBeVisible({timeout: 10000});
--- a/e2e/client/playwright/correct-item-corrections-workflow.spec.ts
+++ b/e2e/client/playwright/correct-item-corrections-workflow.spec.ts
@@ -32,7 +32,9 @@ test('can correct published item using corrections workflow', async ({page}) => 
         page.locator(s('monitoring-group=Sports / Working Stage', 'article-item=Story 5')),
     ).toBeVisible();
     await expect(
-        page.locator(s('monitoring-group=Sports / Working Stage', 'article-item=Story 5')).getByTitle('Correction', {exact: true}),
+        page
+            .locator(s('monitoring-group=Sports / Working Stage', 'article-item=Story 5'))
+            .getByTitle('Correction', {exact: true}),
     ).toBeVisible();
 
     // edit item
@@ -54,7 +56,9 @@ test('can correct published item using corrections workflow', async ({page}) => 
         page.locator(s('monitoring-group=Sports desk output', 'article-item=Story 5.1')).getByTitle('Corrected'),
     ).toBeVisible();
     await expect(
-        page.locator(s('monitoring-group=Sports desk output', 'article-item=Story 5.1')).getByTitle('Correction', {exact: true}),
+        page
+            .locator(s('monitoring-group=Sports desk output', 'article-item=Story 5.1'))
+            .getByTitle('Correction', {exact: true}),
     ).not.toBeVisible();
 
     await page.goto('/#/publish_queue');
--- a/e2e/client/playwright/correct-item-corrections-workflow.spec.ts
+++ b/e2e/client/playwright/correct-item-corrections-workflow.spec.ts
@@ -1,6 +1,6 @@
 import {test, expect} from '@playwright/test';
 import {Monitoring} from './page-object-models/monitoring';
-import {restoreDatabaseSnapshot, s, waitingForToastMsg} from './utils';
+import {restoreDatabaseSnapshot, s, waitForToastMessage} from './utils';
 import {getStorageState} from './utils/storage-state';
 
 test.use({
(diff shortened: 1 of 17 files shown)


tomaskikutis:
Can you reference the ticket?

scripts/core/notify/notify.tsx
```diff
@@ -118,6 +118,7 @@ class NotifyComponent extends React.Component<{}, IState> {
                             className={`alert alert-${msg.type} space-between`}
                             onClick={() => this.removeMessage(i)}
                             data-test-id={`notification--${msg.type}`}
+                            data-test-value={`${msg.msg}`}
```
tomaskikutis: Why are you using string interpolation here?

e2e/client/playwright/page-object-models/authoring.ts
```diff
@@ -25,11 +25,19 @@ export class Authoring {
     field(field: string): Locator {
         return this.page.locator(s('authoring', field)).getByRole('textbox');
     }
+
+    async waitingForToastMsg(type: string, text: string): Promise<void> {
```
tomaskikutis: toast messages are not specific to authoring, thus should not go into this file

e2e/client/playwright/page-object-models/authoring.ts
```diff
@@ -25,11 +25,19 @@ export class Authoring {
...
+        const selector = `notification--${type}=${text}`;
```
tomaskikutis: Naming isn't the best here, since that string is the argument to `s` function, not actual CSS selector. It would be better to store the result of `s` - then you'd have a standard CSS selector - and would avoid calling `s` multiple times.

dzonidoo:
idk how i miss that !?!

e2e/client/playwright/correct-item.spec.ts
```diff
@@ -0,0 +1,33 @@
+import {test, expect} from '@playwright/test';
+import {Monitoring} from './page-object-models/monitoring';
+import {restoreDatabaseSnapshot, s, sleep} from './utils';
```
tomaskikutis, line 3: remove unused imports

e2e/client/playwright/correct-item.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    await page.locator(s('authoring')).locator(s('field--headline')).getByRole('textbox').clear();
+    await page.locator(s('authoring')).locator(s('field--headline')).getByRole('textbox').fill('Story 5.1');
+    await page.locator(s('authoring-topbar')).getByRole('button', {name: 'Send Correction'}).click();
+
+    await expect(
+        page.locator(s('monitoring-group=Sports desk output', 'article-item=Story 5.1')),
+    ).toBeVisible();
+
+    await page.goto('/#/publish_queue');
+    await expect(page.locator(s('publish-queue-item=Story 5.1'))).toBeAttached({timeout: 10000});
```
tomaskikutis: why are you using `toBeAttached`? Why not `toBeVisible`?

e2e/client/playwright/correct-item-corrections-workflow.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+import {Authoring} from './page-object-models/authoring';
+import {Monitoring} from './page-object-models/monitoring';
+import {restoreDatabaseSnapshot, s, sleep} from './utils';
+import {getStorageState} from './utils/storage-state';
+
+test.use({
+    storageState: getStorageState({corrections_workflow: true}),
+});
+
+test('can correct published item', async ({page}) => {
```
tomaskikutis: Test names should be unique. Here set it to `can correct published item using corrections workflow`

e2e/client/playwright/correct-item-corrections-workflow.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    const monitoring = new Monitoring(page);
+    const authoring = new Authoring(page);
+
+    await restoreDatabaseSnapshot();
+    await page.goto('/#/workspace/monitoring');
+
+    await expect(
```
tomaskikutis, line 21: This is unreliable because if it was loading slowly, it would pass as "not visible" before finishing loading and becoming visible.

Before using `not.toBeVisible` ensure the view has finished loading.

e2e/client/playwright/correct-item-corrections-workflow.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    await page.locator(s('authoring')).locator(s('field--headline')).getByRole('textbox').clear();
+    await page.locator(s('authoring')).locator(s('field--headline')).getByRole('textbox').fill('Story 5.1');
+    await page.locator(s('authoring-topbar')).getByRole('button', {name: 'Save'}).click();
+    await authoring.waitingForToastMsg('success', 'Item updated.');
+
+    await page.locator(s('authoring', 'open-send-publish-pane')).click();
+    await page
+        .locator(s('authoring', 'interactive-actions-panel'))
+        .getByRole('button', {name: 'Send correction'})
+        .click();
```
tomaskikutis, line 50: also check whether label "Correction" is not being shown anymore after correction was completed.

e2e/client/playwright/correct-item-corrections-workflow.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    await expect(
+        page.locator(s('monitoring-group=Sports desk output', 'article-item=Story 5.1')),
+    ).toBeVisible();
+
+    await page.goto('/#/publish_queue');
+    await expect(page.locator(s('publish-queue-item=Story 5.1'))).toBeAttached({timeout: 10000});
```
tomaskikutis: why are you using toBeAttached? Why not toBeVisible?

e2e/client/playwright/correct-item-corrections-workflow.spec.ts
tomaskikutis: have you checked this one?

dzonidoo: label 'correction' changes the text to 'corrected'.. so Im checking the visibility of that

e2e/client/playwright/correct-item-corrections-workflow.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    await page
+        .locator(s('authoring', 'interactive-actions-panel'))
+        .getByRole('button', {name: 'Send correction'})
+        .click();
+
+    await expect(
+        page.locator(s('monitoring-group=Sports desk output', 'article-item=Story 5.1')),
+    ).toBeVisible();
+    await expect(
+        page.locator(s('monitoring-group=Sports desk output', 'article-item=Story 5.1')).getByTitle('Corrected'),
```
dzonidoo, line 56: here it is...

e2e/client/playwright/correct-item-corrections-workflow.spec.ts
tomaskikutis: fine, but also check that "correction" is no longer shown

dzonidoo: but that doesnt make sense... one option excludes the other... if the label 'Corrected' exists, that means the text has been replaced on the label component and that label 'Correction' no longer exists...

tomaskikutis: > if the label 'Corrected' exists, that means the text has been replaced on the label component and that label 'Correction' no longer exists...

That is if the application works correctly. If it works incorrectly, both labels could be shown at once and your current test would pass when it shouldn't

tomaskikutis (approved):

PR #4811, opened 2025-05-15
Edit profile - e2e test

dzonidoo:
SDESK-7605


scripts/apps/users/directives/UserEditDirective.ts
```diff
@@ -205,6 +205,7 @@ export function UserEditDirective(api, notify, usersService, userList, session,
                         showConfirmationPrompt({
                             title: 'Do you want to reload the page now?',
                             message: 'The page needs to be reloaded to change the language',
+                            dataTestId: 'confirmation-modal',
```
tomaskikutis: this data test ID is generic so apply it as default inside that function instead of passing here

tomaskikutis (approved):

PR #4813, opened 2025-05-19
Correspondent availability view

tomaskikutis:
SDBELGA-919


scripts/extensions/availability-manager/src/correspondent-availability/participants.ts
```diff
@@ -0,0 +1,56 @@
...
+export function fetchParticipants(): Promise<Set<IUser['_id']>> {
+    const query: ISuperdeskQuery = {
+        filter: {
+            $and: [
+                {[nameof<IDefaultAvailability>('enabled')]: {$eq: true}},
+            ],
+        },
+        page: 1,
+        max_results: 200,
+        sort: [{'versioncreated': 'asc'}],
```
petrjasek: that field is not there

scripts/extensions/availability-manager/src/correspondent-availability/day-view.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+                                return (
+                                    <BoxedListItem
+                                        key={participantId}
+                                        type={(() => {
+                                            if (record == null) {
+                                                return 'default';
+                                            }
+
+                                            switch (record.status) {
```
petrjasek, line 93: looks like that logic is duplicated in https://github.com/superdesk/superdesk-client-core/pull/4813/files#diff-a4368b585f83eeb3681d18f74857ee73a05eefb16b713c8ca8a6247e98b033c2R127-R137

petrjasek (approved):

scripts/extensions/availability-manager/src/correspondent-availability/day-view.tsx
tomaskikutis: It does look like it because style options from 2 different components match at the moment. They are still separate components, so I don't think it's a good idea to couple it. The abstraction would look something like this which doesn't look good:

```
function getAvailabilityStateOrBoxedListItemType(): React.ComponentProps<typeof BoxedListItem>['type']
| React.ComponentProps<typeof CalendarWeekDayItem>['state']
```

PR #4816, opened 2025-05-27
[SDESK-7650]: Update to use async backend version

MarkLark86:


e2e/server/docker-compose.yml
```diff
@@ -9,18 +9,13 @@ services:
...
             - "27017:27017"
         networks:
             - e2e
-        environment:
-            - MONGO_INITDB_DATABASE=superdesk_e2e
         tmpfs:
             - /data/db
-        volumes:
-          - ./mongo-rs-initialize.js:/docker-entrypoint-initdb.d/mongo-rs-initialize.js:ro
-        command: --replSet rs0
```
petrjasek: we need that for recording

MarkLark86: Why is replica set's required for that?
Anyway, it wasn't working on newer pymongo lib, that's why I removed it. I'm guessing might be differences in the connection string or something like that. Will investigate

petrjasek: it's probably not required when running the tests, but it's required when devs are generating new snapshots locally

MarkLark86: @petrjasek I could not get it to work with using docker hostnames. Connection strings for replica sets required hostnames, and I'm guessing the newer pymongo enforces that, yet I could not get it to work (either with MongoDB 4.4 or 6).

I tried combination of the following:
* Setting `MONGODB_ADVERTISED_HOSTNAME` on the mongo service to `mongo` or `localhost`
* Adding `?replicaSet=rs0` to connection stirngs
* Adding `?directConnection=true` to connection string
* Updating the `mongo-rs-initialize.js` file to set host manually (can't use `mongo` as host there, as inside the shell it doesn't resolve to an address)

The only thing that worked was to use `network_most: "host"` on all services and env vars.
Will this cause an issue anywhere for other devs/setups?

MarkLark86: Without setting network to host, I was getting the following (this is was also happening here in github actions, so it's not my env).

```
pymongo.errors.ServerSelectionTimeoutError:
Could not reach any servers in [('127.0.0.1', 27017)].
Replica set is configured with internal hostnames or IPs?,
Timeout: 30s, Topology Description:
<TopologyDescription id: 6836597790ee434780417ddc, topology_type: ReplicaSetNoPrimary, servers: [<ServerDescription ('127.0.0.1', 27017) server_type: Unknown, rtt: None, error=AutoReconnect('127.0.0.1:27017: [Errno 111] Connection refused (configured timeouts: socketTimeoutMS: 20000.0ms, connectTimeoutMS: 20000.0ms)')>]>
```

PR #4822, opened 2025-05-28
Status field that reflects on creation/modification time of article is blinking with every new input value

dzonidoo:
SDESK-7648


tomaskikutis (commented):
To iterate object properties use `for (const key of Object.keys(obj))`. Using for in includes inherited properties which in most cases is not desired.

tomaskikutis (approved):

tomaskikutis:
check this on authoring-react too

dzonidoo:
i will

dzonidoo:
we have this issue on authoring-react as well.. ill fix it and will open new PR

PR #4824, opened 2025-05-30
Config to initialize spellchecker disabled

tomaskikutis:
STT-193

### DIFF
--- a/scripts/core/superdesk-api.d.ts
+++ b/scripts/core/superdesk-api.d.ts
@@ -3499,6 +3499,8 @@ declare module 'superdesk-api' {
             autorefreshContent?: boolean;
 
             elasticHighlight?: any;
+            searchShortcut?: boolean;
+            hideLiveSuggestions?: boolean;
             nestedItemsInOutputStage?: boolean;
             keepMetaTermsOpenedOnClick?: boolean;
             showCharacterLimit?: number;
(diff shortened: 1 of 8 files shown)


scripts/core/superdesk-api.d.ts
```diff
@@ -3674,10 +3679,20 @@ declare module 'superdesk-api' {
 
         userOnlineMinutes: number;
 
-        // e.g. {nl: 'leuven_dutch'}
-        spellcheckers?: {
-            [languageCode: string]: string;
-        };
+        spellchecking?: {
```
petrjasek, line 3682: we should document those api changes somewhere

petrjasek (approved):

scripts/core/superdesk-api.d.ts
tomaskikutis: Yep, I too was thinking how do we keep it all in sync between different versions. I tried to convert `superdesk.config.js` to `superdesk.config.ts` so config would also be compiled and would fail if something's off. I only tested the build and autocomplete which worked well. It remains to see if it fails if wrong types are provided.

Do you think that'd be sufficient or you'd like it to also add it in changelog or somewher else?

petrjasek: I think it might be still useful to keep some changelog for api, you would see it fail but you wouldn't know how to fix it from those errors probably

tomaskikutis: I investigated converting to superdesk.config.ts more and it's not working due to grunt not understanding ts. We may either try grunt forks that support ts or write a custom command and uses local ts to compile that single file.

PR #4848, opened 2025-06-16
improve spellchecker ui - make info message and suggestions scrollable

tomaskikutis:
STT-1204


package.json
```diff
@@ -123,7 +123,7 @@
     "sass-loader": "^10.5.2",
     "shallow-equal": "^3.1.0",
     "shortid": "2.2.8",
-    "superdesk-ui-framework": "4.0.54",
+    "superdesk-ui-framework": "^4.0.56",
```
petrjasek: would rather stick to exact version so when we update ui framework we have to run CI on it

petrjasek (approved):

package.json
tomaskikutis: I wrote a grep-lint rule for this, it fails CI if non-exact is being used ;D https://github.com/superdesk/superdesk-client-core/blob/develop/grep-lint.py#L57 I don't always remember it before that :D

PR #4862, opened 2025-06-23
ui framework 58

tomaskikutis:
SDESK-7654


thecalcc (approved):
@tomaskikutis fix CI, and I guess title should say ui framework 58

PR #4869, opened 2025-06-30
update version of ui-framework

dzonidoo:
link: https://github.com/superdesk/superdesk-planning/pull/2386

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12842,9 +12842,9 @@
       }
     },
     "superdesk-ui-framework": {
-      "version": "4.0.59",
-      "resolved": "https://registry.npmjs.org/superdesk-ui-framework/-/superdesk-ui-framework-4.0.59.tgz",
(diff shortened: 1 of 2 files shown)


tomaskikutis:
I've noticed this test is unstable and is failing from time to time. `can correct published item using corrections workflow `

From the screenshot the issue looks to be the helper that is typing into editor3 field, or maybe no helper is being used.

[image](https://github.com/user-attachments/assets/3dd54e1b-60c8-4243-8643-5355d230b92e)

dzonidoo:
The issue is probably that the clear( ) method is nott called before fill( ) in this test when changing the value of input.
Should we fix that in a separate PR?

dzonidoo:
but i just guessing right now maybe thats not the issue

tomaskikutis:
Let's try - open that PR.

dzonidoo:
i will - Lets merge this one

tomaskikutis (approved):

PR #4874, opened 2025-07-04
Fix embargo date picker

tomaskikutis:
STT-1235

Update from `@superdesk/common` to `@sourcefabric/common` was needed to make sure there's only one version of `date-fns` installed. `@date-fns/tz` requires `date-fns` >= 4


petrjasek (approved):
would that need changes in planning as well?

tomaskikutis:
It should not. For consistency it might be good to have it the same, but in case planning has a different version - it should use its own version and not impact the client core.

PR #4878, opened 2025-07-09
CREATE NEW button in bottom right corner does not react on clicking

dzonidoo:
SDESK-7704

### DIFF
--- a/scripts/apps/authoring/views/opened-articles.html
+++ b/scripts/apps/authoring/views/opened-articles.html
@@ -26,7 +26,7 @@
     </ul>
 
     <div class="opened-articles-bar__create-new" ng-if="privileges.archive">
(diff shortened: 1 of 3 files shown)


tomaskikutis (approved):

tomaskikutis:
It's not the best fix actually, because popover should switch from bottom to top if there's no space available. That's how it was working before I believe.

dzonidoo:
aham i did not know that.. so what now?

tomaskikutis:
I might have broken it with this commit https://github.com/superdesk/superdesk-ui-framework/commit/856a4d119cf32fb3155ecac2dbfa56845ac15176

Double check if that's the case and fix showPopup function to flip the popover in case there's not enough space. You might need to modify `restrictHeightToMaxAvailable`

PR #4882, opened 2025-07-11
add xmp metadata parsing for videos

taylorli2000:
SDCP-959

use exiftool to parse xmp tags

to be completed with [BE pr to superdesk-core](https://github.com/superdesk/superdesk-core/pull/3038) to write xmp to files


scripts/apps/archive/controllers/UploadController.ts
```diff
@@ -361,7 +364,10 @@ export function UploadController(
                 ? Promise.resolve()
                 : Promise.all(acceptedFiles.map(
                     ({file, getThumbnail}) =>
-                        getExifData(file)
+                        (file.type.startsWith('video') ?
+                            processXMP(file) :
+                            getExifData(file)
```
petrjasek: is there any reason to do it differently? is it much slower or something?

txl-cp: main reason is that video files are noticeably slower depending on file size based. I also didn't want to disrupt existing workflow too much in case there are users aware there's no xmp and maybe have workarounds.

petrjasek: I wonder if it would be also noticably slower for images, at the end it would be easier to use the same method for both types if it's supported

txl-cp: performance is dependent on the user's hardware, but I don't notice any slowness as long as the asset is not extremely large. I do agree that it's easier to just use the one fn, do you want to remove the ternary and just use processXMP?

petrjasek: @tomaskikutis what's your take on this?

tomaskikutis: Isn't file metadata stored separately from content? If it is, metadata extraction speed should be similar regardless of file size - given it's implemented correctly and is not reading the entire blob.

It looks to me that the library we're using `@metadata/exif` isn't efficient and is reading the entire blob though. For images slowness might not be noticed, but we shouldn't be using it for video. Even better if we can replace it with a more efficient library and use it for both images and videos.

It appears `@uswriting/exiftool` is efficient so I'd use that for both.

Also do install the library via `npm` instead of using a remote CDN. Let's not use the web worker too - an efficient library should be quick regardless of file size and web worker wouldn't be changing anything. Let me know if I'm wrong.

petrjasek: it's reading metadata embedded in the binary, not stored separately

tomaskikutis: I meant that metadata is in the binary header, so it's not needed to read the entire blob to retrieve it.

petrjasek: not sure about the worker, any reason for it @taylorli2000 ?

txl-cp: It depends where the metadata is stored. sometimes metadata, like XMP, can be stored at the end of the file which would force the lib to read the entire blob.

According to exiftool [docs](https://exiftool.org/exiftool_pod.html#Processing-control) on the -fast option, there is only a substantial benefit to speed when extracting through a network connection.

There is benefit to the web worker here because if a user uploads multiple large videos and the metadata is stored at the end of the file it will cause the user's browser to hang while it's processing. Another reason is that @uswriting/exiftool is esm only so to use the lib via modules the tsconfig module property requires an upgrade.

tomaskikutis: > sometimes metadata, like XMP, can be stored at the end of the file which would force the lib to read the entire blob.

Can you link to a source on this? I'd be inclined to think that even if it was at the end, the code would read from the end and still wouldn't need to load the entire blob into memory.

> According to exiftool [docs](https://exiftool.org/exiftool_pod.html#Processing-control) on the -fast option, there is only a substantial benefit to speed when extracting through a network connection.

`These speed benefits are small when reading images directly from disk, but can be substantial if piping images through a network connection.`

It says the opposite, but I don't see how it's relevant. In either case we won't be reading metadata via network.

> There is benefit to the web worker here because if a user uploads multiple large videos and the metadata is stored at the end of the file it will cause the user's browser to hang while it's processing.

If the library is efficient, it should not load full files in memory and will only check header(or maybe footer). It'd be best if you could test it locally without a worker on a video that's a few GB in size.

I'm not against using a worker in principle, but would like to avoid having to support webpack config for it.

> Another reason is that @uswriting/exiftool is esm only so to use the lib via modules the tsconfig module property requires an upgrade.

That may well be true. The way we handle this at the moment is to fork the library and publish a non-esm version to npm. See https://github.com/superdesk/date-fns-tz/commits/superdesk/

txl-cp: [Here](https://exiftool.org/forum/index.php?topic=6080.msg29931#msg29931) is the owner of exiftool mentioning that it needs to read the whole file for videos if metadata is located at the end. I don't have a source specific to images, but it would make sense if this applies to all valid file formats that exiftool can operate on, which is that it tries to read until it finds the end of the metadata blocks. Presumably, the end of the metadata blocks is near the beginning or end of the file depends on how it was written and the file format.

> It says the opposite, but I don't see how it's relevant. In either case we won't be reading metadata via network.
>
I was just trying to emphasize the fact that it would be a long running process on the main thread if needing to read multiple large video files since it would not be read via network.

>If the library is efficient, it should not load full files in memory and will only check header(or maybe footer). It'd be best if you could test it locally without a worker on a video that's a few GB in size.
>
I can confirm that the time to process a given video file scales with file size, amount of metadata, and hardware by adding some benchmarking to the [demo page of @uswriting/exiftool](https://github.com/6over3/exiftool/blob/main/demo/index.html) on my local. Large videos/files by itself do not require a long processing time, but large files processing time seems to scale with amount of metadata within the file.

>That may well be true. The way we handle this at the moment is to fork the library and publish a non-esm version to npm. See https://github.com/superdesk/date-fns-tz/commits/superdesk/
>
Got it. I can fork or make PR to original lib to output cjs as well, is there preferred option?

andrewmd5: Just to chime in: you don’t need to run our exiftool in a worker. It won’t block the main thread if you’re working with a File object directly since all I/O will be asynchronous. If you pass in a buffer that I/O is synchronous since the entire file is in memory and Chrome/Edge won’t let you allocate more than 2GB so the former is preferred.

tomaskikutis: Thank you for taking your time to comment here @andrewmd5.

@txl-cp I'd double check if Andrew is correct and it really doesn't block the UI thread - and if it doesn't - let's not use the worker.

> Large videos/files by itself do not require a long processing time, but large files processing time seems to scale with amount of metadata within the file.

That's what I would expect with an efficient implementation - that processing time would be proportional to how much metadata there is - not how big the file is. And in most cases metadata would be probably below 5mb.

> Got it. I can fork or make PR to original lib to output cjs as well, is there preferred option?

Making a PR to the original lib would be ideal, but that depends if it gets accepted. If it doesn't - a fork is fine too.

txl-cp: Looks like it's not blocking. I'll remove the worker and use exiftool in place of getExifData

package.json
```diff
@@ -43,6 +42,7 @@
     "@types/react-dom": "16.8.0",
     "@types/react-redux": "7.1.9",
     "@typescript-eslint/parser": "5.57.0",
+    "@uswriting/exiftool": "file:uswriting-exiftool-1.0.5.tgz",
```
txl-cp: noting that this is local package for now while waiting for [PR](https://github.com/6over3/exiftool/pull/16) to add cjs to go through

scripts/apps/archive/controllers/UploadController.ts
```diff
@@ -361,12 +347,16 @@ export function UploadController(
...
+                            transform: (d) => JSON.parse(d),
+                        })
+                            .then(processMetadata, (): Partial<IPTCMetadata> => ({}))
                             .then(
                                 (fileMeta) => mapIPTCExtensions(fileMeta, $scope.currentUser, $scope.parent),
-                                () => ({}), // proceed with upload on exif parsing error
+                                (): Partial<IArticle> => ({}), // proceed with upload on exif parsing error
                             )
-                            .then((meta) => {
+                            .then(async (meta) => {
```
tomaskikutis: async tag shouldn't be used here since nothing is awaited

txl-cp: I put async tag because typescript is giving warning there: `This may be converted to an async function. ts(80006)`.
If it doesn't matter I can remove it or can restructure it a bit
```
.then((meta) => initFile(file, meta, getPseudoId()))
    .then((item) => getThumbnail(file)
        then((htmlString) => item.thumbnailHtml = htmlString),
)
```
```
.then((meta) =>
    getThumbnail(file).then((htmlString) => {
        const item = initFile(file, meta, getPseudoId());

        return item.thumbnailHtml = htmlString;
    }),
),
```

tomaskikutis: Do remove it. Running `npm run lint` will run a typescript check and I just verified that it's passing without `async` tag. If you got a warning - it must have been that your editor doesn't pick up project setup. Maybe it uses a different typescript version for example.

txl-cp: done

scripts/apps/archive/controllers/UploadController.ts
txl-cp: @tomaskikutis 

> Making a PR to the original lib would be ideal, but that depends if it gets accepted. If it doesn't - a fork is fine too.

Might need to fork this if PR does not get approved. Could you create fork of this on superdesk? https://github.com/taylorli2000/exiftool/tree/main

taylorli2000: Nevermind, looks like owner was on vacation and they're back and able to review now

txl-cp: Should be okay now. Owner was on vacation and they're back now

package.json
petrjasek: seems like it's merged now

scripts/apps/archive/constants.ts
```diff
@@ -114,3 +117,44 @@ export const CANCELED_STATES = KILLED_STATES.concat([ITEM_STATE.SPIKED]);
  * KILLED | RECALLED | UNPUBLISHED | SPIKED | SCHEDULED
  */
 export const READONLY_STATES = CANCELED_STATES.concat([ITEM_STATE.SCHEDULED]);
+
+export const IPTC_XMP_TAGS = {
```
tomaskikutis, line 121: Could you adjust this so group names aren't concatenated and don't need to be stripped later. It's not serving any purpose as far as I can see.

use one variable that defines `ipct_to_xmp` and `xmp_to_iptc = invert(ipct_to_xmp)`

use a helper from lodash https://lodash.com/docs/4.17.15#invert

taylorli2000: I didn't use lodash for this because the type was too generic and would need to cast it. The group names are needed because some tags exist in both groups so exiftool would only return the XMP tag, if both exist, without group names. Composite tags needs group name option for output and then would need to strip the group after to get highest priority

webpack.config.js
```diff
@@ -95,6 +95,7 @@ module.exports = function makeConfig(grunt) {
                 'jquery-gridster': 'gridster/dist/jquery.gridster.min',
                 'external-apps': path.join(process.cwd(), 'dist', 'app-importer.generated.js'),
                 'shallow-equal': 'shallow-equal/dist/index',
+                '@uswriting/exiftool/cjs': require.resolve('@uswriting/exiftool/cjs'),
```
petrjasek, line 185: I wonder what does it do?

taylorli2000: webpack keeps resolving the esm build of exiftool so I put an alias here to resolve to cjs correctly. if importing /cjs without this alias, webpack still tries to resolve the esm build.

scripts/apps/archive/constants.ts
tomaskikutis: Sorry for a late response I was on vacation. Could you add unit test(s) to cover tags existing in both groups?

scripts/core/superdesk-api.d.ts
```diff
@@ -4009,7 +4009,7 @@ declare module 'superdesk-api' {
         SubjectReference: string;
         Category: string;
         SupplementalCategories: string;
-        Keywords: string;
+        Keywords: string | string[];
```
tomaskikutis: use `Array<string>` convention

scripts/apps/archive/tests/archive.spec.ts
```diff
@@ -190,4 +192,224 @@ describe('content', () => {
             }, data.item);
         }));
     });
+
+    fdescribe('process item metadata', () => {
+        const exiftoolReturnData = [{
+            "XMP:XMPToolkit": "Image::ExifTool 9.85",
```
tomaskikutis: Could you improve the tests to make them not rely on implementation details? I assume we could have a single function that takes a file and returns extracted metadata.

It'd be good also to keep test data brief and only include samples that are unique in some aspect - for example uses different value - string/number/list.

scripts/apps/archive/tests/archive.spec.ts
```diff
@@ -190,4 +192,224 @@ describe('content', () => {
             }, data.item);
         }));
     });
+
+    fdescribe('process item metadata', () => {
```
tomaskikutis: it shouldn't be focused

scripts/apps/archive/tests/archive.spec.ts
taylorli2000: yes, I redid the tests to use the getMetadata fn and spy/assert on other fns/values depending on video or image test.

scripts/apps/archive/tests/archive.spec.ts
```diff
@@ -190,4 +192,91 @@ describe('content', () => {
             }, data.item);
         }));
     });
+
+    describe('process item metadata', () => {
+        const rawMetadata = [{
+            "XMP:Creator": ["Creator1 (ref2014)","Creator2 (ref2014)"],
```
tomaskikutis: Does metadata have these XMP/IPTC prefixes straight from the file?

tomaskikutis (commented):
Could you upload a sample video file with metadata to github?

scripts/apps/archive/tests/archive.spec.ts
taylorli2000: by default it won't, but I used args to get prefixes.

https://github.com/user-attachments/assets/c15368b9-c753-4a4f-bb6f-a20162b1baf3


https://github.com/user-attachments/assets/12860ff5-6ea0-47e8-ad03-a83339674b74

tomaskikutis:
1. When I asked to make tests implementation independent - I meant library independent too. Now you take library output and test if it gets converted correctly into IPTCMetadata. If we changed the library, those tests would be useless. Update the tests so they do not make assumptions about the library used. Might be best to commit a test file with metadata to the git repo and use it as an input.

2. Simplify the implementation - remove those complex typescript types and use a string dictionary instead. Move `getMetadata` to a separate file and create a separare file for tests too.

taylorli2000:
> 1. When I asked to make tests implementation independent - I meant library independent too. Now you take library output and test if it gets converted correctly into IPTCMetadata. If we changed the library, those tests would be useless. Update the tests so they do not make assumptions about the library used. Might be best to commit a test file with metadata to the git repo and use it as an input.
> 2. Simplify the implementation - remove those complex typescript types and use a string dictionary instead. Move `getMetadata` to a separate file and create a separare file for tests too.

I think I'll need to move the tests to e2e then. exiftool lib parseMetadata doesn't seem to work in karma.

tomaskikutis:
e2e is fine too. Use playwright - add a test file inside `e2e/client/playwright`

taylorli2000:
I figured out how to add it as unit so I did that instead. I simplified the types and use fixtures for the assertions now.

petrjasek:
hi @taylorli2000 can you check those failing tests pls?

tomaskikutis (approved):

PR #4884, opened 2025-07-14
[STT-1241] - As an STT admin I like to add a start and end date to the subscriber config via the UI

BrianMwangi21:
[STT-1241](https://sofab.atlassian.net/browse/STT-1241)

NOTES: Backend part of ticket is in this [PR](https://github.com/superdesk/superdesk-core/pull/2994)

[STT-1241]: https://sofab.atlassian.net/browse/STT-1241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/apps/publish/SubscriberScheduleDatePicker.tsx
+++ b/scripts/apps/publish/SubscriberScheduleDatePicker.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import {DateTimePicker} from 'superdesk-ui-framework/react';
+import {isValid} from 'date-fns';
+import {appConfig} from 'appConfig';
+
+interface IProps {
+  model: Date | null;
+  onChange: (date: Date | null) => void;
+  label?: string;
+  required?: boolean;
+}
+
+export class SubscriberScheduleDatePicker extends React.PureComponent<IProps> {
+    render() {
+        const {model, onChange, label, required = false} = this.props;
+
+        return (
+            <DateTimePicker
+                label={label}
+                value={model}
+                onChange={(val) => {
+                    if (!isValid(val) && val !== null) return;
+                    onChange(val);
+                }}
+                valueType="date"
+                dateFormat={appConfig.view.dateformat}
+                required={required}
+            />
+        );
+    }
+}
--- a/scripts/apps/publish/directives/SubscribersDirective.ts
+++ b/scripts/apps/publish/directives/SubscribersDirective.ts
@@ -9,6 +9,7 @@ import {
     startOfDay,
     parseISO,
 } from 'date-fns';
+import {formatDate} from 'core/get-superdesk-api-implementation';
 
 /**
  * @ngdoc directive
@@ -356,14 +357,14 @@ export function SubscribersDirective(
                     if (mStart != null && isAfter(mStart, now)) {
                         $scope.scheduleErrors.push(gettext(
                             'Schedule starts on {{date}}. Manual activation will override it.',
-                            {date: format(mStart, 'PPP')},
+                            {date: formatDate(mStart)},
                         ));
                     }
 
                     if (mEnd != null && isBefore(mEnd, now)) {
                         $scope.scheduleErrors.push(gettext(
                             'Schedule ended on {{date}}. Manual activation will override it.',
-                            {date: format(mEnd, 'PPP')},
+                            {date: formatDate(mEnd)},
                         ));
                     }
                 } else if (
@@ -374,7 +375,7 @@ export function SubscribersDirective(
                     $scope.scheduleErrors.push(gettext(
                         'Subscriber is inactive but within {{start}} to {{end}} schedule. ' +
                         'It will be activated when you save.',
-                        {start: format(mStart, 'PPP'), end: format(mEnd, 'PPP')},
+                        {start: formatDate(mStart), end: formatDate(mEnd)},
                     ));
                 }
             };
--- a/scripts/apps/publish/directives/SubscribersDirective.ts
+++ b/scripts/apps/publish/directives/SubscribersDirective.ts
@@ -6,6 +6,8 @@ import {
     isAfter,
     isEqual,
     format,
+    startOfDay,
+    parseISO,
 } from 'date-fns';
 
 /**
@@ -187,6 +189,13 @@ export function SubscribersDirective(
                     diff[key] = value;
                 });
 
+                if ($scope.subscriber.schedule?.startDate != null && $scope.subscriber.schedule?.endDate != null) {
+                    diff['schedule'] = {
+                        startDate: format($scope.subscriber.schedule.startDate, 'yyyy-MM-dd'),
+                        endDate: format($scope.subscriber.schedule.endDate, 'yyyy-MM-dd'),
+                    };
+                }
+
                 api.subscribers.save($scope.origSubscriber, diff)
                     .then(
                         () => {
@@ -247,8 +256,8 @@ export function SubscribersDirective(
                     } else {
                         const {startDate, endDate} = $scope.subscriber.schedule;
 
-                        $scope.subscriber.schedule.startDate = startDate ? new Date(startDate) : null;
-                        $scope.subscriber.schedule.endDate = endDate ? new Date(endDate) : null;
+                        $scope.subscriber.schedule.startDate = startDate ? startOfDay(parseISO(startDate)) : null;
+                        $scope.subscriber.schedule.endDate = endDate ? startOfDay(parseISO(endDate)) : null;
                     }
 
                     $scope.destinations = [];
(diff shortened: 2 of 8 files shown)


scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -44,6 +45,10 @@ export function SubscribersDirective(
             $scope.directProducts = null;
             $scope.search = {};
             $scope.highPriorityQueueEnabled = appConfig.high_priority_queue_enabled;
+            $scope.scheduleError = {
```
tomaskikutis: What's the purpose to have a dedicated field per error type? Why not simply have a list of errors?

scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import {gettext} from 'core/utils';
 import {appConfig} from 'appConfig';
+import moment from 'moment';
```
tomaskikutis: We're phasing out usage of `moment`. Use standard `Date` and `date-fns` library if you need helpers.

scripts/apps/publish/SubscriberScheduleDatePicker.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import {DateTimePicker} from 'superdesk-ui-framework/react';
+import {isValid} from 'date-fns';
+import {appConfig} from 'appConfig';
+
+interface IProps {
+  model: Date | null;
```
tomaskikutis: rename to `value` to keep it consistent with standard naming

scripts/apps/publish/SubscriberScheduleDatePicker.tsx
```diff
@@ -0,0 +1,31 @@
...
+    render() {
+        const {model, onChange, label, required = false} = this.props;
+
+        return (
+            <DateTimePicker
+                label={label}
+                value={model}
+                onChange={(val) => {
+                    if (!isValid(val) && val !== null) return;
+                    onChange(val);
```
tomaskikutis: @thecalcc I'm looking forward to see this moved inside the datetime picker component.

scripts/apps/publish/SubscriberScheduleDatePicker.tsx
```diff
@@ -0,0 +1,31 @@
...
+import {appConfig} from 'appConfig';
+
+interface IProps {
+  model: Date | null;
+  onChange: (date: Date | null) => void;
+  label?: string;
+  required?: boolean;
+}
+
+export class SubscriberScheduleDatePicker extends React.PureComponent<IProps> {
```
tomaskikutis: It doesn't contain any code specific to subscriber schedule. Rename to `DateTimePicker`

scripts/apps/publish/views/subscribers.html
```diff
@@ -176,6 +177,38 @@ <h3 class="modal__heading" translate ng-hide="subscriber._id">Add New Subscriber
...
                         </div>
 
+                        <div class="form__row">
+                            <label class="form-label" translate>Subscriber Schedule Settings</label>
+
+                            <div class="form__row-item sd-margin-t--2">
+                                <sd-subscriber-schedule-date-picker
+                                    model="subscriber.schedule.startDate"
+                                    on-change="updateStartDate"
+                                    label="'Start Date'"
```
tomaskikutis: translate the label using the pipe operator

scripts/apps/publish/index.ts
```diff
@@ -35,6 +37,13 @@ export default angular.module('superdesk.apps.publish', ['superdesk.apps.users',
     .controller('publishQueueCtrl', ctrl.PublishQueueController)
     .controller('SubscriberTokenCtrl', ctrl.SubscriberTokenController)
 
+    .component('sdSubscriberScheduleDatePicker',
```
tomaskikutis: move this to where generic components are registered

scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -284,6 +302,103 @@ export function SubscribersDirective(
...
+                            `Subscriber is currently active but the end date ${mEnd.format('LLL')} has passed. 
+                            It will be deactivated when you save.`,
+                        );
+                    }
+                } else if (
+                    mStart && mEnd &&
+                    now.isSameOrAfter(mStart) &&
+                    now.isSameOrBefore(mEnd)
+                ) {
+                    $scope.scheduleError.activeBeforeStartErrorMessage = gettext(
```
tomaskikutis: placeholder usage is incorrect for `gettext` - check existing usages for a good example

scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -284,6 +302,103 @@ export function SubscribersDirective(
...
+
+                if ($scope.subscriber.is_active) {
+                    if (mStart && now.isBefore(mStart)) {
+                        $scope.scheduleError.activeBeforeStartErrorMessage = gettext(
+                            `Subscriber is currently active but will be deactivated until
+                            ${mStart.format('LLL')} when you save.`,
+                        );
+                    }
+
+                    if (mEnd && now.isAfter(mEnd)) {
```
tomaskikutis: Only use booleans inside `if` conditions. In this case you have a date - it should probably be
`if (mEnd != null)`

scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -284,6 +302,103 @@ export function SubscribersDirective(
...
+                        );
+                    }
+                } else if (
+                    mStart && mEnd &&
+                    now.isSameOrAfter(mStart) &&
+                    now.isSameOrBefore(mEnd)
+                ) {
+                    $scope.scheduleError.activeBeforeStartErrorMessage = gettext(
+                        `Subscriber is inactive but the current time is within the active schedule 
+                        (${mStart.format('LLL')} to ${mEnd.format('LLL')}). It will be activated when you save.`,
```
tomaskikutis: do not use newlines in gettext - concatenate multiple strings using a `+` operator instead

scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -284,6 +302,103 @@ export function SubscribersDirective(
...
+                    );
+                }
+            };
+
+            /**
+            * Adjusts the `is_active` flag before saving depending on the values of the schedule settings
+            */
+            $scope.verifyScheduleSettings = function() {
```
tomaskikutis: Update naming - current verb "verify" implies read-only access which is not correct. Use `updateStatus` or something else that makes clear values might be changed.

It would even better to do this on the back-end, so it wouldn't be possible to patch invalid state from the client.

tomaskikutis (commented):
Placeholders in `gettext` are still used incorrectly. Native js interpolation is not allowed. Here's an example:

```
gettext('Accepted by {{user}}', {user: suggestionResolver.display_name})
```

tomaskikutis:
One more thing I forgot to mention - there's a [config](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L3611) in superdesk that controls the format of dates and times. You need to be using that instead of passing hardcoded formating options to date-fns functions.

scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -379,6 +379,37 @@ export function SubscribersDirective(
...
                 }
             };
+
+            /**
+             * Function to generate and set appropriate schedule/label messages
+            */
+            $scope.getScheduleOrStatusLabel = function(subscriber) {
+                if (!subscriber) return '';
+
+                const hasSchedule = subscriber.schedule?.start_date && subscriber.schedule?.end_date;
```
tomaskikutis: Do not implicitly cast to boolean. Use:

`const hasSchedule = subscriber.schedule?.start_date != null && subscriber.schedule?.end_date != null;`

scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -379,6 +379,37 @@ export function SubscribersDirective(
...
+
+                if (hasSchedule) {
+                    const {start_date, end_date} = subscriber.schedule;
+                    const now = startOfDay(new Date());
+                    const mStart = startOfDay(new Date(start_date));
```
tomaskikutis: Do not prefix variables with letters. I assume `mStart` was intended for moment dates, but it's not moment anymore. Even if it was moment - typescript type should be added instead of adding that information to the variable name.

scripts/apps/publish/directives/SubscribersDirective.ts
```diff
@@ -350,32 +350,32 @@ export function SubscribersDirective(
...
                 const {start_date, end_date} = $scope.subscriber.schedule;
                 const now = startOfDay(new Date());
-                const mStart = start_date ? startOfDay(new Date(start_date)) : null;
-                const mEnd = end_date ? startOfDay(new Date(end_date)) : null;
+                const start = start_date ? startOfDay(new Date(start_date)) : null;
+                const end = end_date ? startOfDay(new Date(end_date)) : null;
 
                 if ($scope.subscriber.is_active) {
-                    if (mStart != null && isAfter(mStart, now)) {
+                    if (start != null && isAfter(start, now)) {
```
tomaskikutis, line 357: with native dates you can use standard comparison operators, like

`if (start != null && start > now) {`

I'm not requesting you change it, just letting you know for the future.

tomaskikutis (approved):

scripts/apps/publish/directives/SubscribersDirective.ts
BrianMwangi21: This is noted Tomas. Thanks

PR #4885, opened 2025-07-14
Add support for custom editor tags

thecalcc:
STT-148

Note: As agreed, in a future PR I'll change the color mapping according to use color variables from UI framework, after Fritz comes back.

### DIFF
--- a/scripts/apps/workspace/content/components/get-content-profiles-form-config.tsx
+++ b/scripts/apps/workspace/content/components/get-content-profiles-form-config.tsx
@@ -30,9 +30,12 @@ export const getEditor3PlainTextFormattingOptions = (): Dictionary<PLAINTEXT_FOR
     'lowercase': gettext('lowercase'),
 });
 
+export const getCustomEditorTagId = (id: string) => `EDITOR_TAG_${id}`;
+
 export const getEditor3RichTextFormattingOptions = (): {[MEMBER in RICH_FORMATTING_OPTION]: string} => {
     return {
-        ...Object.fromEntries((appConfig.authoring?.customEditorTags ?? []).map(({id, label}) => [id, label])),
+        ...Object.fromEntries((appConfig.authoring?.customEditorTags ?? [])
+            .map(({id, label}) => [getCustomEditorTagId(id), label])),
         'h1': gettext('h1'),
         'h2': gettext('h2'),
         'h3': gettext('h3'),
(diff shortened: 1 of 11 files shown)


scripts/apps/workspace/content/components/get-content-profiles-form-config.tsx
```diff
@@ -32,6 +32,7 @@ export const getEditor3PlainTextFormattingOptions = (): Dictionary<PLAINTEXT_FOR
 
 export const getEditor3RichTextFormattingOptions = (): {[MEMBER in RICH_FORMATTING_OPTION]: string} => {
     return {
+        ...Object.fromEntries((appConfig.authoring?.customEditorTags ?? []).map(({id, label}) => [id, label])),
```
tomaskikutis: prefix the ID with `EDITOR_TAG_{{tag}}` or similar so configuring a tag with ID `h1` wouldn't break the native `h1` formatting option.

thecalcc: I don't like this solution because I'll need to map IDs to have this prefix in a couple places and that introduces overhead. How about either checking if there's duplicates in `startApp` and if there's duplicates I throw some kind of an error?

tomaskikutis: > I'll need to map IDs to have this prefix in a couple places

How about you use a helper so it's in one place? Checking before app starts could mitigate the issue, but I'd go with prefixing since it encapsules the feaure better and avoids runtime errors.

tomaskikutis (commented):
The helper you made does the job, but isn't great. It's not very clear when it needs to be used, name isn't great - and developers wouldn't know about the helper without checking usages of `id`.

A better abstraction would be to never access `customEditorTags` directly from appConfig and instead define a wrapper that you would use and that wrapper would add a prefixed value. Then after looking at the interface it'd be more clear to developers that ID is used normally and `editor3Style` is only for editor3

```
type IEnhancedTag = ISuperdeskGlobalConfig['authoring']['customEditorTags'][0] & {editor3Style: string}

export const customEditorTags: Array<IEnhancedTag> = (appConfig.authoring?.customEditorTags ?? []).map((item) => {
    return {
        ...item,
        editor3Style: `EDITOR_TAG_${item.id}`,
    };
});
```

tomaskikutis (approved):

PR #4887, opened 2025-07-15
fix embed loading in editor

petrjasek:
we need to execute the external script each time we render the embed,
but the libraries usually create some global object and when it's present
they won't init again, so it's removing all new window keys created on load.

SDAAP-128

### DIFF
--- a/scripts/core/editor3/components/embeds/EmbedBlock.tsx
+++ b/scripts/core/editor3/components/embeds/EmbedBlock.tsx
@@ -41,12 +41,16 @@ export class EmbedBlockComponent extends React.Component<IProps> {
     static propTypes: any;
     static defaultProps: any;
 
+    globalKeys: Array<string>;
+
     constructor(props) {
         super(props);
 
         this.onClickDelete = this.onClickDelete.bind(this);
         this.editEmbedHtml = this.editEmbedHtml.bind(this);
         this.onChangeDescription = this.onChangeDescription.bind(this);
+
+        this.globalKeys = [];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


scripts/core/editor3/components/embeds/EmbedBlock.tsx
```diff
@@ -59,14 +63,22 @@ export class EmbedBlockComponent extends React.Component<IProps> {
...
-                const async = s.hasAttribute('async');
+                const newScript = document.createElement('script');
+                const oldKeys = Object.keys(window);
 
-                if (url.startsWith('http')) {
-                    url = url.substring(url.indexOf(':') + 1);
-                }
+                newScript.setAttribute('src', s.getAttribute('src'));
+                newScript.setAttribute('async', s.hasAttribute('async') ? 'async' : '');
+                document.body.appendChild(newScript);
```
tomaskikutis: Would be good remove `newScript` in case component unmounts before script loads. It could be done using `cleanupFunctionsToRunBeforeUnmounting` that are used elsewhere.

scripts/core/editor3/components/embeds/EmbedBlock.tsx
```diff
@@ -59,14 +63,22 @@ export class EmbedBlockComponent extends React.Component<IProps> {
...
+                newScript.onload = () => {
+                    const newKeys = Object.keys(window);
+
+                    // keep track of new global objects created after the script was loaded
+                    this.globalKeys = newKeys.filter((k) => !oldKeys.includes(k) && isNaN(parseInt(k, 10)));
```
tomaskikutis: I'm not following what's being done with `globalKeys`, but in any case it's not reliable because it might pick up keys that other parts of the app are setting and would interfere. We don't use globals much so impact isn't significant, but still would be good to avoid. What's the purpose of it?

scripts/core/editor3/components/embeds/EmbedBlock.tsx
```diff
@@ -41,12 +41,16 @@ export class EmbedBlockComponent extends React.Component<IProps> {
     static propTypes: any;
     static defaultProps: any;
 
+    globalKeys: Array<string>;
```
tomaskikutis: mark as private

scripts/core/editor3/components/embeds/EmbedBlock.tsx
petrjasek: when you load eg. twitter embed script, it will first time create some `window.twitter` object and handle all objects with some `data-twitter` param. but if you load the script for the second time it will detect that `window.twitter` already exists and won't do anything. so removing such global object will make it initialize on next script load again.

scripts/core/editor3/components/embeds/EmbedBlock.tsx
petrjasek: I think it would still run that onload anyhow, wouldn't it?

tomaskikutis: If you clean up correctly, you would remove the script from DOM too so it should stop loading and I'd expect `newScript.onload` not to be called anymore.

scripts/core/editor3/components/embeds/EmbedBlock.tsx
tomaskikutis: Maybe it's built for loading in a iframe so it acts like that to optimize stuff. Would it make sense to load embeds inside an iframe?

petrjasek: looks like iframe might do it, will update the PR

tomaskikutis (approved):

PR #4889, opened 2025-07-15
Fix field rendering

thecalcc:
STT-1249

### DIFF
--- a/scripts/core/editor3/components/Editor3InitializeSpellchecker.tsx
+++ b/scripts/core/editor3/components/Editor3InitializeSpellchecker.tsx
@@ -45,6 +45,11 @@ export class Editor3InitializeSpellchecker extends React.PureComponent<IProps, I
         const spellcheck = ng.get('spellcheck');
         const language = this.props.spellchecking.language;
 
+        if (language == null) {
+            this.setState({loading: false});
+            return;
+        }
+
         this.getSpellcheckerStatus(language).then((enabled) => {
             spellcheck.isAutoSpellchecker = enabled;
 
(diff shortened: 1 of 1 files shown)


scripts/core/editor3/components/Editor3InitializeSpellchecker.tsx
```diff
@@ -58,7 +58,8 @@ export class Editor3InitializeSpellchecker extends React.PureComponent<IProps, I
                         this.props.dispatch(setExternalOptions({
                             spellchecking: getInitialSpellcheckerData(spellcheck, language),
                         }));
-
+                    }).finally(() => {
```
tomaskikutis: Where does it crash? If language is missing wouldn't be better to simply not call `this.getSpellcheckerStatus` and return?

thecalcc: there's a call to get some abbreviations inside `initializeSpellchecker`, if there's no language provided it returns a promise reject which in turn makes code in `then` never execute

tomaskikutis:
```
if (language == null) {
    this.setState({loading: false});
    return;
}

```
How about this at the top of the file? Seems more reasonable to me.

thecalcc: sounds good, I thought you kept it intentionally how it was since you were doing some changes recently I think

tomaskikutis (approved):

PR #4892, opened 2025-07-16
fix publish schedule time when using different timezone

petrjasek:
SDESK-7688

avoid converting time from local to selected timezone, that's handled on the server, the time should be sent as it is.

### DIFF
--- a/scripts/core/utils.tsx
+++ b/scripts/core/utils.tsx
@@ -398,8 +398,8 @@ export function getUTCOffset(timezoneId: string) {
     return trimStartExact(offsetStr, 'GMT');
 }
 
-export function toIsoStringWithoutTimezoneOffset(date: TZDate) {
-    return date.toISOString().slice(0, 16);
+export function toIsoStringWithoutTimezoneOffset(date: Date) {
+    return date.toISOString().slice(0, 19);
 }
 
 export function correctTimezone(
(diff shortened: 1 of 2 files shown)


tomaskikutis:
It's not caused by your PR, but there's an issue when setting initial schedule or embargo. If browser timezone is Toronto and I set 10 am Prague time - it patches 14h 

https://github.com/user-attachments/assets/ee0e5500-12a2-451a-83a1-039d0c6d4c7f

scripts/core/utils.tsx
```diff
@@ -398,8 +398,8 @@ export function getUTCOffset(timezoneId: string) {
     return trimStartExact(offsetStr, 'GMT');
 }
 
-export function toIsoStringWithoutTimezoneOffset(date: TZDate) {
-    return date.toISOString().slice(0, 16);
+export function toIsoStringWithoutTimezoneOffset(date: Date) {
```
tomaskikutis: I used `TZDate` type on purpose. The output format of `date.toISOString` is a bit different from a native Date object. Until char seconds they are the same so your change is fine.

'2025-07-17T07:42:01'
'2025-07-17T03:41:46.142-04:00'

petrjasek:
will check that

tomaskikutis:
There's now an issue with updating. If I'm in Toronto and set 10:19AM Prague time, all is good, but if I updated to 10:20AM Prague time - it patches incorrectly then.


https://github.com/user-attachments/assets/b2173f93-90e4-4be4-a11d-a3cf4f516a68

petrjasek:
@tomaskikutis should be working now

tomaskikutis (approved):
I found no issues now

PR #4897, opened 2025-07-23
Enable synchronous access of app config

tomaskikutis:


webpack.config.js
```diff
@@ -19,6 +19,90 @@ function countOccurrences(_string, substring) {
     return _string.split(substring).length - 1;
 }
 
+function applyDefaults(appConfig) {
```
petrjasek, line 22: not sure we should have it here, would it work in eg. planning unit tests setup?

petrjasek (approved):

webpack.config.js
tomaskikutis: There are 2 possible places to put it - here or in the new bundle which also wouldn't be loaded in planning unit tests. This one looks good, not sure if it'd work in planning unit tests how it is at the moment. I'll check it out and export a function from here and import in planning's webpack config to make tests work.

petrjasek: I tested it and seems ok https://github.com/petrjasek/superdesk-planning/tree/test-app-config

tomaskikutis: ok, even better if it works as is. I'll merge it then.

PR #4899, opened 2025-07-24
Changes to image metadata are not dynamically reflected [SDESK-7656]

devketanpro:


tomaskikutis:
Could you describe what was the cause of the issue and how your fixes solves it?

devketanpro:
> Could you describe what was the cause of the issue and how your fixes solves it?

Issue:
After editing and saving media (e.g., crop, credit, copyright), the updated metadata was not immediately reflected in the article editor. The editor continued to use outdated data from the DraftJS entity. Changes only became visible after a manual page reload.

Root Cause:
The MediaBlockComponent was rendering media based on the static data embedded in the editor’s contentState. It didn’t react to updates in the actual media object stored in Redux.

Fix:
The component was updated to retrieve the latest media object from state.forms.image.mediaByGuid using the media's guid. If updated data is present, it’s used for rendering. Otherwise, it gracefully falls back to the original entity data.

scripts/core/editor3/components/media/MediaBlock.tsx
```diff
@@ -342,6 +348,7 @@ export class MediaBlockComponent extends React.Component<IProps> {
 const mapStateToProps = (state) => ({
     readOnly: state.readOnly,
     showTitle: state.showTitle,
+    mediaEntities: state.forms?.image?.mediaByGuid || {},
```
tomaskikutis: Type of `state` should be `IEditorStore`. And in `IEditorStore` there's no `forms` field defined. I tried running it and haven't seen it there either so I'm not following completelly how is your fix working.

In any case, I think your fix should be deeper in the callstack. Instead of trying to take media from somewhere else you should see why content block has old data and fix that.

tomaskikutis (commented):
Looks much better.

Could you elaborate on what was the issue with `updateImage` function? Why are you re-creating the content state?

devketanpro:
> Looks much better.
> 
> Could you elaborate on what was the issue with `updateImage` function? Why are you re-creating the content state?

The original `updateImage` function didn’t trigger a UI update because `replaceEntityData` reused the same media object, and Draft.js skipped re-rendering due to shallow comparison. By recreating the `ContentState` and using a new media object, we ensure the editor picks up the change and re-renders the image block correctly. This approach maintains consistency and addresses the rendering issue effectively.

tomaskikutis:
I've looked into it - media object is fine. It was the `contentState` that still had the old reference after calling `replaceEntityData`

tomaskikutis (approved):

PR #4901, opened 2025-07-25
Apply soft limit only if it's defined

thecalcc:
SDESK-7727

Undefined: In the UI it's currently possible to apply a soft limit longer than the hard limit, but then soft limit doesn't render in the editor field. Should that be possible in the first place?

### DIFF
--- a/scripts/core/editor3/components/text-length-overflow-decorator.tsx
+++ b/scripts/core/editor3/components/text-length-overflow-decorator.tsx
@@ -97,16 +97,21 @@ export function getTextLimitHighlightDecorator(hardLimit: number, softLimit?: nu
         }
     }
 
-    const decorator = new CompositeDecorator([
+    const decorators = [
         {
             strategy: strategyForHardLimit,
             component: (props) => <Component {...props} color="#ff0000" />,
         },
-        {
+    ];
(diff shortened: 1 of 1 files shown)


tomaskikutis (approved):
The definition of soft limit is the one that doesn't affect validation/publishing. It's only a visual indicator. Hard limit will likely be greater most times, but if we can avoid limiting it, I'd do so.

thecalcc:
> The definition of soft limit is the one that doesn't affect validation/publishing. It's only a visual indicator. Hard limit will likely be greater most times, but if we can avoid limiting it, I'd do so.

Yep I get what it does, I'm just wondering if we should make the soft limit `maxValue` dynamic, based on the currently set `hardLimit`, so user can't set it so it's higher than `hardLimit` and get confused why it's not showing up in the field.

tomaskikutis:
> Yep I get what it does, I'm just wondering if we should make the soft limit maxValue dynamic, based on the currently set hardLimit, so user can't set it so it's higher than hardLimit and get confused why it's not showing up in the field.

Best would be to fix so it would show up in the field anyway regardless which is greater. Second best fix is to limit as you suggest.

PR #4905, opened 2025-07-29
Support nested property paths in generic list view

thecalcc:
STT-1232


scripts/core/superdesk-api.d.ts
```diff
@@ -4,6 +4,8 @@ declare module 'superdesk-api' {
     // TYPESCRIPT TYPES
 
     type OrderedMap<K, V> = import('immutable').OrderedMap<K, V>;
+    type FormFieldType = import('../core/ui/components/generic-form/interfaces/form').FormFieldType;
```
tomaskikutis: could you rename to `GenericFormFieldType`? There are many other forms.

scripts/core/superdesk-api.d.ts
```diff
@@ -2157,13 +2143,13 @@ declare module 'superdesk-api' {
         openByDefault: boolean;
     }
 
-    export interface IFormGroup { // don't forget to update runtime type checks
+    export interface IFormGroup<T extends object> { // don't forget to update runtime type checks
```
tomaskikutis, line 2146: why are you extending `object` here?

scripts/core/ui/components/generic-form/from-group.tsx
```diff
@@ -42,9 +43,9 @@ export class FormViewEdit<T> extends React.Component<IProps<T>> {
                                     key={i}
                                     formValues={this.props.item}
                                     formField={item}
-                                    value={this.props.item[item.field]}
+                                    value={get(this.props.item, item.field)}
                                     disabled={!this.props.editMode}
-                                    issues={this.props.issues[item.field] || []}
+                                    issues={get(this.props.issues, item.field) || []}
```
tomaskikutis, line 48: what about issues prop in `FormViewEdit` usage above, it doesn't need to be adjusted with lodash.get?

scripts/core/ui/components/ListPage/generic-list-page-item-view-edit.tsx
```diff
@@ -87,13 +88,17 @@ export class GenericListPageItemViewEdit<T> extends React.Component<IProps<T>, I
...
-            nextItem: {
-                ...prevState.nextItem,
-                [field]: nextValue,
-            },
-        }));
+
+        this.setState((prevState) => {
+            const nextItem = {...prevState.nextItem};
+
+            set(nextItem, field, nextValue);
```
tomaskikutis: `lodash.set` mutates and react doesn't re-render if mutated because it uses reference equality to decide whether to re-render. Deep cloning would work around it by always re-rendering which ensures correctness, but messes up performance. Using Immer library would be a proper solution here.

scripts/core/services/modalService.tsx
```diff
@@ -176,44 +176,58 @@ export default angular.module('superdesk.core.services.modal', [
...
+            bodyText: string,
+            headerText: string,
+            okText: string,
+            cancelText: string,
+            additionalCancelText: string,
         ) {
-            var delay = $q.defer();
-
-            $modal.open({
-                templateUrl: asset.templateUrl('core/views/confirmation-modal.html'),
```
tomaskikutis, line 188: that file and related tests would need to be dropped if you're replacing it.

scripts/core/superdesk-api.d.ts
thecalcc: Because the `Paths` type needs an object. You can't get keys of a string

scripts/core/ui/components/generic-form/from-group.tsx
thecalcc: I wasn't sure about the format issues would arrive in, so I added it preemptively rather

scripts/core/ui/components/ListPage/generic-list-page-item-view-edit.tsx
```diff
@@ -88,11 +89,10 @@ export class GenericListPageItemViewEdit<T extends object> extends React.Compone
 
     handleFieldChange(field: string, nextValue: valueof<IProps<T>['item']>) {
         // using updater function to avoid race conditions
-
         this.setState((prevState) => {
-            const nextItem = {...prevState.nextItem};
-
-            set(nextItem, field, nextValue);
+            const nextItem = produce(prevState.nextItem, (item) => {
```
tomaskikutis: Can you rename `item` to `draft`? I think it's a good idea to follow this naming convention so it's clear to all developers where immer specific things start and end. It's also useful for reviews - I need to bring it up as an issue if you mutate stuff in react code, except `immer` and that `draft` naming is a quick way to tell.

scripts/core/ui/components/generic-form/from-group.tsx
tomaskikutis: Then add it preemptively for both cases

tomaskikutis (approved):

e2e/client/specs/helpers/monitoring.ts
```diff
@@ -473,11 +473,20 @@ class Monitoring {
...
             const menu: ElementFinder = this.openItemMenu(group, item);
             const header = menu.element(by.buttonText(action));
-            const btn = menu.element(by.partialButtonText(submenu));
 
-            browser.actions().mouseMove(header, {x: -50, y: -50}).mouseMove(header).perform();
+            /**
+             * PROTRACTOR-BUG:
+             * `menu.element(by.partialButtonText(submenu))` was used before, but stopped working. No idea why, the PR
+             * where it started breaking didn't have any related changes.
+             * It appears chaining doesn't work with text based selectors (by.buttonText, by.partialButtonText).
```
thecalcc, line 481: @tomaskikutis I guess this comment should be changed since you said my changes have something to do with the tests failing right? Let me know if I should remove it alltogether.

tomaskikutis: > you said my changes have something to do with the tests failing right?

That's my theory, not fact. Unless you want to investigate to find the root cause - keeping the comment makes sense.

scripts/core/components/resize-observer-component.tsx
```diff
@@ -33,11 +33,15 @@ export class ResizeObserverComponent extends React.PureComponent<IPropsResizeObs
...
     componentDidMount() {
         this.observerInstance = new ResizeObserver((entries) => {
-            this.setState({
-                dimensions: {
-                    width: Math.floor(entries[0].contentRect.width),
-                    height: Math.floor(entries[0].contentRect.height),
-                },
+            // requestAnimationFrame is used, otherwise on mount of this component
+            // there's undelivered notifications error
+            requestAnimationFrame(() => {
```
tomaskikutis, line 38: what problem are you solving with `requestAnimationFrame`?

tomaskikutis: I mean why is there a `undelivered notifications error`?

thecalcc: React puts an overlay with error message `ResizeObserver loop completed with undelivered notifications` when the highlights view mounts. 

I could not get the true reason behind a test failure because this overlay was causing the test to fail before the real failure. So I had to fix this first, and see where it truly fails. If tests in GitHub CI weren't executed with a headless browser I think the same error would've occurred.

thecalcc: A more correct version of my message is that - when the list view in monitoring unmounts, specifically in the case when going to the highlights view, the error is thrown and react puts an overlay preventing further button clicks, without closing the error overlay beforehand.

PR #4912, opened 2025-08-08
Reapply limit config after publishing

thecalcc:
STT-1199

### DIFF
--- a/scripts/core/editor3/reducers/editor3.tsx
+++ b/scripts/core/editor3/reducers/editor3.tsx
@@ -186,9 +186,7 @@ export const onChange = (
     keepSelection = false,
     skipOnChange = false,
 ): IEditorStore => {
-    // Force update text decorators, since onChange is triggered on item publish.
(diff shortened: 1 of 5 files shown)


scripts/core/editor3/reducers/editor3.tsx
```diff
@@ -186,8 +186,9 @@ export const onChange = (
     keepSelection = false,
     skipOnChange = false,
 ): IEditorStore => {
-    let editorStateNext = updateDecorators(state, newEditorState, 'store-based');
-
+    // Force update text decorators, since onChange is triggered on item publish.
+    // Otherwise text decorators disappear.
+    const editorStateNext = updateDecorators(state, newEditorState, 'store-based', true);
```
tomaskikutis: This reduces a lot of performance optimizations. It will now redecorate on every cursor move. Would be good to understand what's wrong with publishing and do a fix that doesn't degrade performance.

petrjasek (approved):

PR #4914, opened 2025-08-12
Priority and Urgency field: Handle the sorting of the values in the dropdowns alphabetically using the qcode

dzonidoo:
STT-1284

### DIFF
--- a/scripts/apps/authoring/metadata/metadata.ts
+++ b/scripts/apps/authoring/metadata/metadata.ts
@@ -428,10 +428,7 @@ function MetaDropdownDirective($filter, metadata) {
                         scope.list = $filter('sortByName')(scope.list);
                     } else if (scope.field === 'priority' || scope.field === 'urgency') {
                         scope.list.sort((a, b) => {
-                            const valA = a.qcode;
-                            const valB = b.qcode;
-
-                            return String(valA).localeCompare(String(valB));
+                            return String(a.qcode).localeCompare(String(b.qcode));
                         });
                     }
                 }
--- a/scripts/apps/authoring/metadata/metadata.ts
+++ b/scripts/apps/authoring/metadata/metadata.ts
@@ -426,6 +426,8 @@ function MetaDropdownDirective($filter, metadata) {
                         scope.places = _.groupBy(scope.list, 'group');
                     } else if (scope.field === 'genre') {
                         scope.list = $filter('sortByName')(scope.list);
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -426,6 +426,8 @@ function MetaDropdownDirective($filter, metadata) {
                         scope.places = _.groupBy(scope.list, 'group');
                     } else if (scope.field === 'genre') {
                         scope.list = $filter('sortByName')(scope.list);
+                    } else if (scope.field === 'priority' || scope.field === 'urgency') {
+                        scope.list.sort((a, b) => a.qcode - b.qcode);
```
petrjasek: those values won't be always numbers, so would be good to have it work with any values

dzonidoo: something like -> Number(a.qcode) - Number(b.qcode) ?

scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -427,7 +427,12 @@ function MetaDropdownDirective($filter, metadata) {
                     } else if (scope.field === 'genre') {
                         scope.list = $filter('sortByName')(scope.list);
                     } else if (scope.field === 'priority' || scope.field === 'urgency') {
-                        scope.list.sort((a, b) => a.qcode - b.qcode);
+                        scope.list.sort((a, b) => {
+                            const valA = a?.qcode ?? '';
+                            const valB = b?.qcode ?? '';
```
thecalcc: while it's good to have a fallback, in the UI you can't save a custom vocabulary if an item doesn't have a `qcode`. If the user still ends up saving it though the issue should be fixed elsewhere.

scripts/apps/authoring/metadata/metadata.ts
```diff
@@ -428,8 +428,8 @@ function MetaDropdownDirective($filter, metadata) {
                         scope.list = $filter('sortByName')(scope.list);
                     } else if (scope.field === 'priority' || scope.field === 'urgency') {
                         scope.list.sort((a, b) => {
-                            const valA = a?.qcode ?? '';
-                            const valB = b?.qcode ?? '';
+                            const valA = a.qcode;
+                            const valB = b.qcode;
```
thecalcc: why do you need variables?

dzonidoo: ure totally right

thecalcc (approved):

PR #4917, opened 2025-08-15
Sort by username in weekly and daily views

thecalcc:
SDBELGA-964


dzonidoo (approved):

tomaskikutis:
@thecalcc I've made it configurable which is nice to have on its own - and also we won't need to regenerate the screenshots for e2e tests.

tomaskikutis:
I see you added sorting where it wasn't present at all before so I'll have to regenerate the e2e screenshots anyway

thecalcc:
I'd say it would be good to make the screenshots platform independent anyhow. Otherwise we will surely run into the same issue where someone is blocked for a longer period.

PR #4923, opened 2025-08-25
Custom monitoring list filters

thecalcc:
STT-1198

UI Preview:

- Dropdown with some values:

image
image

- Tags component in different states
image
image
image
image


scripts/apps/monitoring/controllers/AggregateCtrl.ts
```diff
@@ -31,18 +48,56 @@ export function AggregateCtrl($scope, desks, workspaces, preferencesService, sto
...
+
+    /**
+     * All active filters currently set
+     */
+    const configsWithValues: Dictionary<string, Array<string>> = listFiltersConfig
+        .reduce((prev, curr) => ({
+            ...prev,
+            [curr.fieldId]: isInMonitoring === false
+                ? []
+                : JSON.parse(localStorage.getItem(curr.fieldId) ?? '[]'),
```
tomaskikutis: Prefix local storage keys when dynamic values are involved. Will it have one local storage key per field? Could it have one key like "filters for fields" and store all of them not to litter the global scope of local storage.

scripts/apps/monitoring/directives/ActiveFilterTags.tsx
```diff
@@ -0,0 +1,115 @@
+import * as React from 'react';
+import {gettext} from 'core/utils';
+import {Button, Spacer, Tag, Text} from 'superdesk-ui-framework/react';
+import {IActiveFilters} from '../controllers/types';
+import {getTagsConfig, getTagsWithValues} from './utils';
+
+interface IProps {
+    activeFilters?: IActiveFilters;
+    removeFilter: (id: string | null, type?: string) => void;
+    dataTestId?: string;
```
tomaskikutis: does it make sense to have a dynamic data-test-id here? If the component is going to be used only once - what about putting it inside this component - maybe to root spacer instead of passing it as a prop?

I see also it's not used.

scripts/apps/monitoring/directives/ActiveFilterTags.tsx
```diff
@@ -0,0 +1,115 @@
+import * as React from 'react';
+import {gettext} from 'core/utils';
+import {Button, Spacer, Tag, Text} from 'superdesk-ui-framework/react';
+import {IActiveFilters} from '../controllers/types';
+import {getTagsConfig, getTagsWithValues} from './utils';
+
+interface IProps {
+    activeFilters?: IActiveFilters;
+    removeFilter: (id: string | null, type?: string) => void;
```
tomaskikutis: what is type? I saw in usage you're passing `fieldId` - maybe it should be renamed here too?

scripts/apps/monitoring/directives/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+            fieldId: 'anpa_category.qcode',
+            getOptions: () => sdApi.vocabularies
+                .getAll().find((x) => x._id === 'categories').items
+                .map((x) => ({id: x.qcode, label: x.name})),
+            selectMultiple: true,
+            operator: 'OR',
+        },
+    ];
+
+const filtersToIgnore = ['fileType', 'customFilters'];
```
tomaskikutis, line 43: could you add a comment describing why ignoring is needed and why for these fields?

scripts/apps/monitoring/directives/MonitoringListFiltersDropdown.tsx
```diff
@@ -0,0 +1,108 @@
...
+        this.props.onChange(fieldId, values, label);
+    }
+
+    render() {
+        return (
+            <WithPopover
+                shouldCloseOnClick={(e) => {
+                    const target = e.target as HTMLElement;
+
+                    return target.closest('[data-test-id="tree-select-popover"]') == null;
```
tomaskikutis: Test IDs should only be used for testing. Use a different attribute - maybe a classname?

scripts/apps/monitoring/directives/MonitoringListFiltersDropdown.tsx
```diff
@@ -0,0 +1,108 @@
+import * as React from 'react';
+import {NavButton, Spacer, TreeSelect, WithPopover} from 'superdesk-ui-framework/react';
+import {ITreeNode, Dictionary, IMonitoringListFilter} from 'superdesk-api';
+import {listFiltersConfig} from './utils';
+
+interface IProps {
+    dataTestId?: string;
```
tomaskikutis: it looks like this component will only be used once - dynamic test id doesn't make sense then

scripts/extensions/monitoringFilters/src/extension.ts
```diff
@@ -0,0 +1,25 @@
+import {IExtension, IExtensionActivationResult, IMonitoringListFilter, ISuperdesk} from 'superdesk-api';
```
tomaskikutis: Is there a purpose to have this as an extension vs as a config option?

scripts/apps/monitoring/controllers/AggregateCtrl.ts
thecalcc: I can partially do it, but since there's coupling with `contentProfile` I won't. It's not much better anyhow. We may argue if there's 50 filters vs 1 and accessing local storage would be faster if they were all under one key but I just kept following the existing pattern which is to have them all separately

tomaskikutis: I don't think there is an existing pattern where dynamic names are used for local storage keys. The purpose to prefix isn't for performance, but for organization and avoiding conflicts which are easy to get into when keys are dynamic and unprefixed.

scripts/apps/monitoring/directives/MonitoringListFiltersDropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+        this.props.onChange(fieldId, values, label);
+    }
+
+    render() {
+        return (
+            <WithPopover
+                shouldCloseOnClick={(e) => {
+                    const target = e.target as HTMLElement;
+
+                    return target.closest('.autocomplete') == null;
```
tomaskikutis: Could you choose a more specific classname that would be more unique and easier to find?

thecalcc:
> I don't think there is an existing pattern where dynamic names are used for local storage keys. The purpose to prefix isn't for performance, but for organization and avoiding conflicts which are easy to get into when keys are dynamic and unprefixed.

I meant to comment only about the single entry for all filters. I forgot to add the prefix to the dynamic keys. Will do now

scripts/apps/monitoring/directives/MonitoringListFiltersDropdown.tsx
thecalcc: I can't just type any class name, I have to type the one that is the most top level in the context of the TreeSelect dropdown. In this case the other option is to narrow it using two classnames that are put on this said element

tomaskikutis: I'd be cleanest to add a class `tree-select-popover` next to `[data-test-id="tree-select-popover"]`.

tomaskikutis (approved):

PR #4926, opened 2025-08-26
Disable user e2e test

dzonidoo:
SDESK-7605

### DIFF
--- a/scripts/apps/users/views/user-list-item.html
+++ b/scripts/apps/users/views/user-list-item.html
@@ -3,6 +3,7 @@
     ng-show="!authorOnlyFilter || authorOnlyFilter && user.is_author"
     ng-class="{selected: user._id === selected._id, 'inactive' : !user.is_enabled || !user.is_active}"
     data-test-id="users-list-item"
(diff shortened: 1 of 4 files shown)


scripts/apps/users/views/user-list-item.html
```diff
@@ -15,7 +15,7 @@
             <span class="disabled-label" ng-if="user.is_enabled && !user.is_active" translate>inactive</span>
             <span class="disabled-label" ng-if="user.is_enabled && user.is_active && user.needs_activation" translate>pending</span>
         </div>
-        <div class="username" data-test-id="username">{{ user.username }}</div>
+        <div class="username" data-test-id="username" data-test-value="{{ user.username }}">{{ user.username }}</div>
```
tomaskikutis: can you move `data-test-value="{{ user.username }}"` on the `<li>` itself?

dzonidoo: I changed it and now Im using the display name.

tomaskikutis (approved):

PR #4934, opened 2025-09-03
Render username instead of signoff in row item (#4932)

thecalcc:
SDBELGA-977


tomaskikutis:
@thecalcc did you try cherry picking my commit where I updated the screenshots before?

tomaskikutis:
fix CI and I'll update screnshots then

thecalcc:
Since i fixed the failing test in https://github.com/superdesk/superdesk-client-core/pull/4936 I'll wait for it to be merged first then update this branch with release/2.10 and see if there's still something else needed than screenshots

thecalcc:
should be only screenshots left @tomaskikutis

PR #4936, opened 2025-09-03
Port prs update UI framework

thecalcc:


tomaskikutis (approved):

petrjasek:
looks good, so the only part failing now are those screenshots tests right? just make sure not to squash it when merging

thecalcc:
> looks good, so the only part failing now are those screenshots tests right? just make sure not to squash it when merging

will do, yeah not quite, some ui-framework/date-fns related stuff isn't compiling so have to fix that beforehand

petrjasek:
I see, was checking only reports from node 22 and there it compiles ok

thecalcc:
> I see, was checking only reports from node 22 and there it compiles ok

yep that was required for node 14 since compilation failed in e2e - https://github.com/superdesk/superdesk-client-core/actions/runs/17432916687/job/49495364493, which I'm not sure exactly why since on MDN it says that optional chaining and nullish coalescing is supported since node 14

thecalcc:
@tomaskikutis this PR also needs screenshot updates

petrjasek:
@thecalcc would be better to push it to the branch and not squash it, it looses the info about those cherry-picked commits

PR #4941, opened 2025-09-05
restrict short field to limit specified in schema

tomaskikutis:
STT-1285


tomaskikutis:
I've restricted values to respect the config in schema. It's already set to 2 on sd-develop and likely elsewhere. I'm thinking what to do with existing values? Probably easiest to write an upgrade script to modify urgency,priority vocabularies - go over entries and trim `List name` to max value defined in schema. What do you think @petrjasek ?

petrjasek:
afaik nobody was using there longer values, so we probably don't need to handle it automatically

petrjasek (approved):

PR #4953, opened 2025-09-18
Urgency is set to Null when opening story

dzonidoo:
STT-1312


tomaskikutis:
why was it `null`?

dzonidoo:
The selected variable was never initialized when loading an existing article for editing.

petrjasek (approved):

PR #4954, opened 2025-09-18
availability manager - increase api limit to 1000

tomaskikutis:
SDBELGA-982


thecalcc (commented):
why another increase? Do we not support pagination inside the availability?

thecalcc (approved):

tomaskikutis:
No, we do not support pagination.

PR #4968, opened 2025-10-03
Update UI framework

thecalcc:
STT-1338

Update UI framework to keep planning `feature/embedded-events-editor-async` UI framework version in sync with client-core develop so STT-next picks up the proper UI framework version


tomaskikutis:
I have https://github.com/superdesk/superdesk-client-core/pull/4967 open doing the same thing. I will fix the tests on ~this one~ my PR. I've modified modals to be appended to the body and it may be causing issues.

PR #4969, opened 2025-10-03
expose date input component for archive search widget

txl-cp:


tomaskikutis:
Have you tried using date picker from ui-framework https://ui-framework.superdesk.org/#/components/date-picker?

The one you're trying to expose should be deprecated.

txl-cp:
> Have you tried using date picker from ui-framework https://ui-framework.superdesk.org/#/components/date-picker?
> 
> The one you're trying to expose should be deprecated.

Yes, that's what we were using at first, but clients wanted navigator so i exposed core date picker instead since it has navigator

tomaskikutis:
What do you mean by navigator? Capability to choose presets like "today"/"tomorrow"? Those are available in ui-framework picker by using `headerButtonBar` prop

txl-cp:
> What do you mean by navigator? Capability to choose presets like "today"/"tomorrow"? Those are available in ui-framework picker by using `headerButtonBar` prop

clicking the month/year label switches into picker for month/year. the presets were not enough

tomaskikutis:
I see. I understand that month/year switches are useful, but we won't be maintaining that component - and thus won't be exposing it via API.

One option you have is to copy-paste that component into CP repository and maintain it yourselves.

Usually, there would be also an option to submit a PR to ui-framework to add missing features there, but I've checked with my colleague and due to release schedule for CP that option isn't available in this case - meaning the release you're currently working on. You may also choose to use the ui-framework component as is for next release, submit a PR to improve ui-framework date picker and it would arrive at some later release.

PR #4977, opened 2025-10-10
fix flaky e2e

petrjasek:
fill with plain text editor inside angular is not stable

### DIFF
--- a/e2e/client/playwright/authoring.picture.spec.ts
+++ b/e2e/client/playwright/authoring.picture.spec.ts
@@ -5,8 +5,6 @@ import {MediaEditor} from './page-object-models/media-editor';
 import {PictureAuthoring} from './page-object-models/authoring';
 import {MediaUpload} from './page-object-models/upload';
 
-test.setTimeout(30000);
-
 /**
  * upload a picture
  * edit metadata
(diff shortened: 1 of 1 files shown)


e2e/client/playwright/authoring.picture.spec.ts
```diff
@@ -29,6 +27,8 @@ test('media metadata editor', async ({page}) => {
 
     await expect(mediaEditor.field('field--headline')).toContainText('The Headline');
 
+    // workaround for flaky fill with plain text editor, it sometimes keeps previous text
+    await mediaEditor.field('field--headline').fill('');
```
tomaskikutis: Was it happening on `input[type=text]` field, not editor3? I was having the same issue with editor3, in the end [wrote a helper](https://github.com/superdesk/superdesk-client-core/blob/develop/e2e/client/playwright/utils/editor3.tsx#L30) that fills the text in a loop until it gets filled correctly :D

tomaskikutis (approved):

e2e/client/playwright/authoring.picture.spec.ts
petrjasek: it's not editor 3 but plain text field using draftjs, so probably the same thing

PR #4978, opened 2025-10-13
The date picker in search filter is not positioned correctly [SDBELGA-995]

devketanpro:
SDBELGA-995

**Purpose:**
Rework the datepicker popover, so it is positioned correctly, and it's content isn't cut off at the bottom.

### DIFF
--- a/scripts/core/ui/ui.ts
+++ b/scripts/core/ui/ui.ts
@@ -410,39 +410,61 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
 
             scope.$watch('open', (value) => {
                 if (value) {
-                    setTimeout(() => {
-                        const popupEl = $popupWrapper[0];
+                    // Use native setTimeout (no Angular digest required here) to wait
+                    // for the popup DOM to be attached and rendered before measuring.
+                    // We also re-run the positioning on the next frame and after a small
+                    const MARGIN = 8;
+                    const RECHECK_DELAY = 50;
+
+                    const positionOnce = () => {
+                        const popupEl = $popupWrapper && $popupWrapper[0];
 
                         if (!popupEl) return;
 
                         // Get position of the active input element
                         const rect = element[0].getBoundingClientRect();
+
+                        // Measure popup size (use defaults if not available)
                         const popupHeight = popupEl.offsetHeight || 320;
                         const popupWidth = popupEl.offsetWidth || 280;
                         const tolerance = 10;
 
                         const viewportHeight = window.innerHeight;
                         const viewportWidth = window.innerWidth;
+                        const scrollX = window.scrollX || window.pageXOffset;
+                        const scrollY = window.scrollY || window.pageYOffset;
 
-                        let top = rect.bottom + window.scrollY;
-                        let left = rect.left + window.scrollX;
+                        // Prefer opening below, leave a small margin so input remains visible.
+                        let top = rect.bottom + scrollY + MARGIN;
+                        let left = rect.left + scrollX;
 
-                        // If not enough space below, open upward
+                        // If not enough space below, open upward (keep margin so input is not overlapped)
                         if (rect.bottom + popupHeight + tolerance > viewportHeight) {
-                            top = rect.top + window.scrollY - popupHeight;
+                            top = rect.top + scrollY - popupHeight - MARGIN;
                         }
 
-                        // Prevent overflow
+                        // Prevent overflow on right
                         if (rect.left + popupWidth + tolerance > viewportWidth) {
-                            left = viewportWidth - popupWidth - tolerance + window.scrollX;
+                            left = viewportWidth - popupWidth - tolerance + scrollX;
+                        }
+
+                        // Prevent overflow on left
+                        if (left < tolerance + scrollX) {
+                            left = tolerance + scrollX;
                         }
 
                         Object.assign(popupEl.style, {
-                            top: `${top}px`,
-                            left: `${left}px`,
+                            top: `${Math.round(top)}px`,
+                            left: `${Math.round(left)}px`,
                             position: 'absolute',
+                            zIndex: '2000',
                         });
+                    };
 
+                    setTimeout(() => {
+                        positionOnce();
+                        requestAnimationFrame(positionOnce);
+                        setTimeout(positionOnce, RECHECK_DELAY);
                         scope.$broadcast('datepicker.focus');
                     }, 0);
                 }
--- a/scripts/core/ui/ui.ts
+++ b/scripts/core/ui/ui.ts
@@ -409,65 +409,78 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
             });
 
             scope.$watch('open', (value) => {
-                if (value) {
-                    // Use native setTimeout (no Angular digest required here) to wait
-                    // for the popup DOM to be attached and rendered before measuring.
-                    // We also re-run the positioning on the next frame and after a small
-                    const MARGIN = 8;
-                    const RECHECK_DELAY = 50;
-
-                    const positionOnce = () => {
-                        const popupEl = $popupWrapper && $popupWrapper[0];
-
-                        if (!popupEl) return;
-
-                        // Get position of the active input element
-                        const rect = element[0].getBoundingClientRect();
-
-                        // Measure popup size (use defaults if not available)
-                        const popupHeight = popupEl.offsetHeight || 320;
-                        const popupWidth = popupEl.offsetWidth || 280;
-                        const tolerance = 10;
-
-                        const viewportHeight = window.innerHeight;
-                        const viewportWidth = window.innerWidth;
-                        const scrollX = window.scrollX || window.pageXOffset;
-                        const scrollY = window.scrollY || window.pageYOffset;
-
-                        // Prefer opening below, leave a small margin so input remains visible.
-                        let top = rect.bottom + scrollY + MARGIN;
-                        let left = rect.left + scrollX;
-
-                        // If not enough space below, open upward (keep margin so input is not overlapped)
-                        if (rect.bottom + popupHeight + tolerance > viewportHeight) {
-                            top = rect.top + scrollY - popupHeight - MARGIN;
-                        }
-
-                        // Prevent overflow on right
-                        if (rect.left + popupWidth + tolerance > viewportWidth) {
-                            left = viewportWidth - popupWidth - tolerance + scrollX;
-                        }
-
-                        // Prevent overflow on left
-                        if (left < tolerance + scrollX) {
-                            left = tolerance + scrollX;
-                        }
-
-                        Object.assign(popupEl.style, {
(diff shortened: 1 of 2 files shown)


thecalcc:
why did we do the padding change? Could you post before/after screenshots

devketanpro:
Before screenshot:
image


After screenshot:
image

devketanpro:
> why did we do the padding change? Could you post before/after screenshots

The padding change was made to improve the spacing and overall layout consistency.

tomaskikutis (commented):
That's not fixing the root cause of the issue. It's positioning code of date picker that should be improved.

thecalcc:
Tomas is correct, fix should be in UI framework, we have some positioning code already there for e.g. `WithPopover` check if you can reuse that so we have dynamic positioning, thus is space on the bottom is not enough to display the whole popover auto positioner puts it on the top or other positions.

scripts/core/ui/ui.ts
```diff
@@ -410,8 +410,19 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
 
             scope.$watch('open', (value) => {
                 if (value) {
-                    $popupWrapper.offset(popupService.position(260, 270, element));
-                    scope.$broadcast('datepicker.focus');
+                    $timeout(() => {
```
thecalcc: why do you need $timeout from angular? you can use `setTimeout` probably. 

Also why is the timeout needed in the first place? Are you waiting for the component to render first? Also can you drop the usage of this legacy `$popupWrapper`?

Could you instead of making a legacy hack replace the whole component with a new one from UI framework?

tomaskikutis: I was thinking the same that it would probably be easier to connect date picker from ui-framework rather than correct positioning code which is not the easiest thing. But if this patch works I'm fine with the existing solution.

tomaskikutis (approved):

thecalcc (changes requested):
let me know if you need help with the positioning

devketanpro:
image

scripts/core/ui/ui.ts
```diff
@@ -410,39 +410,61 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
...
+
+                    const positionOnce = () => {
+                        const popupEl = $popupWrapper && $popupWrapper[0];
 
                         if (!popupEl) return;
 
                         // Get position of the active input element
                         const rect = element[0].getBoundingClientRect();
+
+                        // Measure popup size (use defaults if not available)
```
thecalcc: why would it not be available? if it's not available why does popup get rendered?

scripts/core/ui/ui.ts
```diff
@@ -410,39 +410,61 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
 
             scope.$watch('open', (value) => {
                 if (value) {
-                    setTimeout(() => {
-                        const popupEl = $popupWrapper[0];
+                    // Use native setTimeout (no Angular digest required here) to wait
+                    // for the popup DOM to be attached and rendered before measuring.
```
thecalcc: not sure if we need this comment

scripts/core/ui/ui.ts
```diff
@@ -410,39 +410,61 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
 
             scope.$watch('open', (value) => {
                 if (value) {
```
thecalcc: you can simplify nesting here probably if you flip this condition

scripts/core/ui/ui.ts
```diff
@@ -410,39 +410,61 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
...
-                    setTimeout(() => {
-                        const popupEl = $popupWrapper[0];
+                    // Use native setTimeout (no Angular digest required here) to wait
+                    // for the popup DOM to be attached and rendered before measuring.
+                    // We also re-run the positioning on the next frame and after a small
+                    const MARGIN = 8;
```
thecalcc: this isn't margin, it should be something like `INPUT_HEIGHT`. Also it should be calculated on runtime ideally, not just have a fixed value. If we decide to have different heights for different inputs it would break. 

But that's a big if so we can keep it as is, so if you could just rename it

scripts/core/ui/ui.ts
```diff
@@ -410,39 +410,61 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
 
             scope.$watch('open', (value) => {
                 if (value) {
-                    setTimeout(() => {
-                        const popupEl = $popupWrapper[0];
+                    // Use native setTimeout (no Angular digest required here) to wait
+                    // for the popup DOM to be attached and rendered before measuring.
+                    // We also re-run the positioning on the next frame and after a small
```
thecalcc: is this comment unfinished? also what is recheck delay?

scripts/core/ui/ui.ts
```diff
@@ -410,39 +410,61 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
...
+
+                        // Prevent overflow on left
+                        if (left < tolerance + scrollX) {
+                            left = tolerance + scrollX;
                         }
 
                         Object.assign(popupEl.style, {
-                            top: `${top}px`,
-                            left: `${left}px`,
+                            top: `${Math.round(top)}px`,
```
thecalcc: why the math round here?

scripts/core/ui/ui.ts
```diff
@@ -410,39 +410,61 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
...
+                        const scrollY = window.scrollY || window.pageYOffset;
 
-                        let top = rect.bottom + window.scrollY;
-                        let left = rect.left + window.scrollX;
+                        // Prefer opening below, leave a small margin so input remains visible.
+                        let top = rect.bottom + scrollY + MARGIN;
+                        let left = rect.left + scrollX;
 
-                        // If not enough space below, open upward
+                        // If not enough space below, open upward (keep margin so input is not overlapped)
```
thecalcc: Could you make this comment better? "If there isn't enough space available below the input, position the popup above the input"

scripts/core/ui/ui.ts
```diff
@@ -410,39 +410,61 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
...
                 if (value) {
-                    setTimeout(() => {
-                        const popupEl = $popupWrapper[0];
+                    // Use native setTimeout (no Angular digest required here) to wait
+                    // for the popup DOM to be attached and rendered before measuring.
+                    // We also re-run the positioning on the next frame and after a small
+                    const MARGIN = 8;
+                    const RECHECK_DELAY = 50;
+
+                    const positionOnce = () => {
```
thecalcc: why is the name `positionOnce`?

scripts/core/ui/ui.ts
```diff
@@ -410,39 +410,61 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
...
+                            top: `${Math.round(top)}px`,
+                            left: `${Math.round(left)}px`,
                             position: 'absolute',
+                            zIndex: '2000',
                         });
+                    };
 
+                    setTimeout(() => {
+                        positionOnce();
+                        requestAnimationFrame(positionOnce);
```
thecalcc: why do we need this call here? `requestAnimationFrame` would call `positionOnce` before the next repaint

scripts/core/ui/ui.ts
```diff
@@ -409,65 +409,78 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
...
-                        if (left < tolerance + scrollX) {
-                            left = tolerance + scrollX;
-                        }
+                    // If there isn't enough space below the input, position the popup above the input.
+                    if (rect.bottom + popupHeight + TOLERANCE > viewportHeight && rect.top >= popupHeight + GAP + TOLERANCE) {
+                        top = rect.top + scrollY - popupHeight - GAP;
+                    } else if (rect.bottom + popupHeight + TOLERANCE > viewportHeight) {
+                        // Neither side fits fully: clamp into viewport and prefer below.
+                        top = Math.min(Math.max(scrollY + TOLERANCE, rect.bottom + scrollY + GAP), scrollY + viewportHeight - popupHeight - TOLERANCE);
+                    }
```
thecalcc, line 457: could you make the code here readable?

scripts/core/ui/ui.ts
```diff
@@ -409,65 +409,78 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
             });
 
             scope.$watch('open', (value) => {
-                if (value) {
-                    // Use native setTimeout (no Angular digest required here) to wait
-                    // for the popup DOM to be attached and rendered before measuring.
-                    // We also re-run the positioning on the next frame and after a small
-                    const MARGIN = 8;
-                    const RECHECK_DELAY = 50;
+                if (!value) {
```
thecalcc, line 412: nitpick - could you have this as an explicit check like `value == null`

scripts/core/ui/ui.ts
```diff
@@ -409,65 +409,78 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
...
-                        const popupWidth = popupEl.offsetWidth || 280;
-                        const tolerance = 10;
+                    // Get position of the active input element
+                    const rect = element[0].getBoundingClientRect();
 
-                        const viewportHeight = window.innerHeight;
-                        const viewportWidth = window.innerWidth;
-                        const scrollX = window.scrollX || window.pageXOffset;
-                        const scrollY = window.scrollY || window.pageYOffset;
+                    // Measure popup size (use sensible defaults if not available)
```
thecalcc, line 434: in my understanding, if the popup element is not available, this function shouldn't be called. if so, then we don't need fallback values. But if it's still running when `popupEl` is null, we should fix that first

scripts/core/ui/ui.ts
```diff
@@ -409,65 +409,78 @@ function DatepickerInnerDirective($compile, $document, popupService, datetimeHel
...
+                    const scrollX = window.scrollX || window.pageXOffset;
+                    const scrollY = window.scrollY || window.pageYOffset;
 
-                        // If not enough space below, open upward (keep margin so input is not overlapped)
-                        if (rect.bottom + popupHeight + tolerance > viewportHeight) {
-                            top = rect.top + scrollY - popupHeight - MARGIN;
-                        }
+                    // Dynamic gap based on input height so popup doesn't cover the input.
+                    const inputHeight = rect.height || (element.outerHeight && element.outerHeight()) || 32;
+                    const GAP = Math.max(8, Math.round(inputHeight));
```
thecalcc, line 445: why do we need this? and not just `rect.height`

thecalcc, line 445: this is not a gap, it's input height

thecalcc:
Closing in favour of https://github.com/superdesk/superdesk-client-core/pull/4994

PR #4980, opened 2025-10-15
[SDESK-7783] Add getUniqueClientId to SuperdeskApi

MarkLark86:
### Purpose
In Planning, we need a way to uniquely identify clients based on the browser tab. This is used in the changes for item locking.

### What has changed
Add a new SuperdeskApi function to generate and store a unique ID in the browser tabs session.

Related To: [SDESK-7783](https://sofab.atlassian.net/browse/SDESK-7783)

[SDESK-7783]: https://sofab.atlassian.net/browse/SDESK-7783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/api/user.ts
+++ b/scripts/api/user.ts
@@ -1,6 +1,7 @@
 import ng from 'core/services/ng';
 import {IUser} from 'superdesk-api';
 import {OrderedMap} from 'immutable';
+import {uuid} from 'core/helpers/uuid';
 
 function hasPrivilege(privilege: string): boolean {
     const privileges = ng.get('privileges');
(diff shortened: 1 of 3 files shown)


tomaskikutis:
Wouldn't session ID and user ID be enough? How client ID differs from those?

MarkLark86:
> Wouldn't session ID and user ID be enough? How client ID differs from those?

Session can be shared among browser tabs. Many of our customers use multiple tabs for our apps. This makes it concrete, browser tab based

eos87 (approved):

PR #4981, opened 2025-10-16
fix styles on login page

petrjasek:
SDESK-7450

### DIFF
--- a/scripts/core/auth/login-modal.html
+++ b/scripts/core/auth/login-modal.html
@@ -27,19 +27,12 @@
                 </fieldset>
             </form>
 
-            <fieldset class="buttons" ng-if="methods.google">
-                <button class="btn btn--large btn--hollow btn--sd-green"
-                    ng-click="openLoginPopup('google')" translate>Log In with Google</button>
-            </fieldset>
+            <div class="links" ng-if="features.auth_db">
+                <a href="" ng-click="logout()" class="text" ng-show="identity.Id" translate>Sign in as a different user.</a>
+                <a href="/#/reset-password/" class="text" ng-show="features.auth_db && features.reset_user_password" translate>Forgot password?</a>
+            </div>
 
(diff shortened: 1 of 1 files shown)


scripts/core/auth/login-modal.html
```diff
@@ -54,18 +47,27 @@
                     <span class="icon-unplug"></span>
                     <p translate>Trying to reconnect...<br />Please wait.</p>
                 </div>
-
             </div>
 
-            <div class="links">
-                <a href="" ng-click="logout()" class="text" ng-show="identity.Id" translate>Sign in as a different user.</a>
-                <a href="/#/reset-password/" class="text" ng-show="features.auth_db && features.reset_user_password" translate>Forgot password?</a>
+            <div class="sd-content-divider--with-text sd-content-divider sd-content-divider--horizontal sd-content-divider--margin-small" role="separator" ng-if="(methods.google || methods.saml) && features.auth_db">
```
tomaskikutis, line 52: I suggest to only use react components from ui-framework and not classnames. Classnames are internal implementation details that should be easy to rename and refactor and only component interfaces we treat with care and put effort into not breaking.

You could either create a ~spacer~ divider component  react component in ui-framework, register via angular bridge and use here or copy-paste the CSS into this project and rename classnames to avoid conflicts.

tomaskikutis (approved):

scripts/core/auth/login-modal.html
tomaskikutis: There is actually already a divider component https://github.com/superdesk/superdesk-ui-framework/blob/develop/app-typescript/components/Divider.tsx

petrjasek: this one is different, it has text in it, not sure if we would use it elsewhere

tomaskikutis: In that case I'd copy the styles from the framework and change the classnames to avoid conflicts.

PR #4986, opened 2025-10-20
[STT-1382] - Pinning of the Planning details and AI widget breaks the UI

BrianMwangi21:
[STT-1382](https://sofab.atlassian.net/browse/STT-1382)

Fix pinning of planning details and ai widget breaking the UI. Now they pin without shifting the layout but as expected. Also added a suppression logic to avoid transition on unpin during remount.

### Before
[dWs3_planning-widget-breaks-ui.webm](https://github.com/user-attachments/assets/669ad512-aca8-4dea-9a85-22be5dea83d6)

### After

https://github.com/user-attachments/assets/3f47b2ef-ab4e-4856-8057-46b21a0fce93

[STT-1382]: https://sofab.atlassian.net/browse/STT-1382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/apps/authoring/widgets/widgets.ts
+++ b/scripts/apps/authoring/widgets/widgets.ts
@@ -335,7 +335,7 @@ function WidgetsManagerCtrl(
 
         // Temporarily disable CSS transitions during pin/unpin to prevent sliding animations
         // during remount for the react based widgets
-        angular.element('body').addClass('widget-pin-unpin-operation');
+        angular.element('.widget-wrapper .tabpane').css('transition', 'none');
 
         if ($scope.pinnedWidget) {
             $scope.pinnedWidget.pinned = false;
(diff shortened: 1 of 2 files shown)


scripts/apps/authoring/widgets/widgets.ts
```diff
@@ -330,6 +330,10 @@ function WidgetsManagerCtrl(
     };
 
     $scope.pinWidget = (widget: IWidget) => {
+        // Temporarily disable CSS transitions during pin/unpin to prevent sliding animations
+        // during remount for the react based widgets
+        angular.element('body').addClass('widget-pin-unpin-operation');
```
thecalcc: not sure if it would be better to just set the transitions property instead of adding a class, and then remove it after

scripts/apps/authoring/widgets/widgets.ts
```diff
@@ -365,9 +369,16 @@ function WidgetsManagerCtrl(
...
         IS_WIDGET_PINNED = $scope.pinnedWidget?.pinned ?? false;
+
+        // Then remove the transition suppression class after the DOM updates
+        setTimeout(() => {
+            angular.element('body').removeClass('widget-pin-unpin-operation');
+        }, 50);
     };
 
-    widgetReactIntegration.pinWidget = $scope.pinWidget;
+    widgetReactIntegration.pinWidget = (sideWidget?: IWidget) => {
```
thecalcc: shouldn't you make the fix in `$scope.pinWidget`? that way we can have it working the same way in both angular version and react one

thecalcc (approved):

PR #4991, opened 2025-10-24
[STT-1400] - Adding Department and Tapahtumatyyppi fields to a content profile breaks the content profile dialog

BrianMwangi21:
# Purpose
This PR introduces a null-safety guard around retrieving the type value to prevent runtime errors on some custom vocabularies

## Before
[3muE.webm](https://github.com/user-attachments/assets/8baff198-8811-4e1d-b856-a24f7e8e136e)

## After
https://github.com/user-attachments/assets/6eee3b90-c119-4fdd-832d-dcf8d827d062

### DIFF
--- a/scripts/apps/workspace/content/components/get-content-profiles-form-config.tsx
+++ b/scripts/apps/workspace/content/components/get-content-profiles-form-config.tsx
@@ -170,7 +170,7 @@ export function getContentProfileFormConfig(
     if (
         field?.id != null
         && (
-            schema[field.id].type === 'string'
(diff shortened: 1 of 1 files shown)


thecalcc (approved):
looks good, i'm just wondering why is schema `null` for CVs? Could we not add it if it's null?

BrianMwangi21:
I thought the same at first, but I thought guarding the schema lookups is better long term than editing the CVs cause they can diverge across branches/envs and that could be harder to maintain consistently

PR #4997, opened 2025-11-05
[STT-1368] - Update UI language after user authenticates

thecalcc:
STT-1368

### DIFF
--- a/scripts/core/auth/auth.ts
+++ b/scripts/core/auth/auth.ts
@@ -244,18 +244,17 @@ export default angular.module('superdesk.core.auth', [
                 }
 
                 if (canLogout) {
-                    api.auth.getById(session.sessionId).then((sessionData) => {
-                        api.auth
-                            .remove(sessionData)
-                            .finally(() => {
-                                $rootScope.$broadcast(SESSION_EVENTS.LOGOUT);
-                                localStorage.clear();
-
-                                localStorage.setItem('LOGGED_OUT_LANGUAGE', session.identity.language);
-
-                                $window.location.replace('/'); // reset page for new user
-                            });
-                    });
+                    api.auth.getById(session.sessionId)
+                        .then((sessionData) => {
+                            api.auth
+                                .remove(sessionData)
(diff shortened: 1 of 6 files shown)


scripts/core/auth/auth.ts
```diff
@@ -2,6 +2,7 @@ import {gettext} from 'core/utils';
 import {AuthoringWorkspaceService} from 'apps/authoring/authoring/services/AuthoringWorkspaceService';
 import {appConfig} from 'appConfig';
 import ng from 'core/services/ng';
+import {reloadLanguage} from 'init';
```
petrjasek: imo we shouldn't import that init file anywhere, that might break the assumption that config is loaded before the other code is imported

thecalcc: yep that's right, i'm just thinking how to avoid code duplication in this case. one option is to execute relevant parts only if we're not in a test `env`, other is to just go with duplicate code

scripts/reload-language.ts
```diff
@@ -0,0 +1,67 @@
...
+        user?.language
+        ?? localStorage.getItem('LOGGED_OUT_LANGUAGE')
+        ?? appConfig.default_language
+        ?? window.navigator.language
+        ?? 'en';
+
+    return appConfig.profileLanguages?.includes(language) ? language : 'en';
+}
+
+const language = getUserLanguage();
```
petrjasek: shouldn't this be rather called inside those functions?

scripts/init.ts
```diff
@@ -60,25 +60,27 @@ const appConfig: ISuperdeskGlobalConfig = window['appConfigLoaded'];
 // SETTING UI LANGUAGE
 //
 
-const user: IUser | null = JSON.parse(localStorage.getItem(IDENTITY_KEY));
+function getUserLanguage(): string {
```
petrjasek: why not importing it from `reload-language` ?

petrjasek (approved):

PR #4998, opened 2025-11-06
[STT-1416] - Refactor AnnotationPopup to functional component

eos87:
### Purpose
To fix the bug of annotations not loading. The problem was not in the code but it was due to the missing "Annotation Types" in the system metadata. I took advantage of the ticket to improve this component and the handling of missing metadata and annotation type.

### What has changed
- Converted `AnnotationPopup` from a class-based to a functional React component using hooks
- Added TypeScript interfaces for props and improved handling of missing annotation types with notifications. 
- Removed PropTypes and updated dependency loading logic for annotation types.

### To test
- Create a new item and add an annotation in the body html
- Delete the `annotation_types` record from the database collections `vocabularies`
- Reload your browser
- Try to open the annotation in the editor
- Observe the warnings instead of the app breaking because of the missing types

Thanks for checking! 

Resolves: [STT-1416]

[STT-1416]: https://sofab.atlassian.net/browse/STT-1416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/core/editor3/components/annotations/AnnotationPopup.tsx
+++ b/scripts/core/editor3/components/annotations/AnnotationPopup.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect, useRef} from 'react';
+import React, {useEffect} from 'react';
 import moment from 'moment';
 import {connect} from 'react-redux';
 import {showPopup, PopupTypes} from '../../actions';
@@ -13,6 +13,7 @@ import {FluidRow} from '../../fluid-flex-rows/fluid-row';
 import {gettext} from 'core/utils';
 import {editor3StateToHtml} from 'core/editor3/html/to-html/editor3StateToHtml';
 import {notify} from 'core/notify/notify';
+import {noop} from 'lodash';
 
 interface IAnnotationType {
     qcode: string;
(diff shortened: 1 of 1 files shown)


scripts/core/editor3/components/annotations/AnnotationPopup.tsx
```diff
@@ -13,95 +12,141 @@ import {FluidRows} from '../../fluid-flex-rows/fluid-rows';
...
+}) => {
+    const {author, authorId, date, msg, annotationType} = annotation.data;
+
+    useEffect(() => {
+        if (!annotationTypes || annotationTypes.length === 0) {
+            notify.warning(gettext('Annotation Types information is not available. ' +
+                'Please check your metadata configuration.'));
+            console.warn('Annotation types not available or empty', {annotationType});
+        }
+    }, []);
```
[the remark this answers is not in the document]
eos87: this is fine, the component is rerendered every time anyways.

scripts/core/editor3/components/annotations/AnnotationPopup.tsx
```diff
@@ -13,95 +12,141 @@ import {FluidRows} from '../../fluid-flex-rows/fluid-rows';
...
+    annotation: IAnnotation;
+    editorNode: HTMLElement;
+    highlightId: string;
+    highlightsManager: IHighlightsManager;
+    annotationTypes: Array<IAnnotationType>;
+    close: () => void;
+    showPopup: (type: typeof PopupTypes.Annotation, data: {annotation: IAnnotation; highlightId: string}) => void;
+}
+
+interface IPromiseResults {
```
thecalcc: maybe a better name could be done here? e.g. `IPromiseAnnotationTypes`?

scripts/core/editor3/components/annotations/AnnotationPopup.tsx
```diff
@@ -13,95 +12,141 @@ import {FluidRows} from '../../fluid-flex-rows/fluid-rows';
...
+    const onEdit = () => {
+        showPopup(PopupTypes.Annotation, {annotation, highlightId});
+        close();
+    };
+
+    const onDelete = () => modal
+        .confirm(gettext('The annotation will be deleted. Are you sure?'))
+        .then(() => {
+            highlightsManager.removeHighlight(highlightId);
+        }).catch(noop);
```
thecalcc, line 88: why do we need `noop` in the `catch`? maybe we render an alert telling the user that something went wrong and they need to try again?

thecalcc (commented):
good stuff, minor comments only

scripts/core/editor3/components/annotations/AnnotationPopup.tsx
eos87: it was needed because if the user hits cancel or closes the modal then JS throws an error of un-catch promise (something like that) so no need for user notification in this case.

thecalcc: oh alright

thecalcc (approved):

PR #4999, opened 2025-11-10
[STT-1368] - Reload page on language change after login

thecalcc:
STT-1368


eos87 (commented):
I just worry if this is not creating a race condition between the request and the browser reloading?

eos87 (approved):

PR #5001, opened 2025-11-12
set playwright screenshot diff pixel ratio instead of count

petrjasek:
- to make the tests a bit less strict


eos87 (approved):
Exactly what I needed. Thanks!

PR #5008, opened 2025-11-21
[STT-1466] - Added keyboard combos for ndash and thinspace in editor

BrianMwangi21:
### Purpose
This PR added 2 new keyboard combinations for ndash and thinspace. It also allows for the thinspaces to be seen when the formatting marks are toggled on.

### What has changed
- Added `Cmd/Ctrl + Alt + -` combo for ndash
- Added `Cmd/Ctrl + Alt + Shift + Space` combo for thinspace
- When formatting marks are enabled on the editor and toggled on, the thinspace can be seen as `¤`

### Testing
1. Verify ndash are inserted correctly using `Cmd/Ctrl + Alt + -` combo
2. Verify thinspace are inserted using `Cmd/Ctrl + Alt + Shift + Space` combo
3. When formatting marks are toggled on, verify the thinspace can be seen as `¤` in the editor

### Video

https://github.com/user-attachments/assets/b077e9e6-01d6-46b4-b20f-f1787e5b32a9

### DIFF
--- a/scripts/core/editor3/components/Editor3Component.tsx
+++ b/scripts/core/editor3/components/Editor3Component.tsx
@@ -297,7 +297,7 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
     }
 
     keyBindingFn(e) {
-        const {key, ctrlKey, shiftKey, metaKey} = e;
+        const {key, ctrlKey, shiftKey, metaKey, altKey} = e;
         const selectionState = this.props.editorState.getSelection();
         const modifierKey = isMacOS() ? metaKey : ctrlKey;
 
@@ -349,6 +349,18 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
                 e.preventDefault();
                 return '';
             }
+        }
+
+        // Alt + - for ndash
+        if (key === '-' && altKey && !ctrlKey && !shiftKey) {
+            e.preventDefault();
+            return 'custom-ndash';
+        }
+
+        // Alt + Space for thin space
+        if (key === ' ' && altKey && !ctrlKey && !shiftKey) {
+            e.preventDefault();
+            return 'custom-thin-space';
         }
 
         return getDefaultKeyBinding(e);
@@ -428,6 +440,34 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
             case 'secondary-paste': // this is blocking redo on non-windows systems, should be osx specific
                 newState = EditorState.redo(editorState);
                 break;
+            case 'custom-ndash': {
+                const ndashContentState = Modifier.replaceText(
+                    editorState.getCurrentContent(),
+                    editorState.getSelection(),
+                    '\u2013',
+                );
+
+                newState = EditorState.push(
+                    editorState,
+                    ndashContentState,
+                    'insert-characters',
+                );
+                break;
+            }
+            case 'custom-thin-space': {
+                const thinSpaceContentState = Modifier.replaceText(
+                    editorState.getCurrentContent(),
+                    editorState.getSelection(),
+                    '\u2009',
+                );
+
+                newState = EditorState.push(
+                    editorState,
+                    thinSpaceContentState,
+                    'insert-characters',
+                );
+                break;
+            }
             case 'backspace': {
                 this.setState({contentChangesAfterLastFocus: this.state.contentChangesAfterLastFocus + 1});
 
--- a/scripts/core/editor3/components/Editor3Component.tsx
+++ b/scripts/core/editor3/components/Editor3Component.tsx
@@ -351,16 +351,28 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
             }
         }
 
-        // Alt + - for ndash
-        if (key === '-' && altKey && !ctrlKey && !shiftKey) {
-            e.preventDefault();
-            return 'custom-ndash';
+        // Cmd/Ctrl + Alt + - for ndash
+        if (key === '-' && altKey && !shiftKey) {
+            const isMac = isMacOS();
+            const macCombo = isMac && metaKey && !ctrlKey;
+            const winCombo = !isMac && !metaKey && ctrlKey;
+
+            if (macCombo || winCombo) {
+                e.preventDefault();
+                return 'custom-ndash';
+            }
         }
 
-        // Alt + Space for thin space
-        if (key === ' ' && altKey && !ctrlKey && !shiftKey) {
-            e.preventDefault();
-            return 'custom-thin-space';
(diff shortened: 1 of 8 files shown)


scripts/core/editor3/store/index.ts
```diff
@@ -110,16 +111,23 @@ interface IOptions {
...
-    const {limitConfig, softLimitConfig, spellchecker} = options;
+    const {limitConfig, softLimitConfig, spellchecker, invisibles} = options;
 
     // improve performance by not replacing decorators when possible.
     let mustReApplyDecorators = false;
 
     const decorators: Array<{strategy: any, component: any}> = [LinkDecorator];
 
+    // Add thin space decorator when invisibles are enabled
+    mustReApplyDecorators = true;
```
eos87: do we have to set this to true here without any conditions? it seems to go contrary to the comment in line 120 about performance.

eos87 (commented):
Is there any way to add some test here?

eos87:
One more thing that came to my mind... are we sure these combos do not clash with any built-ins from operating systems? or is that not important?

eos87 (approved):
It looks good to me.

scripts/core/editor3/components/thin-spaces/ThinSpaceDecorator.tsx
```diff
@@ -0,0 +1,65 @@
+import React, {CSSProperties} from 'react';
+import {ContentBlock, ContentState} from 'draft-js';
+import {IPropsDraftDecorator} from 'core/editor3/draftjs-types';
+
+class ThinSpaceComponent extends React.Component<IPropsDraftDecorator> {
+    render() {
+        const wrapperStyle: CSSProperties = {
```
thecalcc, line 8: could you take a look at _helpers.css in UI framework? maybe we have some more that could be applied - e.g. for padding where you're doing 0 2px, that should fit the units from the design system we've got going

scripts/core/editor3/components/thin-spaces/ThinSpaceDecorator.tsx
```diff
@@ -0,0 +1,65 @@
...
+    }
+}
+
+function ThinSpaceStrategy(
+    contentBlock: ContentBlock,
+    callback: (start: number, end: number) => void,
+    contentState: ContentState,
+) {
+    const text = contentBlock.getText();
+    const thinSpaceChar = '\u2009';
```
thecalcc: maybe this could be in some global const or something since it's a specific character

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -428,6 +452,34 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
             case 'secondary-paste': // this is blocking redo on non-windows systems, should be osx specific
                 newState = EditorState.redo(editorState);
                 break;
+            case 'custom-ndash': {
+                const ndashContentState = Modifier.replaceText(
+                    editorState.getCurrentContent(),
+                    editorState.getSelection(),
+                    '\u2013',
```
thecalcc: could be put in a constant

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -428,6 +452,34 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
...
+                    ndashContentState,
+                    'insert-characters',
+                );
+                break;
+            }
+            case 'custom-thin-space': {
+                const thinSpaceContentState = Modifier.replaceText(
+                    editorState.getCurrentContent(),
+                    editorState.getSelection(),
+                    '\u2009',
```
thecalcc: same here

scripts/core/editor3/reducers/editor3.tsx
```diff
@@ -142,9 +142,10 @@ export function updateDecorators(
         },
         limitConfig: stateCurrent.limitConfig,
         softLimitConfig: stateCurrent.softLimitConfig,
+        invisibles: stateCurrent?.invisibles,
     });
 
-    if (result.mustReApplyDecorators !== true) {
+    if (result.mustReApplyDecorators !== true && force !== true) {
```
thecalcc: why this change? we used to have it working for all cases, not sure what changed. this is core editor functionality and we have to be very careful with changes i think

scripts/core/editor3/store/index.ts
```diff
@@ -110,16 +111,22 @@ interface IOptions {
...
-    const {limitConfig, softLimitConfig, spellchecker} = options;
+    const {limitConfig, softLimitConfig, spellchecker, invisibles} = options;
 
     // improve performance by not replacing decorators when possible.
     let mustReApplyDecorators = false;
 
     const decorators: Array<{strategy: any, component: any}> = [LinkDecorator];
 
+    if (invisibles === true) {
+        mustReApplyDecorators = true;
```
thecalcc, line 140: you're setting `mustReApplyDecorators` to `true` here so in what case does it not work to apply decorators without having `force !== true` in `scripts/core/editor3/reducers/editor3.tsx`?

thecalcc (commented):
looks good, just some small comments from my side

scripts/core/editor3/reducers/editor3.tsx
BrianMwangi21: I added this so that “forced” updates like toggling formatting marks on/off still rebuilds the decorator. Without it, the new thin-space decorator never gets removed so the ¤ remains in the editor when you toggle off.

scripts/core/editor3/store/index.ts
BrianMwangi21: Then for this, setting it to true when invisibles is on guarantees we rebuild the decorator when formatting marks turn on. When toggling them off though, `mustReApplyDecorators` flips back to false even though we still have to remove the thin-space decorator. That’s where the force = true from the previous answer comes in. It tells updateDecorators reapply again. So the two work together if that makes sense. 

But if you see better way to handle it you can let me know. Cause only when toggling off the formatting marks did I notice that bug.

scripts/core/editor3/reducers/editor3.tsx
thecalcc: and could we instead then do `mustReApplyDecorators = true` when we're toggling these formatting marks and keep the flow simple?

scripts/core/editor3/store/index.ts
thecalcc: ok I see, it does make sense then to have it given mustReApplyDecorators = false when toggling off, I'm just if that's intended behavior or something, because if there's e.g. some function like `toggleDecorators` it shouldn't set `mustReApplyDecorators` to `false` when toggling them off

thecalcc (approved):

PR #5012, opened 2025-11-27
[SDESK-7522] Add test-ids to Contact form

MarkLark86:
### Purpose
Add `data-test-id` to Contact form rows, to make it easier to select in E2E tests.

### What has changed:
* Form.Row component: Convert PropTypes to typescript interface
* Form.Row component: Add `data-test-id` prop
* Add `data-test-id` to form rows in the Contact form

### DIFF
--- a/scripts/core/ui/components/Form/Row.tsx
+++ b/scripts/core/ui/components/Form/Row.tsx
@@ -1,23 +1,40 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
 import './style.scss';
+
+interface IProps {
+    flex?: boolean;
+    noPadding?: boolean;
+    className?: string | {[classname: string]: boolean};
+    halfWidth?: boolean;
+    enabled?: boolean;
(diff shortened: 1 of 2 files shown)


scripts/core/ui/components/Form/Row.tsx
```diff
@@ -1,23 +1,40 @@
...
+export const Row: React.FC<IProps> = ({
+    children,
+    flex,
+    noPadding,
+    halfWidth,
+    className,
+    enabled,
+    'data-test-id': testId,
+}) => (
+    enabled === false ? null : (
```
MarkLark86, line 29: @thecalcc In the original implementation, `enabled` defaulted to `true`. To keep the changes small without improving this (like using a `disabled` prop instead), I am just testing is false for now until we re-write these forms.

thecalcc (approved):

scripts/core/ui/components/Form/Row.tsx
thecalcc: gotcha, cool

PR #5020, opened 2025-12-05
[SDCP-1031] - Image uploaded to body_html or feature media does not remove the upload dialogue at the end

BrianMwangi21:
### Purpose
This PR adds an extra guard when `getEditorFieldCharactersCount` is called. When a number was passed upstream from a custom iptc extension it would cause an error during the image upload flow. Such a case is this [PR](https://github.com/superdesk/superdesk-cp/pull/527)

### How to test
1. Upload media where slugline is mapped to a number from `object_name` like the image attached below.
2. It should save without issues or errors on the console. It shouldn't stay stuck on the Upload dialogue.

Sample photo to use 
picture

Issue - [SDCP-1031](https://sofab.atlassian.net/browse/SDCP-1031)

[SDCP-1031]: https://sofab.atlassian.net/browse/SDCP-1031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/apps/authoring/authoring/components/CharacterCount.tsx
+++ b/scripts/apps/authoring/authoring/components/CharacterCount.tsx
@@ -10,8 +10,9 @@ interface IProps {
     item: string;
 }
 
-export function getEditorFieldCharactersCount(text: string, html?: boolean) {
-    let input = text?.trim() ?? '';
+export function getEditorFieldCharactersCount(text: string | number, html?: boolean) {
+    const normalized = text == null ? '' : String(text);
+    let input = normalized.trim();
 
     input = html ? cleanHtml(input) : input;
     input = input.replace(/\r?\n|\r/g, '');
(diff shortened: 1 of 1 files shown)


scripts/apps/authoring/authoring/components/CharacterCount.tsx
```diff
@@ -10,8 +10,9 @@ interface IProps {
     item: string;
 }
 
-export function getEditorFieldCharactersCount(text: string, html?: boolean) {
-    let input = text?.trim() ?? '';
+export function getEditorFieldCharactersCount(text: string | number, html?: boolean) {
+    const normalized = text == null ? '' : String(text);
+    let input = normalized.trim();
```
eos87: could probably simplify it with something like:
```suggestion
    const input = `${text ?? ''}`.trim();
```

eos87 (approved):

PR #5021, opened 2025-12-08
[SDESK-7830] - CI Improvements to E2E tests

eos87:
### Purpose
Speed up and streamline our E2E pipelines by introducing Playwright test `sharding`, centralising setup steps and improving how reports are generated and merged in CI.

### What has changed
- Refactored Playwright workflow: introduced test sharding (2 shards), switched CI reporter to blob, and added blob-report artifact uploads.
- Added a new setup-e2e composite action to centralize dependency installation, client build, spec compilation, and server startup for all E2E jobs.
- Added a new merge-playwright-reports job to collect shard outputs, merge them, and generate a single HTML report.
- Updated Protractor workflow to use the new shared setup action and cleaned up artifact naming and logging.

> [!IMPORTANT]
> It's worth mentioning that I've also tried having a unified build step and then reusing the same build for both protractor and playwright but that actually took longer as both runners need to wait for the build before starting, while with this approach they start running in parallel which makes it faster. 
This makes the E2E tests to run up to 15 mins faster. See screenshots below. 

### Before
image

image

### After
image

 [SDESK-7830]

[SDESK-7830]: https://sofab.atlassian.net/browse/SDESK-7830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,95 @@ jobs:
       env:
         TZ: "Europe/Prague"
 
-  e2e:
+  playwright-e2e:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: ['22']
+        shardIndex: [1, 2, 3, 4]
+        shardTotal: [4]
+    defaults:
+      run:
+        working-directory: e2e/client
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-node@v6
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'npm'
+
+    - run: npm ci
+      working-directory: ./
+
+    - run: npm ci
+      working-directory: ./build-tools
+
+    - run: npm ci
+      working-directory: ./end-to-end-testing-helpers
+
+    - run: npm install
+    - run: npm run build
+    - run: npm run specs--compile
+    - run: npm run start-client-server
+
+    - name: Start e2e server
+      run: |
+        docker compose pull
+        docker compose build
+        docker compose up -d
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -43,40 +43,78 @@ jobs:
...
+      run:
+        working-directory: e2e/client
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-node@v6
       with:
-        name: playwright-report-${{ matrix.node-version }}-${{ matrix.suite }}
-        path: e2e/client/test-results
-    # playwright end
+        node-version: '22'
```
thecalcc, line 79: could this be as specific as in the package.json volta pin? Down to minor?

.github/workflows/tests.yml
```diff
@@ -43,40 +43,78 @@ jobs:
...
+        name: playwright-html-report
+        path: e2e/client/playwright-report
+        retention-days: 14
+
+  protractor-e2e:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: ['22']
```
thecalcc, line 103: same here

thecalcc (approved):

.github/workflows/tests.yml
eos87: could be but this picks the latest minor version in `v22` which would give us some head start advantage in case of something broken. So I'd leave it as how it is right now.

PR #5025, opened 2025-12-18
Update ui framework

thecalcc:
Fixes usage of ui framework Modal component in planning

[SDBELGA-1011]

[SDBELGA-1011]: https://sofab.atlassian.net/browse/SDBELGA-1011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/client/specs/highlights_spec.ts
+++ b/e2e/client/specs/highlights_spec.ts
@@ -181,21 +181,21 @@ describe('highlights', () => {
             workspace.selectDesk('Politic Desk');
             monitoring.selectItem(1, 3);
             monitoring.selectItem(2, 2);
-            highlights.multiMarkHighlight('Highlight All Desks (Global)');
-            monitoring.checkMarkedForHighlight('Highlight All Desks (Global)', 1, 3);
-            monitoring.checkMarkedForHighlight('Highlight All Desks (Global)', 2, 2);
+            highlights.multiMarkHighlight('Highlight All Desks');
+            monitoring.checkMarkedForHighlight('Highlight All Desks', 1, 3);
+            monitoring.checkMarkedForHighlight('Highlight All Desks', 2, 2);
 
             // multi mark for highlights, in case of partial mark for selected items
             monitoring.selectItem(1, 0);
             monitoring.selectItem(2, 2);
-            highlights.multiMarkHighlight('Highlight All Desks (Global)');
-            monitoring.checkMarkedForHighlight('Highlight All Desks (Global)', 2, 2);
+            highlights.multiMarkHighlight('Highlight All Desks');
+            monitoring.checkMarkedForHighlight('Highlight All Desks', 2, 2);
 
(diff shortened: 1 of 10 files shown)


e2e/client/specs/highlights_spec.ts
```diff
@@ -181,30 +181,30 @@ describe('highlights', () => {
             workspace.selectDesk('Politic Desk');
             monitoring.selectItem(1, 3);
             monitoring.selectItem(2, 2);
-            highlights.multiMarkHighlight('Highlight All Desks');
-            monitoring.checkMarkedForHighlight('Highlight All Desks', 1, 3);
-            monitoring.checkMarkedForHighlight('Highlight All Desks', 2, 2);
+            highlights.multiMarkHighlight('Highlight All Desks (Global)');
+            monitoring.checkMarkedForHighlight('Highlight All Desks (Global)', 1, 3);
+            monitoring.checkMarkedForHighlight('Highlight All Desks (Global)', 2, 2);
```
eos87: Where is this `(Global)` change coming from?

eos87 (changes requested):
It seems like the only relevant change from UI framework is https://github.com/superdesk/superdesk-ui-framework/pull/1002. It's still not clear to me why is the `appendTo="document.body"` change needed but it seems to be the reason for the tests to be failing. Let's circle back to this one next week to try to figure out a simple way to tackle it as the appendTo body might produce more issues in the future if we backport this change into newer releases.

e2e/client/specs/highlights_spec.ts
thecalcc: the append to document body change I think was needed because of planning - when using a UI framework modal there we need it in order for the modal to show up properly IIRC. 

As for why the tests are failing the way they are - I'm not sure yet, I have to continue digging.

thecalcc:
@eos87 this is ready to be merged if you don't have other feedback. Tests are failing now due to the metadata extraction library CP has implemented for media upload. They did update it in [v1.0.6](https://github.com/6over3/exiftool/releases/tag/v1.0.6) but if we upgrade to that version or a latter one here, the build breaks. Petr told me CP will take care of the fix at some point. I've disabled them for now so we can have the CI in green if everything else is fine

eos87:
> @eos87 this is ready to be merged if you don't have other feedback. Tests are failing now due to the metadata extraction library CP has implemented for media upload. They did update it in [v1.0.6](https://github.com/6over3/exiftool/releases/tag/v1.0.6) but if we upgrade to that version or a latter one here, the build breaks. Petr told me CP will take care of the fix at some point. I've disabled them for now so we can have the CI in green if everything else is fine

E2E tests seem to be failing for a different reason. Please take a look.

petrjasek (approved):

PR #5027, opened 2025-12-19
Release/translation v3 archive v2

txl-cp:
add "enter" key event listener to search widget extensions
run reapply decorators when links of editor3 change
fix xml appearing in metadata when parsed

### DIFF
--- a/scripts/apps/search/components/search-panel-widgets.tsx
+++ b/scripts/apps/search/components/search-panel-widgets.tsx
@@ -24,10 +24,10 @@ export class SearchPanelWidgets extends React.PureComponent<IProps> {
         return (
             <React.Fragment>
                 {widgets.map((Widget, index) => (
-                    <div 
-                        key={index} 
+                    <div
+                        key={index}
                         onKeyDown={(event) => {
-                            if (event.key === "Enter") this.props.searchParameters();
+                            if (event.key === 'Enter') this.props.searchParameters();
                         }}
                     >
                         <Widget
--- a/scripts/apps/search/components/search-panel-widgets.tsx
+++ b/scripts/apps/search/components/search-panel-widgets.tsx
@@ -6,6 +6,7 @@ interface IProps {
     provider: any;
     params: any;
     setParams: (updates: any) => void;
+    searchParameters: () => void;
 }
 
 export class SearchPanelWidgets extends React.PureComponent<IProps> {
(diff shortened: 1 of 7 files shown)


scripts/core/editor3/store/index.ts
```diff
@@ -133,6 +139,8 @@ export const getDecorators = (options: IOptions) => {
         );
     }
 
+    if (isLinksChanged(editorState)) mustReApplyDecorators = true;
```
thecalcc: nitpick - we don't write code in this style, might be better to stick to the codebase style

```js
if (isLinksChanged(editorState)) {
     mustReApplyDecorators = true;
}
```

scripts/core/editor3/store/index.ts
```diff
@@ -377,3 +385,34 @@ export function syncAssociations(item: IArticle, rawState: RawDraftContentState)
...
+            },
+            () => {
+                count += 1;
+            },
+        );
+    });
+    return count;
+};
+
+const isLinksChanged = (editorState: IOptions['editorState']) => {
```
thecalcc: could this have a better name like `hasLinkCountChanged` or something along those lines

scripts/core/editor3/store/index.ts
```diff
@@ -377,3 +385,34 @@ export function syncAssociations(item: IArticle, rawState: RawDraftContentState)
...
+    const {current, next} = editorState || {};
```
thecalcc: i suggest you to use the stricter `??` which would fallback to the empty object if `editorState` is `null` or `undefined` and it would not if for some odd reason its value is falsy

thecalcc (commented):
looks good, just some miner code style comments, also take a look at copilot comments about event propagation

scripts/apps/search/components/search-panel-widgets.tsx
```diff
@@ -23,12 +24,18 @@ export class SearchPanelWidgets extends React.PureComponent<IProps> {
...
                 {widgets.map((Widget, index) => (
-                    <Widget
+                    <div
                         key={index}
-                        provider={this.props.provider}
-                        params={this.props.params}
-                        setParams={(updates) => this.props.setParams(updates)}
-                    />
+                        onKeyDown={(event) => {
+                            if (event.key === 'Enter') this.props.searchParameters();
```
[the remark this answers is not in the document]
txl-cp: this is intended. it needs to bubble up in order to trigger form submission. without it will have stale renders

thecalcc (approved):

PR #5028, opened 2025-12-29
[STT-1530] - Superdesk writing mode: In text link seems empty but is not

BrianMwangi21:
### Purpose
Links created in table cells had different data structure `{url: "...", href: "..."}` instead of correct `{link: {href: "..."}}`. This caused edit link input to be empty, even though links worked when hovering.

### What has changed
Added fallback logic to handle legacy `{url}` format if it exists in other articles and fixed table cell link creation.

### How to test
1. Create an article with a link and then malform the data in the entity map like the prod example [here](https://sofab.atlassian.net/browse/IN-7438?focusedCommentId=147351). As you can see, the structure for entity 0 is 
```
data: {
  href: "",
  title: "",
  url: ""
}
``` 
and not like entity 1
```
data: {
  link: {
    href: ""
  }
}
```
2. Without the new fallbacks, when you edit the link, it doesn't populate the edit field. But with the new fallback and changes to the TableCell link creation, it shouldn't reoccur.
3. The fallbacks have been added incase there exists more articles that have the old format

### DIFF
--- a/scripts/core/editor3/components/links/LinkInput.tsx
+++ b/scripts/core/editor3/components/links/LinkInput.tsx
@@ -35,6 +35,7 @@ const linkTypes = {
 
 export interface ILink {
     href: string;
+    url?: string;
 }
 
 interface IProps extends Partial<IEditorStore> {
--- a/scripts/core/editor3/components/links/LinkInput.tsx
+++ b/scripts/core/editor3/components/links/LinkInput.tsx
@@ -90,7 +90,7 @@ export class LinkInputComponent extends React.Component<IProps, any> {
         const initialValue = '';
 
         this.state = {
-            url: this.props.data ? this.props.data.href || initialValue : initialValue,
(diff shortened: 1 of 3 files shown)


eos87 (changes requested):
I'm not sure about this. Can we instead fix or avoid broken data coming into the editor? It makes me feel like we would end up adding more workarounds for more cases of broken data like this.

BrianMwangi21:
@eos87, the fix actually addresses both concerns. Cause the root cause isn't broken data coming into the editor from the outside. It is how the editor saves the url from the TableCell component. 

1. The `TableCell.tsx` changes now creates links with the `{link: {href}}` structure instead of the `{url}` format going forward. This format was not now being parsed properly in the LinkToolbar component now when the user goes to edit.
2. Then the fallbacks in `LinkToolbar` and `LinkInput` are for legacy articles already in prod that could have the malformed structure. 

Hope this clarifies it.

eos87 (approved):

PR #5032, opened 2026-01-08
add fail-safe binary for exiftool

txl-cp:
Currently there's cors issues with cdn see [issue](https://github.com/6over3/exiftool/issues/26)

maintainer has split off perl5 interpreter into another lib https://github.com/6over3/zeroperl-ts

tried to add the fetch polyfill for exiftool to the bundle recommended [here](https://github.com/6over3/zeroperl-ts?tab=readme-ov-file#browser-usage) but it's not working currently since the wasm version isn't the same.

added a fallback binary wasm to use for the exiftool library for now but would need to upgrade the library or fork and maintain compatibility in the future to remove this.


scripts/apps/archive/parse-metadata.ts
```diff
@@ -11,11 +12,21 @@ type RawMetadata = {
...
 
-const getMetadata = (f: File, options?: ExiftoolOptions) =>
-    (f.type.startsWith('video/')
-        ? getVideoMetadata(f, options)
-        : getPictureMetadata(f, options)
-    ).then(processMetadata, (): Partial<IPTCMetadata> => ({}));
+/**
+ * used as a fail-safe if fetch from exiftool fails
+ */
+export const fetchZeroperl = () => fetch(zeroperl).then((res) => res);
```
petrjasek: wouldn't this work also in tests?

txl-cp: fetch api is mocked globally so fetching the import will timeout

txl-cp: set up fetch for the suite to simplify tests

petrjasek:
hi @taylorli2000 I can't make the tests work locally, does it need some specific setup?

txl-cp:
> hi @taylorli2000 I can't make the tests work locally, does it need some specific setup?

I think i had some issue like this before for something else, but i reinstall and re-link and everything works fine after.

petrjasek:
yep did another `npm ci` and seems working now.
can you pls merge https://github.com/canadianpress/superdesk-client-core/pull/18 ?

petrjasek (approved):

PR #5035, opened 2026-01-14
STT-1509 - As an STT journalist I´d like to know the status of the article I am working on

BrianMwangi21:
### Purpose
This PR adds the same status pill found in the monitoring list view to the editor header so journalists can see the article status as they work.

### What has changed
- Created `StatusInfo` component to display status label in authoring topbar with the same styling as the list view
- Added status widget with priority 0 to the authoring topbar to display before Created/Modified info as per the design in the ticket
- **Bug fix**: Corrected `shouldComponentUpdate()` to properly detect new state keys. This fixed an issue where metadata wasn't displaying for published/archived articles in view mode
- Changed component from `PureComponent` to `Component` to support custom `shouldComponentUpdate()` logic which resolved React console warnings on the same.

### Screenshots

Before : 

After: 

And also after the bug fix, published items also show the metadata too: 

Ticket: [STT-1509](https://sofab.atlassian.net/browse/STT-1509)

[STT-1509]: https://sofab.atlassian.net/browse/STT-1509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/apps/authoring/authoring/authoring-topbar2-react.tsx
+++ b/scripts/apps/authoring/authoring/authoring-topbar2-react.tsx
@@ -7,9 +7,19 @@ import {extensions} from 'appConfig';
 import {dataApi} from 'core/helpers/CrudManager';
 import {CreatedInfo} from './created-info';
 import {ModifiedInfo} from './modified-info';
+import {StatusInfo} from './status-info';
 import {AuthoringToolbar} from 'apps/authoring-react/subcomponents/authoring-toolbar';
 
 const getDefaultToolbarItems = (item: IArticle): Array<ITopBarWidget<IArticle>> => [{
+    availableOffline: true,
+    component: () => (
+        <StatusInfo
+            entity={item}
+        />
+    ),
+    group: 'start',
+    priority: 0,
+}, {
     availableOffline: true,
     component: () => (
         <CreatedInfo
@@ -42,7 +52,7 @@ interface IState {
 /**
  * Only used from angular based authoring view
  */
-export class AuthoringTopbar2React extends React.PureComponent<IProps, IState> {
+export class AuthoringTopbar2React extends React.Component<IProps, IState> {
     constructor(props: IProps) {
         super(props);
 
@@ -72,6 +82,7 @@ export class AuthoringTopbar2React extends React.PureComponent<IProps, IState> {
             this.fetchArticleFromServer();
         }
     }
+
     shouldComponentUpdate(nextProps: IProps, nextState: IState) {
         for (const key of Object.keys(this.props)) {
             if (key === 'children') continue;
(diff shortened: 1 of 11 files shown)


scripts/apps/authoring/authoring/status-info.tsx
```diff
@@ -0,0 +1,33 @@
...
+export class StatusInfo extends React.PureComponent<IProps> {
+    render() {
+        const {entity} = this.props;
+
+        if (entity.state == null) {
+            return null;
+        }
+
+        const label = getStateLabel(entity.state);
+        const cssClass = entity.state === 'correction'
```
thecalcc: so the label component from UI framework won't work?

scripts/apps/authoring/authoring/authoring-topbar2-react.tsx
```diff
@@ -7,9 +7,19 @@ import {extensions} from 'appConfig';
 import {dataApi} from 'core/helpers/CrudManager';
 import {CreatedInfo} from './created-info';
 import {ModifiedInfo} from './modified-info';
+import {StatusInfo} from './status-info';
 import {AuthoringToolbar} from 'apps/authoring-react/subcomponents/authoring-toolbar';
 
 const getDefaultToolbarItems = (item: IArticle): Array<ITopBarWidget<IArticle>> => [{
+    availableOffline: true,
```
thecalcc, line 14: do we want this one configurable or for all users? also this toolbar component is used for authoring angular? if not, then for authoring angular version this status topbar widget won't show up

scripts/apps/authoring/authoring/status-info.tsx
BrianMwangi21: After cross-checking, the label component won't work. Since it supports one style per state while each article state has its own unique css styling. And it keeps the editor label in sync with the list view automatically as requested.

scripts/apps/authoring/authoring/authoring-topbar2-react.tsx
BrianMwangi21: I've asked Gideon whether it will be configurable or for all users. I'm leaning more on all users since it just a label and it's a nice to have. 

On the authoring angular version, I have added the status info too locally (see screenshot) and now it shows up on the upper top bar where the save button is but according to the design on the ticked, it shouldn't be there. It should be on the second topbar where the created/modified info is.

thecalcc: ok then we move it so it's in the second bar?

thecalcc (approved):

scripts/apps/authoring/authoring/status-info.tsx
eos87: If I'm not mistaken, this status label would be used in different places (you could double check with Fritz). Therefore, I'd say we use it from the UI Framework or implement it there. Ideally, the colors are already defined there so we could have a status label component that would take the color as parameter (string name) and perhaps a variant (I'd say "outlined" or "filled/solid"). We need to start implementing this components properly. 
@BrianMwangi21 let me know if you wanna talk more about it, we should involve Fritz too.

eos87 (changes requested):

thecalcc:
@eos87 I already coordinated with Brian, he'll ping Fritz and they'll see what makes sense

eos87 (approved):

PR #5036, opened 2026-01-14
[SDESK-7830] Upgrade project to web pack 5 and related dependencies and configs

eos87:
### Purpose
This pull request upgrades the project to Webpack `v5` and updates related dependencies and configurations for compatibility. It introduces the `node-polyfill-webpack-plugin` to restore Node.js core module polyfills for browser builds, updates asset handling, and adds support for new dependencies such as `react-resizable-panels`. 

### What has changed
**Build System & Dependency Upgrades:**
* Upgraded `webpack`, `webpack-cli`, and `webpack-dev-server` to version 5, and updated `karma-webpack` to v5 for compatibility with the new Webpack version. Also updated `sass` and `superdesk-ui-framework` dependencies.
* Added `node-polyfill-webpack-plugin` to restore Node.js polyfills (such as `Buffer`, `process`, and `stream`) for legacy code, excluding the `console` polyfill. 

**Webpack Configuration Adjustments:**
* Updated the `entry` configuration to use string paths instead of arrays for the `init` and `app` entries, aligning with Webpack 5 best practices.
* Improved module resolution by explicitly adding `node_modules` to the `resolve.modules` array and specifying `mainFields` order for better package resolution.

**Asset Handling Improvements:**
* Replaced deprecated `file-loader` usage with Webpack 5’s built-in `asset/resource` type for handling images, fonts, and WASM files.
* Added a rule to treat `.cjs` files as `javascript/auto` to support packages like `@uswriting/exiftool` that contain embedded Perl code, preventing parsing errors.

**Karma Test Runner Enhancements:**
* Updated `karma.conf.js` to serve static assets (WASM, images, videos) from the `dist/` directory, addressing changes in how `karma-webpack` v5 serves files.
* Set `webpackConfig.entry` to an empty object instead of `null` for compatibility with Webpack 5.

Belongs to: [SDESK-7830]

[SDESK-7830]: https://sofab.atlassian.net/browse/SDESK-7830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webpack.config.js
```diff
@@ -166,15 +167,23 @@ module.exports = function makeConfig(grunt) {
...
+            }),
         ],
 
         resolve: {
             modules: [
                 __dirname,
                 path.join(__dirname, 'scripts'),
                 path.join(__dirname, 'styles', 'sass'),
+                path.join(__dirname, 'node_modules'),
                 'node_modules',
```
thecalcc, line 184: i guess we then don't need the last entry `node_modules` since we've got this one?

webpack.config.js
```diff
@@ -284,12 +294,17 @@ module.exports = function makeConfig(grunt) {
                 },
                 {
                     test: /\.(png|gif|jpeg|jpg|woff|woff2|eot|ttf|svg|mov)(\?.*$|$)/,
-                    loader: 'file-loader',
+                    type: 'asset/resource',
+                },
+                {
```
thecalcc, line 299: does this also fix the automatic metadata extraction when uploading an image? if so we could then re-enable the metadata tests

thecalcc (approved):

webpack.config.js
eos87: ~yeah you're right. I'll remove it.~ Turns out we need it. They serve different purposes. First one is absolute and the second one relative affecting dependencies node_modules folders.

webpack.config.js
eos87: This is part of the fix for those metadata tests but I'm not sure if they are the ones you're referring to. Could you point me to those?

PR #5044, opened 2026-01-21
[STT-1541] - Unsaved changes are lost when exiting Multi-edit mode

BrianMwangi21:
### Purpose
This PR implements and unsaved changes detection and prompting multi-edit on both react and angular paths.

### What has changed
- Added unsaved changed guard to modal close on both angular and react multi-edit paths

### How to test
1. Open any stories in Multi-edit mode    
2. Make one or more changes
3. Exit Multi-edit mode without manually saving. This should show the confirmation dialog modal now.
4. Enable React authoring using `localStorage.setItem('auth-react', 'true')` and reload to test react flow.


scripts/apps/authoring-react/multi-edit-modal.tsx
```diff
@@ -37,6 +41,84 @@ export class MultiEditModal extends React.PureComponent<IProps, IState> {
         this.componentRefs = {};
     }
 
+    private checkUnsavedChanges(): boolean {
```
thecalcc: could we use a better name here since this is checking if there's some article that has unsaved changes?

scripts/apps/authoring/multiedit/multiedit.ts
```diff
@@ -25,6 +29,7 @@ function MultieditService(storage, superdesk, authoringWorkspace: AuthoringWorks
 
     var MIN_BOARDS = 2;
     var STORAGE_KEY = 'multiedit';
+    var articleContexts = {};
```
thecalcc: why not use `let` here?

scripts/apps/authoring/multiedit/multiedit.ts
```diff
@@ -109,6 +118,28 @@ function MultieditService(storage, superdesk, authoringWorkspace: AuthoringWorks
         }
     };
 
+    this.registerArticleContext = function(articleId, context) {
+        if (articleId != null) {
+            articleContexts[articleId] = context;
+        }
+    };
+
+    this.unregisterArticleContext = function(articleId) {
```
thecalcc: could you type the articleId argument here? Also why do we call this function in the first place if `articleId` is `null`?

scripts/apps/authoring/multiedit/multiedit.ts
```diff
@@ -130,13 +161,75 @@ function MultieditController($scope, multiEdit) {
...
+
+                    return Promise.all(dropPromises).then(finalizeExit);
+                }
+
+                if (action === IUnsavedChangesActionWithSaving.save) {
+                    const savePromises = contextsWithUnsaved.map((context) => context.save());
+
+                    return Promise.all(savePromises)
+                        .then(finalizeExit)
+                        .catch(() => undefined);
```
thecalcc: why won't we show some error to the user?

scripts/apps/authoring/multiedit/multiedit.ts
```diff
@@ -130,13 +161,75 @@ function MultieditController($scope, multiEdit) {
...
+                }
+
+                closePromptFn();
+            });
+        }).catch(() => undefined);
```
thecalcc: same here, we shouldn't just silence errors

scripts/apps/authoring-react/multi-edit-modal.tsx
```diff
@@ -37,6 +41,84 @@ export class MultiEditModal extends React.PureComponent<IProps, IState> {
...
+        const hasChanges = this.state.articleIds.map((articleId) => {
+            const ref = this.componentRefs[articleId];
+
+            if (ref == null) {
```
thecalcc: why would the ref be `null`? I think if we remove an article from the multi edit view, we should also remove it from `state.articleIds`

scripts/apps/authoring/multiedit/multiedit.ts
```diff
@@ -130,13 +161,75 @@ function MultieditController($scope, multiEdit) {
     };
 
     $scope.closeMulti = function() {
-        /**
-         * HACK: for some reason the watcher doesn't always fire and multi edit view remains open.
-         * Since authoring-react/multi-edit-modal.tsx will replace this code - I'm not doing a proper fix.
-         */
-        $scope.boards = [];
+        const articleIds = multiEdit.items
```
thecalcc, line 170: same question here, I think we shouldn't have `null` `item.article`s, if there's a bug in the removal where it keeps the article as null we should fix that imo

thecalcc (commented):
it looks good, I'm mostly wondering why we'd have `null` article ids in the state in the first place. If we handle that properly so that we don't have those, code should get simpler. 

As a general thought - could we maybe make a single function that based on the authoring - whether its react or the angular one, handles this functionality? That way we could extract common parts into a single place

BrianMwangi21:
@thecalcc on the null ids, my understanding of the flow is they represent the empty slots that can be populated with articles on the fly so the null checks are more of guards to avoid any unexpected behavior like accessing properties not yet loaded

thecalcc (approved):
if you find a better place for the handler - `handleMultiItemUnsavedChanges` pls move it

thecalcc:
sorry last minute thought but this code I think this should go to release/3.3 because that's the next one for stt

PR #5045, opened 2026-01-21
Update handling of react-resizable-panels in webpack config

eos87:
This pull request makes targeted updates to the Webpack configuration to improve how certain dependencies are handled during the build process. The main focus is on optimizing module resolution and TypeScript loader behavior for pre-built ESM modules.

Dependency resolution and build optimization:

* Removed the custom path resolution for `react-resizable-panels` in the `resolve.alias` section, allowing Webpack to use its default resolution for this package.
* Updated the `ts-loader` exclusion logic to skip TypeScript processing for pre-built ESM modules, specifically `@babel/runtime` and `react-resizable-panels`, which helps prevent unnecessary transpilation and potential build issues.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -200,7 +200,6 @@ module.exports = function makeConfig(grunt) {
                  */
                 react: getModuleDir('react'),
                 'react-dom': getModuleDir('react-dom'),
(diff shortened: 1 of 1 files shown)


thecalcc (approved):

thecalcc:
@eos87 I think we need a version bump also

eos87:
> @eos87 I think we need a version bump also

I guess Petr would do it later once a release is created.

PR #5071, opened 2026-02-12
Fix multiline field

thecalcc:
### Purpose
Fix styles of generic crud form multiline field

#### Before:
image

#### After: 
image
image
image

### Checklist

- [x] I reviewed my code

Resolves: [SDESK-7267]

[SDESK-7267]: https://sofab.atlassian.net/browse/SDESK-7267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/core/ui/components/generic-form/input-types/plain-text.tsx
+++ b/scripts/core/ui/components/generic-form/input-types/plain-text.tsx
@@ -37,8 +37,8 @@ export class PlainText extends React.Component<IInputType<string>> {
                         style={{resize: 'vertical', height: 'auto'}}
                         data-test-id={`gform-input--${this.props.formField.field}`}
                     />
-                    {this.props.issues.map((str, i) => (
-                        <div key={i} className="sd-input__message">{str}</div>
+                    {this.props.issues.map((str) => (
+                        <div key={str} className="sd-input__message">{str}</div>
                     ))}
                 </div>
             );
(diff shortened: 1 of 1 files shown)


scripts/core/ui/components/generic-form/input-types/plain-text.tsx
```diff
@@ -14,32 +14,32 @@ export class PlainText extends React.Component<IInputType<string>> {
...
+                    <textarea
+                        className="sd-input__input pt-1"
+                        disabled={this.props.disabled}
+                        value={valueWithDefaultValue}
+                        onChange={(event) => this.props.onChange(event.target.value)}
+                        rows={3}
+                        style={{resize: 'vertical', height: 'auto'}}
+                        data-test-id={`gform-input--${this.props.formField.field}`}
+                    />
+                    {this.props.issues.map((str, i) => (
```
eos87: should we use something else instead of `i` as the keys for elements in this loop?

thecalcc: it's actually even better to not have that so there's not a full list re-render if just one of the items changes, but the rest stay the same. I'm pretty sure we'll get a list of unique strings in `this.props.issues` thus it's ok to each item from it as a key also.

BrianMwangi21 (approved):

PR #5073, opened 2026-02-16
[SDESK-7305] - Week start day setting not respected globally

BrianMwangi21:
### Purpose
Unify week-start configuration so client-core is the single source of truth while remaining backward compatible with backend `start_of_week`.

### What has changed

Normalized `startingDay` and `start_of_week` during app config initialization so both values stay in sync, with `startingDay` taking precedence.

### Steps to test
1. Set `startingDay: 1` in `superdesk.config.js` and set backend to `START_OF_WEEK=0`.
2. Rebuild and start the client.
3. Verify non-Planning date pickers start on Monday (and match `startingDay`).

### Checklist
- [X] I reviewed my code
- [X] I tested all affected functionality and made sure it works

Note: This PR goes hand in hand with this change in [superdesk-planning](https://github.com/superdesk/superdesk-planning/pull/2736)

Resolves: [SDESK-7305](https://sofab.atlassian.net/browse/SDESK-7305)

[SDESK-7305]: https://sofab.atlassian.net/browse/SDESK-7305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/core/superdesk-api.d.ts
+++ b/scripts/core/superdesk-api.d.ts
@@ -3649,8 +3649,8 @@ declare module 'superdesk-api' {
         };
         apps: any;
         defaultRoute: string;
-        startingDay: any;
-        start_of_week?: any;
+        startingDay?: number | string | null;
+        start_of_week?: number | string | null;
         features: {
             swimlane?: {
                 defaultNumberOfColumns: number;
--- a/scripts/core/superdesk-api.d.ts
+++ b/scripts/core/superdesk-api.d.ts
@@ -3650,6 +3650,7 @@ declare module 'superdesk-api' {
         apps: any;
         defaultRoute: string;
         startingDay: any;
+        start_of_week?: any;
         features: {
             swimlane?: {
                 defaultNumberOfColumns: number;
--- a/scripts/init.ts
+++ b/scripts/init.ts
@@ -47,6 +47,31 @@ fetchSync(
         // apply config from server
         merge(_appConfig, res.config);
 
+        // Normalize and sync week start with startingDay taking precedence.
+        const normalizeWeekStart = (value: any): number | null => {
+            if (value == null) {
+                return null;
+            }
+
+            const normalized = typeof value === 'string' ? parseInt(value, 10) : value;
+
+            return Number.isNaN(normalized) ? null : normalized;
+        };
+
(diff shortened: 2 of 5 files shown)


scripts/init.ts
```diff
@@ -47,6 +47,31 @@ fetchSync(
         // apply config from server
         merge(_appConfig, res.config);
 
+        // Normalize and sync week start with startingDay taking precedence.
+        const normalizeWeekStart = (value: any): number | null => {
```
thecalcc: should we really allow any value here? I think either string or null, or if number is more suitable number.

scripts/init.ts
```diff
@@ -47,6 +47,31 @@ fetchSync(
...
+        if (normalizedStartingDay != null) {
+            _appConfig.startingDay = normalizedStartingDay;
+            _appConfig.start_of_week = normalizedStartingDay;
+        } else if (normalizedStartOfWeek != null) {
+            _appConfig.start_of_week = normalizedStartOfWeek;
+            _appConfig.startingDay = normalizedStartOfWeek;
+        } else {
+            _appConfig.startingDay = 0;
+            _appConfig.start_of_week = 0;
+        }
```
thecalcc: This isn't wrong but it can be shorthanded to two lines using `??`

scripts/core/superdesk-api.d.ts
```diff
@@ -3650,6 +3650,7 @@ declare module 'superdesk-api' {
         apps: any;
         defaultRoute: string;
         startingDay: any;
+        start_of_week?: any;
```
thecalcc: why do we still keep the type as `any` here? I don't get it

scripts/init.ts
```diff
@@ -47,6 +47,28 @@ fetchSync(
         // apply config from server
         merge(_appConfig, res.config);
 
+        // Normalize and sync week start with startingDay taking precedence.
+        const normalizeWeekStart = (value: string | number | null | undefined): number | null => {
```
thecalcc, line 51: would it be a `string | null` or `number | null`? Why do we need both?

thecalcc (approved):

PR #5089, opened 2026-02-25
[STT-1569] Preserve custom tag styles when pasting between editors

eos87:
### Description
When copying text between two editor3 instances (including across separate browser windows), inline styles coming from custom editor tags were lost on paste. 

### What was the problem:
When pasting from another window, editor3 receives raw HTML from the browser clipboard. Custom tag styles are represented as CSS (e.g. `border-block-end: 4px double ...`) in that HTML. The parser had no way to recognize those CSS values and convert them back into the internal DraftJS inline style names, so the styles were silently dropped.

### What has changed:
- The HTML parser now recognizes pasted CSS matching any configured custom tag style and converts it back to the corresponding DraftJS inline style. Both raw CSS variable values (as Firefox preserves them) and browser-computed values (as Chrome resolves them) are handled.
- Because DraftJS's `convertFromHTML` strips inline styles, the recognized spans are first converted to [PUA](https://en.wikipedia.org/wiki/Private_Use_Areas) Unicode marker characters that survive the conversion, then the markers are stripped and the inline styles are applied to the correct character ranges in post-processing.
- The cross-window paste identity check now validates that the clipboard text hash matches the stored copy metadata, preventing stale metadata from interfering.
- Tests cover: the full CSS→inline style pipeline, `extractCustomTagRanges` boundary cases, `editor3StateToHtml` custom tag emission, and the `pasteContentFromOpenEditor` paste path including style preservation.

Resolves: [STT-1569]

[STT-1569]: https://sofab.atlassian.net/browse/STT-1569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/core/editor3/components/handlePastedText.ts
+++ b/scripts/core/editor3/components/handlePastedText.ts
@@ -25,9 +25,22 @@ function removeMediaFromHtml(htmlString): string {
 }
 
 /**
+ * Returns true when the current clipboard text was copied from an editor in this
+ * window. Checks both the content hash (guards against stale metadata from a
+ * previous copy) and that the source editor is still mounted in EDITOR_GLOBAL_REFS.
+ */
+export function isCopySourceInThisWindow(text: string): boolean {
+    const copyMetadata = window[EDITOR_COPY_METADATA];
+
+    return copyMetadata != null
+        && copyMetadata.contentHash === hashString(text)
+        && window[EDITOR_GLOBAL_REFS]?.[copyMetadata.editorKey] != null;
+}
+
+/**
  * Handles paste when the source is another Editor3 instance in the same window.
  *
- * Uses the window-scoped EDITOR_COPY_SOURCE variable set by Editor3Component's
+ * Uses the window-scoped EDITOR_COPY_METADATA variable set by Editor3Component's
  * native copy listener to identify the source editor. The content hash is checked
  * first to ensure the clipboard was not replaced after the copy event fired.
  *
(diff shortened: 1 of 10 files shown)


scripts/core/editor3/components/handlePastedText.ts
```diff
@@ -23,41 +24,80 @@ function removeMediaFromHtml(htmlString): string {
...
-            }
-        }
+    if (internalClipboard == null) {
+        return 'not-handled';
     }
 
-    return 'not-handled';
+    const blocksArray = [];
+
+    internalClipboard.forEach((b) => blocksArray.push(b));
```
BrianMwangi21: Can this be like `const blocksArray = Array.from(internalClipboard);` ?

scripts/core/editor3/html/from-html/index.ts
```diff
@@ -8,8 +8,125 @@ import {
...
+    document.body.appendChild(probe);
+
+    for (const [styleName, cssProps] of Object.entries(customEditorTagStyleMap)) {
+        for (const [propName, rawValue] of Object.entries(cssProps)) {
+            if (!map.has(propName)) {
+                map.set(propName, new Map());
+            }
+
+            const valueMap = map.get(propName);
+            const rawLower = (rawValue as string).toLowerCase();
```
BrianMwangi21, line 65: Suggestion: `const rawLower = String(rawValue).toLowerCase();`

petrjasek (approved):

scripts/core/editor3/html/from-html/index.ts
eos87: I rather keep it this way. If a value in that map is ever nullish, it's a bug in the map definition itself, not something to silently swallow at runtime.

scripts/core/editor3/components/handlePastedText.ts
eos87: done.

BrianMwangi21 (approved):

PR #5100, opened 2026-03-04
STT-1584: As STT journalist I want to add company and person tags with keyboard shortcuts

BrianMwangi21:
### Purpose
This PR adds shortcuts for the Company and Person tags in the editor. 

### What has changed
- Add Ctrl+3 shortcut for Company tags (EDITOR_TAG_company)
- Add Ctrl+6 shortcut for Person tags (EDITOR_TAG_person)
- Update tooltips to show shortcuts: 'Yritys (Ctrl+3)' and 'Henkilö (Ctrl+6)'

### Steps to test
1. Open the editor for a content profile whose body html has company and person tags enabled
2. Start typing and hit Ctrl+3 to enable Company tags or Ctrl+6 to enable Person tags
3. Observe that the icons also are enabled when clicked
4. Observe that also on hovering, the tooltip on the icons also show the shortcuts

### Checklist
- [x] I reviewed my code

Resolves: [STT-1584](https://sofab.atlassian.net/browse/STT-1584)

[STT-1584]: https://sofab.atlassian.net/browse/STT-1584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/core/editor3/components/Editor3Component.tsx
+++ b/scripts/core/editor3/components/Editor3Component.tsx
@@ -416,6 +416,34 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
             onCreateChangeStyleSuggestion,
         } = this.props;
 
+        const canEditInSuggestingMode = (editorState, author) =>
+            Suggestions.allowEditSuggestionOnLeft(editorState, author) ||
+            Suggestions.allowEditSuggestionOnRight(editorState, author);
+
+        const toggleInlineStyleOrSuggest = (
+            editorState,
+            style: string,
+            suggestingMode: boolean,
+            author,
+            onCreateChangeStyleSuggestion: (style: string, active: boolean) => void,
+        ) => {
+            const active = editorState.getCurrentInlineStyle().has(style);
+
+            if (suggestingMode) {
+                if (!canEditInSuggestingMode(editorState, author)) {
+                    return {handled: true as const};
+                }
+
+                onCreateChangeStyleSuggestion(style, active);
+                return {handled: true as const};
+            }
+
+            return {
+                handled: false as const,
+                newState: RichUtils.toggleInlineStyle(editorState, style),
+            };
+        };
+
         if (singleLine && command === 'split-block') {
             return 'handled';
         }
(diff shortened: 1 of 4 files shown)


scripts/core/editor3/components/toolbar/StyleButton.tsx
```diff
@@ -66,7 +66,20 @@ export default class StyleButton extends React.Component<IPropsStyleButton> {
...
+                // Add keyboard shortcuts to tooltips for company and person tags
+                let tooltip = label;
+
+                if (editor3Style === 'EDITOR_TAG_company') {
+                    tooltip = `${label} (Ctrl+3)`;
+                } else if (editor3Style === 'EDITOR_TAG_person') {
+                    tooltip = `${label} (Ctrl+6)`;
+                }
+
+                return [editor3Style, tooltip];
```
eos87: This should come from the configuration of the custom tag/button instead of hardcoding it here. So we could define any tooltip from the superdesk.config.js

eos87: perhaps, even the keyboard shortcut could be configurable? Not sure if too tricky, but worth considering it because I feel this is not going to be the first time we have a request to have a custom style button with custom shortcut. Thoughts?

scripts/core/editor3/components/Editor3Component.tsx
```diff
@@ -488,6 +504,40 @@ export class Editor3Component extends React.Component<IPropsEditor3Component, IS
...
+                        && !Suggestions.allowEditSuggestionOnRight(editorState, author)) {
+                        return 'handled';
+                    }
+                    onCreateChangeStyleSuggestion(style, active);
+                    return 'handled';
+                }
+
+                newState = RichUtils.toggleInlineStyle(editorState, style);
+                break;
+            }
```
eos87, line 564: there is significant duplication in these two cases. I quickly asked Copilot for a suggestion and came with this:

Extract the logic into a separate helper
```tsx
const canEditInSuggestingMode = (editorState, author) =>
  Suggestions.allowEditSuggestionOnLeft(editorState, author) ||
  Suggestions.allowEditSuggestionOnRight(editorState, author);

const toggleInlineStyleOrSuggest = (
  editorState,
  style: string,
  suggestingMode: boolean,
  author,
  onCreateChangeStyleSuggestion: (style: string, active: boolean) => void,
) => {
  const active = editorState.getCurrentInlineStyle().has(style);

  if (suggestingMode) {
    if (!canEditInSuggestingMode(editorState, author)) {
      return {handled: true as const};
    }

    onCreateChangeStyleSuggestion(style, active);
    return {handled: true as const};
  }

  return {
    handled: false as const,
    newState: RichUtils.toggleInlineStyle(editorState, style),
  };
};
```

Then your switch cases become minimal:

```tsx
case 'toggle-company-tag': {
  const res = toggleInlineStyleOrSuggest(
    editorState,
    'EDITOR_TAG_company',
    suggestingMode,
    author,
    onCreateChangeStyleSuggestion,
  );

  if (res.handled) return 'handled';
  newState = res.newState;
  break;
}

case 'toggle-person-tag': {
  const res = toggleInlineStyleOrSuggest(
    editorState,
    'EDITOR_TAG_person',
    suggestingMode,
    author,
    onCreateChangeStyleSuggestion,
  );

  if (res.handled) return 'handled';
  newState = res.newState;
  break;
}
```

Looks pretty good. What do you think?

eos87 (changes requested):

scripts/core/editor3/components/toolbar/StyleButton.tsx
BrianMwangi21: Fair point and great idea! 

I don't think it would be tricky but I'd recommend a separate ticket since we'd need to refactor the entire keybinding system and not just the 2 added here. Cause there are a couple in the file. So its a bigger scope and would require proper time I think. Plus proper testing too.

eos87: yeah make sense. The changes look good, I would just like to test them tomorrow.

eos87 (approved):
Unfortunately I didn't have time to test it. Code-wise looks good!

PR #5108, opened 2026-03-11
Revert "Fix authoring react vocabulary field moving on a different row if selected item is long"

eos87:
Reverts superdesk/superdesk-client-core#5065


eos87:
Reverting this one because of STT planning going blank.

petrjasek (approved):

PR #5109, opened 2026-03-12
use picture profile in media metadata editor modal

petrjasek:
as a fallback for backward compatibility

### Checklist

- [ ] I reviewed my code
- [ ] I tested all affected functionality and made sure it works

Resolves: [issue-number]


eos87 (approved):
Looks good

PR #5113, opened 2026-03-20
refactor media editor XSS fix

petrjasek:
sanitize it before rendering instead of when typing, so it doesn't encode entities while typing

[STT-1527]

[STT-1527]: https://sofab.atlassian.net/browse/STT-1527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/apps/archive/directives/HtmlPreview.ts
+++ b/scripts/apps/archive/directives/HtmlPreview.ts
@@ -3,6 +3,7 @@ import {getAnnotationsFromItem} from 'core/editor3/helpers/editor3CustomData';
 import {META_FIELD_NAME} from 'core/editor3/helpers/fieldsMeta';
 import ng from 'core/services/ng';
 import {gettext} from 'core/utils';
+import {sanitizeHtmlContent} from 'core/helpers/sanitize-html-input';
 import {IArticle} from 'superdesk-api';
 
 function getAnnotationTypesAsync(scope) {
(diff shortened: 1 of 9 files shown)


scripts/core/helpers/sanitize-html-input.ts
```diff
@@ -1,8 +1,5 @@
 import DOMPurify from 'dompurify';
 
-export function sanitizeHtmlInput(value: string): string {
-    return DOMPurify.sanitize(value || '', {
-        FORBID_TAGS: ['script', 'style', 'iframe', 'object', 'embed', 'base', 'form', 'meta', 'link'],
-        FORBID_ATTR: ['srcdoc'],
-    });
```
eos87, line 7: what is the reason to remove the forbidden tags and attributes?

scripts/apps/archive/directives/HtmlPreview.ts
```diff
@@ -57,13 +58,19 @@ export function HtmlPreview($sce, $timeout) {
...
         },
         templateUrl: 'scripts/apps/archive/views/html-preview.html',
         link: function(scope, elem, attrs) {
-            scope.$watch('sdHtmlPreview', (html) => {
-                scope.html = $sce.trustAsHtml(adjustHTMLForPreview(html));
+            scope.$watch('sdHtmlPreview', (html: string) => {
+                const isBodyHtml = scope.type === 'body-html';
+                const safe = isBodyHtml
+                    ? adjustHTMLForPreview(html)
+                    : sanitizeHtmlContent(html);
```
eos87, line 69: shouldn't we sanitize the content always? what's the difference here? 

Also, thinking that this will prevent rendering custom tags, like the case of STT company and person, if the condition gets to false.

petrjasek: we need to keep the scripts there when rendering embeds

scripts/core/helpers/sanitize-html-input.ts
petrjasek: it should be safe default, this initial list was generated, not sure what's more secure

eos87 (approved):

PR #5133, opened 2026-04-01
[SDANSA-576] Show seconds in publish schedule when configured

BrianMwangi21:
### Purpose
This PR now allows the publish schedule date/time picker to show seconds when the configured view time format includes seconds.

### What has changed
- Updated the publish schedule and embargo DateTimePicker components to pass allowSeconds only when `appConfig.view.timeformat.includes('ss')`.

### Steps to test
1. Set `appConfig.view.timeformat` to `HH:mm:ss` like the way it is set for ANSA
2. Open the article publish schedule panel.
3. Verify the time picker shows seconds.
4. Change `appConfig.view.timeformat` to `HH:mm`.
5. Reload the app and verify the seconds field is no longer shown.

### Checklist

- [x] I reviewed my code

Resolves: [SDANSA-576](https://sofab.atlassian.net/browse/SDANSA-576)

[SDANSA-576]: https://sofab.atlassian.net/browse/SDANSA-576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/core/interactive-article-actions-panel/subcomponents/publishing-date-options.tsx
+++ b/scripts/core/interactive-article-actions-panel/subcomponents/publishing-date-options.tsx
@@ -127,6 +127,7 @@ export class PublishingDateOptions extends React.PureComponent<IProps> {
                         <DateTimePicker
                             value={embargo}
                             valueType="date"
+                            allowSeconds={appConfig.view.timeformat.includes('ss')}
                             locale={{
                                 type: 'full',
                                 payload: getLocaleForDatePicker(),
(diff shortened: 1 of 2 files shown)


petrjasek (commented):
would rather add an explicit config for that, they might change `viewFormat` but would still expect this to work

BrianMwangi21:
Makes sense. Let me do that

petrjasek (approved):

PR #5199, opened 2026-06-01
[SDESK-7940] - Selecting CV items parent not working properly first time in desks with preffered items set

BrianMwangi21:
### Purpose
This PR fixes the CV metadata dropdown so subgroup results behave consistently when searching even for desks with preferred CV items configured.

### What has changed
- Updated the metadata terms dropdown to render searched group items the same way as browsed group items.
- Added the group `>` indicator for matching subgroup results in search mode.
- Reset dropdown state on close so reopening returns to the expected desk preferred view or root list.

### Steps to test
1. Import the `Products.json` in the ticket into a content profile and assign it to a desk.
2. Configure some preferred CV items for that desk.
3. Open authoring and use the CV dropdown/search.
4. Search for a subgroup like `Motori`.
5. Verify `Motori` appears as a group in search results with the `>` indicator.
6. Click `Motori` and verify it opens the subgroup list with the proper header/navigation.
7. Close and reopen the dropdown and verify it returns to the expected preferred/root state without losing the search input behavior.

#### Before
image

### Checklist

- [x] I reviewed my code

Resolves: [SDESK-7940](https://sofab.atlassian.net/browse/SDESK-7940)

[SDESK-7940]: https://sofab.atlassian.net/browse/SDESK-7940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/core/directives/TypeaheadDirective.ts
+++ b/scripts/core/directives/TypeaheadDirective.ts
@@ -114,7 +114,6 @@ export default angular.module('superdesk.core.directives.typeahead', [])
             link: function(scope, element, attrs, controller) {
                 var $input = element.find('.input-term > input');
                 var $list = element.find('.item-list');
-                let wasVisible = false;
 
                 $input.on('focus', () => {
                     scope.$applyAsync(() => {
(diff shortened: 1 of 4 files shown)


scripts/core/directives/TypeaheadDirective.ts
```diff
@@ -203,6 +207,12 @@ export default angular.module('superdesk.core.directives.typeahead', [])
                 });
 
                 scope.$watch('isVisible()', (visible) => {
+                    if (wasVisible && !visible && !scope.focused && typeof scope.close === 'function') {
+                        scope.close();
+                    }
+
+                    wasVisible = visible;
```
petrjasek: you can get the previous value, but not sure actually why it's needed? isn't this identical to `! visible` ?

BrianMwangi21: makes sense! thanks

petrjasek (approved):

PR #5243, opened 2026-06-30
[SDESK-7802] - Group actions in the tree-dot menu in React Actions bar

BrianMwangi21:
### Purpose
This PR update the React authoring actions menu to use the newer ui-framework api that supports grouping

### What has changed
- Bumped to `superdesk-ui-framework` to 7.0.0-dev1 which has new grouped menu capabilities
- Swapped the authoring actions menu from using Dropdown to use Menu

### Steps to test
1. Open an article in authoring and open the actions menu.
2. Verify actions are grouped correctly and separators appear between sections similar to the Angular view

### Screenshots
| Before    | After |
| -------- | ------- |
| image |  image |

### Checklist

- [X] I reviewed my code
- [X] I tested all affected functionality and made sure it works

Resolves: [SDESK-7802](https://sofab.atlassian.net/browse/SDESK-7802)

[SDESK-7802]: https://sofab.atlassian.net/browse/SDESK-7802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/client/playwright/editor3.spellchecker.authoring-react.spec.ts
+++ b/e2e/client/playwright/editor3.spellchecker.authoring-react.spec.ts
@@ -51,12 +51,12 @@ test('toggling spellchecker on/off shows/clears warnings (authoring-react)', asy
     await expect(warnings).toHaveCount(2);
 
     await page.getByRole('button', {name: 'Actions menu'}).click();
-    await page.getByRole('menuitem', {name: 'Run automatically'}).click();
+    await page.getByText('Disable spellchecker', {exact: true}).click();
 
     await expect(warnings).toHaveCount(0);
 
     await page.getByRole('button', {name: 'Actions menu'}).click();
-    await page.getByRole('menuitem', {name: 'Run automatically'}).click();
+    await page.getByText('Enable spellchecker', {exact: true}).click();
 
     await expect(warnings).toHaveCount(2);
 });
--- a/e2e/client/playwright/editor3.spellchecker.authoring-react.spec.ts
+++ b/e2e/client/playwright/editor3.spellchecker.authoring-react.spec.ts
@@ -51,12 +51,12 @@ test('toggling spellchecker on/off shows/clears warnings (authoring-react)', asy
     await expect(warnings).toHaveCount(2);
 
     await page.getByRole('button', {name: 'Actions menu'}).click();
-    await page.getByRole('switch', {name: 'Run automatically'}).click();
+    await page.getByRole('menuitem', {name: 'Run automatically'}).click();
 
     await expect(warnings).toHaveCount(0);
 
     await page.getByRole('button', {name: 'Actions menu'}).click();
-    await page.getByRole('switch', {name: 'Run automatically'}).click();
+    await page.getByRole('menuitem', {name: 'Run automatically'}).click();
 
     await expect(warnings).toHaveCount(2);
 });
--- a/e2e/client/playwright/editor3.spellchecker.authoring-react.spec.ts
+++ b/e2e/client/playwright/editor3.spellchecker.authoring-react.spec.ts
@@ -51,12 +51,12 @@ test('toggling spellchecker on/off shows/clears warnings (authoring-react)', asy
     await expect(warnings).toHaveCount(2);
 
     await page.getByRole('button', {name: 'Actions menu'}).click();
-    await page.getByText('Disable spellchecker', {exact: true}).click();
+    await page.getByRole('switch', {name: 'Run automatically'}).click();
 
     await expect(warnings).toHaveCount(0);
 
     await page.getByRole('button', {name: 'Actions menu'}).click();
-    await page.getByText('Enable spellchecker', {exact: true}).click();
+    await page.getByRole('switch', {name: 'Run automatically'}).click();
 
     await expect(warnings).toHaveCount(2);
 });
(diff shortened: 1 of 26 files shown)


e2e/client/playwright/editor3.spellchecker.authoring-react.spec.ts
```diff
@@ -51,12 +51,12 @@ test('toggling spellchecker on/off shows/clears warnings (authoring-react)', asy
...
 
     await page.getByRole('button', {name: 'Actions menu'}).click();
-    await page.getByText('Disable spellchecker', {exact: true}).click();
+    await page.getByRole('menuitem', {name: 'Run automatically'}).click();
 
     await expect(warnings).toHaveCount(0);
 
     await page.getByRole('button', {name: 'Actions menu'}).click();
-    await page.getByText('Enable spellchecker', {exact: true}).click();
+    await page.getByRole('menuitem', {name: 'Run automatically'}).click();
```
eos87: let's leave this one as it was (without the checkbox and Disable/Enable spellchecker). We don't want to carry the complexity and kind of the inconsistency of this specific option in the legacy menu.

scripts/apps/authoring-react/subcomponents/authoring-actions-menu.tsx
```diff
@@ -35,6 +40,82 @@ export class AuthoringActionsMenu extends React.PureComponent<IProps, IState> {
...
+                value: action.icon === 'toggle-on',
+                onChange: (value) => {
+                    action.icon = value ? 'toggle-on' : 'toggle-off';
+                    action.onTrigger();
+                    this.setState((state) => ({
+                        actions: state.actions == null ? null : [...state.actions],
+                    }));
+                },
+            };
+        }
```
eos87: let's not worry about the switch, as I mentioned before, I rather keep consistency in the UI so you can remove it.

eos87 (changes requested):
Let's keep the dynamic grouping we had instead of hardcoding the groups into `ACTION_GROUPS`. While the `groupId` is not enough, maybe we should introduce an additional (bit more complex) `group` attribute that would hold the priority and the label together with the id. Something like this 

```tsx
interface IAuthoringActionGroup {
    id: string;
    label?: string;      // header text; omit => separator-only (current behavior)
    priority?: number;   // group order across the menu
}
```

and update the references to use `group` instead of `groupId`  (perhaps print a warning if an action still has `groupId` and insert it into an empty-labeled group "general"). This way also we don't need to rely on translatable strings as ids.

eos87:
ahh one more thing, please add an end to end tests to cover the functionality. You can use the `superdesk-e2e` skill 😁

scripts/apps/authoring-react/subcomponents/authoring-actions-menu.tsx
```diff
@@ -35,6 +32,111 @@ export class AuthoringActionsMenu extends React.PureComponent<IProps, IState> {
...
     }
 
+    toMenuItem(action: IAuthoringAction): IMenuItem {
+        return {
+            label: action.label,
+            onClick: action.onTrigger,
+            shortcut: (action.group?.id ?? action.groupId) === 'spellchecker'
+                && action.label === gettext('Check spelling')
+                ? 'Ctrl+Shift+Y'
+                : undefined,
```
eos87: two questions here: 
- does it work when translations are on in another language different than English? 
- why is the shortcut hardcoded here?

scripts/apps/authoring-react/subcomponents/authoring-actions-menu.tsx
```diff
@@ -35,6 +32,111 @@ export class AuthoringActionsMenu extends React.PureComponent<IProps, IState> {
...
+        } else if (action.groupId != null) {
+            console.warn(
+                'AuthoringAction "' + action.label + '" uses deprecated groupId "' + action.groupId + '". ' +
+                'Migrate to the "group" attribute.',
+            );
+
+            // TODO: Remove this fallback once superdesk-planning migrates from groupId to group.
+            const isPlanning = action.groupId === 'planning-actions';
+            const fallbackLabel = isPlanning ? gettext('Planning') : undefined;
+            const fallbackPriority = isPlanning ? 10 : 0;
```
eos87, line 88: I'm wondering if we can already migrate superdesk-planning

scripts/apps/authoring-react/subcomponents/authoring-actions-menu.tsx
BrianMwangi21: Two answers here:
- Yes it does
- I've removed the hard coded shortcut. now it is got from the keyBindings property.

scripts/apps/authoring-react/subcomponents/authoring-actions-menu.tsx
BrianMwangi21: We can migrate for sure but once this PR is merged into develop. Cause I'd need to use the new `IAuthoringActionGroup` interface which only exists in this PR for now. 

And then once we also update planning, I can come back and remove this fallback cause it won't be needed anymore.

eos87 (approved):
Looks good. One last thing is: can you adjust the e2e spec to check the group labels are there too? I think at the moment only checks the menu items are visible but not the group labels. It would be also great to verify the dividers. Thanks!

PR #5254, opened 2026-07-21
Fix e2e base image caching failing on forks due to missing trailing newline

eos87:
### What has changed
Two fixes:

- `e2e/server/base-images.txt` was missing a trailing newline on its last line (`marlonb/mailcrab:latest`). The pull loop uses `while IFS= read -r image`, which skips a final line with no newline, so mailcrab was never pulled. `mapfile` however keeps that line, so `docker save "${base_images[@]}"` was asked to save an image that had never been pulled, and failed. Added the trailing newline.
- Hardened the pull loop in `.github/workflows/tests.yml` to `while IFS= read -r image || [ -n "$image" ]; do` so a future missing newline cannot silently drop the last image again.


eos87:
Closing in favor of 
https://github.com/superdesk/superdesk-client-core/commit/c30c827109971137a9dfa5badab31d6a00db3a9c

PR #5294, opened 2026-08-06
[SDESK-7804] Render the send to / publish widget dark and remove the leftover panel

eos87:
### Purpose

A user opens an article in authoring-react and hits the purple door to send or publish it. The panel that slides in is white, while everything around it is dark. It reads as a different application bolted onto the editor.

The white panel is also a leftover. SDESK-7447 moved these actions into a side widget back in February, but its first bullet, "remove the interactive article actions panel", never happened. So authoring-react has been carrying two hosts for the same five actions, and the one the ticket screenshotted is the one that should not be there any more.

This makes the widget dark and finishes the removal.

### What has changed

- The send to / publish widget renders on a dark surface. `Layout.Panel` already accepts a `theme`, which emits `data-theme="dark-ui"`; the widget host just never passed it. The custom properties inherit to the whole subtree, so nothing else needed styling.
- The panel is no longer rendered inside authoring-react, and the `markupV2` branches that existed only to serve it are gone from `panel-main`, `panel-header`, `panel-content`, `panel-footer`, `index-ui` and the six `*-tab` files.
- **The panel component itself is untouched.** Monitoring, search and Angular authoring still render it, and they still dispatch `interactiveArticleActionStart`.

Two gaps turned up while removing it. Both would have shipped as regressions, so they are fixed here rather than filed:

- The widget never actually opened on `interactiveArticleActionStart`. Its listener only updated tabs on an already-mounted widget; nothing selected it. A "Send to" from the monitoring list against an item open in authoring-react would have gone nowhere once the panel was removed.
- Publishing validation errors were not reaching the fields. The panel fed them back through `setValidationErrors`; the widget's `handleError` was an empty stub, so field-level highlighting was silently lost. Toasts still fired, which is why it was easy to miss.

### Steps to test

With `auth-react` enabled:

1. Open an article, click the purple door. The widget should be dark, matching the rest of authoring.
2. Send the article to another desk from the widget and confirm it arrives.
3. With that article open, trigger "Send to" on the same item from the monitoring list. The widget should open on the right tab.
4. Try to publish an article that fails validation. The offending fields should highlight, not just show a toast.
5. With `auth-react` off, check the purple door in monitoring still behaves exactly as before.

### Known narrowing

For an archived or killed item that is open in authoring-react **and** actioned from the monitoring list, the action is now a no-op where the panel previously opened. The widget's `isAllowed` returns false for those states. Fixing it properly means changing `triggeredFromAuthoring` semantics in `index-hoc.tsx`, which affects Angular authoring too, so it is deliberately out of scope here.

### Checklist

- [x] I reviewed my code

Resolves: SDESK-7804

### DIFF
--- a/e2e/client/index.js
+++ b/e2e/client/index.js
@@ -31,6 +31,10 @@ setTimeout(() => {
                 id: 'availability-manager',
                 load: () => import('superdesk-core/scripts/extensions/availability-manager'),
             },
+            {
+                id: 'publishing-sections',
+                load: () => import('./test-extensions/publishing-sections'),
+            },
         ],
         {}
     );
--- a/e2e/client/playwright/page-object-models/authoring.ts
+++ b/e2e/client/playwright/page-object-models/authoring.ts
@@ -71,6 +71,18 @@ export class Authoring {
         await this.page.waitForTimeout(2000);
     }
 
+    /**
+     * Opens an authoring-react side widget from the sidebar.
+     * The sidebar buttons all share the `widget-icon` test id and are told apart by
+     * `data-test-value`, which carries the widget id.
+     */
+    async openSideWidget(widgetId: string): Promise<void> {
+        await this.page
+            .getByTestId('widget-icon')
+            .and(this.page.locator(`[data-test-value="${widgetId}"]`))
+            .click();
+    }
+
     field(field: string): Locator {
         return this.page.locator(s('authoring', field)).getByRole('textbox');
     }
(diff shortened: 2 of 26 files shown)


eos87:
Closing unmerged. Investigating the light panel surfaced a larger problem: send to / publish is hosted as a side widget, and the fixes here all exist to make a widget behave like a panel.

SDESK-7804 has been rewritten to move these actions back into a panel for authoring-react, and the work continues on `auth-react/SDESK-7804-publish-panel`. The `publishingSections` test fixture from this branch is being carried over; the rest is superseded.

PR #5319, opened 2026-08-08
Automate Underline (1313669372)

eos87:
### Purpose
Automates the QA case [Underline](https://sofab.atlassian.net/wiki/spaces/SET/pages/1313669372) as a Playwright spec so it runs in CI instead of being tested by hand.

Stacked on the Bold branch (#5304): it reuses that branch's formatting-option-button test id, so this PR is based on hg/automate-editor3-bold and GitHub will retarget it to develop when #5304 merges. Review only this PR's delta.

### What has changed
- New spec `e2e/client/playwright/editor3.underline.spec.ts` covering the scenario below
- Annotated with `{type: 'confluence', description: '1313669372 partial'}` per the coverage convention

### Steps to test
**Given** an article created from the Sports default template with a headline

**When**
1. The body's toolbar Underline button is toggled on and text is typed
2. The button is toggled off and more text is typed
3. Underline is toggled on over a selected part of the un-underlined run and off over a selected part of the underlined run
4. The article is closed and saved through the unsaved-changes prompt, then reopened

**Then**
- Each run's computed `text-decoration-line` matches the toggle that produced it
- The toolbar button's active state tracks the caret/selection
- The mix of underlined and plain runs is unchanged after reopen

Run it: `cd e2e/client && npx playwright test editor3.underline.spec.ts`

The spec passed 3 consecutive local runs with no changes between them.

### Checklist
- [x] I reviewed my code


eos87:
Generated with AI. Reviewed and ran it locally. Looks good.

BrianMwangi21 (approved):

PR #5325, opened 2026-08-08
Add editor3-formats and authoring-extras e2e snapshots, roles and `samgamgee` user

eos87:
### Purpose
Second fixture batch from the QA automation campaign, closing the gaps that blocked or limited eleven specs. The big one: no snapshot enabled strikethrough, subscript, superscript, preformatted, formatting marks, h1/h3-h6, or list options on any editor3 field, which fully parked five Editor3 QA cases and left six others `partial` on their "custom text fields" clause.

### What has changed
- New `editor3-formats` record snapshot: the Story profile's `body_html` gains the twelve missing format options, plus a "Sample rich text" custom editor3 field (vocabulary `sample_rich_text`) carrying the same options
- New `authoring-extras` record snapshot: "test sports story" carries `expiry` and a "Rivendell" keyword for the Info widget scenarios (the API write re-versions the item, which is documented: it sorts to the top of Sports / Working Stage under this snapshot)
- New `saved-search-private` record snapshot: admin-owned non-global "Shire drafts" (kept out of `main` because an open campaign PR counts private searches there)
- `main` gains a roles collection (Administrator with every registered privilege, and "Sub Editor" lacking `send_to_personal` and `unlock`; neither is default) and `samgamgee` (password `samgamgee`, Sub Editor, Sports desk member), the campaign's second known-credentials user alongside `frodobaggins`
- `WRITING_TESTS.md` and the records README document all of it, including the locator gotchas (option keys with spaces, the formatting-marks button lacking `data-test-value`, the mislabelled keywords row in the metadata widget)

### Steps to test
1. Restore `editor3-formats` and open a new article: the Body toolbar shows all the new options, and the "Sample rich text" field renders with its own editor3 toolbar
2. Restore `main`: both revert (toolbar back to the original nine options, custom field gone)
3. Log in as `samgamgee`/`samgamgee`: lands on Sports, holds Sub Editor privileges minus `send_to_personal` and `unlock`
4. Restore `authoring-extras`: the Info widget on "test sports story" shows the Expiry line and the Rivendell keyword

Existing user/monitoring/saved-search specs pass unchanged (verified against a pristine branch-point baseline, `--workers=1`).

### Checklist
- [x] I reviewed my code
- [x] I tested all affected functionality and made sure it works

PR #5310, opened 2026-08-07
Add media-items e2e snapshot and `frodobaggins` desk-member user

eos87:
### Purpose
The e2e fixture backlog from the QA automation campaign: several documented scenarios cannot be automated because the base snapshot lacks the state they need. This adds the two most-requested pieces: media items of every type, and a second desk-member user with known credentials and zero privileges. Together they unblock positive type-filter assertions, image editing scenarios, and multi-user / privilege-branch scenarios (e.g. the "no Unlock privilege" branch of Lock item).

### What has changed
- New `media-items` record snapshot layered on `main` (restore with `restoreDatabaseSnapshot({snapshotName: 'media-items'})`): "Rivendell picture" (picture), "Moria graphic" (graphic), "Isengard video" (video), "Lothlorien audio" (audio) and "Shire package" (plain composite) in Sports / Working Stage. Kept out of `main` on purpose: existing specs assert exact item counts against it (see the new note in the records README).
- New user `frodobaggins` (password `frodobaggins`, matching the `admin`/`admin` convention) in `main`: Sports desk member, `user_type: user`, null role, no privileges field, so it holds zero privileges. Documented in `WRITING_TESTS.md`.
- `e2e/server/settings.py` derives `MEDIA_PREFIX` from `SERVER_NAME` and rewrites the recorded `:5000` rendition prefix to whatever port the backend answers on (no-op on CI, which runs on 5000). Without this, every thumbnail in the snapshot 404s on a local `:5002` stack or a slot. `docker-compose.yml` now bind-mounts `settings.py` like the slot compose already did, so prebuilt server images pick it up.
- `fixtures/audio.wav` added (no audio fixture existed).

Caveat documented in both docs: "Moria graphic" is a mongo-flipped picture (the API cannot create graphics) and keeps `profile: picture`, so type filtering works but graphic-specific authoring behaviour is not represented.

### Steps to test
1. `./e2e/scripts/e2e-up.sh`, then `POST /api/restore_record {"name": "media-items"}`: all five items appear in Sports / Working Stage in Monitoring, with thumbnails rendering (rendition URLs rewritten to your port)
2. Restore `main`: the media items are gone, and `frodobaggins`/`frodobaggins` logs in via the login form
3. As admin, lock an item; `frodobaggins` gets 403 on unlock (no privilege), while admin can unlock

[...]

PR #5327, opened 2026-08-08
Automate the Editor3 formatting style QA cases as a consolidated spec

eos87:
### Purpose
Automates five Editor3 formatting style QA cases as one consolidated Playwright spec. They exercise the same toolbar mechanics and differ only in the style applied, so they share one spec file with per-test `confluence` annotations. The styles are only enabled by the `editor3-formats` snapshot, so this PR is stacked on the fixture chain.

Stacked on the fixture branch (#5325): review only this PR's delta. Merge order: #5310, #5325, then this (GitHub retargets automatically as bases merge).

Covered cases:
- [Strikethrough](https://sofab.atlassian.net/wiki/spaces/SET/pages/1313669365) - `complete`
- [Subscript](https://sofab.atlassian.net/wiki/spaces/SET/pages/1313669368) - `complete`
- [Superscript](https://sofab.atlassian.net/wiki/spaces/SET/pages/1313669370) - `complete`
- [Preformatted text](https://sofab.atlassian.net/wiki/spaces/SET/pages/1313669359) - `partial`
- [Toggle formatting marks](https://sofab.atlassian.net/wiki/spaces/SET/pages/1313669377) - `complete`

### What has changed
- New spec `e2e/client/playwright/editor3.formatting-styles.spec.ts` with per-test `confluence` annotations
- Replicates the `formatting-option-button` test id on `StyleButton.tsx` byte-identically from the Bold branch (#5304), so the merges dedupe
- Every style is also exercised on the "Sample rich text" custom editor3 field, covering the cases' custom-text-fields clause

## Editor3 formatting styles (`editor3.formatting-styles.spec.ts`)

**Given** the `editor3-formats` snapshot is restored (full format toolbar on the Story profile `body_html` plus the "Sample rich text" custom editor3 field), the Sports monitoring view is open and a new article is created from the default desk template

### Inline styles: strikethrough (1313669365), subscript (1313669368), superscript (1313669370)

**When**
1. the caret is placed in the Body field and the style's toolbar button is clicked, then text is typed
2. the button is clicked again and more text is typed
3. the second half of the regular run is selected and the button is clicked
4. the second half of the styled run is selected and the button is clicked
5. the article is closed and saved through the "Save changes?" prompt, then reopened from Monitoring
6. the same toggle-on-and-type step is repeated in the "Sample rich text" custom field, and the article is saved, closed and reopened again

**Then**
- the button carries `Editor3-activeButton` exactly while the style is on, and loses it on every toggle off
- text typed with the button active renders styled (`text-decoration-line: line-through`, `vertical-align: sub`, `vertical-align: super`) and text typed with it inactive does not
- toggling over a selection restyles only the selected half, in both directions, and re-syncs the button's active state to the selection
- all four runs come back with their styles after the save/reopen round-trip, and the headline is preserved
- the custom field behaves the same way and its styled run also survives save and reopen

### Preformatted text (1313669359, partial)

**When**
1. the caret is placed in the empty Body field and the Preformatted text button is clicked
2. text is typed, then the button is clicked again and more text is typed
3. a new paragraph is started, regular text is typed, part of it is selected and the button is clicked
4. a further paragraph is started (inheriting the preformatted style), text is typed, part of it is selected and the button is clicked
5. the article is closed and saved, then reopened
6. the toggle is repeated in the "Sample rich text" custom field, and the article is saved, closed and reopened again

**Then**
- the block under the caret becomes a `pre` block, the caret sits inside it, and the wrapper editor3 puts around preformatted blocks carries the `Roboto Mono` family
- text typed with the button active lands in the `pre` block; text typed with it inactive lands in a `div` block
- toggling over a partial selection converts the whole paragraph in both directions, because preformatted is a block style
- after reopen the first and last paragraphs are regular and the middle one is still preformatted, and the custom field's preformatted block survives too
- **not covered** (asserted as the product actually behaves, not as the case describes): toggling the option off does not move the caret out into a new paragraph, it reverts the current block (SDESK-4253, linked from the case); and preformatted text does not render in a courier face, because a global `:not([class^="icon-"], ...)` font reset re-declares Arial on every descendant of the monospace wrapper

### Toggle formatting marks (1313669377)

**When**
1. the caret is placed in the Body field and the "Toggle formatting marks" button is clicked
2. three paragraphs are typed
3. the button is clicked off and on again
4. text is typed in the "Sample rich text" custom field and that field's own marks button is clicked
5. the article is closed and saved, then reopened

**Then**
- the button carries `Editor3-activeButton` while marks are on, and the editor renders an `::after` mark on every block whose computed content is the icon glyph in `rgb(64, 153, 191)` (`--sd-colour-interactive`, the blue the case names)
- every one of the three paragraphs carries a mark, all marks disappear when the option is toggled off, and all reappear when it is toggled back on
- the toggle is per editor instance: turning it on in the custom field leaves the body's marks as they were
- after reopen the three body paragraphs and the custom field text are preserved, and the marks are off again, since the toggle is editor UI state rather than part of the item

Run it: `cd e2e/client && npx playwright test editor3.formatting-styles.spec.ts`
The spec passed 3 consecutive local runs with no changes between them, plus a green verification run after the review polish.

### Checklist
- [x] I reviewed my code


eos87 (commented):
This was implemented fully by AI. I reviewed and ran it locally. It looks good.

BrianMwangi21 (approved):

PR #5350, opened 2026-08-10
Add publishing e2e snapshot with a NINJS subscriber and enable associated-item publishing

eos87:
### Purpose
Third fixture batch: unlocks the publishing-dependent QA cases (the locked-items publishing/correcting cases, the NINJS output case, and the publishing family in #5349).

### What has changed
- New `publishing` record snapshot: `main` plus an unfiltered "All content" product and subscriber "Public API" with a "NINJS Email" destination. HTTP push cannot succeed inside the E2E slot network (nothing answers, entries hang in `retrying`), so delivery rides the email transmitter's verbatim fallback into the stack's mailcrab, which keeps a genuine NINJS payload and reaches queue state `success` (verified at the API and in the UI). Kept out of `main` because `subscribers.spec.ts` asserts the one-subscriber count there. Note: under this snapshot an untargeted publish queues two rows, so queue assertions narrow by Subscriber or Destination.
- `PUBLISH_ASSOCIATED_ITEMS = True` in `e2e/server/settings.py`. This is app-level config that can't be toggled per spec, so it's on for the whole suite. That's the same approach we use for `VALIDATOR_MEDIA_METADATA`, and it's how superdesk-core's own test suite runs. It's safe right now because no snapshot item has `associations` and no spec (on develop or any open campaign branch) publishes one. The first spec that does will see the associated items published along with the story; the settings comment and `WRITING_TESTS.md` explain this. I checked both behaviours it enables against a running stack: the locked-item publish errors and the associated-items publishing work as the QA cases describe.
- Docs in `WRITING_TESTS.md` (the publishing snapshot, the two-queue-rows caveat, why publish assertions need no polling given `CELERY_ALWAYS_EAGER`) and the records README.

### Steps to test
1. Restore the `publishing` snapshot and publish "test sports story": the Publish Queue page (`/#/publish_queue`) shows a `success` NINJS row for "Public API", and mailcrab holds the ninjs body
2. Publish an article whose associated item is locked by another user: refused with the documented error; unlock and it publishes, association included
3. Existing publish-flow specs pass (22 green; the 2 failures reproduce identically at the branch point)

### Checklist
- [x] I reviewed my code


eos87:
Created entirely with AI. I reviewed the code, tested it locally, CI passes. Looks good to me!

BrianMwangi21 (approved):

PR #5266, opened 2026-07-24
[STT-1694] Add spell checker toggle to authoring header with configurable shortcut

BrianMwangi21:
### Purpose
This PR add a dedicated spell checker toggle button in the authoring header, next to the proof-read theme toggle, and gives it a simple, configurable keyboard shortcut.

### What has changed
- Added a spell checker toggle button in the header (icon-font "A" glyph) with an active/dark state when on.
- Added a shortcut to toggle spell checking, defaulting to `Alt/Option+S`, configurable via `spellchecking.shortcut`.
- The button, shortcut, and "…" menu switch now share the same `spellcheck` service state.
- Dropped the old `Ctrl+Shift+Y` alias.
- Extracted shared cross-platform shortcut helpers and added unit tests.

### Steps to test
1. Open a story in the authoring view.
2. Click the spell checker button in the header — the button darkens (active) and spell checking turns on; click again to turn it off.
3. Press `Alt/Option+S` — it toggles the same state and the button reflects it.
4. Optionally set `spellchecking.shortcut` in the config (e.g. `{key: 's', modifiers: ['alt', 'shift']}`) and confirm the new shortcut works and the tooltip label updates.

### Checklist
- [x] I reviewed my code
- [x] I tested all affected functionality and made sure it works

Resolves: https://sofab.atlassian.net/browse/STT-1694

PR #5355, opened 2026-08-24
Merge release/3.5 into develop

eos87:
### Purpose
Forward-port `release/3.5` into `develop`. 3.5 had three commits that never made it across: the STT-1694 spell checker toggle with configurable shortcut (#5266), the 3.5 e2e server build fix (#5273), and the STT-1673 backport (#5257). The last two are already effectively on develop, so this is mostly about STT-1694.

### What has changed
- Merge of `origin/release/3.5`. The only conflicts were three Protractor specs that 3.5 touched and develop has since deleted in the Playwright migration. Kept them deleted: 3.5 only renamed a selector in them and nothing on develop uses the old one.
- Second commit fixes the new `spellcheck-shortcut.spec.ts` from 3.5 so it doesn't depend on the host OS. It failed on macOS with `Option+S` vs `Alt+S`. Moved the `withPlatform` helper into `core/tests` and reused it in both shortcut specs.


petrjasek (approved):
looks good to me

PR #5357, opened 2026-08-25
Content lists extension

IvanJelicSF:
# Content lists extension

Adds a new `content-lists` client extension: a page for curating manually ordered lists of articles, plus management of the webhooks that notify external consumers when those lists change. It is the client side of the content-lists API in superdesk-core; the page is only contributed when the user has the `content_lists` privilege.

## What it does

**Lists grid** (`src/lists-grid/`)
- Cards for every manual content list, with a preview of the first few articles, an "items updated" timestamp, and a search field over list names.
- Create a list inline; edit name, article limit, description and an active/inactive toggle in a settings modal; delete with confirmation.
- Inactive lists are marked with a disabled label on the card.

**List editor** (`src/list-editor/`)
- Two panes: the list itself and an article picker. Articles are dragged from the picker into the list (react-beautiful-dnd) and reordered by dragging.
- Picker searches Published / Scheduled / In progress sources with paging; results already present in the list are dimmed so they can't be added twice.
- Items can be pinned (sticky position, drag-disabled) or removed; the article limit surfaces a notification when exceeded.
- Changes are batched and sent as one PATCH; saving uses the list's `content_list_items_updated_at` token for optimistic concurrency and reports a conflict when another user saved in the meantime. Navigating away with unsaved changes prompts for confirmation.
- Article references that no longer resolve render as "no longer available" rows instead of breaking the list.

**Webhooks** (`src/webhooks/`)
- Modal to list, search and filter (all/enabled/disabled) webhooks, and a side panel to create or edit one: URL, enabled toggle, and excluded lists.

**Live updates** (`src/live-updates.ts`)
- Subscribes to the backend's `content_list:*` websocket events plus `content:update`, so grid cards, list contents and article previews refresh when someone else changes them. Requires the backend to address those notifications to the extension (`extension="content-lists"`).

## Changes outside the extension

- `scripts/core/helpers/match-page-url.ts` (new) + `extension-page.tsx`: extension page URLs may now contain `:param` segments; exact matches still win.
- `scripts/core/register-extensions.tsx`: `reloadOnSearch: false` on extension page routes, so a page can keep state in query parameters without reloading the whole view. The content-lists page uses this to keep the selected list in the URL without the top menu flickering on every navigation.
- `.github/workflows/tests.yml`: install the extension's dependencies and run its unit tests in CI.
- `e2e/client/index.js`: register the extension in the e2e client.
- `e2e/server/docker-compose.yml`: raise mongo's `nofile` ulimit to 64000. Each spec's `restore_record` drops and recreates every collection and index, and WiredTiger holds a descriptor per file, so a full suite run exhausted the default 1024 and mongod aborted with `WT_PANIC: Too many open files`.

## Tests

- 167 unit specs covering the API layer, article search queries, utils, live updates and every component. They run under Karma with a webpack/swc setup that mirrors the root config (`scripts/extensions/content-lists/karma.conf.js`), resolving tooling from the repo root's `node_modules`.
- 20 Playwright specs under `e2e/client/playwright/extensions/content-lists/` covering the grid, editor, picker, dimmed already-added results and webhooks, with shared API helpers for fixture setup.

## Notes for the reviewer

- The extension has its own `package.json` (only direct dependency: `react-beautiful-dnd`) and an `.npmrc` with `legacy-peer-deps=true`. React, lodash, superdesk-ui-framework and @sourcefabric/common are host-provided peers; installing them locally would pull in a second React copy and break rendering, and a plain `npm install` would otherwise fail with ERESOLVE.

Resolves [SDESK-8011](https://sofab.atlassian.net/browse/SDESK-8011)

[SDESK-8011]: https://sofab.atlassian.net/browse/SDESK-8011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/tests.yml
```diff
@@ -33,11 +33,21 @@ jobs:
         restore-keys: |
           webpack-${{ runner.os }}-node${{ matrix.node-version }}-
 
+    - name: Install content-lists extension dependencies
+      run: npm ci --legacy-peer-deps
```
eos87: why is `--legacy-peer-deps` needed here?

scripts/extensions/content-lists/src/utils.ts
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * Re-inserts every pinned entry at its recorded sticky position
+ * after any mutation of the list.
+ */
+export function fixPinnedItemsPosition(entries: Array<IListEntry>): Array<IListEntry> {
+    let result = entries;
+
+    result.forEach((entry, index) => {
+        if (entry.sticky && entry.stickyPosition != null) {
+            result = reorder(result, index, entry.stickyPosition);
```
eos87: This breaks with two or more pinned items. The loop iterates over the original array's indexes but reorders `result`, which has already shifted after the first `reorder`, so the second pinned entry is looked up at a stale index.

Repro: list `A, S1 (pinned at 1), S2 (pinned at 2), B`, drag A to the bottom. We render `S2, B, S1, A`, but the server (and what you see after save + reload) gives `B, S1, S2, A`. Any further drags then record positions from the wrong on-screen order.

Suggest iterating over the pinned entries sorted by `stickyPosition` and re-finding each one's current index in `result` before reordering. Worth adding a unit test with two adjacent pinned items; the current specs only cover one.

scripts/core/register-extensions.tsx
```diff
@@ -36,6 +36,10 @@ export function registerExtensions(
             template: '<sd-extension-page '
                 + 'data-setup-full-width-capability="setupFullWidthCapability">'
                 + '</sd-extension-page>',
+
+            // allow pages to store state in query parameters
+            // without the route (and the entire view) reloading
+            reloadOnSearch: false,
```
eos87: I think this is not necessary and has no effect. `superdesk.activity()` already defaults every activity to `reloadOnSearch: false` (see `scripts/core/activity/activity.ts:88`), so extension pages never reloaded on query-param changes to begin with.

e2e/client/playwright/extensions/content-lists/webhooks.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    // not an exact match: the item's icon is a font glyph rendered through a
+    // ::before pseudo-element, and Chromium folds it into the accessible name
+    await page.getByRole('menuitem', {name: 'Webhooks'}).click();
+    await expect(page.locator(s('manage-webhooks'))).toBeVisible();
+}
+
+// the row element; hover actions live outside the content wrapper carrying the test id
+function webhookRow(page: Page, url: string): Locator {
+    return page.locator('li').filter({has: page.locator(s(`webhook-item=${url}`))});
```
eos87: `page.locator('li')` is a structural selector, which we should avoid (see `WRITING_TESTS.md`).  Any markup change in `BoxedListItem` will break it silently. I'd suggest to put a `data-test-id` on the row itself, either on the `BoxedListItem` wrapper or by moving the test id up one level, and select that directly. Same for `.locator('button')` on the settings menu at line 7: a test id on the `NavButton` would do.

e2e/client/playwright/extensions/content-lists/picker-dims-added-articles.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        const list = await createContentList('dim already-added test');
+        const article = FIXTURE_ARTICLES.published;
+
+        await page.goto(`/#/content-lists?list=${list._id}`);
+
+        const pickerResults = page.getByTestId('content-list--picker-results');
+        const pickerArticle = articleRow(page, pickerResults, article.id);
+
+        // Before adding: the published article is offered normally.
```
eos87: Nit: 😬 
these "Before adding / Add it to the list and persist / After adding" comments narrate what the code already says; `WRITING_TESTS.md` asks to keep comments for the non-obvious why and let the test title carry the scenario. The `waitForResponse` comment further down is exactly the kind we want, keep that one. Same applies to `// re-open to verify persistence` in `lists-grid.spec.ts:55`.

eos87 (changes requested):
Thanks for this work, looking great really!


A few things: 

**e2e specs: please align them with `e2e/WRITING_TESTS.md`**

The specs work and CI is green, but they don't follow our conventions from #5223. 

Most of this is mechanical:

- `s()` is the legacy helper, new specs use `getByTestId`. Some of the specs use `s()`; `picker-dims-added-articles.spec.ts` already does it right. The doc has a translation table.
- The inline helpers (`openWebhooksModal`, `webhookRow`, `articleRow`, the repeated "Actions > Settings" sequence, and `dragAndDrop`) belong in a `page-object-models/content-lists.ts`. The drag helper especially: it carries the known-flaky workaround, so the next spec should inherit it instead of copying it.
- Most clicks go through `getByRole(name)` / `getByPlaceholder`. Role and text are fine for assertions, not as the primary selector for actions (localization). Adding data-test-id to the dropdown items and menu buttons is easy here, since the components are part of this PR.

Finally, there are some conflicts, please take a look. Let me know if you need a hand with these.

.github/workflows/tests.yml
IvanJelicSF: Actually not needed, removed

eos87 (approved):
Looks great! 🚢 it.

eos87 (approved):
Looks great! 🚢 it.

PR #5358, opened 2026-08-28
[SDESK-7988] - Firefox: Disconnected from Notification server message blinks every time after items are exported

BrianMwangi21:
### Purpose
In Firefox, exporting items flashed a "Disconnected from Notification Server!" warning on every export. Clicking the export download link navigated the top-level document, which made Firefox tear down the notification websocket (reconnecting ~5s later). Chrome was unaffected.

### What has changed
- Export now downloads via a fetched blob URL instead of pointing an anchor at the export URL and clicking it, avoiding the top-level navigation. Falls back to the direct link if the fetch fails.
- Added a Playwright e2e regression test for the export → download flow.

### How to test
- Firefox: Monitoring / Global search → select a few items → Export → pick a formatter → confirm. The file downloads and no "Disconnected from Notification Server" warning appears. 

Resolves: https://sofab.atlassian.net/browse/SDESK-7988

### DIFF
--- a/e2e/client/playwright/export.spec.ts
+++ b/e2e/client/playwright/export.spec.ts
@@ -1,0 +1,36 @@
+import {test, expect} from '@playwright/test';
+import {Monitoring} from './page-object-models/monitoring';
+import {restoreDatabaseSnapshot} from './utils';
+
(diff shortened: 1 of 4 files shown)


eos87 (approved):
Looks good

PR #5369, opened 2026-09-09
Match authoring-react field and panel styling to authoring-angular

eos87:
### Purpose

Authoring-react drifted visually from authoring-angular in a number of small ways that add up: labels in the wrong place, missing controls, colours and shadows that were never ported, and a collapse animation that only played in one direction. This closes the gap so the two views look the same.

Stacked on #5366, so review that one first.

### What has changed

**Header**
- Field labels render inline in a 90px right-aligned column, as in angular, instead of stacked above the input.
- The content profile icon is back, to the left of PROFILE.
- The item state badge ("IN PROGRESS") is back in the secondary toolbar. It was simply never registered.
- Fields now grow to fill their row. Angular treats the configured width as a `flex-basis`; react used a fixed `width`, so a row of quarter-width fields (PRIORITY, URGENCY) left half the row empty.
- The collapse animation now plays in both directions over 150ms. The framework transitions `max-height` against a 1400px ceiling on a ~400px header, so collapsing never visibly travelled.

**Body**
- Field labels are grey by default, darken on hover and turn blue on focus. React had dark labels by default and no focus colour.

**Top bar**
- Painted `--sd-colour-panel-bg--200` to match angular.
- The divider under it now renders. The shadow was missing, and the bar is a static grid item under three positioned boxes, so declaring it alone painted nothing.
- The close button swaps its label for an icon below 880px, matching angular's `sd-media-query`. It is now a square hollow button rather than a round one.

**Panels**
- Every side widget now matches the Related Items panel from #5366: white header, `--sd-colour-panel-bg--100` body, 16px inset.
- The column's left edge gets the monitoring pane's shadow back. React's authoring root sat at `z-index: 100`, the same as the pane, so its opaque background swallowed it.

**One bug found along the way**: `authoring-section.tsx` read a field's configured width in two places but defaulted it in only one, so a field without one rendered `width: "undefined%"` and collapsed to its content. A package's headline was rendering at 57% of its row and a picture's description at 61%.

### Steps to test

1. Open an article in Sports / Working Stage, with and without the `auth-react` localStorage flag, and compare the two side by side.
2. Header: labels sit inline, the profile icon is left of PROFILE, the state badge is left of "Created", and PRIORITY/URGENCY fill their row.
3. Click the chevron between header and body: it should slide both ways, not snap shut.
4. Hover the body: labels darken. Focus a field: its label turns blue.
5. Narrow the window below ~880px: the CLOSE button becomes an X.
6. Open any side widget: white header, grey body, cards inset.

### Checklist
- [x] I reviewed my code

Resolves:

### DIFF
--- a/scripts/apps/authoring-react/authoring-section/authoring-section.tsx
+++ b/scripts/apps/authoring-react/authoring-section/authoring-section.tsx
@@ -1,5 +1,11 @@
 import React, {RefObject} from 'react';
-import {IAuthoringSectionTheme, IFieldsV2, IVocabularyItem, IAuthoringValidationErrors} from 'superdesk-api';
+import {
+    IAuthoringFieldV2,
+    IAuthoringSectionTheme,
+    IFieldsV2,
+    IVocabularyItem,
+    IAuthoringValidationErrors,
+} from 'superdesk-api';
 import {Map} from 'immutable';
 import {IToggledFields} from '../authoring-react';
 import {AuthoringSectionField} from './authoring-section-field';
@@ -69,6 +75,15 @@ function groupItemsToRows<T>(items: Array<T>, getWidth: (item: T) => number) {
 }
 
 /**
+ * Percentage of the row the field asks for. A field with no configured width takes the whole row:
+ * `getContentProfile` defaults it that way too, but profiles built in code (packages, the
+ * description field on media items) leave it unset.
+ */
+function getFieldWidth(field: IAuthoringFieldV2): number {
+    return field.fieldConfig.width ?? 100;
+}
+
+/**
  * A variable is needed in order to use the same object reference
  * and allow PureComponent to skip re-renders.
  */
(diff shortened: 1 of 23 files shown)


scripts/apps/authoring-react/toolbar-components/angular-integration/minimize-button.tsx
```diff
@@ -1,18 +1,29 @@
...
     return (
-        <NavButton
-            text={gettext('Minimize')}
-            onClick={() => exposed?.keepChangesAndClose()}
-            icon="minimize"
-            iconSize="big"
-        />
+        <Tooltip content={gettext('Minimize')}>
+            {({attributes}) => (
+                <button
```
petrjasek: shouldn't we use our `Button` instead?

petrjasek (approved):

scripts/apps/authoring-react/authoring-section/authoring-section.tsx
```diff
@@ -112,8 +125,16 @@ export class AuthoringSection<T> extends React.PureComponent<IPropsAuthoringSect
...
+                                    // The width is a basis, so fields stretch to fill a row the
+                                    // widths leave short. In the header that matches angular
+                                    // (`.authoring-header__item {flex-grow: 1}`). In the body it
+                                    // deliberately does not: angular's `sd-width` sets only the
+                                    // basis on `.field`, which does not grow, and leaves the gap.
                                     return (
-                                        <div key={field.id} style={{width: `${field.fieldConfig.width}%`}}>
+                                        <div
+                                            key={field.id}
+                                            style={{flex: `1 1 ${getFieldWidth(field)}%`}}
```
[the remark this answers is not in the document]
eos87: this is intentional, body fields grow to fill the row on purpose

scripts/apps/authoring-react/toolbar-components/angular-integration/minimize-button.tsx
eos87: in this case NavButton is the right component, but it doesn't have a tooltip. I'll add support in the ui-framework side and then replace this. I'll add a TODO comment so I don't forget.

